PR #3, opened 2017-10-19
Add a break hint before "constraint" in a type def

### LINKED ISSUE #1: Fails on type with "constraint" applied
There is a space missing before the `constraint` keyword on output.

For example:
```ocaml
type 'a t = 'a list constraint 'a = [< `X]
```
gives:
```
$ ocamlformat test.ml -o test_formatted.ml
File "./test_formatted2055e4.ml", line 1, characters 30-31:
Error: Syntax error
```
with the following output:
```ocaml
type 'a t = 'a listconstraint 'a = [< `X]
```

hcarty:
Fixes #1


jberdine:
Thanks!

PR #5, opened 2017-10-19
Add spacing around = in records

### LINKED ISSUE #4: Record literal syntax--asymmetrical spacing around '='
# Observed behaviour

The following:

```ocaml
let x = {a = 1; b = true}
```

is reformatted as:

```ocaml
let x = {a= 1; b= true}
```

# Expected behaviour

One space on either side of the equals sign.

hcarty:
Fixes #4

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -34,6 +34,8 @@ let protect =
       raise exc
 
 
+let assignment (c: Conf.t) : s = if c.sparse then "@ =@ " else "=@ "
+
 let rec sugar_arrow_typ ({ast= typ} as xtyp) =
   let ctx = Typ typ in
   let {ptyp_desc; ptyp_loc} = typ in
@@ -528,7 +530,7 @@ and fmt_pattern (c: Conf.t) ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
             cbox 2 (fmt_longident txt)
         | _ ->
             cbox 2
-              ( fmt_longident txt $ fmt "@ =@ "
+              ( fmt_longident txt $ fmt (assignment c)
               $ cbox 0 (fmt_pattern c (sub_pat ~ctx pat)) )
       in
       hvbox 0
@@ -621,7 +623,8 @@ and fmt_fun_args c args =
       , Some xexp )
       when String.equal l txt ->
         cbox 0
-          (fmt "?(" $ str l $ fmt "= " $ fmt_expression c xexp $ fmt ")")
+          ( fmt "?(" $ str l $ fmt (assignment c) $ fmt_expression c xexp
+          $ fmt ")" )
     | Optional l, xpat, Some xexp ->
         cbox 0
           ( fmt "?" $ str l $ fmt ":(" $ fmt_pattern c xpat $ fmt " = "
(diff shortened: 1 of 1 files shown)


jberdine:
Considering the discussion in #4 , for this to proceed this change should be guarded behind an option in Conf.

hcarty:
@jberdine What do you think of using the additional spacing when the sparse configuration option is `true`?  See the latest commit on this PR for a potential implementation.

If you're ok with this approach I can rebase so the PR can merge cleanly.

hcarty:
And @yawaramin, same question.

jberdine:
Technically, that would still not pass `make test`. :-) But more thoughtfully, I don't think that it is clear what sets of alternatives people will gravitate to. My instinct is that, if the implementation is separate, it would be better to have a separate option so that everyone can select the set of choices that work for them. Over time, some patterns may emerge, and those can be encoded into 'styles' that aggregate a number of finer-grained options. Does that make sense? But admittedly the name of `sparse` is currently pretty bad, sorry about that.

hcarty:
Yeah, no tests added/checked/corrected yet :-)

I'll think on this a bit more.  The appropriate configuration option may have something to do with whitespace padding around lots of syntactic elements, ex `{x= 1}` vs `{ x = 1 }`.

jberdine:
Also, infer uses the sparse config, and `make test` checks that it doesn't change the formatting of infer and ocamlformat itself. :-)

yawaramin:
IMHO, configurable styles defeat the purpose of a code formatter. I am happy with one, consistent, style.

hcarty:
Thanks for the reviews!  I'll close this for now. If/when I come up with something more complete and separate I'll submit an updated PR.

I agree that, ideally, there will be one format used for projects. Even if that format evolves over time. That said, having some configurable options available helps when figuring out what that format should be! So there's some value in this early stage of the project in having more configuration options than most projects would or should use.

jberdine:
I agree that there would be value in all OCaml projects adopting the same style. But realistically that is very very far away. And more importantly, I think that there is a lot of value in auto-formatting a project even if the style is not exactly the same as some other projects. I think we can realize this second value long before the entire community agrees on a single style. And I agree that we need alternatives to evaluate the alternatives.

yawaramin:
Yes, agree that a young project needs some options to help judge between different styles. IMO a style configuration file (a 'style file') would be better than adding a bunch of command-line flags to the ocamlformat tool. This way in the future you don't need to break compatibility with previous versions if you want to remove options. You can just ship a standard style file and let that be the default for everyone who is OK with following the default style.

Btw, I'm not asserting that all OCaml projects should follow the same style, just that ocamlformat should try really hard to enforce a single consistent style. I think style files would allow a happy middle ground.

jberdine:
Currently there are both command line options and a config file, although there has been little consideration for accepting config files of previous versions. That's a good point. Or are you saying that it would be better to have only the file?

yawaramin:
Exactly, it would be best to allow configuration with only a style file. This way you don't need to worry about adding or removing command-line flags, you have one command-line flag (e.g., `--style`) and you make the tool forward-compatible by ignoring style file configuration it doesn't understand, and backward-compatible by supporting configuration of previous versions. Best of both worlds.

PR #24, opened 2017-10-22
Fix the printing of  infix operators starting with '#'

Octachron:
This brief PR fixes the printing of infix operators starting with `#`.

Before this patch, `;; x #$ y` was reformatted as ` #$ x y` due to the fact that `#` had not been added to the list of starting character for infix operators.


jberdine:
Thanks! (And sorry for the commit churn, I resolved and merged, but github seems to need the web ui to be used to recognize that things have been done.)

PR #27, opened 2017-10-22
Improve opam installation

jpdeplaix:
This PR improve the installation process by replacing the ```install``` and ```remove``` fields from the opam file by an .install file which describes to opam where to install the files and where they can be found in case of uninstallation.

It also fixes this commit https://github.com/ocaml-ppx/ocamlformat/commit/b074dd9331e3580d1218951b64abd8c7f094713c which puts the emacs script under ```%{share}%``` and not under ```%{share_root}%```: the first one is located in ```~/.opam/<switch>/share/ocamlformat``` but the second one is located in ```~/.opam/<switch>/share```. It allows the script to be in the default emacs script directory (also used by almost every emacs script available on opam).

This new file can also be used in https://github.com/ocaml/opam-repository/pull/10502 (just move ocamlformat.install to a files directory in your PR, remove the ```install``` and ```remove``` fields and it should do the trick).


jberdine:
Thanks! Regarding `share` vs `share_root`, I think that there is some discrepancy between the install file sections and the opam variables. In particular, the result of `opam config var share` for me is `~/.opam/<switch>/share`, and that is what `%{share}%` seems to expand to. I don't know if this is sorted out in opam 2.

jpdeplaix:
> and that is what %{share}% seems to expand to

Are you sure ? it seems rather weird. It should expend to ```~/.opam/<switch>/share/ocamlformat```.
It's normal for ```opam config var share``` to give ```~/.opam/<switch>/share```. I suppose what you wanted to do is ```opam config var ocamlformat:share``` which gives ```~/.opam/<switch>/share/ocamlformat``` (if the package exists).

Note: I use opam 1.2.

jberdine:
Sure? No, not entirely. But after `opam pin add ocamlformat .` from the root of my ocamlformat clone, I see:
```
$ cd; find .opam -name 'ocamlformat.el'
.opam/4.05.0-ocamlformat/build/ocamlformat.0.1/emacs/ocamlformat.el
.opam/4.05.0-ocamlformat/packages.dev/ocamlformat/emacs/ocamlformat.el
.opam/4.05.0-ocamlformat/share/emacs/site-lisp/ocamlformat.el
```
The only explanation I can see is that `%{share}%` in opam files corresponds to `share_root:` in install files.
I'm using opam 1.2.2.

dbuenzli:
> The only explanation I can see is that %{share}% in opam files corresponds to share_root: in install files.

AFAIR this is correct, `%{ocamlformat:share}%` will expand to what @jpdeplaix thinks (see `opam config list` for the list of variables and meaning). In `.install` files what is in a `share:` field gets installed in `%{ocamlformat:share}%` and what is in `share_root` in `%{share}%`.

jpdeplaix:
Ok fair enough. Sorry for the misconception.

dbuenzli:
Well to be fair the naming in the `.install` file and `install:` field feel a bit inconsitent (but I never realized it).

rgrinberg:
How come there's a need for manual install instructions or a manually written .install file in the first place? This project uses jbuilder which generates an appropriate .install file.

@jberdine any reason why that doesn't work?

jberdine:
Thanks for the suggestion @rgrinberg , but I had tried that before without luck. In particular, if I invoke `jbuilder install ocamlformat`, then jbuilder does not perform a build according to the `build:` field of the `ocamlformat.opam` file, but instead seems to try to perform the actions of `jbuilder build`. Since the build field of the opam file performs some steps before invoking jbuilder, this build fails.

mseri:
Some of the `Makefile`'s step (e.g. the generation of `version.ml` or the preparation of the patches could be moved in the `jbuild` file). @rgrinberg is there any way to modify the `jbuild` workspace file as done in the `Makefile` via `jbuilder`?

jberdine:
I would love to completely remove the Makefile, but have not been able to so far.

The generation of `Version.ml` is done in the Makefile since I don't know how to convince jbuilder to update it every time the output of `git describe --tags --dirty --always` changes (and something else does too).

For the workspace generation hack, it may be more productive to consider what it is achieving to see if there is a better/easier way. There are two issues.
- First, because jbuilder requires some switch to be specified in the jbuild-workspace file, some mechanism of looking up the current switch is needed. If the switch was optional, this sed hacking could be removed.
- Second, multiple jbuilder 'contexts' are used to implement multiple build modes, which have different compilation flags. This is analogous to a non-hardcoded version of `jbuilder build` versus `jbuilder build --dev`. I would be very happy if there was a configurable way to do this without resorting to generating jbuilder-workspace files and using OCaml instead of sexp jbuild files.

jberdine:
In the time until jbuilder can be used to control the installation, I would like to merge this PR. But I can't before the CLA is signed.

jberdine:
Thanks!

PR #37, opened 2017-10-27
Rename 'opt' jbuilder context to 'default' to ease building without opam

jberdine:


jberdine:
Subsumed by b887f7b7951ee001ba63429ce34994b8acc8fba0

PR #53, opened 2017-12-04
Fix missing "as" in Ptyp_alias formatting

### LINKED ISSUE #51: Failure to parse "as" in value's mli entry
ocamlformat 0.2 from opam is unable to process this syntax:
```ocaml
val run :
  unit -> (unit -> ('a, [> `Msg of string] as 'b) result) ->
  ('a, 'b) result
```
The error:
```
$ ocamlformat /tmp/x.mli -o /tmp/x2.mli
File "/tmp/x27d8987.mli", line 2, characters 44-45:
Error: Syntax error
```

The issue seems to be the `as 'b` - if I remove that then ocamlformat processes the code without issues.

hcarty:
Fixes #51


jberdine:
Thanks!

PR #58, opened 2017-12-15
base v0.10.0 compatibility

hcarty:
These changes allow ocamlformat to build against the recently released base v0.10.0.


jberdine:
Thanks!

PR #63, opened 2018-01-05
Formatting of %ext on if/while/for/match/try/;

hcarty:
This is not ready to merge yet because:

1. There is a large chunk of code commented out because it generates incorrectly formatted code with this change.  I haven't had a chance to find out why yet.
2. No formatting has been applied to the new code. **EDIT:** Formatting has been applied.
3. It should really have some tests.

I'm pushing this out for feedback on better ways to handle this.

This PR adds sugared formatting of: `if%ext`, `while%ext`, `for%ext`, `match%ext` and `try%ext`.  This covers the syntax covered by Lwt's ppx, for example.

Some example formatting with this PR:
```ocaml
let () =
  if%ext true then () else () ;
  if%ext true then () else if true then () else () ;
  let%ext x = () in
  for%ext i = 1 to 10 do () done ;
  while%ext false do () done ;
  match%ext x with _ -> ()


let () =
  let%ext x = () in
  try%ext x with _ -> ()


let () =
  if%ext true then () else () ;
  if%ext true then () else if true then () else () ;
  if%ext true then () else ()


let () =
  (match%ext x with _ -> ()) ;
  match%ext x with _ -> ()


let () =
  () ;
  () ;%ext
  () ;
  () ;%ext ()
```

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -410,6 +410,8 @@ end = struct
         | Pexp_construct
             ({txt= Lident "::"}, Some {pexp_desc= Pexp_tuple [e1; e2]}) ->
             assert (e1 == exp || e2 == exp)
+        | Pexp_extension (_, PStr [{pstr_desc= Pstr_eval (e, _)}]) ->
+            assert (e == exp)
         | Pexp_constant _ | Pexp_extension _ | Pexp_ident _ | Pexp_new _
          |Pexp_object _ | Pexp_pack _ | Pexp_unreachable ->
             assert false
@@ -856,6 +858,7 @@ end = struct
          |Pexp_construct
             ({txt= Lident "::"}, Some {pexp_desc= Pexp_tuple [_; e]})
          |Pexp_construct (_, Some e)
+         |Pexp_extension (_, PStr [{pstr_desc= Pstr_eval (e, _)}])
          |Pexp_fun (_, _, _, e)
          |Pexp_ifthenelse (_, e, None)
          |Pexp_ifthenelse (_, _, Some e)
--- a/src/Fmt.mli
+++ b/src/Fmt.mli
@@ -58,6 +58,10 @@ val list_pn : 'a list -> (?prev:'a -> 'a -> ?next:'a -> t) -> t
 (** Format a list using provided function for the elements, which is passed
     the previous and next elements, if any. *)
 
+val list_k : 'a list -> t -> ('a -> t) -> t
+(** Format a list using the first function for the separators between
+    elements. *)
+
 (** Conditional formatting ----------------------------------------------*)
 
 val fmt_if : bool -> s -> t
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -359,6 +359,10 @@ let fmt_docstring ?pro ?epi doc =
              $ Option.call ~f:epi ) ) )
 
 
+let fmt_extension_suffix ext =
+  opt ext (fun {txt; loc} -> str "%" $ Cmts.fmt loc (str txt))
+
+
 let rec fmt_attribute c pre = function
   | ( {txt= ("ocaml.doc" | "ocaml.text") as txt}
     , PStr
@@ -1102,8 +1106,11 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
                          | Some xcnd ->
                              hvbox 0
                                ( hvbox 2
-                                   ( fmt_if (not first) "else " $ fmt "if "
-                                   $ fmt_expression c xcnd )
+                                   ( fmt_if (not first) "else "
+                                   $ ( if first then
+                                         fmt "if" $ fmt_extension_suffix ext
+                                     else fmt "if" )
+                                   $ str " " $ fmt_expression c xcnd )
                                $ fmt "@ then" )
                          | None -> fmt "else" )
                        $ fmt_if parens " (" $ fmt "@ "
@@ -1194,7 +1201,7 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
           hvbox 0
             (wrap_fits_breaks_if parens "(" ")"
                ( hvbox 0
-                   ( str keyword
+                   ( str keyword $ fmt_extension_suffix ext
                    $ fmt_attributes c (fmt "") ~key:"@" pexp_attributes
                        (fmt "")
                    $ fmt "@;<1 2>" $ fmt_expression c (sub_exp ~ctx e0)
@@ -1204,7 +1211,7 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
           wrap_fits_breaks_if parens "(" ")"
             (hovbox 2
                ( hvbox 0
-                   ( str keyword
+                   ( str keyword $ fmt_extension_suffix ext
                    $ fmt_attributes c (fmt "") ~key:"@" pexp_attributes
                        (fmt "")
                    $ hvbox 0
@@ -1263,10 +1270,15 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
                  $ hvbox (-2) (list flds "@,; " fmt_field) ) ))
         $ fmt_atrs )
   | Pexp_sequence _ ->
+      let fmt_sep =
+        match ext with
+        | None -> fmt " ;@ "
+        | Some {txt; _} -> fmt " ;%%" $ str txt $ fmt "@ "
+      in
       hvbox 0
         ( wrap_if parens "(" ")"
             (list (sugar_sequence c width xexp) " ;@;<1000 0>" (fun grp ->
-                 list grp " ;@ " (fmt_expression c) ))
+                 list_k grp fmt_sep (fmt_expression c) ))
         $ fmt_atrs )
   | Pexp_setfield (e1, {txt}, e2) ->
       hvbox 0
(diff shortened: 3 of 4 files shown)


hcarty:
Some good and bad output examples when using this patch.

The following formats successfully:
```ocaml
let () =
  if%ext true then () else () ;
  if%ext true then () else if true then () else () ;
  let%ext x = () in
  for%ext i = 1 to 10 do () done ;
  while%ext false do () done ;
  match%ext x with _ -> ()


let () =
  let%ext x = () in
  try%ext x with _ -> ()


let () =
  if%ext true then () else () ;
  if%ext true then () else if true then () else () ;
  if%ext true then () else ()
```

The following produces an error (`Internal Error: formatting changed ast`)
Input:
```ocaml
let () =
  (match%ext x with
   | _ -> ());
  match%ext x with
  | _ -> ()
```
Output:
```ocaml
let () =
  match%ext x with _ -> () ;
  match%ext x with _ -> ()
```

jberdine:
I have not had a chance to look at this in detail yet, but the last example where formatting fails indicates that there is probably a problem with the relative precedences of extension and sequence expressions.  I suspect that the `Pexp_extension` case of `exposed` in `Ast` needs to be changed to consider the expression which is wrapped in an extension.

hcarty:
@jberdine Thanks!  I pushed a commit which fixes that issue at least.

hcarty:
As of 03cc831 the code is formatted and seems to work, at least for the limited local checks I've made.  There is still that big commented-out block in `fmt_expression` that I'm haven't had a chance to look at more closely.  If you get time to do so and/or have suggestions for what may be going wrong I'd appreciate any insight you can provide.

hcarty:
cbafe02 supports formatting `;%ext`

src/Ast.ml
```diff
@@ -852,6 +852,7 @@ end = struct
          |Pexp_construct
             ({txt= Lident "::"}, Some {pexp_desc= Pexp_tuple [_; e]})
          |Pexp_construct (_, Some e)
+         |Pexp_extension (_, PStr [{pstr_desc= Pstr_eval (e, _)}])
```
jberdine: I think that this is the source of the failed formatting you observe. This clause causes `continue` to be called on `e`, which leads to `sub_exp ~ctx:(Exp exp) e`. Calling `sub_exp` is effectively a (runtime) assertion that `e` is an immediate sub-expression of `exp`. The implementation of `sub_exp` checks this (in debug mode). The parenthesization code relies heavily on the property that 'subterms' are physically equal to an 'immediate' component value. Exactly what qualifies as 'immediate' is determined by the implementation of `check_exp` above. So in order to add this clause, an analogous one needs to be added to `check_exp`. I think that adding
```
| Pexp_extension (_, PStr [{pstr_desc= Pstr_eval (e, _)}]) -> assert (e == exp)
```
before line 413 should do it. Note that this is uncovering a collection of existing bugs, since `check_exp` currently assumes that there are no expression sub-terms of `Pexp_extension` nodes.

src/Fmt_ast.ml
```diff
@@ -1252,10 +1261,15 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
                  $ hvbox (-2) (list flds "@,; " fmt_field) ) ))
         $ fmt_atrs )
   | Pexp_sequence _ ->
+      let pp_sep =
+        match ext with
+        | None -> fun fs () -> Format.fprintf fs " ;@ "
+        | Some {txt; _} -> fun fs () -> Format.fprintf fs " ;%%%s@ " txt
```
jberdine: A minor point, but it would be more consistent with the existing code to change this to e.g.:
```
| None -> fmt " ;@ "
| Some {txt; _} -> fmt " ;%%" $ str txt $ fmt "@ "
```
and probably name it `fmt_sep` instead of `pp_sep`.

src/Fmt.mli
```diff
@@ -58,6 +58,11 @@ val list_pn : 'a list -> (?prev:'a -> 'a -> ?next:'a -> t) -> t
 (** Format a list using provided function for the elements, which is passed
     the previous and next elements, if any. *)
 
+val list_sep :
+  'a list -> (Format.formatter -> unit -> unit)
+  -> ('a -> Format.formatter -> unit) -> Format.formatter -> unit
+(** Format a list using the first function for the separators between elements *)
```
jberdine: Nit: the other docstrings end with a period, and then it will probably need to be re-wrapped.

src/Fmt_ast.ml
```diff
@@ -357,6 +357,10 @@ let fmt_docstring ?pro ?epi doc =
              $ Option.call ~f:epi ) ) )
 
 
+let fmt_keyword_with_extension keyword ext =
+  str keyword $ opt ext (fun {txt; loc} -> str "%" $ Cmts.fmt loc (str txt))
```
jberdine: I'm not sure if it is simpler to include `str keyword $` in this function, or if it might be better to shorten this to a function that just formats `ext`. Maybe consider both and see what you think.

src/Fmt_ast.ml
```diff
@@ -1099,8 +1105,11 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
                          | Some xcnd ->
                              hvbox 0
                                ( hvbox 2
-                                   ( fmt_if (not first) "else " $ fmt "if "
-                                   $ fmt_expression c xcnd )
+                                   ( fmt_if (not first) "else "
+                                   $ ( if first then
+                                         fmt_keyword_with_extension "if" ext
+                                     else fmt "if" )
```
jberdine: I don't really understand the branch on `first` here. Is it desired to have
```
if%lwt e0 then e1 else if e2 then e3 else e4
```
instead of
```
if%lwt e0 then e1 else if%lwt e2 then e3 else e4
```
?
I guess my real concern is that `sugar_ite` may need to be extended to collect up the extensions.

hcarty:
Thanks for the comments and suggestions - I'll try to take care of them later this week.

src/Fmt_ast.ml
hcarty: That is the intention - only having `%lwt` on the first `if`.  Now that I write that, though, I honestly don't know what the proper syntax would be.  I can check.

jberdine: I suspect that each ifthenelse expression can have an extension or not independently of the sub-expressions in the arms. In that case, `sugar_ite` will need to detect `Pexp_extension` super-expressions of `Pexp_ifthenelse` expressions, and return an additional optional extension with each `cnd` expression.

hcarty:
Leaving a note here to myself as a found a bug while using this branch which may or may not be related to these changes.

The parens around `while%lwt ... done` are dropped in the following code leading to a syntax error.
```ocaml
let () =
  Lwt_main.run (
    while%lwt true do
      Lwt.return_unit
    done
  )
```

src/Ast.ml
hcarty: I tried this and it didn't seem to fix the issue.  The change is in the latest commit though (
5bf3e38).

hcarty:
I think I addressed the main style feedback.  The `while`-wrapped-in-parens [issue](https://github.com/ocaml-ppx/ocamlformat/pull/63#issuecomment-358202328) is still there and that big block of code is still commented out.

hcarty:
The `while` issue exists without extension points attached so I think that's something entirely separate.  I'll file an issue.

jberdine:
I think the issue with `while` is fixed now, rebase?

hcarty:
I'll make the `Fmt.mli` change and rebase now - thanks for your patience on this PR!

jberdine:
Thanks, don't worry, my patience is isn't being tried!  I'm the one who took a week to fix a blocking bug after all.

hcarty:
Squashed and rebased with the `list_k` rename included.

hcarty:
Testing to make sure the commented block can be removed...

hcarty:
Ok - as far as I can tell formatting works without those two cases commented out in `Fmt_ast.ml`!  Potentially final rebase and forced push done.

src/Ast.ml
hcarty: I think this is fixed now thank you!

jberdine:
This is looking good, thanks! The only question I have is related to the nested `if%ext` [comments](https://github.com/ocaml-ppx/ocamlformat/pull/63#discussion_r160296261). Do you think we should merge this as-is, or do something more to handle nested `if%ext` expressions? I'm happy either way, but it would be good to open an issue to track if we merge now.

hcarty:
I did a bit of testing.  As of now, nested `if%ext` are left alone.  For example, the formatting the following code with this branch will result in identical output:
```ocaml
let () =
  if%ext true then () else if%ext true then () else if true then () else ()
```
Both existing `%ext`s are kept and the inner-most `if` does not have an `%ext` applied.  If I remove the check in the formatting code then the outer-most `%ext` propagates through all of the inner `if`s.

That `if%ext` handling code was definitely a bit of a "I'll try this and see if it works" on my part - is there a better way to achieve the same result?

jberdine:
Ok, I see, the nested `if%ext` expressions are currently correctly formatted, by splitting them into separate "nests" as far as `sugar_ite` is concerned. This means that formatting with vs without extensions changes a bit:
```
let () =
  if%ext true then "aaaaaaaaaaaaaaaaaaaa"
  else
    if%ext true then "bbbbbbbbbbbbbbbbbbbb"
    else if true then "cccccccccccccccccccc"
    else "dddddddddddddddddddd"


let () =
  if%ext true then "aaaaaaaaaaaaaaaaaaaa"
  else if true then "bbbbbbbbbbbbbbbbbbbb"
  else if true then "cccccccccccccccccccc"
  else "dddddddddddddddddddd"
```
In order to format these two cases uniformly, `sugar_ite` would need to be extended to detect and return the extensions. That can be saved for a separate PR I think.

jberdine:
Thanks again, nice one!

Issue #65, opened 2018-01-05
Skip formatting of appropriately annotated files


hcarty:
ocamlformat will have bugs from time to time.  It would be nice if it could skip formatting of files (or portions of a file) based on something like an annotation.  For example:
```ocaml
[@@@ocamlformat disable]

let this_won't_be_formatted =
  1
```

hcarty:
Skipping portions of a file would be nice to have but I suspect skipping the entire file would be much simpler to implement and would still be useful when formatting bugs affect a project.

PR #66, opened 2018-01-05
Disable formatting with [@@@ocamlformat.disable]

hcarty:
This approach is somewhat crude as the formatter simply fails with a
message when `[@@@ocamlformat.disable]` is found in a source file.

Relevant to #65

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2016,6 +2016,12 @@ and fmt_structure_item c ~sep ~last:last_item {ctx; ast= si} =
   wrap_k fmt_cmts_before fmt_cmts_after
   @@
   match si.pstr_desc with
+  | Pstr_attribute ({txt= "ocamlformat.disable"; loc}, _) ->
+      internal_error
+        (Format.asprintf
+           "Not formatting because ocamlformat is disabled on line %d"
+           loc.loc_start.pos_lnum)
+        []
   | Pstr_attribute atr ->
       let doc, atrs = doc_atrs [atr] in
       fmt_docstring ~epi:(fmt "") doc
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -19,6 +19,8 @@ open Parsetree
 open Ast
 open Fmt
 
+exception Formatting_disabled
+
 (* Debug: catch and report failures at nearest enclosing Ast.t *)
 
 let protect =
@@ -1617,6 +1619,8 @@ and fmt_signature_item c {ast= si} =
   @@
   let ctx = Sig si in
   match si.psig_desc with
+  | Psig_attribute ({txt= "ocamlformat.disable"; _}, _) ->
+      raise Formatting_disabled
   | Psig_attribute atr ->
       let doc, atrs = doc_atrs [atr] in
       fmt_docstring ~epi:(fmt "") doc
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2016,12 +2016,8 @@ and fmt_structure_item c ~sep ~last:last_item {ctx; ast= si} =
   wrap_k fmt_cmts_before fmt_cmts_after
   @@
   match si.pstr_desc with
-  | Pstr_attribute ({txt= "ocamlformat.disable"; loc}, _) ->
-      internal_error
-        (Format.asprintf
-           "Not formatting because ocamlformat is disabled on line %d"
-           loc.loc_start.pos_lnum)
-        []
+  | Pstr_attribute ({txt= "ocamlformat.disable"; _}, _) ->
+      raise Formatting_disabled
   | Pstr_attribute atr ->
       let doc, atrs = doc_atrs [atr] in
       fmt_docstring ~epi:(fmt "") doc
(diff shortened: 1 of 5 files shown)


jberdine:
This would indeed be a useful feature!

It would be better I think to define a new exception to raise which could be handled in e.g. `Translation_unit.parse_print` so that it could either skip or copy the input file as appropriate given the command line arguments.

I also wonder whether files which do not parse ought to be scanned for this attribute textually, so that e.g. files that use camlp4 or some other unsupported preprocessing could also be annotated and skipped.

hcarty:
I can work on the new exception part.  I'm not very familiar with this part of the code - should the new code path log something to `stderr` and end in `exit 1`, similar to the other exception handlers?  Or should it output the original source verbatim, acting like it completed successfully?

Scanning for the annotation textually would be nice, but maybe in a separate PR?  Making it robust could get a little tricky.

hcarty:
Commit dd0853e adds a custom `Formatting_disabled` exception as you suggested @jberdine.  When this exception is raised the original bytes are spit back out as they are.  As an example, the following code makes a successful round-trip through the formatter:
```ocaml 
[@@@ocamlformat.disable]

let fancy = "🔥"

let    f
x
                     =
 `Unfortunate_formatting
```

src/Translation_unit.ml
```diff
@@ -138,3 +138,5 @@ let parse_print (XUnit xunit) (conf: Conf.t) iname ifile ic ofile =
   | Syntaxerr.Error _ as exc ->
       Location.report_exception Caml.Format.err_formatter exc ;
       Caml.exit 1
+  | Formatting_disabled ->
+      Stdio.Out_channel.output_string Stdio.stdout source
```
jberdine: IIUC, this will work only when outputting to stdout, but not if an output file is specified, or if formatting in-place.  A match analogous to the one on lines 123-9 above is probably needed, so that nothing is done in the `Inplace` case, and the file is copied in the `In_out` case.

And minor style nit: I would put this new match case before the others as it is an expected successful termination case, while the others are fatal errors.

src/Fmt_ast.ml
```diff
@@ -2016,6 +2016,8 @@ and fmt_structure_item c ~sep ~last:last_item {ctx; ast= si} =
   wrap_k fmt_cmts_before fmt_cmts_after
   @@
   match si.pstr_desc with
+  | Pstr_attribute ({txt= "ocamlformat.disable"; _}, _) ->
+      raise Formatting_disabled
```
jberdine: An analogous case for signatures is also needed I think.

src/import/Import.mli
```diff
@@ -34,6 +34,10 @@ val ( $ ) : ('a -> unit) -> ('a -> 'b) -> 'a -> 'b
 (** Sequential composition of functions: [(f $ g) x] is exactly equivalent
     to [(f x) ; (g x)]. Left associative. *)
 
+exception Formatting_disabled
+    (** Raise this exception to Indicate that formatting has been disabled for the
+        current file *)
```
jberdine: Not for this PR, but looks like there is a separate existing issue with the indentation of docstrings on exception declarations.

src/import/Import.ml
```diff
@@ -21,6 +21,8 @@ include (
 include Option.Monad_infix
 include Stdio
 
+exception Formatting_disabled
+
```
jberdine: Could this exception be moved to `Fmt_ast` instead since it is really part of the interface of its `fmt_structure` (and `fmt_signature`) functions.

jberdine:
I think the terminating successfully as you have done is the right choice, otherwise all automated or integrated workflows will still be broken.

Yes, scanning textually should be a separate PR.  Unless it seems prevalent enough that it should be done _instead_ of an attribute.  I am not sure.

hcarty:
Thanks for the comments and suggestions - I'll try to take care of them later this week.

src/Fmt_ast.ml
hcarty: 👍

src/Translation_unit.ml
hcarty: 👍

src/import/Import.mli
hcarty: I noticed that too - I can add an issue for it if you'd like

hcarty:
Commit 94975b1 should address the code comments you made - please let me know if you see any other issues!

As far as textual scanning vs annotation - I like the annotation version because it lives in the language.  Other AST-based tools can use it to if/when appropriate.  It may be worth changing the annotation name to something more generic so that the Reason folks could use it as well for files which should not be `refmt`'d.

src/import/Import.mli
jberdine: Fixed in 108a9d3388b21bd9668b80ae9bc9fcae35d513ce

jberdine:
Thanks!

The argument for an annotation makes sense.

For now I wouldn't worry about converting Reason code with `[@@@ocamlformat.disabled]`, as that is a bit of an inconsistent request.

hcarty:
Thank you for cleaning it up and merging!

Issue #11, opened 2017-10-20
Automatically wrap long strings


edwintorok:
Taking an example from https://ocaml.org/learn/tutorials/guidelines.html

```ocaml
let universal_declaration =
  "-1- Programs are born and remain free and equal under the law;\n\
   distinctions can only be based on the common good."
```

gets converted to this and exceeds line length:
```ocaml
let universal_declaration =
  "-1- Programs are born and remain free and equal under the law;\ndistinctions can only be based on the common good."
```

An alternative is to use `{| |}` where newlines are preserved:
```ocaml
 {| -1- Programs are born and remain free and equal under the law;
distinctions can only be based on the common good.|}
```

jberdine:
The current situation is indeed suboptimal. Continuation escapes are eagerly interpreted by the parser, so that by the time the formatter gets the parsetree, they have already been expanded.

It should be possible to implement a re-wrapping procedure to automatically insert line breaks with continuation escapes into long strings in order to stay within the margin.

jberdine:
AFAIU this should be supported now by 56d48075752c8370b533a06550f1cbdfbe949deb. Please reopen with more info if not.

Issue #10, opened 2017-10-20
Preserve original escape sequences in strings


edwintorok:
`"\xff\xaa\x55"`  gets converted to `"\255\170U"`, and `"A\xffB"` to `"A\255B"`
There should be a configuration field to specify whether this conversion is desired, and I'd prefer if `ocamlformat` wouldn't change the choice of escape sequence when all bytes in the string use the same escape sequence.

OTOH if someone writes "\xff\170U" .... I wouldn't mind if ocamlformat chose one.

jberdine:
The current situation is indeed suboptimal. The problem is that the parser eagerly interprets the escape sequences, so that by the time the formatter gets the parsetree, they have already been expanded. IIUC, the situation will improve for comments in 4.06.

edwintorok:
Could the output be reformatted to hex if a project's configuration specifies it prefers hex?

jberdine:
Potentially, I think so. Currently String.escaped is called to ensure the generated strings parse back to the same string. Some other function satisfying that same spec could be used instead. See Fmt_ast.fmt_constant.

vjeux:
For prettier (a JavaScript formatter), I've played with many variants around string escaping and the solution that ended up working was to read the original source (I hope that you have the node location) and just convert ' to " and fix the escaping and remove useless `\`. All the attempts at finding a consistent way to print those unicode characters / emoji ended up people creating issues about them.

jberdine:
I've heard rumblings that keeping the original strings in the parsetree is coming soon, so up to now I've been holding out on going back to the original source. Mostly a matter of prioritization.

mbarbin:
@vjeux's solution seems great.  I did not notice this issue, opened another one and seems like the discussion there reached the same conclusion.  Closing the other issue. #147

jberdine:
A hybrid solution would be to add a string escape mode for reusing the bytes of the input, and applying it selectively using attributes as per #81.

hcarty:
Thank you everyone involved!

PR #70, opened 2018-01-25
Add options to format chars and break strings

smondet:
Cf. issues #10, #11.

```
--break-string-literals=VAL (absent=newlines or
OCAMLFORMAT_BREAK_STRING_LITERALS env)
    Break string literals. Can be set in a config file with a
    `break-string-literals {never,newlines}` line.

--escape-chars=VAL (absent=hexadecimal or OCAMLFORMAT_ESCAPE_CHARS
env)
    How to escape chars. Can be set in a config file with a
    `escape-chars {hexadecimal,octal,minimal}` line.
```

I used `Cmdliner.Term.enum`s to leave space for future improvements, like
detecting UTF-8? or breaking long lines after spaces (or `/` for URIs) ….
Or when OCaml gets the string literals *literally* in the parse tree we can add
a `keep` option.

**⁂**

To play with it one can use/modify this script:

```bash
cat > /tmp/tt.ml <<'EOF'
module Test = struct
let x = "hello\nworldœd\
d→\n\
\t« »\
\ deij\n jl\\dsle ldjsellkj dse\n\x42 dsleij djlekjdsliejj ASEDE1092"

let _ = ('a', '\\', '\n', '\x00')
end
EOF
cat /tmp/tt.ml
testopt () {
    echo "=============== Test $1"
    _build/default/src/ocamlformat.exe --debug \
                                       --escape-chars $1 \
                                       --break-string $2 /tmp/tt.ml > /tmp/tt_$1.ml
    if [ $? -ne 0 ] ; then
        echo "ERROR: "
        cat /tmp/tt.1.ml
    fi
    cat /tmp/tt_$1.ml
    ocaml /tmp/tt_$1.ml
}
testopt min  nev
testopt hex  nev
testopt oct  nev
testopt min  new
testopt hex  new
testopt oct  new
```

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,1 +1,2 @@
 margin 77
+break-string-literals never
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -60,10 +60,35 @@ let rec is_sugared_list exp =
   | _ -> false
 
 
-let rec is_trivial exp =
+let looks_like_format str =
+  String.fold str ~init:(false, true) ~f:(fun (prev_is_at, acc) -> function
+    | '@' when prev_is_at -> (false, acc)
+    | '@' -> (true, acc)
+    | '\n' when prev_is_at -> (false, true && acc)
+    | '\n' -> (false, false)
+    | _ -> (false, acc) )
+  |> snd
+
+
+let would_force_break (c: Conf.t) s =
+  match c.break_string_literals with
+  | `Newlines ->
+      let n = String.length s in
+      let looks_like_format = lazy (looks_like_format s) in
+      let rec would_force_break_ i =
+        i < n
+        && ( Char.equal s.[i] '\n' && not (Lazy.force looks_like_format)
+           || would_force_break_ (i + 1) )
+      in
+      would_force_break_ 0
+  | _ -> false
+
+
+let rec is_trivial c exp =
   match exp.pexp_desc with
+  | Pexp_constant Pconst_string (s, None) -> not (would_force_break c s)
   | Pexp_constant _ | Pexp_field _ | Pexp_ident _ -> true
-  | Pexp_construct (_, exp) -> Option.for_all exp ~f:is_trivial
+  | Pexp_construct (_, exp) -> Option.for_all exp ~f:(is_trivial c)
   | _ -> false
 
 
@@ -480,8 +505,9 @@ end = struct
   let rec is_simple (c: Conf.t) width ({ast= exp} as xexp) =
     let ctx = Exp exp in
     match exp.pexp_desc with
-    | Pexp_array _ | Pexp_constant _ | Pexp_field _ | Pexp_ident _
-     |Pexp_record _ | Pexp_tuple _ | Pexp_variant _
+    | Pexp_constant _ -> is_trivial c exp
+    | Pexp_array _ | Pexp_field _ | Pexp_ident _ | Pexp_record _
+     |Pexp_tuple _ | Pexp_variant _
      |Pexp_construct (_, None) ->
         true
     | Pexp_construct
@@ -491,7 +517,7 @@ end = struct
     | Pexp_construct (_, Some e0) -> is_simple c width (sub_exp ~ctx e0)
     | Pexp_apply ({pexp_desc= Pexp_ident {txt= Lident ":="}}, _) -> false
     | Pexp_apply (e0, e1N) ->
-        is_trivial e0 && List.for_all e1N ~f:(snd >> is_trivial)
+        is_trivial c e0 && List.for_all e1N ~f:(snd >> is_trivial c)
         && width xexp * 3 < c.margin
     | _ -> false
 
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -60,28 +60,15 @@ let rec is_sugared_list exp =
   | _ -> false
 
 
-let looks_like_format str =
-  String.fold str ~init:(false, true) ~f:(fun (prev_is_at, acc) -> function
-    | '@' when prev_is_at -> (false, acc)
-    | '@' -> (true, acc)
-    | '\n' when prev_is_at -> (false, true && acc)
-    | '\n' -> (false, false)
-    | _ -> (false, acc) )
-  |> snd
-
-
 let would_force_break (c: Conf.t) s =
-  match c.break_string_literals with
-  | `Newlines ->
-      let n = String.length s in
-      let looks_like_format = lazy (looks_like_format s) in
-      let rec would_force_break_ i =
-        i < n
-        && ( Char.equal s.[i] '\n' && not (Lazy.force looks_like_format)
-           || would_force_break_ (i + 1) )
-      in
-      would_force_break_ 0
-  | _ -> false
+  let contains_internal_newline s =
+    match String.rindex s '\n' with
+    | None -> false
+    | Some i when i = String.length s - 1 -> false
+    | _ -> true
+  in
+  Poly.equal c.break_string_literals `Newlines
+  && contains_internal_newline s
 
 
 let rec is_trivial c exp =
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -169,7 +169,7 @@ let escape_chars =
              [ ("hexadecimal", `Hexadecimal)
              ; ("decimal", `Decimal)
              ; ("minimal", `Minimal) ])
-          `Minimal
+          default
       & info ["escape-chars"] ~doc ~env)
 
 
@@ -182,7 +182,7 @@ let break_string_literals =
   mk ~default
     Arg.(
       value
-      & opt (enum [("newlines", `Newlines); ("never", `Never)]) `Newlines
+      & opt (enum [("newlines", `Newlines); ("never", `Never)]) default
       & info ["break-string-literals"] ~doc ~env)
 
 
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -157,7 +157,7 @@ let sparse =
 
 let escape_chars =
   let doc =
-    "How to escape chars. Can be set in a config file with a `escape-chars {hexadecimal,octal,minimal}` line."
+    "How to escape chars. Can be set in a config file with a `escape-chars {hexadecimal,decimal,minimal}` line."
   in
   let env = Arg.env_var "OCAMLFORMAT_ESCAPE_CHARS" in
   let default = `Minimal in
@@ -167,7 +167,7 @@ let escape_chars =
       & opt
           (enum
              [ ("hexadecimal", `Hexadecimal)
-             ; ("octal", `Octal)
+             ; ("decimal", `Decimal)
              ; ("minimal", `Minimal) ])
           `Minimal
       & info ["escape-chars"] ~doc ~env)
@@ -220,7 +220,7 @@ type t =
   { margin: int
   ; sparse: bool
   ; max_iters: int
-  ; escape_chars: [`Hexadecimal | `Minimal | `Octal]
+  ; escape_chars: [`Hexadecimal | `Minimal | `Decimal]
   ; break_string_literals: [`Newlines | `Never] }
 
 let update conf name value =
@@ -233,7 +233,7 @@ let update conf name value =
         escape_chars=
           ( match value with
           | "hexadecimal" -> `Hexadecimal
-          | "octal" -> `Octal
+          | "decimal" -> `Decimal
           | "minimal" -> `Minimal
           | other ->
               user_error
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -155,6 +155,37 @@ let sparse =
   mk ~default Arg.(value & flag & info ["sparse"] ~doc ~env)
 
 
+let escape_chars =
+  let doc =
+    "How to escape chars. Can be set in a config file with a `escape-chars {hexadecimal,octal,minimal}` line."
+  in
+  let env = Arg.env_var "OCAMLFORMAT_ESCAPE_CHARS" in
+  let default = `Hexadecimal in
+  mk ~default
+    Arg.(
+      value
+      & opt
+          (enum
+             [ ("hexadecimal", `Hexadecimal)
+             ; ("octal", `Octal)
+             ; ("minimal", `Minimal) ])
+          `Hexadecimal
+      & info ["escape-chars"] ~doc ~env)
+
+
+let break_string_literals =
+  let doc =
+    "Break string literals. Can be set in a config file with a `break-string-literals {never,newlines}` line."
+  in
+  let env = Arg.env_var "OCAMLFORMAT_BREAK_STRING_LITERALS" in
+  let default = `New_lines in
+  mk ~default
+    Arg.(
+      value
+      & opt (enum [("newlines", `New_lines); ("never", `Never)]) `New_lines
+      & info ["break-string-literals"] ~doc ~env)
+
+
 let no_version_check =
   let doc =
     "Do no check version matches the one specified in .ocamlformat."
@@ -185,13 +216,37 @@ let validate () =
 
 ;; parse info validate
 
-type t = {margin: int; sparse: bool; max_iters: int}
+type t =
+  { margin: int
+  ; sparse: bool
+  ; max_iters: int
+  ; escape_chars: [`Hexadecimal | `Minimal | `Octal]
+  ; break_string_literals: [`New_lines | `Never] }
 
 let update conf name value =
   match name with
   | "margin" -> {conf with margin= Int.of_string value}
   | "max-iters" -> {conf with max_iters= Int.of_string value}
   | "sparse" -> {conf with sparse= Bool.of_string value}
+  | "escape-chars" ->
+      { conf with
+        escape_chars=
+          ( match value with
+          | "hexadecimal" -> `Hexadecimal
+          | "octal" -> `Octal
+          | "minimal" -> `Minimal
+          | other ->
+              Printf.ksprintf failwith "Unknown escape-chars value: %S"
+                other ) }
+  | "break-string-literals" ->
+      { conf with
+        break_string_literals=
+          ( match value with
+          | "never" -> `Never
+          | "newlines" -> `New_lines
+          | other ->
+              Printf.ksprintf failwith
+                "Unknown break-string-literals value: %S" other ) }
   | "version" when not !no_version_check ->
       if String.equal Version.version value then conf
       else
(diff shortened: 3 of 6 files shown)


src/Fmt_ast.ml
```diff
@@ -303,18 +303,41 @@ let rec fmt_longident (li: Longident.t) =
       cbox 0 (fmt_longident li1 $ wrap "(" ")" (fmt_longident li2))
 
 
-let fmt_constant const =
+let fmt_constant (conf: Conf.t) const =
+  let ocaml_whitespace = function
+    | '\n' | '\t' | '\r' | ' ' -> true
+    | _ -> false
+  in
```
jberdine: `ocaml_whitespace` seems to be the same as `Char.is_whitespace`, replace?

src/Fmt_ast.ml
```diff
@@ -303,18 +303,41 @@ let rec fmt_longident (li: Longident.t) =
...
+  let escape_char c =
+    let code = Char.to_int c in
+    match conf.Conf.escape_chars with
```
jberdine: Could you change this to use a type ascription `(c: Conf.t)` on the argument above to be uniform with the rest of the code.

src/Fmt_ast.ml
```diff
@@ -303,18 +303,41 @@ let rec fmt_longident (li: Longident.t) =
...
+    | _ when ocaml_whitespace c || Char.equal c '\\' -> str (Char.escaped c)
+    | _ when 0x20 <= code && code <= 0x7E -> char c
```
jberdine: For readability, how about defining a named function for this test. Something like `is_printing c` or `not (is_control_code c)`.

src/Fmt_ast.ml
```diff
@@ -303,18 +303,41 @@ let rec fmt_longident (li: Longident.t) =
...
       str lit $ opt suf char
-  | Pconst_char c -> wrap "'" "'" @@ str (Char.escaped c)
+  | Pconst_char c -> wrap "'" "'" @@ escape_char c
   | Pconst_string (s, delim) ->
+      let escape_literal string =
+        String.fold string ~init:(false, fmt "@[") ~f:
+          (fun (freshline, prev) ch ->
+            match (ch, conf.Conf.break_string_literals) with
+            | ' ', _ when freshline -> (false, prev $ str "\\ ")
+            | '\n', `New_lines -> (true, prev $ fmt "\\n\\@\n")
```
jberdine: I think that this case will apply for e.g. format strings such at the one on this line, which is probably not desirable since in a format string `@\n` is semantically an atomic unit.

src/Fmt_ast.ml
jberdine: And when a char is part of a string, then the `"` char needs to be escaped.

src/Conf.ml
```diff
@@ -155,6 +155,37 @@ let sparse =
   mk ~default Arg.(value & flag & info ["sparse"] ~doc ~env)
 
 
+let escape_chars =
+  let doc =
+    "How to escape chars. Can be set in a config file with a `escape-chars {hexadecimal,octal,minimal}` line."
+  in
+  let env = Arg.env_var "OCAMLFORMAT_ESCAPE_CHARS" in
+  let default = `Hexadecimal in
```
jberdine: I would guess that minimal would be the best default, thoughts?

smondet: I prefer `` `Minimal` too :+1:

src/Conf.mli
```diff
@@ -16,8 +16,11 @@ type t =
   ; sparse: bool  (** Generate more sparsely formatted code if true. *)
   ; max_iters: int
         (** Fail if output of formatting does not stabilize within
-            [max_iters] iterations. *)
-  }
+      [max_iters] iterations. *)
```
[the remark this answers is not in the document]
smondet: :+1:  Indeed I'm already relying too much on `ocamlformat` :)

src/Fmt_ast.ml
smondet: OK, not sure that is what you mean, but I'll try not outputing `@\n` when the `\n` is the last character

To get

```
  let x fmt = Format.fprintf fmt "@[Hello@\n\
                                  World@]@\n"
```

instead of

```
  let x fmt = Format.fprintf fmt "@[Hello@\n\
                                  World@]@\n\
                                  "
```

jberdine:
Actually, I was thinking that `@\n` should not be treated as a newline at all, so yielding
```
let x fmt = Format.fprintf fmt "@[Hello@\nWorld@]@\n"
```
(assuming a wide enough margin)

smondet:
@jberdine But we can't differentiate at the AST-level between `Format.format` strings and other string and `Printf.printf` literals right?

jberdine:
Right, format strings can't be distinguished before type inference. But I expect that statistically, there are way more occurrences of `@\n` in format strings than others. And the workaround if those are not broken seems not so hard, while breaking them can't easily be worked around AFAIK.

smondet:
I think doing statistics should be out of scope for this first `newlines` flavor (which is quite dumb after all) :).
seems to me like making it less predictable for a single very "niche" use-case would make it more annoying to use (?)

jberdine:
I don't really know, but I think that there are very few cases where `\n` in an ocaml string does not mean newline, maybe only format strings. For one concrete case, I don't think I can use the newlines flavor for any of my code in its current state, but I could (and would like to) if it treated `@\n` as not-a-newline.

smondet:
`\n` always *means* newline that's why we cannot differentiate them in
the AST; it's just that the implementation of `Format` module
interprets `@` followed by a newline slightly differently

I've pushed an attempt at a basic format string detection, let me know
if it's enough.

hcarty:
Would it be sufficient to limit the check to look back one character when a `\n` is encountered?  If the previous character is `@`, print the literal `\n`.  Otherwise emit a newline.  That avoids the complexity of trying to detect "is this a format string?" while not breaking `@\n` in format strings.

src/Fmt_ast.ml
```diff
@@ -303,18 +303,54 @@ let rec fmt_longident (li: Longident.t) =
...
+        let looks_like_format =
+          String.fold string ~init:(false, true) ~f:
+            (fun (prev_is_at, acc) -> function
+            | '@' when prev_is_at -> (false, acc)
+            | '@' -> (true, acc)
+            | '\n' when prev_is_at -> (false, true && acc)
+            | '\n' -> (false, false)
+            | _ -> (false, acc) )
+          |> snd
+        in
```
jberdine: How about
```
let looks_like_format index =
  index <> 0 && Char.equal string.[index - 1] '@'
```
or am I missing something?

jberdine: Cross-posted with @hcarty

smondet:
I chose to detect `@\n` as "all `\n` are preceded by `@`" because if not a random `@` before a `\n` in any literal would make the pretty-printing annoyingly unpredictable.
(≥ 12 years of OCaml as main language and this PR is the first time I use `@\n`; the documentation actually [says](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Format.html#VALfprintf) *"not the normal way of pretty-printing"* for that one!)

src/Fmt_ast.ml
```diff
@@ -303,18 +303,54 @@ let rec fmt_longident (li: Longident.t) =
...
+        String.foldi string ~init:(false, fmt "@[") ~f:
+          (fun index (freshline, prev) ch ->
+            match (ch, c.Conf.break_string_literals) with
```
jberdine: nit: can remove `Conf.`

src/Fmt_ast.ml
smondet: wouldn't that bring too many "false positives"?

hcarty:
This PR may also fix string literals with unicode characters included, at least in some cases.  For example:
```ocaml
let fancy = "This is some text 🔥 and newlines \n
\n\\n"

let formatish = "@@@.@\n"
```
formats to the following when using this PR and `--break-string-literals=newlines` (which I really hope remains the default!):
```ocaml
let fancy = "This is some text 🔥 and newlines \n\
             \n\
             \n\
             \\n"

let formatish = "@@@.@\n"
```

hcarty:
Unfortunately it also seems to eat escaped characters in string literals.

Input:
```ocaml
let s = "\x00\x01\x02\x03"
```
Output:
```ocaml
let s = ""
```

jberdine:
I think this last issue with escaped chars is only for non-printing control characters. E.g.:
```
let s = "\x00\x01\x02\x03\x19\x20\xfe\x7f\x80\xe2\x8a\x95"
```
formats to
```
let s = "^@^A^B^C^Y \376^?\200\342\212\225"
```
So I guess that escaping chars < ' ' and = \x7f even in minimal mode is needed.

@hcarty, I think the 'eating' behavior you saw is due to a terminal not printing control codes, the bytes seem to be in output files.

hcarty:
@jberdine Ah good catch, thanks!

smondet:
@jberdine OK I've escaped those characters and made the `looks_ike_format` lazy.

Do you have an example of code messed up by `is_simple`? I've been trying different cases and they output the same.
(this one right?
https://github.com/smondet/ocamlformat/blob/string-and-char-literals/src/Fmt_ast.ml#L640)

Also in
https://github.com/smondet/ocamlformat/blob/string-and-char-literals/src/Fmt_ast.ml#L640
is the absence of space between `|` and some of the patterns a bug or is it really intended?

hcarty:
I think formatting like
```
| A
 |B -> expr1
| C -> expr2
```
is intended to highlight that `A` and `B` result in the same `expr1` code path.

hcarty:
@smondet Is `octal` the right for that option?  For example `\019` isn't octal.

smondet:
@hcarty Oh right! it's decimal :)

jberdine:
Yes, @hcarty is correct about spacing of or-patterns. Note that there is the `sparse` option to always use `| Constructor`, but then there are more line breaks after `->` in order to keep from hiding the multiple patterns of one case structure. Overlooking these has been one of the syntactic gotchas we have hit in practice.

Regarding `is_simple`, I'm sorry, I was not clear and forgot that there are several `is_simple` functions. What I meant was the one called in [`fmt_args_grouped`](https://github.com/smondet/ocamlformat/blob/string-and-char-literals/src/Fmt_ast.ml#L769) which is defined in [Ast](https://github.com/smondet/ocamlformat/blob/string-and-char-literals/src/Ast.ml#L480).

For code that repros the issue, I ran `make fmt` (with a version of `looks_like_format` that did not use your 'all newlines are escaped' criterion) and in the resulting diff there were several cases where an arg that did not start its own line contained a break within, and then following arg also did not start its own line.

smondet:
With this:

```
  let rec is_simple (c: Conf.t) width ({ast= exp} as xexp) =
    let ctx = Exp exp in
    match exp.pexp_desc with
    | Pexp_constant Pconst_string (s, _)
      when Polymorphic_compare.equal c.break_string_literals `Newlines ->
        not (String.mem s '\n')
(* ... *)
```

it indeed transforms:

```
  let x =
    Printf.printf "%s\n\
                   %s\n"
      "delijsde\n\
       dsleijd\n\
       dsesldije\n\
       ljdselijdsed\n\
       dsleijdsedslidjselidjseldijse"
```

into:


```
  let x =
    Printf.printf
      "%s\n\
       %s\n"
      "delijsde\n\
       dsleijd\n\
       dsesldije\n\
       ljdselijdsed\n\
       dsleijdsedslidjselidjseldijse"
```

but with labelled arguments, it still behaves quites weirdly:

```
let x = StringLabels.concat ~sep:"some big\n sep" ["bouh"; "bah"]
```

into:

```
  let x = StringLabels.concat ~sep:"some big\n\
                                    \ sep" ["bouh"; "bah"]
```

(or


```
-    Cmts.fmt_before ~epi:(fmt "\n@\n") ~eol:(fmt "\n@\n") ~adj:(fmt "@\n")
+    Cmts.fmt_before ~epi:(fmt "\n\
+                               @\n") ~eol:(fmt "\n\
+                                                @\n") ~adj:(fmt "@\n")

-  and fmt_cmts_after = Cmts.fmt_after ~pro:(fmt "\n@\n") si.pstr_loc in
+  and fmt_cmts_after = Cmts.fmt_after ~pro:(fmt "\n\
+                                                 @\n") si.pstr_loc in
````

)

The only difference in treatment I see is the `"@,"` after the `:`?

File: src/Conf.ml [hunk 9]
```diff
@@ -155,6 +155,37 @@ let sparse =
...
+  let default = `Minimal in
+  mk ~default
+    Arg.(
+      value
+      & opt
+          (enum
+             [ ("hexadecimal", `Hexadecimal)
+             ; ("decimal", `Decimal)
+             ; ("minimal", `Minimal) ])
+          `Minimal
```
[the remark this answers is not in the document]
smondet: I thought it would be better to keep the word `default` available for a future better solution (?)

src/Fmt_ast.ml
```diff
@@ -303,18 +303,61 @@ let rec fmt_longident (li: Longident.t) =
       cbox 0 (fmt_longident li1 $ wrap "(" ")" (fmt_longident li2))
 
 
-let fmt_constant const =
+let fmt_constant (c: Conf.t) const =
+  let is_printable char_code = 0x20 <= char_code && char_code <= 0x7E in
+  let is_not_term_friendly char_code =
```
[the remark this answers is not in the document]
smondet: Then it's not minimal any more :)

By me, we don't need to escape any of those since text editors (and tools/compilers) can handle them properly. 

The special case of `\x00` is indeed maybe better escaped because it breaks some Unix tools (`sed` & co are in undefined behavior when there is a NUL).

AFAIK, we're lucky control-codes cannot be part of an UTF-8 character :)

src/Fmt_ast.ml
```diff
@@ -303,18 +303,61 @@ let rec fmt_longident (li: Longident.t) =
...
+    (* See discussion on PR #70 *)
+    char_code < 4 || char_code = 0x7F
```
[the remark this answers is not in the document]
smondet: I don't know where to put the bound (I think escaping only `\x00` would fit me fine).

it would be something like `Char.(' ' <= char)` since Base overrides `<=`. Order on ascii-codes seems more logical to me, since characters are weirdly defined once we also have unicode in mind.

jberdine:
I'm not sure the difference is because of labels, it could be lengths, etc. But rereading the definition of `is_simple`, I think that `is_trivial` earlier in `Ast` will need to be similarly adjusted.

src/Conf.ml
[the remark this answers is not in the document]
jberdine: I mean the `default` local variable defined just above, so that the default can be changed in only one place in the code.

src/Fmt_ast.ml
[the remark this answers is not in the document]
jberdine: I guess that I'm thinking of the mode where ocamlformat produces output to stdout, and I'm not at all confident that sending control codes to stdout will work as expected everywhere. But I'm not an expert in such topics. I was just thinking that in standard (7 bit) ascii there are only control codes and printing characters, and it would be reasonable to escape everything except the printing characters. But I don't have a strong position if you think that sending control codes to stdout will work.

src/Fmt_ast.ml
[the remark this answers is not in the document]
jberdine: Ok, I don't have a strong position, though I think that wrt unicode it's a long way from `char`.

src/Fmt_ast.ml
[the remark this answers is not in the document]
smondet: :+1:  let's play safe :)

smondet:
I've tried to fix the "simple/trivial" thing but still run into weird cases:
See https://github.com/smondet/ocamlformat/commit/bc0edc9808899d668be59d9e20bf5b662e1392d9

Things like:

```
  let x = StringLabels.concat ~sep:"some big\n\
                                    \ sep" ["bouh"; "bah"]

  let f = function
    | "simple" ->
        true
    | "not\n\
       simple" | "not\n\
                  simple\n\
                  at all" ->
        false
    | other ->
        assert false
```

.ocamlformat
```diff
@@ -1 +1,2 @@
 margin 77
+break-string-literals never
```
jberdine: To preserve the current formatting

src/Ast.ml
```diff
@@ -60,10 +60,22 @@ let rec is_sugared_list exp =
...
+    | Some i when i = String.length s - 1 -> false
+    | _ -> true
+  in
+  Poly.equal c.break_string_literals `Newlines
+  && contains_internal_newline s
+
+
+let rec is_trivial c exp =
   match exp.pexp_desc with
+  | Pexp_constant Pconst_string (s, None) -> not (would_force_break c s)
```
jberdine: To prevent string literals that will force a line break from being considered 'trivial', and thereby 'simple', which the application argument formatting is not prepared to handle.

src/Fmt_ast.ml
```diff
@@ -328,28 +328,16 @@ let fmt_constant (c: Conf.t) const =
...
-            match (ch, c.Conf.break_string_literals) with
-            | ' ', _ when freshline -> (false, prev $ str "\\ ")
-            | '"', _ -> (false, prev $ str "\\\"")
-            | '\n', `Newlines
-              when not (Lazy.force looks_like_format)
-                   && index <> String.length string - 1 ->
-                (true, prev $ fmt "\\n\\@\n")
-            | other, _ -> (false, prev $ escape_char other) )
-        |> snd $ fmt "@]"
+        vbox 0
```
jberdine: Wrap in a vertical box so that all breaks are newlines

src/Fmt_ast.ml
```diff
@@ -328,28 +328,16 @@ let fmt_constant (c: Conf.t) const =
...
+          ( String.foldi string ~init:(false, fmt "") ~f:
+              (fun index (freshline, prev) ch ->
+                match (ch, c.break_string_literals) with
+                | ' ', _ when freshline -> (false, prev $ str "\\ ")
+                | '"', _ -> (false, prev $ str "\\\"")
+                | '\n', `Newlines when index <> String.length string - 1 ->
+                    (true, prev $ fmt "\\n\\@;<1000 0>")
```
jberdine: Use a break hint asking for a huge amount of space to force all enclosing boxes to break, not just the immediately enclosing one. Otherwise short strings with internal newlines can leave non-immediately enclosing boxes unbroken, leading to the bad formatting cases in the discussion.

smondet:
:+1: Indeed looks good!

```
  let x =
    StringLabels.concat
      ~sep:
        "some big\n\
         \ sep"
      ["bouh"; "bah"]


  let f = function
    | "simple" ->
        true
    | "not\n\
       simple"
    | "not\n\
       simple\n\
       at all" ->
        false
    | other ->
        assert false
```

Thanks for the comments explaining your changes btw!!
(especially useful to learn more on the `Format` magic)

jberdine:
Thanks!

hcarty:
@smondet Indeed, thank you!  These are both very useful features!  Time to update my local pin.

Issue #13, opened 2017-10-20
Add support for the OCaml object language


jberdine:
The object language of OCaml is largely unimplemented. Designing the formatting of these constructs would be best done by someone with a lot of experience with them, who knows the readability pitfalls to avoid, etc.

emillon:
A non-intrusive first step could be to support the `a#b` syntax. That would support code that uses object APIs without too much effort. My best guess is that `a#b` can be formatted with the same rules as `a.b`, but this is just a suggestion.

hhugo:
Fyi, I'm working on this.

PR #72, opened 2018-01-30
Handle the `Pexp_send _` case

smondet:
Just like the `Pexp_field _` one.
Cf. **@emillon**'s [comment][1] on #13.

```ocaml
  let f x = (x ()).contents

  let x obj = obj#hello ()

  let x obj_f = (obj_f ())#hello ()

  let f obj = obj#hello_some_pretty_long_one ~with_labels:true ()

  let f obj =
    obj#hello_some_pretty_long_one ~with_labels:true
      "desjd\n\
       dijsde\n" {md|
In **markdown**
|md}
```

[1]: https://github.com/ocaml-ppx/ocamlformat/issues/13#issuecomment-344237707

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -74,7 +74,7 @@ let would_force_break (c: Conf.t) s =
 let rec is_trivial c exp =
   match exp.pexp_desc with
   | Pexp_constant Pconst_string (s, None) -> not (would_force_break c s)
-  | Pexp_constant _ | Pexp_field _ | Pexp_send _ | Pexp_ident _ -> true
+  | Pexp_constant _ | Pexp_field _ | Pexp_ident _ | Pexp_send _ -> true
   | Pexp_construct (_, exp) -> Option.for_all exp ~f:(is_trivial c)
   | _ -> false
 
@@ -493,8 +493,8 @@ end = struct
     let ctx = Exp exp in
     match exp.pexp_desc with
     | Pexp_constant _ -> is_trivial c exp
-    | Pexp_array _ | Pexp_field _ | Pexp_send _ | Pexp_ident _
-     |Pexp_record _ | Pexp_tuple _ | Pexp_variant _
+    | Pexp_array _ | Pexp_field _ | Pexp_ident _ | Pexp_record _
+     |Pexp_send _ | Pexp_tuple _ | Pexp_variant _
      |Pexp_construct (_, None) ->
         true
     | Pexp_construct
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -74,7 +74,7 @@ let would_force_break (c: Conf.t) s =
 let rec is_trivial c exp =
   match exp.pexp_desc with
   | Pexp_constant Pconst_string (s, None) -> not (would_force_break c s)
-  | Pexp_constant _ | Pexp_field _ | Pexp_ident _ -> true
+  | Pexp_constant _ | Pexp_field _ | Pexp_send _ | Pexp_ident _ -> true
   | Pexp_construct (_, exp) -> Option.for_all exp ~f:(is_trivial c)
   | _ -> false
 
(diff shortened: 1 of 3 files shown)


src/Ast.ml
```diff
@@ -74,7 +74,7 @@ let would_force_break (c: Conf.t) s =
 let rec is_trivial c exp =
   match exp.pexp_desc with
   | Pexp_constant Pconst_string (s, None) -> not (would_force_break c s)
-  | Pexp_constant _ | Pexp_field _ | Pexp_ident _ -> true
+  | Pexp_constant _ | Pexp_field _ | Pexp_send _ | Pexp_ident _ -> true
```
jberdine: nit: was in alphabetic order before

src/Ast.ml
```diff
@@ -493,8 +493,8 @@ end = struct
     let ctx = Exp exp in
     match exp.pexp_desc with
     | Pexp_constant _ -> is_trivial c exp
-    | Pexp_array _ | Pexp_field _ | Pexp_ident _ | Pexp_record _
-     |Pexp_tuple _ | Pexp_variant _
+    | Pexp_array _ | Pexp_field _ | Pexp_send _ | Pexp_ident _
+     |Pexp_record _ | Pexp_tuple _ | Pexp_variant _
```
jberdine: nit: was in alphabetic order before

src/Ast.ml
```diff
@@ -647,6 +647,7 @@ end = struct
       | Pexp_setfield (_, _, e0) when e0 == exp -> Some (LessMinus, Non)
       | Pexp_setinstvar _ -> Some (LessMinus, Non)
       | Pexp_field _ -> Some (Dot, Left)
+      | Pexp_send _ -> Some (HashOp, Left)
```
jberdine: Isn't `#` non-associative?

smondet: My understanding of this <http://caml.inria.fr/pub/docs/manual-ocaml/expr.html> is that it *behaves* like `.`.

Also it was already the decision there: https://github.com/smondet/ocamlformat/blob/pexp-send/src/Ast.ml#L171-L172 :)

Testing does not throw ast-diff-errors for all these:

```ocaml
let f x = (x ()).contents
let f x = (x ()).contents#method_of_ref ()
let f x = ((x ()).contents ())#method_of_ref ()
let f x = x#hello.contents.contents
let f x = x#hello.contents#bye.contents
let f x = x.contents#hello
let f x = x.contents#hello.contents
```

jberdine: I'm not sure whether the manual is referring to the `#` send expression or the prefix ops that start with a `#` char, but the [parser](https://github.com/ocaml/ocaml/blob/trunk/parsing/parser.mly#L598) seems to treat them differently.

But you may be right that once code parses, then it might not be possible to trigger a formatting error. I don't know, but would rather match the parser if possible.

smondet: With `Non` instead of `Left` for `HashOp` it gives something like this:

```oaml
let f x = (x ()).contents

let f x = (x ()).contents#method_of_ref ()

let f x = ((x ()).contents ())#method_of_ref ()

let f x = (x#hello).contents.contents

let f x = ((x#hello).contents#bye).contents

let f x = x.contents#hello

let f x = (x.contents#hello).contents
````

My eyes/brain are trained to find parentheses helpful in those cases but I know it's not everyone's taste :)

Tell me if you prefer with `Non` and I'll commit that.

(PS the ops that start with `#` are not supposed to be part of the language itself, they are REPL-only, so in another chapter:
<http://caml.inria.fr/pub/docs/manual-ocaml/toplevel.html> 
There is still another `#` for type definitions and for pattern mattching though ... OCaml syntax :heart: )

jberdine: I think that the extra parens are due to the current order of `Dot` and `HashOp` being wrong in the definition of `type prec`. Fixing that and going with `Non`-associative seems like the right option.

(I think I saw somewhere that the  `#` ops were intended/allowed for syntax extensions, not just the repl directives. Not sure. But yeah, the OCaml syntax is not exactly small...)

smondet: (Ah right there is the `js_of_ocaml` ppx thing that converts `##` and `##.`!)

OK I'm changing the order in `pred` also :+1:

jberdine:
Thanks!

PR #73, opened 2018-01-30
Improve char escaping to ascii / uniform hexa / utf8

jberdine:
The lower-128 chars are escaped as standard by Char.escaped

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -62,7 +62,7 @@ let rec is_sugared_list exp =
 
 let would_force_break (c: Conf.t) s =
   let contains_internal_newline s =
-    match String.rindex s '\n' with
+    match String.index s '\n' with
     | None -> false
     | Some i when i = String.length s - 1 -> false
     | _ -> true
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -157,20 +157,38 @@ let sparse =
 
 let escape_chars =
   let doc =
-    "How to escape chars. Can be set in a config file with a `escape-chars {hexadecimal,decimal,minimal}` line."
+    "Escape encoding for character literals. Can be set in a config file with an `escape-chars {decimal,hexadecimal,preserve}` line. `hexadecimal` mode escapes every character, `decimal` produces ASCII printable characters using decimal escape sequences as needed, and `preserve` escapes ASCII control codes but leaves the upper 128 characters unchanged."
   in
   let env = Arg.env_var "OCAMLFORMAT_ESCAPE_CHARS" in
-  let default = `Minimal in
+  let default = `Decimal in
   mk ~default
     Arg.(
       value
       & opt
           (enum
-             [ ("hexadecimal", `Hexadecimal)
-             ; ("decimal", `Decimal)
-             ; ("minimal", `Minimal) ])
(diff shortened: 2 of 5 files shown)


jberdine:
@smondet, when thinking about how to describe what the character escaping does, it occurred to me that the behavior is almost the same as using `Char.escape` on the ascii chars, and only adjusting behavior for the non-ascii ones. I find the simplification in the description, and of the code, attractive. Would this approach work in your context? Are there other considerations I'm overlooking?

smondet:
@jberdine the problem then is that in `hexadecimal` mode we get a mix of decimal and hexa:

with this branch:

```
let magic_number = "\x00\x01\xb3\xf0"
```

becomes:

```
let magic_number = "\000\001\xb3\xf0"
```

which can become quite a headache for users parsing network packets :)
(but for that use case nothing is ideal anyway)

jberdine:
@smondet I see. But on master isn't there a similar problem with the printable characters?
```
let magic_number = "\x00\x01\x20\xf0"
```
formats to
```
let magic_number = "\x00\x01 \xf0"
```

smondet:
@jberdine yes that's why I said that no option was ideal, still 2 kinds of chars at once seems more manageable than 3, esp. since the "≥ 128" condition is quite arbitrary.
(but as long as the AST doesn't have the original string, I think don't there will be a perfect solution :/)

jberdine:
@smondet Ok, good point about 2 vs 3.

When *re*formatting source code, it is technically possible to look up the string from the original source based on the location information. Eventually I think ocamlformat should also become a library for code generation, e.g. for things like `deriving_inline`, and looking up original strings would introduce an unfortunate inconsistency between the reformat and generate use cases. But that might be ok. I guess it would also be possible in the code generation case to add an attribute to the AST to specify the encoding or explicit string even.

I don't think that the "≥ 128" condition is arbitrary though, since ASCII specifies exactly the 7-bit characters, and it is those for which ASCII and UTF-8 coincide. So by escaping only the 7-bit characters, we get something that precisely preserves UTF-8 strings.

In the short term, it seems to make sense to support 3 scenarios:
- escape as little as possible to map every char into ASCII
- preserve UTF-8
- uniformly encode every char as `\x..`

I have a suspicion that your original proposal was closer to this and my commentary pushed it away.

smondet:
(I meant "≥ 128" is arbitrary for the  "binary/byte-array" parsing/mangling use-case :) )

With the new commits it's more uniform indeed but the escaping of the strings is still catching `\n` and `"`:
https://github.com/ocaml-ppx/ocamlformat/blob/escape-chars/src/Fmt_ast.ml#L327-L328
so the hexadecimal does not turn out fully hexadecimal.


```ocaml
let mix = "\x00\x01\xb3\xf0\x0b\x09\x0a\x20\x22"
(* becomes *)  
  let mix =
    "\x00\x01\xb3\xf0\x0b\x09\n\
     \ \""
```

jberdine:
Ah yes, now I see what you meant. :-)

Yes, good point, strings are still not uniform. I'm working on that, but for discussion suppose that strings were escaped analogously to chars. (I'm integrating this escaping stuff with changes I have been working on to do paragraph-fill-style string and comment wrapping.)

PR #75, opened 2018-01-31
Fix formatting of [%ext e1]; e2

### LINKED ISSUE #74: Unsugared extension point formatted badly
Maybe related to some of the issues found during #63 with commenting/uncommenting of code sections.

Given:
```ocaml
let () = [%ext expr]; ()
```
ocamlformat `master` errors out, outputting:
```ocaml
let () = [%ext fun -> expr] ; ()
```
when run with `-g`.

hcarty:
Fixes #74 

Thanks you @jberdine for pointing out how to fix this!

The following now formats correctly:
```ocaml
let () =
  [%ext expr] ;
  ()
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -822,8 +822,11 @@ and fmt_expression c ?(box= true) ?eol ?parens ?ext ({ast= exp} as xexp) =
       ( { pexp_desc=
             Pexp_extension
               ( {txt}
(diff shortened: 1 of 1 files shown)


hcarty:
Also tested on some larger internal projects using this syntax and it works properly for those.

jberdine:
Thanks!

PR #76, opened 2018-02-02
Handle `Pexp_new` case (#13)

smondet:
Seems like a simple case.

```
let x = new Objects.one ~hello:true ()
let x = (new Objects.one) ~hello:true ()
```

become

```ocaml
let x = new Objects.one ~hello:true ()

let x = new Objects.one ~hello:true ()
```

Together with #72 I got to make `ocamlformat` succeed on some `js_of_ocaml-ppx`-heavy code of mine:

```ocaml
sprintf "Date: %s"
      (Js.to_string (new%js Js.date_now)##toString)
```

becomes

```ocaml
    sprintf "Date: %s" (Js.to_string [%js new Js.date_now] ## toString)
```

I didn't manage to test the `Lapply` case of:

```ocaml
type t =
  Longident.t =
    Lident of label
  | Ldot of Longident.t * label
  | Lapply of Longident.t * Longident.t
```

UTop seems to syntax-err when there is a functor application in the module path (?)


jberdine:
Thanks!

PR #83, opened 2018-03-05
Fix repeated ppx extension on mutual/recursive let-bindings

mbarbin:
Repeating the extension on each of the and causes a parse error.
```
let%ext () = () and%ext () = () in ()
```
  The extension is only needed on the first let of the letand group.
```
let%ext () = () and () = () in ()
```
(Context is trying ocamlformat on code using Core.Command.Parm's as well as Async.Deferred's monads).

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1255,7 +1255,9 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
         (vbox 0
            ( hvbox 0
                (list_fl bindings (fun ~first ~last binding ->
-                    fmt_value_binding c ~rec_flag ~first ?ext ctx binding
+                    fmt_value_binding c ~rec_flag ~first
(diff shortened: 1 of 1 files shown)


hcarty:
Good catch, thanks!

mbarbin:
Sorry I did not notice while doing the change through github.com the change currently makes the line of code too long in comparison to other lines in the file.  I am not sure how to indent, because the file is not stable through my local config of ocp-indent, I fear I will make undesirable whitespace changes.

hcarty:
@mbarbin If you run `make fmt` the source code will be formatted appropriately by `ocamlformat`.  It's very helpful for exactly that case!

jberdine:
Looks good, thanks! Its good to merge once you `make fmt` and update.

mbarbin:
Done, thanks!

jberdine:
Thanks!

Issue #85, opened 2018-03-05
Missing parens in patterns


mbarbin:
The syntax requires patterns **Ppat_unpack** and **Ppat_constraint** to be parenthesized.
On the following example:
```
let f (module X) = X.x
```
and
```
let f = function
  | `A { x = (_ : int) } -> ()
```
ocamlformat remove the parens, resulting in a syntax error:

```
let f module X = X.x
```
and
```
let f = function
  | `A { x= _ : int } -> ()
```

I toyed around locally in [srcAst.ml](https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Ast.ml) and made the function **parenze_pat** return **true** for those, which seems to fix it.  However, this seems too naive as a proposed fix, the code seems to have precise intends that I haven't understood yet.

```diff
diff --git a/ocamlformat/src/Ast.ml b/ocamlformat/src/Ast.ml
--- 27f85f1c5a6f/ocamlformat/src/Ast.ml
+++ working-dir/ocamlformat/src/Ast.ml
@@@@@@@@@@ -779,18 +779,18 @@@@@@@@@@
      | ( Pat {ppat_desc= Ppat_construct _}
        , Ppat_construct ({txt= Lident "::"}, _) ) ->
          true
      | Exp {pexp_desc= Pexp_let (_, bindings, _)}, Ppat_tuple _ ->
          List.exists bindings ~f:(function
            | {pvb_pat; pvb_expr= {pexp_desc= Pexp_constraint _}} ->
                pvb_pat == pat
            | _ -> false )
-|    | Pat {ppat_desc= Ppat_constraint _}, Ppat_unpack _
-|     |Pat {ppat_desc= Ppat_construct _ | Ppat_variant _}, Ppat_constraint _
+|    | _, Ppat_unpack _
+|    | _, Ppat_constraint _
       |( Pat
            { ppat_desc=
                ( Ppat_alias _ | Ppat_array _ | Ppat_constraint _
                | Ppat_construct _ | Ppat_variant _ ) }
        , Ppat_tuple _ )
       |( ( Pat
              { ppat_desc=
                  ( Ppat_construct _ | Ppat_exception _ | Ppat_or _
@@@@@@@@@@ -800,23 +800,18 @@@@@@@@@@
        , Ppat_alias _ )
       |( Pat {ppat_desc= Ppat_lazy _}
        , (Ppat_construct _ | Ppat_variant (_, Some _) | Ppat_or _) )
[...]

PR #86, opened 2018-03-06
Missing parens in patterns

mbarbin:
#85 

This pull request is not complete, it is showing a patch too general followed by an example of regression.  Needs to tweak it some more.


jberdine:
I made a small change in 767d9f7c6af30293ef4f1e237191d17f160a34d5 that seems to fix the cases in #85, but there may be more bad cases and a more involved patch might be needed.

jberdine:
@mbarbin, should we close this?

mbarbin:
@jberdine some code I was lookig at was still missing a few parens as of the last update, but I need to update this PR to something more refined.  Closing for now, I will adjust.  Thanks!

PR #92, opened 2018-03-10
Apply no transformation on empty files or files with only comments

mbarbin:
The current behavior is to drop the comments, and then failing due to detecting ast changes.

Instead we can apply no translation and preserve the input file as is.

Empty files do come in practice, as an empty mli for a main.exe gives you warnings for unused values in the main.ml.

A more refine behavior would be to process the comments if any.  This is left as future work.


jberdine:
Thanks!

PR #93, opened 2018-03-10
use sugar syntax to avoid repeating record field name in Pexp_constraint

mbarbin:
Some sugar was added to the parser at some point for constrained field in records:

That is one can write:
```
[%sexp { x : int; y : string }]
```
Instead of:
```
[%sexp { x = (x : int); y = (y : string) }]
```
Recognize this case, and produce it.


mbarbin:
I don't remember from which Ocaml version this was added, but this would mean the code produced would need to be parsed by a recent enough ocaml version.  Not sure whether this is worth guarding by some configurable option.

jberdine:
Thanks!

PR #94, opened 2018-03-10
Write extension on the let at toplevel: e.g. let%expect_test _

mbarbin:
Opening the discussion to change:
```
[%expect_test
 let _ = ...
]
```
into:
```
let%expect_test _ = ...
```

Seems it would fit nicely with the new ppx_let style, let%bind, let%map, etc.

Example [here](https://github.com/janestreet/core/blob/9f38c7fc8b9b2121b7934f575fb3af73db5c2a4c/test/src/cidr_tests.ml)

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -470,8 +470,11 @@ let rec fmt_attribute c pre = function
            (wrap "[" "]" (str pre $ str txt $ fmt_payload c (Pld pld) pld))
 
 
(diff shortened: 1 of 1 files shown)


hcarty:
This is useful for Lwt too, at least in limited form, as it supports top-level `let%lwt v = expr` as an alias for `let v = Lwt_main.run expr`.

jberdine:
Looks good! I think that `let%ext` is much more prevalent in human-written code.

PR #96, opened 2018-03-22
Use original filename when given in error messages.

mbarbin:
```
$ jbuilder exec src/ocamlformat.exe -- /tmp/tmp.1234.ml --name '/path/to/original_file.ml'
File "/path/to/original_file.ml", line 6, characters 4-7:
Error: Syntax error
```

Without the patch this is reporting:
```
$ jbuilder exec src/ocamlformat.exe -- /tmp/tmp.1234.ml --name '/path/to/original_file.ml'
File "/tmp/tmp.1234.ml", line 6, characters 4-7:
Error: Syntax error
```

### DIFF
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -17,11 +17,12 @@ open Migrate_ast
 
 (** Operations on translation units. *)
 type 'a t =
-  { input: string -> In_channel.t -> 'a * (string * Location.t) list
+  { input:
+      input_name:string -> In_channel.t -> 'a * (string * Location.t) list
   ; init_cmts: string -> 'a -> (string * Location.t) list -> unit
   ; fmt: Conf.t -> 'a -> Fmt.t
   ; parse:
-      ?warn:bool -> string -> string -> In_channel.t
+      ?warn:bool -> input_name:string -> In_channel.t
       -> 'a * (string * Location.t) list
   ; equal:
       'a * (string * Location.t) list -> 'a * (string * Location.t) list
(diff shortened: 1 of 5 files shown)


src/Translation_unit.ml
```diff
@@ -17,11 +17,10 @@ open Migrate_ast
 
 (** Operations on translation units. *)
 type 'a t =
-  { input: string -> In_channel.t -> 'a * (string * Location.t) list
```
jberdine: Note that `input` and `parse` are different in the reason support. Is removing `input` necessary for this change?

mbarbin: Ah, OK sorry about this.  I did not see it while compiling.  Is the reason support in another branch or another repo?

Removing input is not necessary for this change, I will make another version where this does not happen.

mbarbin: I see it now, it's:
```
make reason
```
which is not built by default (and btw, it is currently broken as I missed the no_translation change in f00efadca380e94f38db320d8ac61d52b0c97815).

Shall the target  *reason* be added to the default make target?

jberdine: It could be added to the default target. It isn't now only to avoid waiting for the extra link step on each build when developing. Breaking compilation of the reason stuff has been rare. But mainly it has not been tested for some time, last time with reason < v2, so I'm unsure if it works reasonably well, or if anyone is using it.

mbarbin: Understood, thank you.

I updated the commit to not break the compilation this time.

src/ocamlformat.ml
```diff
@@ -15,7 +15,7 @@
 let impl : _ Translation_unit.t =
   let parse = Translation_unit.parse Migrate_ast.Parse.implementation in
   { parse
-  ; input= parse Location.none.loc_start.pos_fname
+  ; input= (fun ~input_name ic -> parse ~input_name ic)
```
jberdine: Would it be clearer to use `parse ?warn:None` here instead? Or maybe there is some other reason to prefer the eta-expansion?

src/ocamlformat.ml
```diff
@@ -28,7 +28,7 @@ let impl : _ Translation_unit.t =
 let intf : _ Translation_unit.t =
   let parse = Translation_unit.parse Migrate_ast.Parse.interface in
   { parse
-  ; input= parse Location.none.loc_start.pos_fname
+  ; input= (fun ~input_name ic -> parse ~input_name ic)
```
jberdine: Same question as above

src/ocamlformat.ml
mbarbin: Agreed, I'd prefer ``parse ?warn:None``.  Thank you.  I'd be grateful if you want to polish the PR further, otherwise I'd push an update an Monday.  Thanks!

jberdine:
Thanks!

PR #102, opened 2018-03-22
Include attributes on extension points

### LINKED ISSUE #101: Extension points with attributes
The following fails to format:
```ocaml
let _ =
  [%ext match x with
         | () -> ()]
  [@attr y]
```
Output with `-g` is:
```ocaml
let _ = (match%ext x with () -> ())
```

hcarty:
Fixes #101

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1478,8 +1478,9 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
                             | Pexp_try _ | Pexp_let _ | Pexp_ifthenelse _
                             | Pexp_sequence _ ) } as e1 )
                     , _ ) } as str ) ] ) ->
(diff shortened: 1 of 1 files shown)


src/Fmt_ast.ml
```diff
@@ -1479,6 +1479,7 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
                             | Pexp_sequence _ ) } as e1 )
                     , _ ) } as str ) ] ) ->
       fmt_expression c ~box ?eol ~parens ~ext (sub_exp ~ctx:(Str str) e1)
+      $ fmt_atrs
```
jberdine: I have not tested, but I suspect that the whole body of this case would be better wrapped in `hvbox 0 ( ... )`.

hcarty: Added!

jberdine:
Thanks!

PR #104, opened 2018-03-28
Handle the case `Ptyp_object _` (#13)

smondet:
The following cases work well:

```ocaml
type t =
  < hello: string  (** some doc *)
  ; world: int
  ; more: int * float
  ; make: int
  ; it: string
  ; long: float [@default 42.] >
[@@deriving make]

type 'a u = <hello: string  (** more doc *) ; world: int ; .. > as 'a

type 'a v = < .. > as 'a

let x : unit -> <bouh: string ; .. > = fun () -> assert false

let lookup_obj : < .. > -> (< .. > as 'a) list -> 'a = fun _ -> assert false
```

Then, the new "`Oinherit`" syntax-case
(added last summer https://github.com/ocaml/ocaml/pull/1118)

It works with OCaml 4.06.1, but fails to parse with `ocamlformat` for some reason I don't understand:

```ocaml
type with_inherit = < stuff: int; t>
```

```
File "/tmp/tt.ml", line 1, characters 35-36:
Error: Syntax error
```

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -303,8 +303,12 @@ end = struct
               | Rtag (_, _, _, t1N) -> List.exists t1N ~f
               | Rinherit t1 -> typ == t1 ) )
       | Ptyp_package (_, it1N) -> assert (List.exists it1N ~f:snd_f)
-      | Ptyp_object _ | Ptyp_class _ ->
-          internal_error "objects not implemented" [] )
+      | Ptyp_object (fields, _) ->
+          assert (
+            List.exists fields ~f:(function
+              | Otag (_, _, t1) -> typ == t1
+              | Oinherit t1 -> typ == t1 ) )
+      | Ptyp_class _ -> internal_error "objects not implemented" [] )
     | Pat ctx -> (
       match ctx.ppat_desc with
       | Ppat_constraint (_, t1) -> assert (typ == t1)
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -569,8 +569,29 @@ and fmt_core_type c ?(box= true) ({ast= typ} as xtyp) =
               $ list ls "@ " (fmt "`" >$ str)
           | Open, Some _ -> impossible "not produced by parser" )
         $ fits_breaks "]" "@ ]" )
-  | Ptyp_object _ | Ptyp_class _ ->
-      internal_error "classes not implemented" [] )
+  | Ptyp_object (fields, closedness) ->
+      hvbox 0
+        (wrap_fits_breaks "<" ">"
+           ( list fields "@ ; " (function
+               | Otag (lab_loc, attrs, typ) ->
+                   (* label loc * attributes * core_type -> object_field *)
+                   let doc, atrs = doc_atrs attrs in
+                   let fmt_cmts = Cmts.fmt c lab_loc.loc in
+                   fmt_cmts
+                   @@ hvbox 4
+                        ( hvbox 2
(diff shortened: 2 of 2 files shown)


jberdine:
Thanks, this progress is great to see!

First, make sure that you're testing the debug build (i.e. run `_build/dbg/src/ocamlformat.exe -g`), as most of the sanity checking is not compiled into the default build (it's very slow).

I'm not sure it is the problem, but the code checking that "subexpressions" are correct will need to be extended too: https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Ast.ml#L306 . This code is checking that for values `{ctx; ast}` of type `Ast.xt`, the `ast` is really an "immediate" subterm of `ctx`. The parenthesization, etc. code heavily relies on this notion of "subterm", and it's easy to get a little wrong, especially when syntactic sugar is involved, and debugging the consequent stange behavior is not obvious.

smondet:
@jberdine I use `--debug` indeed and the syntax error happens before any debug output.

OK, I'll add the `Ast.check` code, do you have an example of code where this case can be triggered?

smondet:
Also, about the syntax error, maybe it is in the migrate-parsetree dependency?

I get an error with a JaneSt ppx:


```
         │ Welcome to utop version 2.1.0 (using OCaml version 4.06.1)! │                                                                                                                         

Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads


Type #utop_help for help about using utop.

ocaml# #require "ppx_sexp_conv";;
ocaml# type u = < world: int ; t>;;inheritance in object type are not supported before OCaml 4.06
ocaml# type u = < world: int ; t>;;
Error: Error while running external preprocessor
Command line: /home/smondet/.opam/4.06.1/lib/ppx_deriving/./ppx_deriving '/tmp/camlppx03e0ba' '/tmp/camlppx3414d1'

ocaml# 
```

jberdine:
Note that because of conditional compilation, `--debug` does not do much in the default build, while it does in the dbg one.

Re the parse error, it seems that ppxlib is behind. Compare https://github.com/ocaml-ppx/ppxlib/blob/master/ast/ast.ml#L182 and https://github.com/ocaml/ocaml/blob/trunk/parsing/parsetree.mli#L96 .

jberdine:
But OMP is ok I think: https://github.com/ocaml-ppx/ocaml-migrate-parsetree/blob/master/src/ast_406.ml#L159

smondet:
@jberdine I filled the `Ast.check_typ` part like for records, but I don't know how to construct a piece of code that triggers the parentization :)

jberdine:
@smondet, I'm actually not sure that parenthesization is needed for object types, since they are always embraced by `<` and `>`. But anyhow, the code in [`fmt_core_type`](https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Fmt_ast.ml#L519) will wrap parens if [`Ast.parenze_typ`](https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Ast.ml#L745) returns true. So if more parens are needed, that would be the code to change.

For me this PR works on all the code in the given examples. Looks good to go to me, or is there anything left you are aware of?

smondet:
@jberdine works for me too 👍

It seems that the `master` branch of `ppxlib` indeed contains the syntax parsing fix.

jberdine:
Thanks!

PR #108, opened 2018-03-29
Add parens around module value types in signatures

### LINKED ISSUE #107: Missing parens on first class module value's signature
```ocaml
module type M = sig
  val storage : (module S with type t = t)
end
```
formats to
```ocaml
module type M = sig
  val storage : module S with type t = t
end
```
resulting in a type error due to the missing `(` + `)` around the module type.

hcarty:
Fixes #107


hcarty:
This always adds wrapping parens as I _think_ they're always required in this context.

jberdine:
Thanks!

PR #122, opened 2018-04-09
fix for tuples type inside variant defs

hhugo:
```type t = A of (int * int) * int```


jberdine:
@hhugo, could you have a look at this version, which tries to fix this issue by changing the parenthesization logic rather than special case during formatting.

hhugo:
+1

hhugo:
works for me

Issue #120, opened 2018-04-08
missing parents with type annotation inside patterns


hhugo:
```
let f = match None with
  | (_ : int option) -> true
```

PR #124, opened 2018-04-09
fix constraint in pattern

hhugo:
I'm not sure this is the right fix.
Test case is:
```
let f = 
  match None with
  | Some (_ : int) -> e 
  | (_ : int option) -> e
```


hhugo:
fix https://github.com/ocaml-ppx/ocamlformat/issues/120

PR #134, opened 2018-04-11
 Fix: parens around lists in local opens

hhugo:


hhugo:
`M.([a;b;c])` becomes `M.[a;b;c]`

hhugo:
Ack, cc @adept

PR #135, opened 2018-04-13
Add some regtests (WIP)

hhugo:
Is is based on the ocp-indent `test.sh` script available https://github.com/OCamlPro/ocp-indent/blob/rewind/tests/test.sh.

I did a quick scan over github issues and collected a bunch of test case.

Run `make regtests` to see the current state of regtests.


jberdine:
Thanks for kicking this off!  A few comments for discussion:

- It would probably be useful to run these in CI by either incorporating `make regtests` into `make test` or adding it separately to `.travis.yml`.

- If the error output is to be checked in, we should sanitize it so that e.g. generated filenames and line numbers in stack backtraces are not shown.

- Is it intended to commit the html report file, I'm worried about that causing a lot of churn.

- I expect that it would be more useful to distinguish the cases where a test fails (non-zero exit code) and those where it exits with zero but produces output that is unexpected. The first kind, abstract syntax failures, are uncontroversial bugs that should block the CI, while the second kind are a mix of bugs, limitations, and are likely to be controversial or sensitive to configuration.

- It would probably better to use the `dbg` build when running tests, to enable the internal assertion checking.

- Currently `make test` checks that the ocamlformat source itself is formatted by passing `-n 1`, which fails on code which is not invariant under formatting. I wonder if it would be easier to do such an invariance check instead of checking in both test input and output. Thoughts?

hhugo:
- I don't think `[FAILED]` test should block CI (even if exit code is non-zero). Only regressions (failing test in the passing directory) should. This allows to check in tests for bug that are not yet fixed (like test for the object language).
- I'll remove failing.html

hhugo:
- CI now runs regtests and will fail in case of unexpected changes. it also uses the `dbg` mode

test/test.sh
```diff
@@ -0,0 +1,297 @@
...
+done
+
+TMP=$(mktemp -d /tmp/ocamlformat-test.XXXXX)
+trap "rm -rf /tmp/ocamlformat-${TMP#/tmp/ocamlformat-}" EXIT
+
+ocamlformat() {
+    [ $# -eq 1 ]
+    opts=$(cat $1.opts 2>/dev/null || true)
+    tmpfile=$TMP/$(basename $1)
+    "$OCAMLFORMAT" $opts "$1" >$tmpfile 2>&1 || echo "NON-ZERO EXIT CODE" > $tmpfile
```
jberdine: Note that the `--output` option could be used instead of redirecting stdout for slight perf gain since ocamlformat will generate a temp file when outputting to stdout. But I guess that it is desirable to redirect all output in case of errors.

jberdine:
Thanks!

PR #138, opened 2018-04-15
Improve: no parens for { !e with a }

hhugo:


src/Ast.ml
```diff
@@ -931,6 +931,15 @@ end = struct
       | Pexp_record (flds, _)
         when List.exists flds ~f:(fun (_, e0) -> e0 == exp) ->
           exposed Non_apply exp (* Non_apply is perhaps pessimistic *)
+      | Pexp_record
+          ( _
+          , Some
+              ( { pexp_desc=
+                    Pexp_apply ({pexp_desc= Pexp_ident {txt= Lident i}}, _)
```
jberdine: I think this should force there to be a single arg, or else cases like
```
{(( ! ) x y) with a; b= c}
```
would fail (thinking of a custom prefix op).

src/Ast.ml
```diff
@@ -931,6 +931,15 @@ end = struct
...
+                } as e0 ) )
+        when (e0 == exp) && is_prefix_id i ->
```
jberdine: Using `is_prefix` could shorten the pattern a little.

PR #139, opened 2018-04-15
 Improve: indent rhs of ref update

hhugo:


src/Fmt_ast.ml
```diff
@@ -1010,6 +1010,12 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
         ( fmt_expression c (sub_exp ~ctx s) $ fmt ".["
         $ fmt_expression c (sub_exp ~ctx i) $ fmt "]@ <- "
         $ fmt_expression c (sub_exp ~ctx e) )
+  | Pexp_apply
+      ( {pexp_desc= Pexp_ident {txt= Lident ":="}; pexp_attributes= []}
+      , [(Nolabel, r); (Nolabel, v)] ) ->
```
jberdine: You probably need to add `hovbox 0` here analogous to the general case below, or else get cases like
```
;; if cccccccccccccccccccc then x :=
       f aaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbb
```

hhugo: indeed

test/passing/refs.ml
```diff
@@ -5,6 +5,10 @@ let _ =
   xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx :=
     something very
       looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong ;
+  xxxxxxxxxxxxx xxxxxxxxxxx xxxxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx xxxxxxxxxx
+    xxxxxxxxxxxxx :=
+    something very
+      looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong ;
```
jberdine: @hhugo what do you think of this one? To me this would be clearer with `:=` on the left margin.

hhugo: I'm with you, the following look better to me. @mbarbin, any feedback ?
```
xxxxxxxxxxxxx xxxxxxxxxxx xxxxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx xxxxxxxxxx
  xxxxxxxxxxxxx
:=
  something very
    looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong ;
```

mbarbin: +1.  I find it is easier to spot the operator `:=` when it is on the left margin.

jberdine: Ok, it sounds like using the previous formatting in case `is_simple` does not hold of the lhs might work.

hhugo:
now only applying the style when lhs is simple.

hhugo:
Thanks

jberdine:
You did the work!

hhugo:
You've pressed the merge button :)

Issue #115, opened 2018-04-06
make precedence explicit for `a || b && c`


hhugo:
I would prefer `a || b && c` to be printed like `a || (b && c)`

hhugo:
cc @mbarbin

smondet:
I also prefer explicit parentization, just noting that `ocamlformat` already shows some precedence when it uses indentation:

```ocaml

let _ =
  something_long "actually, very very very very very very very very long, ...."
  && lorem
       "ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, \
        ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, "
  || lorem
       "ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, \
        ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, "
     && lorem
          "ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, \
           ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, ipsum, \
           ipsum, ipsum"
```

jberdine:
https://github.com/ocaml-ppx/ocamlformat/commit/ea5ed9ffeae42fe9caaa1005786f3d83da802157 adds parens in this case, based on one possible generalization of the given example. Please reopen with more comments/examples if this is not what was intended.

PR #141, opened 2018-04-15
Improve: no parens around infix comparison ops (refines #115)

hhugo:


src/Fmt_ast.ml
```diff
@@ -1077,9 +1077,21 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
         $ fmt_if_k (not last) (break_unless_newline 1 0)
       in
       let is_nested_diff_prec_infix_ops =
+        let skip_parent_for_some_infix x =
+          match x.pexp_desc with
+          | Pexp_ident {txt= Lident i} -> (
+            match i with
+            | "==" | "=" | "<" | ">" | "<=" | ">=" | "<>" | "!=" -> true
```
jberdine: I'm a little wary of what seems like hard-coding a special case. Is there a more general rule that this is implementing, that could be used as an explanation for which infix ops get parenthesized and which don't? I wonder if it would make sense to omit parens for (assignment and) infix ops of precedence InfixOp0 (that is, ["!="] and symbols starting with ['=' | '<' | '>' | '|' | '&' | '$'])? I guess the question is whether it is really just these ops that you think shouldn't have parens, or whether these are just the only ones of the low precedence which are defined in the default environment?

Probably best would be to omit parens in the case where the type of the infix op arg differs from that of the return type, but the formatter (currently) operates on the untyped tree.

src/Fmt_ast.ml
```diff
@@ -1077,9 +1077,21 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
...
+            | ":=" -> true
+            | _ -> false )
+          | _ -> assert false
```
jberdine: If it doesn't complicate things, I'd prefer to avoid [assert false] and move the [is_infix] check here. I expect that would be clearer.

src/Fmt_ast.ml
```diff
@@ -1077,9 +1077,21 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
...
+        in
+        let may_need_parent e =
+          is_infix e && not (skip_parent_for_some_infix e)
+        in
         match (xexp.ctx, xexp.ast.pexp_desc) with
         | Exp {pexp_desc= Pexp_apply (f, _)}, Pexp_apply (e, _) ->
-            is_infix e && is_infix f
+            may_need_parent e && may_need_parent f
```
jberdine: "parent" or "parens"?

hhugo:
reworked

PR #142, opened 2018-04-15
Improve: support (type a b c)

hhugo:


hhugo:
@jberdine, i do not manage to reproduce the CI failure at home.

src/Fmt_ast.ml
```diff
@@ -1432,11 +1442,6 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
                    $ fmt "@ ->" )
                $ fmt "@ " $ cbox 0 (fmt_expression c (sub_exp ~ctx pc_rhs))
                )) )
-  | Pexp_newtype ({txt}, exp) ->
-      hvbox 0
-        (wrap_if parens "(" ")"
-           ( fmt "fun (type " $ str txt $ fmt ") ->@ "
-           $ fmt_expression c (sub_exp ~ctx exp) $ fmt_atrs ))
```
jberdine: I wonder if there is an arrangement of attributes which is not expressible in the sugared form. But thinking about that now, the current `sugar_fun` is already incorrect.

jberdine:
Curious, I can't repro the CI failure either. I don't know what's going on there.

Issue #133, opened 2018-04-11
Add an option to not add extra newlines


samoht:
It would be nice to not add extra newlines everywhere :-) 

In https://github.com/ocaml-ppx/ocamlformat/issues/46#issuecomment-343607985 @jberdine mentions that having an option to have less newlines would be desirable, so I am opening an issue to keep track of that. Ideally, 1 newline between long functions and 0 between one-liner should be enough :p

gpetiot:
Do you have some examples of extra newlines with the current version of ocamlformat ?

hhugo:
https://github.com/ocaml-ppx/ocamlformat/pull/285 is addressing the issue by preserving from the source up to one new line.

jberdine:
There is a difference between preserving the formatting and auto-formatting it in a "no extra newlines" style though. In general I consider all "preserve" behavior to be basically a failure of the auto-formatting algorithms. Some things are just too hard, and so I don't feel bad for failing. But @samoht's proposed rule seems simple enough I wouldn't write this off as impossible yet. (Although my first quick hack attempt did quickly get broken and complicated.)

samoht:
@gpetiot here is an example:

```ocaml
let f x = x + 1
let g = ()

let f = function
  | `xxxxxxxxxxxxxxxxxx ->
      yyyyyyyyyyyyyyyyyyyyyyyy
        zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
        zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
        zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

let x = 1
let y = 2
```

This style is quite common if you look at @dbuenzli libraries, and I know a few people using it.

jberdine:
Yes, I also favored this style in my own code before...

One thing to stay aware of is what happens when there are docstrings on any of the bindings in the example. If one-liners with docstrings don't have newlines between, then warning 50 very quickly gets triggered and then ocamlformat doesn't know where to put the docstrings and dies.

samoht:
I'm fine to add more lines if you add a docstring. "morally" the doc string is attached to the function and so it's no more a one-liner.

PR #144, opened 2018-04-15
Improve: two open lines following multiline definition only with --sparse

jberdine:
Generating two open lines following multi-line function definitions only if an option is set is easy. This PR reuses `--sparse`. Generating no open lines following single-line definitions is harder, I expect mostly as some work is needed to avoid ambiguous docstring attachment and the current implementation relies on always having an open line following a definition.


jberdine:
link #133

samoht:
That looks like a very good first step :-) Thanks!

jberdine:
I'm tempted to merge this as-is as a first step since the more sparse style is still easily available.

Any feedback from others?

PR #148, opened 2018-04-17
Align arrows in function signatures.

hhugo:
@jberdine, are you happy with that style or would you rather have a config for it ?

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -225,8 +225,11 @@ end
 (** Operations determining precedence and necessary parenthesization of
     terms based on their super-terms. *)
 and Requires_sub_terms : sig
-  val is_simple :
-    Conf.t -> (expression In_ctx.xt -> int) -> expression In_ctx.xt -> bool
+  val is_simple
+    :  Conf.t
+    -> (expression In_ctx.xt -> int)
+    -> expression In_ctx.xt
+    -> bool
 
   type cls = Let_match | Match | Non_apply | Sequence | Then | ThenElse
 
--- a/src/Cmts.mli
+++ b/src/Cmts.mli
@@ -26,29 +26,43 @@
 module Format = Format_
 open Migrate_ast
 
-val init_impl :
-  string -> Parsetree.structure -> (string * Location.t) list -> unit
+val init_impl
+  :  string
+  -> Parsetree.structure
+  -> (string * Location.t) list
+  -> unit
 (** [init_impl source_text structure comments] associates each comment in
     [comments] with a source location appearing in [structure], using
     [source_text] to help resolve ambiguities. Initializes the state used by
     the [fmt] functions. *)
 
-val init_intf :
-  string -> Parsetree.signature -> (string * Location.t) list -> unit
+val init_intf
+  :  string
+  -> Parsetree.signature
+  -> (string * Location.t) list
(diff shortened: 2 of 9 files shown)


src/Ast.ml
```diff
@@ -225,8 +225,11 @@ end
 (** Operations determining precedence and necessary parenthesization of
     terms based on their super-terms. *)
 and Requires_sub_terms : sig
-  val is_simple :
-    Conf.t -> (expression In_ctx.xt -> int) -> expression In_ctx.xt -> bool
+  val is_simple
+    :  Conf.t
+    -> (expression In_ctx.xt -> int)
+    -> expression In_ctx.xt
+    -> bool
```
jberdine: @hhugo can you confirm that cases like this are intended? An alternative would be to align the arrows when the type itself does not fit on one line, rather than the entire `val ... : ...` signature item.

hhugo: This was not intended and it just a side effect of me trying to have `:  typ` on its own line when the type does not fit.

The current PR will format to either (1) or (3). If you know how to support (2) as well, please amend this PR.

1.
```
val x : a -> b -> c
```
2.
```
val x :
  a -> b -> c
```
3.
```
val x
  :  a
  -> b
  -> c
```

jberdine: Would 1 and 2 with the following variant of 3 work for you?
```
val x :
     a
  -> b
  -> c
```

jberdine: The distinction between 1 and 2 is just whether it fits on one line, if so, 1, if not break after the `:` and see if it fits, if so, 2, else 3.

I see the code churn problem with my variant 3. I can see how to implement the original 3, but it would entail using the following variant of 2:
```
val x
  : a -> b -> c
```

The trouble is that the implementation does not actually backtrack through options of whether a line fits or not (that gets badly exponential quickly, and would involve replacing the `Format` module with something of a significantly different design). That makes it problematic to e.g. place the `:` on the same line as `x` if the later type fits on a line but place the `:` on the next line otherwise. There might be a way, but I haven't seen it yet.

hhugo: The main issue I see with your variant is the interaction with ocp-indent: 
```
val x :
  a
  -> b
  -> c
```

jberdine: :-)  Now what was I saying about ocp-indent vs flexibility... ;-)

What about the variant of 2, ocp-indent is ok with that isn't it?

hhugo: the variant of 2 seems fine to me.

mbarbin: @jberdine do you feel strongly having 2 is needed?  I think the rule with 3 variation is needlessly complex.  1 if it fits, either just go to 3, seems just fine to me.  I would think that optimizing for the class of definitions that just happens to fall in the 2 range is not worth it.

jberdine:
There are two separable things going on in this PR, one is aligning arrow types and the other is fiddling the position of the `:` for ocp-indent. I've tried to implement the first part in #161 in a simple way for all occurrences of arrow types, without changing the box structure too much (which may interact badly with comments, etc.). I've also added a config option in #162 to use for ocp-indent compatibility changes, and tried to adapt this PRs changes. Feedback on this approach, and how usable the results of #162 are, would be helpful.

PR #149, opened 2018-04-18
change formatting of let module/open/exception/pats

hhugo:
- force break after `let ... in`
- use long form for pexp_open if followed by a let keyword or a ppx extensions

### DIFF
--- a/test/passing/let_in_constr.ml
+++ b/test/passing/let_in_constr.ml
@@ -1,1 +1,4 @@
-let _ = Some (let open! List in 1)
+let _ =
+  Some
+    (let open! List in
+    1)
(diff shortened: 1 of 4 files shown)


test/passing/let_in_constr.ml
```diff
@@ -1 +1,4 @@
-let _ = Some (let open! List in 1)
+let _ =
+  Some
+    (let open! List in
+    1)
```
hcarty: Should `1` be aligned with the `l` in `let` here?

hcarty:
I think this is a nice improvement in general.

test/passing/let_in_constr.ml
jberdine: Quite possibly, but that is unrelated so won't delay this PR.

Issue #69, opened 2018-01-22
Splitting pipeline operators on newlines


lpil:
Hello! First of all thank you for this tool, I'm very much enjoying it.

I like to use a pipeline operator in order to structure my code so that it reads from left to right (similar to F#, Elm, Elixir, ReasonML etc)

```ocaml
let ( |> ) x f = f x
```

When using this ocamlformat wraps long lines like so:

```ocaml
let sum_of_squares num =
  num + 1 |> List.range 0 |> List.map ~f:square
  |> List.fold_left ~init:0 ~f:( + )
```

It's more commonly formatted like so:

```ocaml
let sum_of_squares num =
  num + 1 
  |> List.range 0 
  |> List.map ~f:square
  |> List.fold_left ~init:0 ~f:( + )
```

Is it possible to get a format closer to this, with one function on each line? Alternatively is there another operator that I could to get this style?

Thanks,
Louis

jberdine:
Currently `|>` is treated just like any other infix operator, so one question is whether this suggestion to always break is exclusively about `|>` or should apply to some/all other infix operators as well?

Note that the current implementation tries to ensure that the cases which are not broken are both short and simple in an attempt to avoid generating difficult to read code. I would be very interested to see examples of real code where this fails.

I would be happy in principle with a PR that added a config option to always break `|>`.

I'm more hesitant about changing the default, since it isn't clear to me why current problems with `|>` won't be equally applicable to user-defined operators (such as those defined and used extensively in ocamlformat itself). It would be better to see examples of hard to read formatting and fix the issue in general. Note that I think there are two distinct situations, one where the current output is "ok" but unconventional or otherwise not the desired style, in which case adding a config would be preferred. In the other case, where the current output is objectively hard to read, it would be better to fix the underlying problem.

lpil:
I agree with your reasoning.

I'm new to OCaml but interested in contributing this feature if I find time. If I open a PR with a first iteration would you be happy to provide feedback and guidance? I'm sure my code isn't very idiomatic yet :)

Thanks,
Louis

jberdine:
I'm very happy to give feedback, etc. on PRs! One thing to keep in mind, at least initially, is that it's usually best to err on the side of submitting too early and too often. :-) This is especially important since there is essentially no test suite.

lpil:
Oh dear! I feel a little less brave now

jberdine:
[...]

PR #150, opened 2018-04-18
Align (all) infix operators (if break)

hhugo:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/69.

I personally find the code more readable this way. Feedback welcome.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -147,6 +147,22 @@ let escape_strings =
           default
       & info ["escape-strings"] ~doc ~env)
 
+let break_infix =
+  let doc =
+    "Break sequence of infix operators. Can be set in a config file with a \
+     `break-infix {wrap,fit-or-vertical}` line. `wrap` will group simple \
+     expressions and try to format them in a single line."
+  in
+  let env = Arg.env_var "OCAMLFORMAT_BREAK_INFIX" in
+  let default = `Wrap in
+  mk ~default
+    Arg.(
+      value
+      & opt
+          (enum [("wrap", `Wrap); ("fit-or-vertical", `Fit_or_vertical)])
+          default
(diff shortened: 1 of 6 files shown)


jberdine:
Motivated by this, I implemented the argument grouping heuristic used for general applications also for infix op arguments, see https://github.com/ocaml-ppx/ocamlformat/commit/cf82be1837727e37dc08f6887ea8ee00976efc4c. That change gives an intermediate between the previous behavior and the "always break" behavior of this PR. Can I ask you to try https://github.com/ocaml-ppx/ocamlformat/commit/cf82be1837727e37dc08f6887ea8ee00976efc4c and see if you still think the always break approach is needed?

hhugo:
I had a look and I still prefer breaking always. I find (1) much easier to read.
Should we add a config ?

(1)
```
let sum_of_squares num =
  num + 1
  |> List.range 0
  |> List.map ~f:square
  |> List.fold_left ~init:0 ~f:( + )
```
(2)
```
let sum_of_squares num =
  num + 1 |> List.range 0 |> List.map ~f:square
  |> List.fold_left ~init:0 ~f:( + )
```

jberdine:
Ok, yes, lets add an option. Independently, I wonder if `|>` is something of a special case, or at least wonder if there are two classes of infix ops: those that are primarily about control flow and those that are primarily about value computation. My impression is that for e.g. arithmetic, there is much less desire to always break. Thoughts?

hhugo:
I'm not happy the option name, any suggestion ?

I had the same impression initially but I didn't manage to identify these two classes.

jberdine:
For the name, perhaps `--break-infix=wrap` or `--break-infix=single-line-or-vertical` or perhaps `fit` instead of `single-line`?

PR #151, opened 2018-04-18
Improve: more parens in patterns

hhugo:
Put parens around complex pattern on the lhs of let binding.

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -764,9 +764,6 @@ end = struct
       , ( Ppat_construct (_, Some _)
         | Ppat_variant (_, Some _)
         | Ppat_or _ | Ppat_alias _ ) ) ->
-        (* let (Some p) = e *)
-        (* let (P1 | P2) = e *)
-        (* let (pat as ident) = e *)
         true
     | Exp {pexp_desc= Pexp_let (_, bindings, _)}, Ppat_tuple _ ->
         List.exists bindings ~f:(function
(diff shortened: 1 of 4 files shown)


jberdine:
Motivation question: would it make sense to add these parens only if the pattern does not fit on one line, or you think its better even for the one-liners?

hhugo:
I personally like it better even for one-liners.
One of the motivation is to make the difference with function declaration more apparent
```
let f x = e
let `f x = e
let F x = e
```

jberdine:
Ok, I see. The point about function definitions never occurred to me because colors:
image

src/Ast.ml
```diff
@@ -749,6 +749,14 @@ end = struct
     | ( Pat {ppat_desc= Ppat_construct _}
       , Ppat_construct ({txt= Lident "::"}, _) ) ->
         true
+    | ( (Exp {pexp_desc= Pexp_let _; _} | Str {pstr_desc= Pstr_value _; _})
+      , ( Ppat_construct (_, Some _)
+        | Ppat_variant (_, Some _)
+        | Ppat_or _ | Ppat_alias _ ) ) ->
+        (* let (Some p) = e *)
+        (* let (P1 | P2) = e *)
+        (* let (pat as ident) = e *)
```
jberdine: Nit: Do you think that it is valuable to have only a few such comments indicating the concrete syntax corresponding to the parsetree constructors? I was basically assuming that anyone reading to code would either be somewhat familiar with the parsetree, or willing to read it as they went.

hhugo: It was only meaningful internally to help understand what diffs can be upstream. Now removed.

jberdine:
Ok, I'm not really convinced about the cases except for Ppat_alias, but it's not very invasive so lets see if there are complaints and then we can see-saw back, add an option, etc.

PR #153, opened 2018-04-18
Improve: allow to customize position of doc comments

hhugo:


jberdine:
I need to do some testing. The last time I tried this, there were many problems with formatted code triggering warning 50 (docstring attachment).

src/Conf.ml
```diff
@@ -93,6 +93,18 @@ let debug =
   let default = false in
   mk ~default Arg.(value & flag & info ["g"; "debug"] ~doc ~env)
 
+let doc_comments_position =
```
jberdine: Minor point, but note that the existing options use the same name (modulo '_' vs '-') for the ref set by parsing, the field name of Conf.t, and the command line option itself.

PR #154, opened 2018-04-18
Improve: formatting for multiple attributes

hhugo:
Multiple attributes not fitting on a single line will format as one per line.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -481,7 +481,7 @@ and fmt_extension c ctx key (({txt} as ext), pld) =
       fmt_structure_item c ~sep:"" ~last:true ~ext (sub_str ~ctx si)
   | _ -> wrap "[" "]" (str key $ str txt $ fmt_payload c ctx pld)
 
-and fmt_attributes c pre ?(box= true) ~key attrs suf =
+and fmt_attributes c ?(pre= fmt "") ?(box= true) ~key attrs suf =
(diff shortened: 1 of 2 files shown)


src/Fmt_ast.ml
```diff
@@ -1627,7 +1632,7 @@ and fmt_value_description c ctx vd =
         $ fmt " :@ " $ fmt_core_type c (sub_typ ~ctx pval_type)
         $ list_fl pval_prim (fun ~first ~last:_ s ->
               fmt_if first "@ =" $ fmt " \"" $ str s $ fmt "\"" ) )
-    $ fmt_attributes c (fmt "@;<2 2>") ~key:"@@" atrs (fmt "")
+    $ fmt_attributes c (fmt "@;<2 2>") ~box:false ~key:"@@" atrs (fmt "")
```
jberdine: Instead of removing the box, I suspect that `fmt "@;<2 2>"` should be evaluated before `fmt_attributes`:
```
$ fmt "@;<2 2>" $ fmt_attributes c (fmt "") ~key:"@@" atrs (fmt "")
```
(and I guess the `pre` arg of `fmt_attributes` ought to be optional)

hhugo: It causes issue with the docstring bellow. Do you mind giving it a try ?

jberdine: Sorry, you're right, what you had is better.

PR #155, opened 2018-04-18
Improve: alternative format for ite construct

hhugo:
This will probably not reach consensus, I'll put that change under a config.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -147,6 +147,21 @@ let escape_strings =
           default
       & info ["escape-strings"] ~doc ~env)
 
+let if_then_else =
+  let doc =
+    "If-then-else formatting. Can be set in a config file with an \
+     `if-then-else {keyword-first,compact}` line."
+  in
+  let env = Arg.env_var "OCAMLFORMAT_IF_THEN_ELSE" in
+  let default = `Compact in
+  mk ~default
+    Arg.(
+      value
+      & opt
+          (enum [("keyword-first", `Keyword_first); ("compact", `Compact)])
+          default
+      & info ["if-then-else"] ~doc ~env)
+
 let inplace =
   let doc = "Format in-place, overwriting input file(s)." in
   let default = false in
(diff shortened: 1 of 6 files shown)


jberdine:
Yes, I think a config for this one is a good idea.

src/Conf.ml
```diff
@@ -134,6 +134,19 @@ let escape_strings =
           default
       & info ["escape-strings"] ~doc ~env)
 
+let ifthenelse =
+  let doc =
+    "Format for if-then-else. Can be set in a config file with an \
+     `if-then-else {kw-first,compact}` line."
```
jberdine: I'm not sure it is worth it to abbreviate keyword to kw, but don't feel strongly either way.

src/Fmt_ast.ml
```diff
@@ -1337,24 +1337,45 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
...
+                           ( ( if first then
+                                 fmt "if" $ fmt_extension_suffix c ext
+                             else fmt "else if" )
+                           $ str " " $ fmt_expression c xcnd )
+                         $ fmt "@ "
+                     | None -> fmt "" )
+                     $ hvbox 2
+                         ( ( match xcnd with
+                           | Some _ -> fmt "then"
+                           | None -> fmt "else" )
```
jberdine: Likewise, this `match` could be written with `fmt_or` if you prefer.

src/Fmt_ast.ml
```diff
@@ -1337,24 +1337,45 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
...
+                           )
+                       $ fmt_if parens " )" )
+                     $ fmt_if (not last) "@ "
+                 | `Kw_first ->
+                     ( match xcnd with
+                     | Some xcnd ->
+                         hvbox 2
+                           ( ( if first then
+                                 fmt "if" $ fmt_extension_suffix c ext
+                             else fmt "else if" )
```
jberdine: Just FYI, this if-then-else could be written using `fmt_or_k` if you prefer.

test/passing/ite_kw_first.ml.opts
```diff
@@ -0,0 +1 @@
+--if-then-else kw-first
```
jberdine: Nice capability

PR #156, opened 2018-04-19
Fix: parens around unpack module

hhugo:


hhugo:
close in favour of https://github.com/ocaml-ppx/ocamlformat/pull/160

PR #157, opened 2018-04-19
Improve: less parens for tuples

hhugo:


hhugo:
note that this PR is based on top of https://github.com/ocaml-ppx/ocamlformat/pull/149

src/Fmt_ast.ml
```diff
@@ -1502,15 +1501,19 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
             $ fmt_expression c (sub_exp ~ctx e2) )
         $ fmt_atrs )
   | Pexp_tuple es ->
-      let parens =
+      let force_no_parens =
         match xexp.ctx with
-        | Str {pstr_desc= Pstr_eval _} -> false
-        | _ -> true
```
jberdine: I'm nervous about "manually" forcing no parens here. I think that the effect of not emitting discretionary/redundant parens around tuple expressions can be achieved by changing this `true` to `parens`. Probably a few little things also need to be sorted out (like `assoc_of_prec` should be `Non` on `Comma` I expect), but doing it this way would use the precedence-based parenthesization algorithm uniformly with the rest of the code. So far the approach has been to (try to) systematically ensure that the parenthesization functions in Ast return safe results, which are sometimes overridden in Fmt_ast to add _more_ parens in ad hoc ways.

hhugo:
I will add a switch in the config

src/Fmt_ast.ml
```diff
@@ -1424,8 +1424,7 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
       let opn, cls =
         let can_skip_parens =
           match e0.pexp_desc with
-          | Pexp_array _ | Pexp_constraint _ | Pexp_record _ | Pexp_tuple _ ->
-              true
+          | Pexp_array _ | Pexp_constraint _ | Pexp_record _ -> true
```
jberdine: I'm not sure about this change, shouldn't the two modes behave the same in this case?

src/Fmt_ast.ml
```diff
@@ -1498,6 +1497,11 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
           $ Option.call ~f:epi )
       $ fmt_atrs
   | Pexp_record (flds, default) ->
+      let field_alias (li1: Longident.t) (li2: Longident.t) =
+        match (li1, li2) with
+        | Ldot (_, x), Lident y -> String.equal x y
+        | _ -> Poly.equal li1 li2
+      in
```
jberdine: Redundant with the definition above?

src/Fmt_ast.ml
```diff
@@ -1553,15 +1557,21 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext
...
         match xexp.ctx with
-        | Str {pstr_desc= Pstr_eval _} -> false
-        | _ -> true
+        | Exp {pexp_desc= Pexp_extension _} -> true
+        | Pld _ -> true
+        | Str {pstr_desc= Pstr_eval _} -> true
+        | _ -> false
+      in
+      let wrap =
+        if parens || Poly.(c.Conf.parens_tuple = `Always) then
```
jberdine: Nit: other functions have a type constraint on the `Conf.t` argument when needed and use type-based disambiguation.

jberdine: This changes the behavior in `Always` mode, before the `let parens` disables them for `Pstr_eval`.

src/Fmt_ast.ml
hhugo: With can only skip Pexp_tuple when parens_tuple = `Always

PR #159, opened 2018-04-20
Improve: keep the original string escaping

### LINKED ISSUE #10: Preserve original escape sequences in strings

hhugo:
Fixes #10

### DIFF
--- a/src/Source.ml
+++ b/src/Source.ml
@@ -1,0 +1,73 @@
+(** Concrete syntax, set by [init]. *)
+let source = ref ""
+
+let init s = source := s
+
+let string_between (l1: Location.t) (l2: Location.t) =
+  let pos = l1.loc_end.pos_cnum + 1 in
+  let len = l2.loc_start.pos_cnum - l1.loc_end.pos_cnum - 1 in
+  if len >= 0 then Some (String.sub !source ~pos ~len)
+  else
+    (* can happen e.g. if comment is within a parenthesized expression *)
+    None
+
+let normalize_string mode s =
+  let lexbuf = Lexing.from_string s in
+  let rec loop () =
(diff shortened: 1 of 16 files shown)


src/Source.ml
```diff
@@ -0,0 +1,73 @@
...
+  let s = String.sub s start (stop - start) in
+  let s = Literal_lexer.char (Lexing.from_string s) in
+  String.sub s 1 (String.length s - 2)
+
+let string_at (l: Location.t) =
+  let pos = l.loc_start.pos_cnum in
+  let len = l.loc_end.pos_cnum - pos in
+  String.sub !source ~pos ~len
+
+let string_litteral mode (l: Location.t) =
```
jberdine: s/litteral/literal/g

src/Translation_unit.ml
```diff
@@ -79,7 +79,8 @@ let parse_print (XUnit xunit) (conf: Conf.t) ~input_name:iname
     in
     if Conf.debug then
       dump xunit dir base ".old" ".ast" (xunit.normalize (ast, cmts)) ;
-    xunit.init_cmts source ast cmts ;
+    Source.init source ;
+    xunit.init_cmts ast cmts ;
```
jberdine: It would be good if the implicit dependence on the order of initialization was either hidden (can Cmts call Source.init) or at least resulted in a reasonable exception if gotten wrong.

jberdine:
Excellent, thanks a lot!

PR #161, opened 2018-04-23
Improve: align arrow type args and do not wrap them

jberdine:
Format all arrow types that do no fit on one line with vertically aligned `->`s.


hhugo:
I'm in favour of merging this one.

PR #162, opened 2018-04-23
Add: option to attempt to indent the same as ocp-indent

jberdine:
Based on #161, add an option to branch on to special case formatting for post-processing with ocp-indent. This PR only changes the position of the `:` in value declaration signature items, but there are likely to be more special cases.


jberdine:
I expect that this is suboptimal, but will merge this for now and the deficiencies can be fixed in other PRs as they are encountered.

hhugo:
does this PR actually changes the position of the `:` ? I dont see any diff when I reformat the code of ocamlformat..

hhugo:
ignore me

PR #171, opened 2018-04-29
`>]` is invalid.

hhugo:
- `[%ext: < .. >]`
- ```type t = [`A of <a; b>]]```

I've tried a few fix but I'm not happy. Do you have any suggestion ?

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -225,6 +225,12 @@ and Requires_sub_terms : sig
 
   val exposed : cls -> expression -> bool
 
+  type ltgt = LT | GT
+
+  val typ_exposed_left : ltgt -> core_type -> bool
+
+  val typ_exposed_right : ltgt -> core_type -> bool
+
   val prec_ast : T.t -> prec option
 
   val parenze_typ : core_type In_ctx.xt -> bool
@@ -971,6 +977,20 @@ end = struct
           | _ ->
               if is_right_infix_arg pexp_desc exp then is_sequence exp
               else exposed Non_apply exp )
+    | _ -> false
+
+  type ltgt = LT | GT
+
+  let rec typ_exposed_left c typ =
+    match (typ.ptyp_desc, c) with
+    | Ptyp_arrow (_, t, _), _ -> typ_exposed_left c t
+    | Ptyp_object _, LT -> true
+    | _ -> false
+
+  let rec typ_exposed_right c typ =
+    match (typ.ptyp_desc, c) with
+    | Ptyp_arrow (_, _, t), _ -> typ_exposed_right c t
+    | Ptyp_object _, GT -> true
     | _ -> false
 end
 
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -223,13 +223,11 @@ and Requires_sub_terms : sig
 
   type cls = Let_match | Match | Non_apply | Sequence | Then | ThenElse
 
-  val exposed : cls -> expression -> bool
+  val exposed_right_exp : cls -> expression -> bool
 
-  type ltgt = LT | GT
+  val exposed_left_typ : core_type -> bool
 
-  val typ_exposed_left : ltgt -> core_type -> bool
-
-  val typ_exposed_right : ltgt -> core_type -> bool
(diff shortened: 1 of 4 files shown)


hhugo:
What you describe matches my understanding.
I went ahead and implement a (surely partial) hack.

src/Fmt_ast.ml
```diff
@@ -56,6 +56,15 @@ let rec sugar_or_pat c ({ast= pat} as xpat) =
...
   | _ -> [xpat]
 
+let rec protect_token c = function
+  | Typ typ -> (
+    match (typ.ptyp_desc, c) with
+    | Ptyp_arrow (_, _, t2), _ -> protect_token c (Typ t2)
+    | Ptyp_object _, (']' | '}') -> true
+    | _ -> false )
+  | Pld (PTyp t) -> protect_token c (Typ t)
+  | _ -> false
```
jberdine: This function has a similar functionality to `exposed` in `Ast`, but for types instead of expressions. It might make sense to move it there.

src/Fmt_ast.ml
```diff
@@ -589,6 +601,16 @@ and fmt_core_type c ?(box= true) ?pro ({ast= typ} as xtyp) =
   | Ptyp_var s -> fmt "'" $ str s
   | Ptyp_variant (rfs, flag, lbls) ->
       let row_fields rfs = list rfs "@ | " (fmt_row_field c ctx) in
+      let protect_token =
+        match (lbls, rfs) with
+        | _, [] -> impossible "not produced by parser"
```
jberdine: I seem to hit this case with
```
let ( = ) (v1: [> ]) (v2: [> ]) = Core.Polymorphic_compare.( = ) v1 v2
```

src/Ast.ml
```diff
@@ -225,6 +225,12 @@ and Requires_sub_terms : sig
 
   val exposed : cls -> expression -> bool
 
+  type ltgt = LT | GT
+
+  val typ_exposed_left : ltgt -> core_type -> bool
```
jberdine: nit: the other functions, e.g. parenze_*, have "type" suffixes rather than prefixes

src/Ast.ml
```diff
@@ -225,6 +225,12 @@ and Requires_sub_terms : sig
...
+
+  val typ_exposed_right : ltgt -> core_type -> bool
```
jberdine: For my understanding, is it the case that currently there is some redundancy since exposed_left is always concerned with LT and ignores GT, and similarly exposed_right is always concerned with GT and ignores LT? Do you have a generalization in mind that will be needed, or would it make sense to add the most specialized version for now and generalize later when it becomes clear what will be needed?

src/Fmt_ast.ml
```diff
@@ -1808,7 +1830,11 @@ and fmt_type_declaration c ?(pre= "") ?(suf= ("" : _ format)) ?(brk= suf)
         $ fmt "@ "
         $ hvbox 0
             (wrap_fits_breaks "{" "}"
-               (list lbl_decls "@,; " (fmt_label_declaration c ctx)))
+               (list_fl lbl_decls (fun ~first ~last x ->
+                    fmt_if_k (not first) (fmt "@,; ")
```
jberdine: could write `fmt_if (not first) "@,; "` if you want

src/Fmt_ast.ml
```diff
@@ -599,12 +620,12 @@ and fmt_core_type c ?(box= true) ?pro ({ast= typ} as xtyp) =
...
               $ list ls "@ " (fmt "`" >$ str)
           | Open, Some _ -> impossible "not produced by parser" )
-        $ fits_breaks "]" "@ ]" )
+        $ fits_breaks (if protect_token then " ]" else "]") "@ ]" )
   | Ptyp_object ([], Open) -> fmt "< .. >"
   | Ptyp_object ([], Closed) -> fmt "< >"
   | Ptyp_object (fields, closedness) ->
       hvbox 0
-        (wrap_fits_breaks "<" ">"
+        (wrap "< " " >"
```
jberdine: With this, can probably change `"@ ; .. "` below to `"@ ; .."`

PR #172, opened 2018-04-30
Various changes

hhugo:
Based on top of https://github.com/ocaml-ppx/ocamlformat/pull/171
- add parens around nested constraint
- add parens around constraint inside aliases
- improve formatting of `new c x y`
- add parens around constraint in exception pat
- add parens around nested ( :: ) in pats
- fix extensible gadt
- fix attribute in let bindings
- fix [S with type lident = ty]
- fix attribute on polymorphic variant
- fix private on extensible variants


hhugo:
Dont merge yet. I wanna add regtests

PR #181, opened 2018-05-06
Improve: semisemi at the end

hhugo:
It's unclear to me whether the current behaviour was intended or not.


src/Conf.ml
```diff
@@ -321,8 +321,9 @@ let validate () =
   else if not !inplace && List.length !inputs > 1 then
     `Error (false, "Must specify only one input file without --inplace")
   else `Ok ()
+;;
 
-;; parse info validate
+parse info validate
```
jberdine: The current behavior is actually intended. :-) A while ago I found that I was not even the only one who preferred this style, see https://inbox.ocaml.org/caml-list/55885170.9070501@cis.upenn.edu/ .

How would you feel about formatting this  as:
```
  else `Ok ()

;;
parse info validate
```

The trouble I have with the output of this PR currently is that it leads to people asking why some `let` bindings have `;;` terminators and others do not, etc. I have found that prefixing module-level expression evaluations with `;;` completely eliminates this confusion, and also highlights potentially side-effecting computations (here the `;;` token is colored with maximum contrast).

hhugo: [%%ext 1;;2;;3] is currently broken

hhugo: I remember now.
At janestreet, we put `;;` at the end of multiple structure item. We do so for multiple reasons that @mbarbin might remember better than I do.

For example,
```
 let is_prefix exp =
   match exp.pexp_desc with
   | Pexp_ident {txt= Lident i} -> is_prefix_id i
   | _ -> false
;;

;; assert (is_prefix e) 
```
The formatting looks good to me and is valid ocaml. However, it fails with `ocamldep`. I've just opened a mantis issue ( https://caml.inria.fr/mantis/view.php?id=7793 )

hhugo:
I'll close this PR. I'm happy with the current format. We just need a fix in ocamldep

src/Conf.ml
jberdine: I think it would also be ok to uniformly separate structure items by `;;`, but that seems to be a style that has fallen out of fashion. IIUC, it does sometimes lead to better type error messages though.

Issue #184, opened 2018-05-06
missing `;;` for structure items in extensions


hhugo:
``
[%%ext 1;;2;;3]
```

PR #185, opened 2018-05-06
Fix: extensions with multiple module-level eval expressions

jberdine:


test/passing/extensions.ml
```diff
@@ -23,3 +23,10 @@ let invariant t =
        , xxxxxxxxxxxxxx
        , xxxxxxxxxxx
        , xxxxxxxxxxxxxxxxxxxx ) when a < b]
+
+[%%ext
+;; 11111111111111111111
+
+;; 22222222222222222222
+
+;; 33333333333333333333]
```
jberdine: @hhugo, does this fix the problematic case you ran into with #181 ?

hhugo: I believe so

jberdine:
should fix #184

test/passing/extensions.ml
hhugo: Could we special case when there is one item only?

jberdine: #195 ?

PR #186, opened 2018-05-06
Build: switch to dune, use `dev` and `release` instead of `dbg` and `opt`

jberdine:
Preview of simplifications enabled by ocaml/dune#419


hhugo:
lgtm

PR #187, opened 2018-05-06
Build: do not test if bisect_ppx is available

jberdine:


jberdine:
If `make` does not try to build the `coverage` context, then is it still necessary to conditionally add the `preprocess` clause? It seems better to fail if someone executes `make coverage` and `bisect_ppx` is not installed.

jberdine:
@hhugo, would this way of building work in your setting?

hhugo:
works for me

PR #191, opened 2018-05-07
Yet another batch of fixes

hhugo:


hhugo:
tests  should pass once rebased on top of other pending PRs

test/passing/shortcut_ext_attr.ml
```diff
@@ -0,0 +1,147 @@
+(* TEST flags = "-dparsetree" ocamlc_byte_exit_status = "2" *
+   setup-ocamlc.byte-build-env ** ocamlc.byte *** check-ocamlc.byte-output *)
```
jberdine: ?

Re `-dparsetres`, note that `ocamlformat -g test.ml -o test.fmt.ml` will dump `test.old.ast` (and `test.new.ast` if different) using `Pprintast` (but after being normalized).

hhugo: This file was taken untouched from the OCaml testsuite.

hhugo:
note that with that PR, most(1) of the valid(2) sources from the OCaml repo format correctly(3).

(1) all but files under ocamldoc/
(2) files that parse without triggering warning 50.
(3) ocamlformat run without error.

PR #200, opened 2018-05-22
revert: Improve: remove redundant parens around application operators

hhugo:


test/passing/apply.ml
```diff
@@ -0,0 +1,2 @@
+(* parens have the effect of erasing optional arguments *)
+let _ = (f x y) z
```
jberdine: Just for my understanding, do you have a more detailed test which type-checks with the parens but does not without them?

hhugo: I do not fully understand what's going on but here is one: 
```
# let id x = x;;
val id : 'a -> 'a = <fun>
# let plus a ?(b=0) c = a + b + c;;
val plus : int -> ?b:int -> int -> int = <fun>
# id (plus 1);;
- : ?b:int -> int -> int = <fun>
# id (plus 1) 1;;
- : int = 2
# id (plus 1) ~b:1;;
Error: This expression has type ?b:int -> int -> int
       but an expression was expected of type b:'a -> 'b
# (id (plus 1)) ~b:1;;
- : int -> int = <fun>
```

jberdine: Thanks! I asked about this on [mantis](https://caml.inria.fr/mantis/view.php?id=7832).

jberdine:
rebased and added @hhugo's detailed test

PR #218, opened 2018-06-24
Improve: support for toplevel files

hhugo:


src/Migrate_ast.ml
```diff
@@ -40,6 +42,23 @@ module Printast = struct
...
+              let open Parsetree in
+              let b =
+                match b with
+                | Pdir_none -> Pdir_none
+                | Pdir_string s -> Pdir_string s
+                | Pdir_int (s, c) -> Pdir_int (s, c)
+                | Pdir_ident i -> Pdir_ident i
+                | Pdir_bool b -> Pdir_bool b
+              in
+              Ptop_dir (a, b) ) )
```
jberdine: Compilation fails here when the current compiler version is different from the one used internally (currently 406). It seems we need `to_current.copy_directive_argument`.

jberdine: Fixed in master now.

PR #220, opened 2018-06-30
switch to OCaml 4.07

hhugo:


jberdine:
@hhugo, please look this over, I rebased and refactored a dependency on the ast version

hhugo:
Lgtm

Issue #213, opened 2018-06-21
Unattached documentation command (ignored)


mseri:
The following snippet
```ocaml
type blocklist =
  { blocksize: int [@version 1, 1, 0] (** Size of the individual blocks *)
  ; ranges: (int64 * int64) list
    (** list of block ranges, where a range is
        a (start,length) pair,measured in units
        of [blocksize] *)
  }
[@@deriving rpcty]
```
results in the following failure
```
$ git ls-files '*.ml*' | xargs ocamlformat -i
File "example/example3_idlfc3a3a.ml", line 24, characters 20-56:
Error (warning 50): unattached documentation comment (ignored)
make: *** [reformat] Error 1
```

The snipped in `example3_idlfc3a3a.ml` is the following
```ocaml
type blocklist =
  { blocksize: int  (** Size of the individual blocks *) [@version 1, 1, 0]
  ; ranges: (int64 * int64) list
        (** list of block ranges, where a range is
        a (start,length) pair,measured in units
        of [blocksize] *)
  }
[@@deriving rpcty]
```

PR #221, opened 2018-07-01
Fix: attributes and doc comments

hhugo:


hhugo:
fix #213

jberdine:
@hhugo, please look this over, I rebased and reverted the cosmetic change that is independent of the fix. Was that also intentionally changed? If so, can we discuss that change in a separate PR?

hhugo:
I was not sure the double space was intended. Happy to merge without that diff.

PR #227, opened 2018-07-10
Improve: no space around parens for module unpack

hhugo:


jberdine:
@hhugo, what do you think about #230 instead?

hhugo:
Go for it

PR #241, opened 2018-07-30
Fix: comments in empty (), []

hhugo:
formatting would stabilise otherwise

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -773,8 +773,8 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
         let opn = txt.[0] and cls = txt.[1] in
         Cmts.fmt c.cmts loc
           (hvbox 0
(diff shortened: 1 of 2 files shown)


src/Fmt_ast.ml
```diff
@@ -773,8 +773,10 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
         let opn = txt.[0] and cls = txt.[1] in
         Cmts.fmt c.cmts loc
           (hvbox 0
-             (wrap_fits_breaks (String.of_char opn) (String.of_char cls)
-                (Cmts.fmt_within c.cmts ~pro:(fmt "@ ") ~epi:(fmt "@ ")
+             (wrap_k
+                (fun f -> Format.fprintf f "%c" opn)
+                (fun f -> Format.fprintf f "%c" cls)
```
jberdine: Could use `Fmt.char` here

PR #244, opened 2018-07-30
Fix: fix string_literal when location include the attribute.

hhugo:


src/Source.ml
```diff
@@ -49,23 +52,54 @@ let tokens_at t ?(filter= fun _ -> true) (l: Location.t) :
   loop []
 
 let string_literal t mode (l: Location.t) =
+  (* the location of a [string] might include surrounding comments and
+     attributes because of [reloc_{exp,pat}] and a [string] can be found in
+     attributes payloads. {[ f ((* comments *) "c" [@attributes]) ]} *)
```
jberdine: Does this make you think that the `reloc` functions are themselves broken?

hhugo: It's unclear to me.

PR #246, opened 2018-07-30
Improve comments formatting

gpetiot:
Some improvements on the formatting of comments.
Some comments were attached to the wrong piece of code (next one instead of previous one), and the formatting added unnecessary linebreaks.
A new regression test has been added: passing/comments_in_types.ml

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -1239,7 +1239,7 @@ end = struct
     prec_ast ast
     >>| fun prec_ast ->
     let cmp = Poly.compare prec_ctx prec_ast in
-    if cmp < 0 then (* ast higher precedence than context: no parens *)
+    if cmp < 0 (* ast higher precedence than context: no parens *) then
       false
     else if cmp > 0 then (* context higher prec than ast: add parens *)
       true
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -300,12 +300,17 @@ end
 let is_adjacent t (l1 : Location.t) (l2 : Location.t) =
   Option.value_map (Source.string_between t.source l1 l2) ~default:false ~f:
     (fun btw ->
-      match String.strip btw with
-      | "" -> true
-      | "|" ->
-          Source.begins_line t.source l1
-          && Position.column l1.loc_start <= Position.column l2.loc_start
-      | _ -> false )
+      let i_cr = String.index btw '\n' in
+      let j_cr = String.rindex btw '\n' in
+      match (i_cr, j_cr) with
+      | Some i, Some j when i <> j -> false
+      | _ -> (
+        match String.strip btw with
+        | "" -> true
+        | "|" ->
+            Source.begins_line t.source l1
+            && Position.column l1.loc_start <= Position.column l2.loc_start
+        | _ -> false ) )
 
 (** Heuristic to choose between placing a comment after the previous loc or
     before the next loc. Places comment after prev loc only if they are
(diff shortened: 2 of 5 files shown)


src/Cmts.ml
```diff
@@ -309,10 +314,39 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
           not (is_adjacent t l1 l2) )
     with
-    | [cmtl] when is_adjacent t (snd (List.last_exn cmtl)) next ->
+    | [cmtl]
+      when is_adjacent t (snd (List.last_exn cmtl)) next
+           && not (is_adjacent t prev loc) ->
         (CmtSet.empty, cmts)
     | after :: befores ->
-        (CmtSet.of_list after, CmtSet.of_list (List.concat befores))
+        let loc_prev = prev.Warnings.loc_start in
```
hhugo: `prev.loc_start` or `prev.Location.loc_start`

src/Cmts.ml
```diff
@@ -309,10 +314,39 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
+        else if
+          loc_cmt.pos_lnum - loc_prev_end.pos_lnum
+          < loc_next.pos_lnum - loc_cmt_end.pos_lnum
+        then (CmtSet.of_list after, CmtSet.of_list (List.concat befores))
+        else if
+          is_adjacent t prev loc
+          && is_adjacent t (snd (List.last_exn cmtl)) next
+          && loc_prev_end.pos_lnum = loc_cmt.pos_lnum
+          && loc_cmt_end.pos_lnum = loc_next.pos_lnum
+        then (CmtSet.of_list after, CmtSet.of_list (List.concat befores))
```
hhugo: you have  5 branches returning `(CmtSet.of_list after, CmtSet.of_list (List.concat befores))`, any way you can simply this a bit ? or at least have comments describing cases handle by each branch.

src/Fmt_ast.ml
```diff
@@ -524,17 +524,18 @@ let rec fmt_attribute c pre = function
...
   | PStr [({pstr_desc= Pstr_value _; _} as si)], (Pld _ | Str _ | Top) ->
       fmt_structure_item c ~sep:"" ~last:true ~ext (sub_str ~ctx si)
   | _ ->
       let protect_token =
         match pld with PTyp t -> exposed_right_typ t | _ -> false
       in
-      wrap "[" "]"
-        ( str key $ str txt $ fmt_payload c ctx pld
-        $ fmt_if protect_token " " )
+      Cmts.fmt c.cmts loc
```
hhugo: can you add a regression test for this ?

test/passing/comments_in_types.ml
```diff
@@ -0,0 +1,63 @@
...
+  s    (* correct, because this doesn't look like a constructor to ocp-indent *)
+type t =
+    S                         (* correctly indented a little more, because... *)
+type t =
+  | S                                   (* we leave room for the vertical bar *)
+
+
+
+(* analogous value expressions, analogous to lists, some different from now *)
+let _ =
```
hhugo: Why is this test here ?

Makefile
```diff
@@ -42,7 +42,7 @@ reason: setup
 
 .PHONY: clean cleanbisect
 clean: cleanbisect
-	rm -rf _build dune-workspace
+	rm -rf _build dune-workspace src/*[0-9]*.ml
```
jberdine: I am uncomfortable with such wildcards in arguments to rm. It would be easy to add a source file that this would delete.

src/Ast.ml
```diff
@@ -1218,10 +1218,9 @@ end = struct
     prec_ast ast
     >>| fun prec_ast ->
     let cmp = Poly.compare prec_ctx prec_ast in
-    if cmp < 0 then (* ast higher precedence than context: no parens *)
+    if cmp < 0 (* ast higher precedence than context: no parens *) then
       false
-    else if cmp > 0 then (* context higher prec than ast: add parens *)
-      true
+    else if cmp > 0 then (* context higher prec than ast: add parens *) true
```
jberdine: These two changes seem to indicate that something isn't quite right.

test/passing/comments_in_types.ml
gpetiot: Originally this file comes from ocp-indent regtests (https://github.com/OCamlPro/ocp-indent/blob/rewind/tests/passing/js-type.ml). Before my changes the comments « we leave room ... » and « analogous ... » were stuck together and linked to the first list.
Maybe I should only keep the first list and remove the others.

Makefile
```diff
@@ -42,7 +42,7 @@ reason: setup
 
 .PHONY: clean cleanbisect
 clean: cleanbisect
-	rm -rf _build dune-workspace src/*[0-9]*.ml
+	rm -rf _build dune-workspace
```
gpetiot: Don't know why this commit broke the CI build

jberdine: Did you locally merge master into your branch? I suspect there is an interaction between this and #244.

test/passing/comments_in_types.ml
jberdine: I think that it would be easier to understand the intent of the tests going forward if the comments were changed to indicate the expected formatting, such as indicating to which syntactic phrase each comment is intended to refer.

src/Ast.ml
jberdine: I think this is a consequence of the issue fixed in #253. I locally rebased this PR on #253 and reverted the change to Ast.ml and then it reformats without change.

gpetiot:
(I will reboot this PR).
I am wondering if the current hashtables for comments are enough. Suppose we are formatting this:
```
let x = 42 (* comment on 42 *)
let x = 43 (* comment on 43 *)
let z = 44
```

The current formatting (on master) is:
```
let x = 42

(* comment on 42 *)
let x = 43

(* comment on 43 *)
let z = 44
```

I guess a nice output would be:
```
(* comment on 42 *)
let x = 42

(* comment on 43 *)
let x = 43

let z = 44
```

There will be a blank line between the comments of cmts_before and the code. Maybe we need to have an additional hashtbl for Cmts.t, `cmts_just_before` or something, those comments would be printed before the corresponding code without a blank line.
The general formatting could be:

```
(* comments before *)

(* comments just before *)
some (* comments within *) code

(* comments after *)
```

Do you agree this change is necessary ? I don't think it could be fixed with only modifying the heuristic function that chooses between after_prev/before_curr.

jberdine:
It is unclear to me whether it is necessary to add the notion of "just 
before".

First, it has been a goal, though not fully achieved, for the comment 
attachment to avoid moving comments with respect to code. During 
development, it is very aggravating if commenting out some code and 
reformatting moves the code somewhere else so that it is then difficult 
to uncomment to the original state.

For
```
let x = 42 (* comment on 42 *)
```
I expect that the comment is attached to the following binding, not 42. 
If the comment was attached to 42, then the formatting might be 
acceptable. This change would involve changing the heuristic for 
choosing whether to attach a comment to the parsetree node before it 
versus after it, but would use the existing tables to store that choice.

Note that the current heuristic for this choice was implemented prior to 
the support for querying the concrete syntax token stream. It could be 
that using that, a significantly better heuristic could be defined. For 
example, it might work well to prefer before in case the token before is 
not a keyword or punctuation, while the token after is.

hhugo:
maybe rebase this one once https://github.com/ocaml-ppx/ocamlformat/pull/339 is merged.

Issue #12, opened 2017-10-20
Add option to always break pattern match cases


edwintorok:
Currently `ocamlformat` reformats this to a single line:
```ocaml
let f x = function
  | C -> 1
  | D -> 2
```

becomes:

`let f x = function C -> 1 | D -> 2`

This is nice for aesthetic reasons, but if this code is checked into version control then it'll suddenly get all reformatted to multiple lines whenever one of the expressions becomes longer, making diffs harder to check for correctness. The programmer might've foreseen the expression becoming more complicated and separated each case on a line of its own in advance.
OTOH it could be that there are only ever 2 values there (like for `'a result`), in which case using the multiline pattern matching would be wasteful and harder to read.

Would be nice if there was a configuration field to keep the original singleline vs multiline choice, provided that it can still meet the column width limit.

jberdine:
It would be interesting to have a mode which attempts to format code in a way that minimizes diff sizes. This has not been high on the list of constraints so far since there is a very strong tension with having a target margin that the formatter attempts to reach but not overshoot. The optimal choice for minimizing diff size is probably to ignore the margin.

Note that ocamlformat formats the abstract syntax, as opposed to tidying up the concrete syntax. This means that any two pieces of code that the parser interprets the same, must format the same. In other words, a goal of ocamlformat is to take control of laying out the concrete syntax away from the programmer.

edwintorok:
Minimizing diff size (ignoring whitespace though when computing the diff) could be an interesting way to gradually introduce `ocamlformat` in largish projects without too much initial churn.

jberdine:
AFAIU, it wouldn't help with the initial conversion unless the project had an extremely uniform style already, and it just happened to be the same as that produced by the auto-formatter. Note that ocamlformat takes the result of the parser and completely ignores the concrete syntax that was parsed. (Small lie, placement of comments may look at the concrete syntax in a limited way.)

vjeux:
For prettier, a JavaScript formatter, one thing I found is that there are patterns that you always want expanded, even if they fit in 80 columns.

For example function definition or class definitions.

```js
function f() {
  return true;
}
```

It seems like for function pattern matching, it's very unlikely you want to inline it. #my2cents

jberdine:
[...]

PR #251, opened 2018-07-31
Improve: add option --break-cases to break each pattern-matching case

gpetiot:
This change implements the feature suggested in issue #12


test/passing/break_cases.ml
```diff
@@ -0,0 +1,25 @@
+let f x = function
```
hhugo: can you add examples with or_pattern ?

src/Fmt_ast.ml
```diff
@@ -2269,11 +2269,14 @@ and fmt_cases c ctx cs =
             (fmt_or_k paren_body (fmt "@;<1 2>->") (fmt " ->@;<0 3>"))
             (fmt_or_k paren_body (fmt "@;<1 -2>-> (") (fmt " ->@;<0 -1>"))
         in
+        let pro =
+          fmt_or_k c.conf.break_cases
+            (break_unless_newline 100 0 $ fmt "| ")
```
jberdine: Elsewhere in the code, 1000 is used to force-break a box. It may be that 100 is close enough to a width someone will actually use that it could fit and then look very bad.

test/passing/break_cases.ml
```diff
@@ -0,0 +1,25 @@
...
+  | C ->
+      1
+  | D ->
+      2
```
samoht: I would have expected `--break-case` to format this into:

```ocaml
let f x = function
  | C -> 1
  | D -> 2
```

Also I am wondering if we can do something about alignements of `->` (not in scope for this PR), e.g:

```ocaml
let f x = function
  | C (x, _) -> x
  | D        -> 2
```

gpetiot: I agree it would look better, I don't know why the line breaks after the arrow but I will look into it.

jberdine: @samoht I think this is due to the way this patch interacts with the two prior modes. The current `--sparse` mode adds breaks after `->` in order to avoid the formatting of nested or-patterns which changes the indentation of `|` to indicate the nesting, which some people find odd. When nested or-patterns are not visually distinguished, there is a source of bugs for code that has a long sequence of nested or-patterns with some short body partway though. Without breaking after `->` or indicating the nesting structure with the indentation of `|`, it is very easy to visually mis-read the code, concluding that a later body covers an earlier pattern, when in fact there is a "hidden" body in between. Manually formatting code can add a break for the hidden body to avoid this, but it is harder with auto-formatting. Given the current implementation of ocamlformat in terms of the stdlib Format module, I don't think it is possible to force two non-nested boxes to either both break or both not break. Without that implementation possibility, I don't know of other formatting algorithms for code like this that avoids the hidden body danger. Other ideas?

Vertical alignment is tricky in general, at least when what appears to the left of the desired alignment point (e.g. the patterns before the `->`s) does not have a statically fixed length. To implement such alignment, I think it seems necessary to first format all the patterns, then compute their lengths and whether they broke over multiple lines, and then backtrack to format the entire list of cases. The question of whether each pattern fits on a line ought to be sensitive to the starting indentation, which isn't possible with the current design using Format I think. (Format does not support much backtracking, and doing so quickly makes the core algorithms exponential, which might be survivable, but maybe not.) Something rougher could be done instead, such as requiring the patterns to be short enough to fit in e.g. one third of the margin width in order to align them, and then just hope that the starting indentation does not exceed that in practice. I have been meaning to try that, but have not had time, and am not sure how optimistic to be about how well it might work.

test/passing/break_cases.ml
```diff
@@ -0,0 +1,28 @@
+let f x = function
+  | C
+  | P
+  | A
+  | K -> 1
```
jberdine: Here, imagine that the patterns above were long and complex, with nesting, parens, etc. In such cases spotting the `-> 1` can be very hard, leading to misreading the code thinking that the body for `D` also applies to e.g. `A`. I don't know how to avoid this danger without breaking after `->` or changing the indentation of `|` for the nested cases.

samoht: I don't think that's a problem in practice (at least I haven't met that case yet). It's rare to have long complex patterns and or clause (I think?). 

Also to be fair, I would tend to indent this function as:

```ocaml
let f x = function
  | A | B | C -> foo
  | D             -> bar
```

But I am not sure an automated tool can handle this case properly... (and I am to change my "code style" to something more standard)

jberdine: It could very well be a rare occurrence, but we have hit it enough that I would avoid using such a mode on our code.

Modulo alignment, it should be possible to get your formatting by force-breaking between the cases, but not between the nested or-patterns. I think it would make sense to force-break both the cases and or-patterns with `--sparse --break-cases`, and to force-break the cases but not the nested or-patterns with just `--break-cases`.

test/passing/break_cases.ml
jberdine: I think it would be good to also add some tests to cover the case where there is a nested or-pattern that is wide enough that it breaks even without `--sparse`. And to add a couple tests that include `when` guards in both modes as there is some interaction with breaking.

gpetiot:
My last commit e0de4c8 should fix the problem with parenthesized or-patterns you pointed out in your change request (can't directly reply to the request).

test/passing/break_cases.ml
gpetiot: Thoses cases are now covered in the break_cases(_sparse) files.

jberdine:
Getting closer!

Try cases like:
```
;; match x with
   | true -> (
     match y with
     | true -> "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
     | false -> "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" )
   | false -> "cccccccccccccccccccccccccccccc"
```
It seems that the open paren on the inner match is getting dropped with this patch.

Also, check cases like:
```
;; match x with
   | "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", yyyyyyyyyy
     when fffffffffffffff bbbbbbbbbb yyyyyyyyyy ->
       ()
   | _ -> ()
```
It seems that there is a missing break hint before `when`.

gpetiot:
I just added new tests for the problems you pointed out, it should be fine now.

src/Conf.ml
```diff
@@ -319,6 +319,11 @@ let wrap_comments =
   let default = false in
   mk ~default Arg.(value & flag & info ["wrap-comments"] ~doc ~env)
 
+let break_cases =
+  let doc = "Break pattern match cases." in
+  let default = false in
```
hhugo: @jberdine, what do you think about changing the default to true ?

src/Ast.ml
```diff
@@ -1292,8 +1292,7 @@ end = struct
               Ppat_construct
                 ({txt= Lident "::"}, Some {ppat_desc= Ppat_tuple [_; tl]})
           }
-      , Ppat_construct ({txt= Lident "::"}, _) )
-      when tl == pat ->
+      , Ppat_construct ({txt= Lident "::"}, _) ) when tl == pat ->
```
jberdine: There shouldn't be (significant) formatting changes like this when the option is not enabled. It is definitely intended that when a pattern does not fit on a line, the `when` clause should start at the beginning of a line.

src/Conf.ml
jberdine: I don't know... choosing defaults is hard. Given the config files, I can change our code bases to disable it if we turn it on by default, of course. My default inclination is to leave the defaults unchanged where the choice seems to be largely one of preference, to avoid unnecessarily churning code where the config files do not specify all settings.

hhugo: > to avoid unnecessarily churning code where the config files do not specify all settings.

I don't think I understand what you mean

jberdine: I'm just thinking that it is in general easier for users if upgrading does not needlessly change behavior, rather than add new behavior that can be enabled asynchronously with upgrading. Probably not a big deal, but that way it is more likely that the version and config files do not need to be changed atomically. Depending on how things are deployed (e.g. if binaries use something like rpms and config files are in source control), that can help.

hhugo: ack

test/passing/break_cases.ml
```diff
@@ -0,0 +1,35 @@
...
+;; match x with
+   | true -> (
+     match y with
+     | true -> "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+     | false -> "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" )
+   | false -> "cccccccccccccccccccccccccccccc"
+
+;; match x with
+   | "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", yyyyyyyyyy
+     when fffffffffffffff bbbbbbbbbb yyyyyyyyyy -> ()
```
jberdine: Without `--break-cases`, if the pattern + when guard if present does not fit on a line, then a break is added after `->` in order to avoid "hiding" a short body at the end of a multi-line pattern. Similarly, hit by reformatting ocamlformat but not shown by the tests, if the pattern doesn't fit, then the when guard is bumped to the next line. Can you confirm that the present behavior with less breaking is as you intend?

gpetiot: Yes I intended to limit the breaks. When this option is set, the break is forced for `|` so as you state the guard is bumped to the next line, no matter the size of the pattern.
So we have:
```
| K -> x
| A
  when b -> y
```

I prefer having this : 
```
| K -> x
| A when b -> y
```

But if you insist of having the first formatting I just have to revert my last commit.

jberdine: I'm not insisting, just checking that the behavior is as intended. But to be more explicit, depending on the margin, we will see e.g.:
```
;; match x with
   | ( "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
     , yyyyyyyyyy )
     when fffffffffffffff bbbbbbbbbb yyyyyyyyyy -> ()
```
and
```
;; match x with
   | ( "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
     , yyyyyyyyyy ) when f y -> ()
```
In these sorts of cases, there are possibly less breaks with `--break-cases` than without, which is what makes me wonder if this is intended as it does not fit with my understanding of the informal intent of this option.

jberdine: I have seen cases where people had a harder time reading code because of such hidden when clauses or short bodies, so I would suggest keeping those breaks that the current implementation inserts. But to achieve the output you want otherwise might require adding more boxes, if changing the `fmt` you mention doesn't do it.

jberdine:
Maybe add a test similar to:
```
let is_sequence exp =
  match exp.pexp_desc with
  | Pexp_sequence _
   |Pexp_extension
      ( _
      , PStr [{pstr_desc= Pstr_eval ({pexp_desc= Pexp_sequence _}, []); _}]
      ) ->
      true
  | _ -> false
```
Note that without `--break-cases` there is a break between `->` and `true` while with `--break-cases` there is not.

test/passing/break_cases_sparse.ml
```diff
@@ -0,0 +1,68 @@
...
+;;
+match x with
+| true -> (
+  match y with
+  | true ->
+      "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+  | false ->
+      "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" )
+| false ->
+    "cccccccccccccccccccccccccccccc"
```
jberdine: Looks good! Are you happy with this now and ready for this to be merged, or is there a plan to try to avoid breaking in cases like the line above while preserving the breaks like in cases below?

test/passing/break_cases_sparse.ml
```diff
@@ -0,0 +1,68 @@
...
+    ()
+
+let is_sequence exp =
+  match exp.pexp_desc with
+  | Pexp_sequence _
+  | Pexp_extension
+      ( _
+      , PStr [{pstr_desc= Pstr_eval ({pexp_desc= Pexp_sequence _}, []); _}]
+      ) ->
+      true
```
jberdine: Here, referred to above.

test/passing/break_cases_sparse.ml
gpetiot: If there was an easy way to avoid breaking in the case above I think it would look even better, but I did not manage to do that without breaking the other cases. If someone thinks it should be done in this PR (@samoht ?) I am open to suggestions.

samoht: No opinion, as I will probably not use that sparse option :-)

samoht: Actually, I am a bit lost. How do I get the tool to format:

```ocaml
let f x = function
  | C -> 1
  | D -> 2
```

without adding line breaks?

samoht: Ok so I am very confused now.  `ocamlformat` always formats this into:

```ocaml
let f x = function
  | x ->
      1
  | y ->
      2
```

even without using `--sparse` option.

But maybe this should be done with an other option in a different PR...

gpetiot: Yes I went back with the formatting of arrows because I did not manage to:
- break after `->` when the patterns or the when clause or the rhs expression breaks
- not break otherwise
When I tried to remove the break hint, I had results like:
```
let is_sequence exp =
  match exp.pexp_desc with
  | Pexp_sequence _
   |Pexp_extension
      ( _
      , PStr [{pstr_desc= Pstr_eval ({pexp_desc= Pexp_sequence _}, []); _}]
      ) -> true
```
`) -> true` fits on the line, but since the pattern breaks, we also would like to breka after `->`

So maybe it is better to stick with the default formatting of arrows for now.

jberdine: @samoht, the cause of your confusion is why I asked if this is ready, since as far as I understand it doesn't completely match the initial goal.

There is some interaction between the breaking after `->` of individual cases and the breaking between cases. The current patch force-breaks the box enclosing the cases, but that has the additional effect of also force-breaking after `->`. I haven't thought through it in detail, but I suspect that it is necessary to add an additional level (probably two, actually) of boxing around individual case patterns + `->` as well as around whole cases. That way the box enclosing the cases can be force-broken while letting the individual cases break naturally, with a box around the pattern and `->` so that if the pattern breaks, there is also a break after the `->`.

gpetiot: I tried what you suggested but did not manage to make it work (tried several combinations of box types). I don't know if it is worth investigating.
By the way I squashed the commits in case you want to merge.

jberdine:
With the current treatment of breaking around `->` and `when` in this PR, I am concerned with the overlap between `sparse` and `break-cases`. Both disabled makes sense, just sparse makes sense, both enabled (breaks after `when` more than optimal but) makes sense as a 'more sparse' mode, but sparse disabled and break-cases enables does not seem to make sense to me. Do others agree? If so, I think it would be best to combine the sparse and break-cases options into a 3 state option, maybe `break-cases = {fit, nested, all}`.

gpetiot:
I like the idea of a 3 state option  `break-cases = {fit, nested, all}`.
Just to be sure:
- fit --> default 
- nested --> present `sparse`  (remove the existing option)
- all --> `sparse` + `break-cases`

gpetiot:
I did not see any particular issue when running `make -C test test test_diff` (minor changes due to recent commits, but nothing caused by option break-cases), but the .ocamlformat files have to be updated.
infer/.ocamlformat:
```
margin=100
break-cases=nested
version=0.5
```
infer/sledge/.ocamlformat:
```
break-string-literals=wrap
margin=77
break-cases=fit
wrap-comments=true
```

jberdine:
The .ocamlformat syntax change should result in lots of spewed warnings but still work, no?

Also note that you can checkout master, run the test, run `make test_stage`, checkout your branch, and then run test and test_diff to see the changes of only your branch.

gpetiot:
Yes the old syntax still works but it is necessary to replace `sparse true` with `break-cases nested` and `sparse false` with `break-cases fit` because the option sparse is removed.

jberdine:
Thanks!

src/Fmt_ast.ml
```diff
@@ -3585,7 +3601,8 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
                ?ext:(if first then ext else None)
                ctx binding
                ?epi:
-                 (Option.some_if c.conf.sparse
+                 (Option.some_if
+                    (break_cases_level c > 0)
```
hhugo: This ident  does not really make sense anymore. This has nothing to do with  break_cases.

jberdine: Ah yes, this should probably be its own option, perhaps `extra-line-after-multiline-binding` or something. I'm actually tempted to remove this extra space entirely and see if anyone complains.

PR #253, opened 2018-08-01
Improve: formatting of comments

jberdine:
Ensure that comments are removed in before-within-after order as the
formatting code relies on. PR #223 happened to have the order
within-before-after, causing some regressions.

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -554,11 +554,15 @@ let fmt_within t ?(pro= Fmt.break_unless_newline 1 0)
     ?(epi= Fmt.break_unless_newline 1 0) =
   fmt_cmts t t.cmts_within ~pro ~epi ~eol:(Fmt.fmt "")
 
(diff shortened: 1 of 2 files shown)


src/Cmts.ml
```diff
@@ -547,11 +547,15 @@ let fmt_within t ?(pro= Fmt.break_unless_newline 1 0)
...
   let before = fmt_before t ?pro ?epi ?eol ?adj loc in
-  let inner = k in
-  let after = fmt_after t ?pro ?epi loc in
-  before $ inner $ after
+  (* remove the within comments from the map by accepting the continuation *)
+  fun k ->
+    (* delay the after comments until the within comments have been removed *)
+    let after = fmt_after t ?pro ?epi loc in
+    let inner = k in
+    before $ inner $ after
```
jberdine: Apologies... this mix of higher-order combinators for formatting and state is subtle and quite suboptimal.

PR #254, opened 2018-08-01
[RFC] Remove bitrotten Reason support

jberdine:
This PR is a request for comments from potential users of the Reason conversion support. The support for converting Reason to OCaml has not been actively used or maintained for a year. Much has changed in that time, not to mention that ocamlformat depends on an old version of reason. Unless someone steps forward to actively maintain this functionality, it makes more sense to remove it for the time being.


hhugo:
Copy-pasting comment from @osener

>Re: #254, I think it is fair to remove Reason support from ocamlformat. I had tried using `refmt` to convert Reason sources to OCaml (and going the other way) a while back and saw some loss of syntax sugar such as `let%lwt`

> I just tried again, this time using ocamlformat and refmt together, and it works really well except the annotations Refmt inserts such as `[@reason.raw_literal ...]` and `[@explicit_arity]`. If Refmt provides an option to disable those, something like the command below would be almost lossless:
```sh
refmt --parse=ml --print=re source.ml | refmt --parse=re --print=ml | ocamlformat
```

hhugo:
@jberdine, do you want to merge this based on the comment above ?

jberdine:
I'm wary of the reason support being something like false advertising. Basically I don't currently have a use for the Reason conversion, and don't use it, and don't have time to fix the bugs that must be in it. On the other hand, it might be more approachable to someone who wants to get it working to leave the code there. Maybe it would be enough to be more explicit that it is there, but untested and largely unsupported?

jberdine:
(Replying to https://github.com/ocaml-ppx/ocamlformat/issues/289#issuecomment-410990795 but keeping the discussion here.)

It is not so much work to keep the existing code in the repo, and to keep it compiling. It would be a lot of work to test it. I think it would also likely take some work to keep up to date with changes to refmt, which is a moving target, and ocamlformat's Reason to OCaml conversion support relies tightly on one of refmt's internal representations. (The way ocamlformat_reason works is to pipe a binary serialized internal representation from refmt to ocamlformat.)

The last I used ocamlformat_reason, it handled comments and things like explicit_arity reasonably well, much better than outputting ocaml from refmt (which uses Pprintast from ocaml's compiler-libs) and then reformatting that ocaml.

But ok, it seems that the best plan is to not remove this code for now.

hcarty:
I'm willing to help at least with the testing side of it.  I think it's better for both Reason and OCaml to have a reasonably clean two way conversion process.  Having Reason -> OCaml provides an escape hatch for OCaml users who experiment with Reason (lowering the risk involved in trying out Reason's syntax).

@jordwalke who may be interested in chiming in and/or may know someone who could help with maintenance.

jberdine:
@hcarty it would be great to have some help testing! Even just if someone pays attention to whether it gets broken would be helpful.

I agree about lowering the risk to try Reason, I just can't dedicate the time to it.

gpetiot:
We are thinking again about removing the Reason code.
@osener: Now that the Reason ecosystem is more mature, is there a better way to achieve what you wanted outside of ocamlformat?

PR #255, opened 2018-08-01
Improve: preserve comments formatted as block ..

hhugo:
.. such as file headers.


src/Cmts.ml
```diff
@@ -516,24 +516,51 @@ let fmt_cmt t cmt =
...
+              ( a.Location.loc_start.pos_lnum = a.Location.loc_end.pos_lnum
+              && b.Location.loc_start.pos_lnum = b.Location.loc_end.pos_lnum
+              && Position.column a.Location.loc_start
+                 = Position.column b.Location.loc_start
+              && Position.column a.Location.loc_end
+                 = Position.column b.Location.loc_end
+              && abs
+                   ( a.Location.loc_start.pos_lnum
+                   - b.Location.loc_start.pos_lnum )
+                 = 1 ) )
```
jberdine: Maybe outline these tests into a few functions with descriptive names?

hhugo: done

src/Cmts.ml
```diff
@@ -516,24 +524,56 @@ let fmt_cmt t cmt =
...
+      list_pn groups (fun ?prev group ?next ->
+          fmt_or_k (Option.is_none prev)
+            (Option.call ~f:pro $ open_vbox 0)
+            (fmt "@ ")
+          $ ( match group with
+            | [] -> impossible "previous match"
+            | [cmt] -> fmt_cmt t cmt $ maybe_newline ~next cmt
+            | group ->
+                fmt_if_k (Option.is_none prev)
+                  (Fmt.break_unless_newline 0 0)
```
jberdine: Do you have examples where the 2 lines above are needed? With `--sparse` they lead to an extra open line before the comments in the following example:
```
let f x =
  match x with
  | `A ->
      ()
  | `B ->
      (* Proin ipsum nunc, finibus et finibus, semper et mi. Aenean *)
      (* pretium fermentum tellus, a faucibus sagittis et. Cras non *)
      ()
```
The `Format` operations that are conditional on newlines are pretty fragile to use, since leading whitespace for indentation counts as being not starting a new line.

This is the last little issue I have seen in my testing.

test/passing/comment_header.ml
```diff
@@ -0,0 +1,45 @@
...
+(*             Xavier Leroy, projet Gallium, INRIA Rocquencourt           *)
+(*                                                                        *)
+(*   Copyright 2014 Institut National de Recherche en Informatique et     *)
+(*     en Automatique.                                                    *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
```
jberdine: I realize it is just an example, but it might be better if this didn't talk about copyright, licenses, etc.

hhugo: updated

src/Cmts.ml
hhugo: sorry, I've pushed -force and removed your commit.

hhugo: These two lines are not need and are now removed.
I've added back your example.

PR #260, opened 2018-08-02
Fix: many fixes regarding attributes

hhugo:
Note that we now include in our regtest  a huge files from the ocaml compiler testsuite.
see test/passing/source.ml


hhugo:
thanks. looks good.

PR #261, opened 2018-08-02
Fix: comments, around Longident

hhugo:


src/Fmt_ast.ml
```diff
@@ -344,6 +344,10 @@ let rec fmt_longident (li: Longident.t) =
   | Lapply (li1, li2) ->
       cbox 0 (fmt_longident li1 $ wrap "(" ")" (fmt_longident li2))
 
+let fmt_longident_loc c ?(pre= Caml.format_of_string "")
```
jberdine: is this preferable to using `("" : _ format)` ?

hhugo: fixed

hhugo:
I've added the test you've mentioned and fixed it.

test/passing/comments_in_record.ml
```diff
@@ -24,3 +24,8 @@ let { (* cmts *)
     ; (* e *)
       e : loooooooooooooooooooooooooooooooooooooooooooooooooooooooong_int } =
   exp
+
+let x =
+  { (* Xxxx xxxxxxxx xxxxx xx xx xx xxxx xxxxxx - XXxx_xxxxx xxx'x. *)
+  Irure_sed_a.in_nisi_sed= Irure_sed_fugiat.LaboRum sint_sed
```
jberdine: This shows a change that isn't a regression in this patch, it's just different than before, but still not consistent with the behavior for the case where the field is not module-qualified.

hhugo: fix

Issue #49, opened 2017-11-25
Line endings changed to \r\n on Windows.


leviroth:
On Windows, I have some files with the Unix `\n` line endings. However, when running them through ocamlformat, the line endings are changed to the Windows `\r\n` line endings.

Ocamlformat version: 0.2

Sample file (any file should work):

```ocaml
let () =
  let x = () in
  x
```

jberdine:
Thanks for the report. Nothing specific is done with line endings by ocamlformat, it just writes a file in standard text mode. On windows this results in "native" `\r\n` line endings. It may be that switching the file to binary mode would work to prevent newlines from being translated. But there may be other consequences as well, I don't know. There is also the question of whether windows users in general expect windows-specific line endings or not. It would be useful to get input from more windows users on this.

leviroth:
>It may be that switching the file to binary mode would work to prevent newlines from being translated.

If I'm understanding things right, the newline behavior can be modified via the `Formatter.formatter_out_functions` type. Something like:

```ocaml
  let old_out_functions = Format.pp_get_formatter_out_functions fs () in
  let new_out_functions =
    let open Format in
    { old_out_functions with
      out_newline= (fun () -> old_out_functions.out_string "\r\n" 0 2) }
  in
  Format.pp_set_formatter_out_functions fs new_out_functions
```

>There is also the question of whether windows users in general expect windows-specific line endings or not. It would be useful to get input from more windows users on this.

Personally, I think I would expect the line ending type of a given file to be preserved, regardless of platform. So on Windows, formatting a Unix-ending file results in a Unix-ending file, while a Windows-ending file remains a Windows-ending file even if the formatting happens to be done on Unix.

jberdine:
Note that the line ending type used in the original file is not available in the parsetree, AFAIK. Adding a config option to specify the desired line endings seems like a good solution.

leviroth:
Sounds good to me. I assume the Emacs integration would be able to pass that configuration along automatically based on the current buffer?

jberdine:
I guess that the emacs integration could do that, but I don't know emacs very thoroughly.

jberdine:
Question: is there a use case where ocamlformat is run under unix, on files with unix line endings, and output files with windows line endings are desired? @leviroth

leviroth:
I can't think of any, personally.

jberdine:
[...]

PR #274, opened 2018-08-03
[WIP] Option ending type

gpetiot:
Attempt to fix issue #49 
For now only the `Format.pp_set_formatter_out_functions` have been changed. The `@\n` in formats should be fine, but  should the `\n` in formats all be replaced by `!Conf.current_ending_token` ? (maybe they are not `@\n` for a reason ?)


hhugo:
turning `fmt "\n@\n"` into `fmt "@\n@\n"` would generate leading white space after the first newline

jberdine:
Yes, see https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Fmt_ast.ml#L342 . This is a hack that I'm not proud of, but I don't know a better way.

src/Conf.ml
```diff
@@ -238,6 +239,19 @@ module Formatting = struct
...
+      "Type of line endings printed. Can be set in a config file with an \
+       `ending-type {preserve,unix,windows}` line."
+    in
+    let env = Arg.env_var "OCAMLFORMAT_ENDING_TYPE" in
+    let names = ["ending-type"] in
+    let all =
+      [("preserve", `Preserve); ("unix", `Unix); ("windows", `Windows)]
+    in
+    C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
+        {conf with ending_type= x} )
```
jberdine: inline isn't going to work is it?

src/Conf.ml
```diff
@@ -606,4 +621,22 @@ let action =
...
+      let line = Stdio.In_channel.input_line ~fix_win_eol:false chan in
+      Stdio.In_channel.close chan ;
+      match line with
+      | Some line when String.is_substring line ~substring:"\r\n" ->
+          `Windows
+      | _ -> `Unix )
+  | `Unix -> `Unix
+  | `Windows -> `Windows
+
+let ending_token file =
```
jberdine: Naming suggestion: s/ending_token/eol_sequence/

src/Conf.ml
```diff
@@ -606,4 +621,22 @@ let action =
...
+  match ending_type file with `Unix -> "\n" | `Windows -> "\r\n"
+
+let current_ending_token = ref "\n"
```
jberdine: I don't see why this can't be removed, it seems to be only written but never read?

src/Conf.ml
```diff
@@ -606,4 +621,22 @@ let action =
 
 and debug = !debug
 
+let ending_type file =
+  match config.ending_type with
+  | `Preserve -> (
+      let chan = Stdio.In_channel.create ~binary:false file in
+      let line = Stdio.In_channel.input_line ~fix_win_eol:false chan in
+      Stdio.In_channel.close chan ;
```
jberdine: I tend to prefer `In_channel.with_file` when it fits

src/Conf.ml
gpetiot: I intended to use it in place hard-coded '\n' in Fmt_ast, don't know if it is necessary at all (need some tests).

gpetiot:
I added some examples, the `preserve` and the conversions are tested and they seem to work fine.

src/Fmt_ast.ml
```diff
@@ -2048,12 +2048,12 @@ and fmt_class_structure c ~ctx ~parens ?ext self_ fields =
            $ ( match fields with
              | ({pcf_desc= Pcf_attribute a}, _) :: _
                when Option.is_some (fst (doc_atrs [a])) ->
-                 fmt "\n"
+                 fmt (eol_fmt ())
```
jberdine: Given the changes in Translation_unit that set the formatter output function, are these changes in Fmt_ast strictly needed? Can you explain what goes wrong without them? These changes are much more invasive than just those in Translation_unit, and make the code harder to change, and harder to keep right as it's non-obvious to know when \n is ok and when it isn't. It might even be better to just post-process the output with `tr` or whatever to get the desired line endings.

src/Fmt.ml
```diff
@@ -29,6 +29,16 @@ let break n o fs = Format.pp_print_break fs n o
...
 
+let eol_str () =
+  match !Conf.current_line_endings with `Unix -> "\n" | `Windows -> "\r\n"
+
+let eol_fmt () =
+  match !Conf.current_line_endings with
+  | `Unix -> Caml.format_of_string "\n"
+  | `Windows -> Caml.format_of_string "\r\n"
+
+let double_eol_fmt () = Caml.( ^^ ) (eol_fmt ()) "@\n"
```
jberdine: I'm wary of dynamic format string construction as I have had performance problems with it in the past.

src/Fmt_ast.ml
gpetiot: Yes these changes are needed, otherwise there are many '\n' remaining in the output when line-endings = windows. The only '\n' we have to keep are those in literal strings (`"a\nb"`).
If I can get the same results with post-processing the output I will do it.

gpetiot:
I replaced my formatting changes with a post-processing. There are still minor issues on the source_windows test (a few additional blank lines are generated).

jberdine:
Note that [normally](https://github.com/ocaml/ocaml/blob/trunk/parsing/lexer.mll#L328) the lexer emits an EOL token for any 'newline' (which is 0 or more \r and 1 \n). But for [quoted strings](https://github.com/ocaml/ocaml/blob/trunk/parsing/lexer.mll#L676), the lexer preserves the bytes unchanged. I suspect that this is the source of the difference in behavior you see.

But can we step back a moment and recall the use cases that we know people want?
1. running on unix, with input files with unix line endings, producing output files with unix line endings
2. running on windows, with input files with windows line endings, producing output files with windows line endings
3. running on windows, with input files with unix line endings, producing output files with unix line endings

In particular, I do not know that anyone would use

4. running on unix, with input files with windows line endings, producing output files with windows line endings

As far as I know, cases 1 and 2 are handled in master, and 3 would be achieved by opening output files in binary mode. Is that wrong?

gpetiot:
I agree the last use case may not be very useful. I will do as you suggest for the third case.

leviroth:
>running on unix, with input files with windows line endings, producing output files with windows line endings

Suppose a repository is created on Windows, with Windows line endings, and then later cloned and edited on Unix. It would probably be desirable to keep the original line endings, rather than create a diff with pervasive line ending changes.

leviroth:
Nb. In #49, @jberdine asked if "there a use case where ocamlformat is run under unix, on files with **unix** line endings, and output files with windows line endings are desired[.]" That's different from the use case I quoted above. I'm not sure if that was mis-worded, or if it was an unrelated question.

jberdine:
Ok thanks, so it seems that there is a use case for producing windows EOLs on unix. Let's consider all the cases:

Host OS | Input EOLs | Output EOLs | Implementation Strategy
------- | ---------- | ----------- | -----------------------
unix | unix | unix | as master
unix | windows | unix | as master
windows | unix | unix | open output files in binary mode
windows | windows | unix | open output files in binary mode
unix | unix | windows | override `out_newline` or post-process with sed
unix | windows | windows | override `out_newline` or post-process with sed
windows | unix | windows | as master
windows | windows | windows | as master

It would be preferable to override `out_newline` instead of relying on sed or any other binary on the host.

The only issue with that was related to quoted strings, correct? I am not sure that the behavior observed before is incorrect. In particular, quoted strings are intended to be interpreted as raw byte sequences, with no translation, escaping, etc. So if there is an input file with X EOLs that includes a quoted string, then I think that producing an output file with Y EOLs should not change the byte sequences that look like EOLs that appear within quoted strings. Thoughts?

src/Conf.ml
```diff
@@ -808,3 +823,13 @@ let action =
...
 let parse_line_in_attribute = parse_line ~from:`Attribute
+
+let line_endings c file =
+  match c.line_endings with
+  | `Unix -> `Unix
+  | `Windows -> `Windows
+  | `Preserve -> (
+      let read_line = Stdio.In_channel.input_line ~fix_win_eol:false in
+      match Stdio.In_channel.with_file ~binary:false file ~f:read_line with
+      | Some l when String.is_substring l ~substring:"\r" -> `Windows
```
jberdine: String.mem instead of String.is_substring

src/Conf.ml
```diff
@@ -808,3 +823,13 @@ let action =
...
 
 let parse_line_in_attribute = parse_line ~from:`Attribute
+
+let line_endings c file =
+  match c.line_endings with
+  | `Unix -> `Unix
+  | `Windows -> `Windows
+  | `Preserve -> (
+      let read_line = Stdio.In_channel.input_line ~fix_win_eol:false in
+      match Stdio.In_channel.with_file ~binary:false file ~f:read_line with
```
jberdine: Shouldn't binary be true

test/passing/string_unix_to_windows.ml
```diff
@@ -0,0 +1,21 @@
...
+  "aaaaaaaaaaaaaaaaaaaaaaaaa\n	\
+   bbbbbbbbbbbbbbbbbbbbbbbbbb\n	\
+   cccccccccccccccccc\n	\
+   "
+
+let _ = ('\xff', '\255', '\n')
+
+let f = function '\xff' .. '\255' -> ()
+
+let f ("test"[@test "test"]) = 2
```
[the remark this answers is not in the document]
gpetiot: The windows-> unix conversion is ok. But I encounter the same problem as I did on my other branch when I run make test, it seems the option is not set (via .opts or attributes) but it is ok when I run the test manually with the options.
The unix-> windows still uses inadequate line endings sometimes, I am looking into it.

src/Fmt_ast.ml
```diff
@@ -3653,9 +3653,13 @@ and fmt_structure c ctx itms =
...
     )
   in
+  let unix_line_endings =
+    Poly.(Conf.line_endings c.conf (Source.input_filename c.source) = `Unix)
+  in
   hvbox 0
     (list_fl grps (fun ~first ~last grp ->
-         fmt_if (not first) "\n@\n" $ fmt_grp ~last grp ))
+         fmt_if_k (not first) (fmt_or unix_line_endings "\n@\n" "@\n@\n")
+         $ fmt_grp ~last grp ))
```
jberdine: If you have to go to such lengths as this, for such a corner case, I think it would be better to just post-process with sed in the cases where generating windows endings when running on unix. It would be nice to only override the out_newline function, but if that doesn't work, it's going to be too much work and be too invasive. All the other cases seem to have working nice implementations now though.

gpetiot: I replaced the last commits with a basic post-processing. I still have an issue with `make test`, looks like the options in attributes are broken (same problem in other PRs as well).

src/Conf.ml
jberdine: @gpetiot did you see this? Have you tested on a windows machine?

src/Conf.ml
```diff
@@ -471,6 +472,19 @@ module Formatting = struct
...
+    let doc = "Line endings used." in
+    let all =
+      [ ( "preserve"
+        , `Preserve
+        , "$(b,preserve) preserves the original line endings." )
+      ; ("unix", `Unix, "$(b,unix) uses Unix line endings.")
+      ; ("windows", `Windows, "$(b,windows) uses Windows line endings.") ]
+    in
+    let env = Arg.env_var "OCAMLFORMAT_LINE_ENDINGS" in
+    C.choice ~names:["line-endings"] ~all ~doc ~env ~allow_inline:true
```
jberdine: Ok, `~allow_inline:true` isn't (always) going to work, since the output file is opened before the input attributes are read and processed. This seems ok, just less convenient for us updating test results I suppose.

src/Translation_unit.ml
```diff
@@ -132,7 +135,12 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
     xunit.fmt source cmts_t conf ast fs ;
     Format.pp_print_newline fs () ;
     Out_channel.close oc ;
-    let fmted = In_channel.with_file tmp ~f:In_channel.input_all in
+    ( match Conf.line_endings conf input_file with
+    | `Windows when Sys.unix ->
+        let command = Format.sprintf "sed -i 's/$/\r/' %s" tmp in
+        ignore (Unix.system command)
```
jberdine: nit: In general I'd prefer to have type annotations on ignore for safety: `Unix.system command |> (ignore : Unix.process_status -> _)`

It would probably be good to give an understandable error message if this fails. In particular, users might not expect ocamlformat to run sed.

src/Conf.ml
gpetiot: I don't have a windows machine at hand to test this.
I have no clue for the binary mode opening, if you are confident that it should be true I will change it.

jberdine: In non-binary (that is, text) mode, the runtime system on windows will helpfully translate \r\n to \n. So without ~binary:true, this code will never see a \r

src/Conf.ml
```diff
@@ -808,3 +823,13 @@ let action =
...
 let parse_line_in_attribute = parse_line ~from:`Attribute
+
+let line_endings c file =
+  match c.line_endings with
+  | `Unix -> `Unix
+  | `Windows -> `Windows
+  | `Preserve -> (
+      let read_line = Stdio.In_channel.input_line ~fix_win_eol:false in
+      match Stdio.In_channel.with_file ~binary:false file ~f:read_line with
+      | Some l when String.mem l '\r' -> `Windows
```
hhugo: It should be enough to check the last char only.

test/passing/string_unix_to_windows.ml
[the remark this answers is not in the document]
hhugo: This option cannot be configured with attributes. we should remove attributes from the test file and use `~allow_inline:false`.

jberdine:
I pushed a commit that changes to use `ex` and `awk` instead of `sed` since `sed` helpfully converts to unix line endings, does not portably support `\r` in expressions, and `-i` is also not portable. Also the opts files were forgotten I think.

jberdine:
I think this is ready to go, but it would really be good if someone could test on a windows and a cygwin machine at least once. @leviroth maybe (just clone this branch and run `make test`)?

leviroth:
I won't be able to get to it tonight, but I should be able to tomorrow.

hhugo:
What about we implement this sed/awk in ocaml instead ? I Would expect it to take only few lines.

jberdine:
Ok, I think I successfully overrode the output function to do the translation directly.

jberdine:
Ok, thanks a lot. That doesn't look to me like a failure of this PR, but still. Are there other options for testing on windows @gpetiot @hhugo? I wonder if the easiest thing is to set up Appveyor, but that is probably not good for debugging.

gpetiot:
I managed to setup a cygwin environment (I had the same issue as @leviroth, (tried many things) I think running `sh tools/gen_version.sh src/Version.ml` is what fixed it).
`make fixpoint` does not pass on this branch (it does not on master either), I don't know what I missed here.

jberdine:
Note that cygwin is neither unix nor windows, so I don't know what to expect the behavior to be here.

gpetiot:
I tried to cross-compile ocamlformat to windows, but there is no package ocaml-migrate-parsetree-windows >= v1.0.10, I submitted this issue: https://github.com/ocaml-cross/opam-cross-windows/issues/79

jberdine:
Thanks a lot for trying to test this. I don't understand though, how are the people who requested this feature using ocamlformat if it doesn't even build? @leviroth, any hints?

leviroth:
I've never had trouble installing ocamlformat via Cygwin on Windows. It's just running the tests that I couldn't get to work.

Note that [this Cygwin-based version of opam](https://fdopen.github.io/opam-repository-mingw/) seems to be by far the best-supported way of using OCaml on Windows; for me, "OCaml on Windows" usually refers to running that distribution.

jberdine:
Thanks @leviroth, that is useful to know. If you know, it would be useful to know if when you have successfully build ocamlformat with cygwin it was a released version or an opam pin or a manual clone (since the `gen_version` business is different in those cases).

But for this PR, when thinking about it it seems I had purged cygwin from my memory, and it complicates things. Since cygwin can itself either do eol translation or not, I expect that to get a reliable implementation for this PR we need to change from just testing Sys.unix to, when Sys.unix is false, actually writing a crlf newline to a file and reading it back to detect whether the crlf has been translated to lf.

leviroth:
I'm fairly certain that I built a released version, as via `opam install ocamlformat`.

src/Translation_unit.ml
```diff
@@ -115,6 +115,19 @@ let lf_to_crlf =
     (String.Escaping.escape_gen_exn ~escapeworthy_map:[('\n', '\n')]
        ~escape_char:'\r')
 
+let crlf_translated_to_lf () =
+  let filename = "ocamlformat_crlf_testing" in
+  let binary = true and append = false in
+  let write oc = Stdio.Out_channel.output_string oc "\r\n" in
+  Stdio.Out_channel.with_file ~binary ~append filename ~f:write ;
```
jberdine: It would be better to put this file in the temp dir like the others.

src/Translation_unit.ml
```diff
@@ -115,6 +115,19 @@ let lf_to_crlf =
...
+  let read = Stdio.In_channel.input_line ~fix_win_eol:false in
```
jberdine: Since we are more concerned with the output, I wonder if it would make sense to test by writing in text mode and reading in binary.

src/Translation_unit.ml
```diff
@@ -110,6 +110,25 @@ type result =
...
 
+let lf_to_crlf =
+  Staged.unstage
+    (String.Escaping.escape_gen_exn ~escapeworthy_map:[('\n', '\n')]
+       ~escape_char:'\r')
+
+let crlf_translated_to_lf () =
+  let prefix = "ocamlformat_crlf_testing" and ext = ".ml" in
+  let temp_dir = Filename.get_temp_dir_name () in
+  let filename, oc = Filename.open_temp_file ~temp_dir prefix ext in
```
jberdine: nit: specifying temp_dir is redundant

src/Translation_unit.ml
```diff
@@ -110,6 +110,25 @@ type result =
...
+  Stdio.Out_channel.output_string oc "\r\n" ;
```
jberdine: nit: `Stdio.` prefixes not needed as it's opened

gpetiot:
I just tried on windows (with the ocaml for windows tool, did not manage to cross compile).
It still does not work. When applied on src/Ast.ml for example we get
- line endings = Unix
- binary mode = true

but the formatting did not stabilize after 10 iterations. Don't know what we missed here.

src/Translation_unit.ml
```diff
@@ -122,25 +140,39 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
...
+        in
         (tmp, oc)
     in
     dump xunit dir base ".old" ".ast" ast ;
     let source = Source.create source_txt in
     let cmts_t = xunit.init_cmts source conf ast comments in
     let fs = Format.formatter_of_out_channel oc in
+    ( if
+      (Sys.unix && Poly.(Conf.line_endings conf input_file = `Windows))
+      || ((not Sys.unix) && crlf_translated_to_lf ())
```
hhugo: I don't understand the rational for this one

gpetiot: The idea was to translate \n to \r\n when we are not on unix and the \r\n are not preserved when printing the formatted output (because we are not sure how they are handled on cygwin or windows).
I guess the `line_endings = windows` condition has been forgotten for this case.
Something like this should be better:
- on unix: if we want to produce windows output
- on windows/cygwin: if we want to produce windows output and \r\n are not preserved

gpetiot:
(The last commit did not changed the result on cygwin by the way.)

@dra27 can we ask for your help on this? Are there common mistakes that we could have done? Do you know where I should start looking?

PR #277, opened 2018-08-03
Test: add ocaml compiler to test suite, and improve `make -C test`

jberdine:
From the `test` directory:

- `make test` reformats the source files under `DIRS`

- `make test_extra` reformats the source files under `XDIRS`, using
  looser checking options

- `make test_status` runs `git status` in each test dir

- `make test_diff` runs `git diff` in each test dir

- `make test_stage` runs `git add` in each test dir, in order to
  baseline the current state

- `make test_unstage` reverts the above

- `make test_clean` resets the test dirs to the upstream or staged state

- `make test_pull` runs `git pull` in each test dir

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -14,6 +14,11 @@
 # By default, test projects used as regression tests
 DIRS=code/ocamlformat code/infer
 
+# Extra test directories, for which looser checking is done
+XDIRS=code/ocaml
+
+ALL_DIRS=$(DIRS) $(XDIRS)
+
 # To test using the dev or release executable
 #       make MODE=<mode>
 MODE?=dev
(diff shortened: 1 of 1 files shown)


test/Makefile
```diff
@@ -30,6 +30,8 @@ test:
 	  || git -C code clone ../../../ocamlformat
 	test -d code/infer \
 	  || git -C code clone "https://github.com/facebook/infer.git"
+	test -d code/ocaml \
+	  || git -C code clone "https://github.com/hhugo/ocaml.git"
```
hhugo: Why not take the official one git repo ?

jberdine: I didn't know if you had made some additional changes wrt the official one.  I'm happy to change it.

hhugo: I've formatted the code using -no-comment-check to get a larger portion of the codebase formatted.

jberdine: I tried that, and realized that no-comment-check isn't parsed in config files

hhugo: This was on purpose. I didn't want that behavior to trigger without the flag appearing in the cmd line.

PR #279, opened 2018-08-04
Improve: parens and nested match

hhugo:
- Always parens nested match, even the trailing one.
The reason for this change is that it ease adding new cases to the parent match.
- Add an option to choose where to put the parens in nested matches. It can either be right at the beginning of the parens' rhs or be pushed down to the nested match.

I'm not happy with option name, suggestion welcome.

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -410,8 +410,6 @@ end = struct
 
   let ( == ) = Caml.Pervasives.( == )
 
-  let ( != ) = Caml.Pervasives.( != )
-
   let dump fs ctx ast =
     Format.fprintf fs "ast: %a@\nctx: %a@\n" T.dump ast T.dump ctx
 
@@ -1539,7 +1537,6 @@ end = struct
       match pexp_desc with
       | Pexp_function cases | Pexp_match (_, cases) | Pexp_try (_, cases) ->
           List.exists cases ~f:(fun {pc_rhs} -> pc_rhs == exp)
-          && (List.last_exn cases).pc_rhs != exp
           && exposed_right_exp Match exp
       | Pexp_ifthenelse (cnd, _, _) when cnd == exp -> false
       | Pexp_ifthenelse (_, thn, None) when thn == exp ->
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -14,18 +14,22 @@
 open Migrate_ast
 open Parsetree
 
-let reset, register_reset =
+let init, register_reset, nested_match =
   let l = ref [] in
+  let nested_match = ref `Parent in
   let register f = l := f :: !l in
-  let reset () = List.iter !l ~f:(fun f -> f ()) in
-  (reset, register)
+  let init conf =
+    nested_match := conf.Conf.nested_match ;
+    List.iter !l ~f:(fun f -> f ())
+  in
+  (init, register, nested_match)
 
 (** Predicates recognizing special symbol identifiers. *)
 
 let is_prefix_id i =
   match i with
   | "!=" -> false
-  | _ -> match i.[0] with '!' | '?' | '~' -> true | _ -> false
+  | _ -> ( match i.[0] with '!' | '?' | '~' -> true | _ -> false )
 
 let is_prefix exp =
   match exp.pexp_desc with
@@ -67,10 +71,10 @@ let index_op_get i =
 let index_op_set i =
   match String.chop_suffix i ~suffix:"<-" with
   | None -> None
-  | Some i ->
+  | Some i -> (
     match (i.[0], parens_kind i) with
     | '.', Some (s, o, c) -> Some (s, o, c)
-    | _ -> None
+    | _ -> None )
 
 let index_op_string = (".", '[', ']')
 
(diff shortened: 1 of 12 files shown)


hhugo:
credit to @mbarbin for the initial implementation

src/Ast.ml
```diff
@@ -1270,19 +1273,19 @@ end = struct
...
-      | _ -> false
+      | _ -> false )
 
   (** [parenze_cty {ctx; ast}] holds when class type [ast] should be
       parenthesized in context [ctx]. *)
   let parenze_cty ({ctx; ast= cty} as xcty) =
     assert (check_cty xcty ; true) ;
     match xcty with _ ->
-      match ambig_prec (sub_ast ~ctx (Cty cty)) with
+      ( match ambig_prec (sub_ast ~ctx (Cty cty)) with
```
jberdine: Is the open paren here and not at the end of the previous line intentionally?

hhugo: no

hhugo: fixed

jberdine:
The current formatting uses the "docked" trailing dangling open parens in an attempt to match the style for e.g. if-then-else and assert expressions. It turned out to be more implementation effort than I could afford to use that style uniformly for other expression forms though. The advantage of the docked form is that is saves one level of indentation for the matched-on expression when it breaks:
```ocaml
match
  something
    that
    breaks
with
| _ ->
```
versus
```ocaml
( match
    something
      that
      breaks
  with
| _ ->
```
For that reason, I'm hesitant to remove the current behavior, but I realize that overall there is some inconsistency due to the implementation not managing to produce the docked form in many instances.

I wonder if it would make sense to combine various options into an umbrella one which sets options in a way that does not try to conserve indentation, and another to not worry about consuming vertical space, or perhaps even combine the two?

Re naming, maybe make the option a bool and use `--leading-nested-match-parens`. I don't know, nothing good comes to mind at the moment.

jberdine:
What I'd like to keep is how e.g.:
```
... -> (
  match
    something
      that
      breaks
  with
  | _ ->
```
can indent the code between `match` and `with` one level less:
```
... ->
  ( match
      something
        that
        breaks
    with
  | _ ->
```
without becoming unclear or having to play tricks. Maybe it would also be ok to go with
```
... ->
  ( match
    something
      that
      breaks
  with
  | _ ->
```

src/Ast.ml
```diff
@@ -1441,6 +1444,11 @@ end = struct
     | {pcl_desc= Pcl_fun _}, Non_apply -> true
     | _ -> false
 
+  let memo_parenzed_inner_nested_match =
```
jberdine: Nit: I found the name `memo_` momentarily confusing since this is also used outside `mark_parenzed_inner_nested_match`, so it is more real state than just memoization.

src/Ast.ml
```diff
@@ -1512,13 +1520,70 @@ end = struct
...
+       |Pexp_poly _ | Pexp_record _ | Pexp_send _ | Pexp_unreachable
+       |Pexp_variant (_, None)
+       |Pexp_while _ ->
+          false
+    in
+    let _ : bool =
+      Hashtbl.Poly.find_or_add memo_parenzed_inner_nested_match exp
+        ~default:exposed_
+    in
+    ()
```
jberdine: Just FMI, do you prefer this style over `... |> (ignore : bool -> _)` for any particular reason?

src/Ast.ml
```diff
@@ -1563,7 +1644,7 @@ end = struct
       | Pexp_record (flds, _)
         when List.exists flds ~f:(fun (_, e0) -> e0 == exp) ->
           exposed_right_exp Non_apply exp
-          (* Non_apply is perhaps pessimistic *)
+      (* Non_apply is perhaps pessimistic *)
```
jberdine: I think if you revert this it will be preserved, and would be a better attachment

src/Conf.ml
```diff
@@ -305,6 +306,17 @@ module Formatting = struct
     C.int ~names:["m"; "margin"] ~default:80 ~doc ~docv ~env
       ~allow_inline:false ~update:(fun conf x -> {conf with margin= x} )
 
+  let nested_match =
+    let doc =
+      "Nested match formatting. Can be set in a config file with an \
+       `nested-match {parent,child}` line."
```
jberdine: missing `=`

src/Ast.ml
hhugo: renamed

src/Ast.ml
hhugo: none. Changed.

src/Ast.ml
hhugo: done.

src/Conf.ml
hhugo: ack

hhugo:
I've done another pass, stashed and rebase.

hhugo:
@jberdine, I've merged a bit fast. I forgot the change the name/type of the cmd option. I'll do another PR later tonight.

PR #280, opened 2018-08-04
Improve: break in type declaration for variant and record

hhugo:
I could add an option to configure that but I'd only do it if some people prefer the old behavior.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -163,6 +163,20 @@ let break_infix =
           default
       & info ["break-infix"] ~doc ~env)
 
+let type_decl =
+  let doc =
+    "Style of type declaration. Can be set in a config file with a \
+     `break-infix {wrap,fit-or-vertical}` line. `wrap` will group simple \
+     expressions and try to format them in a single line."
+  in
+  let env = Arg.env_var "OCAMLFORMAT_TYPE_DECL" in
+  let default = `Auto in
+  mk ~default
+    Arg.(
+      value
+      & opt (enum [("sparse", `Sparse); ("auto", `Auto)]) default
+      & info ["type-decl"] ~doc ~env)
+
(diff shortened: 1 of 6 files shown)


jberdine:
Yes, as an option please.

hhugo:
option addded

src/Conf.ml
```diff
@@ -163,6 +163,20 @@ let break_infix =
           default
       & info ["break-infix"] ~doc ~env)
 
+let type_decl =
+  let doc =
+    "Style of type declaration. Can be set in a config file with a \
+     `break-infix {wrap,fit-or-vertical}` line. `wrap` will group simple \
+     expressions and try to format them in a single line."
```
jberdine: copypasta

src/Conf.ml
```diff
@@ -163,6 +163,20 @@ let break_infix =
...
+  in
+  let env = Arg.env_var "OCAMLFORMAT_TYPE_DECL" in
+  let default = `Auto in
+  mk ~default
+    Arg.(
+      value
+      & opt (enum [("sparse", `Sparse); ("auto", `Auto)]) default
```
jberdine: Nit: I think most cases order the alternatives with the default first.

src/Conf.ml
```diff
@@ -358,6 +372,7 @@ type t =
   ; if_then_else: [`Compact | `Keyword_first]
   ; infix_precedence: [`Indent | `Parens]
   ; break_infix: [`Wrap | `Fit_or_vertical]
+  ; type_decl: [`Sparse | `Auto]
```
jberdine: Same default-first nit.

hhugo:
I've renamed `auto` into `compact` similar to `if-then-else`

PR #281, opened 2018-08-04
Always break for let_and contruct 

hhugo:
Note that this could be an option

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -21,6 +21,7 @@ type t =
   ; extension_sugar: [`Preserve | `Always]
   ; if_then_else: [`Compact | `Keyword_first]
   ; infix_precedence: [`Indent | `Parens]
+  ; let_and: [`Compact | `Sparse]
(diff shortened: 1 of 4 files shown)


jberdine:
Also as an option please.

hhugo:
I'd like to keep the number of options under control / be able to explain why such option exists.
To me, let-and on a single line is harder to read probably because of the `and` keyword lost in the middle.

Could you explain why you would want to keep the current behaviour ?

jberdine:
First, perhaps in the case where `let` or `and` has an extension, it would make sense to always break. Otherwise it is (I think) always possible to use multiple let-in bindings instead of let-and. So making let-and always break takes away the option of writing a couple very short abbreviation-style bindings as a one-liner. I have seen people instead write `let (x,y) = (e1, e2) in` as a workaround, which seems much worse. Are you seeing problems even in situations where the margin is reasonably low, and with syntax coloring?

hhugo:
I'll add an option for this one.

PR #283, opened 2018-08-04
Style of functions and applications

hhugo:
I usually find it easier to read code when things are either on a single line or on a single column.
This PR tries to apply that logic to applications and functions.

This PR also fixes a bug where label would be totally disconnected from their value.
e.g.
```ocaml
let _ = 
  call something ~f:
    fun x -> x
```

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -298,8 +298,8 @@ end
     column of the first location is not greater than that of the second
     location. *)
 let is_adjacent t (l1 : Location.t) (l2 : Location.t) =
-  Option.value_map (Source.string_between t.source l1 l2) ~default:false ~f:
-    (fun btw ->
+  Option.value_map (Source.string_between t.source l1 l2) ~default:false
+    ~f:(fun btw ->
       match String.strip btw with
       | "" -> true
       | "|" ->
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -15,7 +15,7 @@ type t =
   { break_cases: [`Fit | `Nested | `All]
   ; break_infix: [`Wrap | `Fit_or_vertical]
   ; break_string_literals: [`Newlines | `Never | `Wrap]
-  ; no_comment_check: bool
+  ; comment_check: bool
   ; doc_comments: [`Before | `After]
   ; escape_chars: [`Decimal | `Hexadecimal | `Preserve]
   ; escape_strings: [`Decimal | `Hexadecimal | `Preserve]
@@ -31,7 +31,8 @@ type t =
   ; parens_tuple: [`Always | `Multi_line_only]
   ; quiet: bool
   ; type_decl: [`Compact | `Sparse]
-  ; wrap_comments: bool }
+  ; wrap_comments: bool
+  ; wrap_fun_args: bool }
 
 (** Extension of Cmdliner supporting lighter-weight option definition *)
 module Cmdliner : sig
@@ -97,7 +98,7 @@ module C : sig
 
   val choice : all:(string * 'a) list -> 'a option_decl
 
-  val flag : bool option_decl
+  val flag : default:bool -> bool option_decl
 
   val int : default:int -> docv:string -> int option_decl
 
@@ -122,7 +123,7 @@ end = struct
     ; parse: string -> 'a
     ; update: config -> 'a -> config
     ; allow_inline: bool
-    ; r: 'a ref }
+    ; get: unit -> 'a }
 
   type 'a option_decl =
        names:string list
@@ -154,17 +155,25 @@ end = struct
             []
     in
     let r = mk ~default term in
-    let opt = {names; parse; update; r; allow_inline} in
+    let get () = !r in
+    let opt = {names; parse; update; get; allow_inline} in
     store := Pack opt :: !store ;
     opt
 
-  let flag ~names ~update ~env ~doc ~allow_inline =
+  let flag ~default ~names ~update ~env ~doc ~allow_inline =
     let open Cmdliner in
-    let term = Arg.(value & flag & info names ~doc ~env) in
+    let invert_flag = default in
+    let names_for_cmdline =
+      if invert_flag then
+        List.filter_map names ~f:(fun n ->
+            if String.length n = 1 then None else Some ("no-" ^ n) )
+      else names
+    in
+    let term = Arg.(value & flag & info names_for_cmdline ~doc ~env) in
     let parse = Bool.of_string in
-    let default = false in
     let r = mk ~default term in
-    let opt = {names; parse; update; r; allow_inline} in
+    let get () = if invert_flag then not !r else !r in
+    let opt = {names; parse; update; get; allow_inline} in
     store := Pack opt :: !store ;
     opt
 
@@ -173,13 +182,14 @@ end = struct
     let term = Arg.(value & opt int default & info names ~doc ~docv ~env) in
     let parse = Int.of_string in
     let r = mk ~default term in
-    let opt = {names; parse; update; r; allow_inline} in
+    let get () = !r in
+    let opt = {names; parse; update; get; allow_inline} in
     store := Pack opt :: !store ;
     opt
 
   let update ~config ~name ~value ~inline =
-    List.find_map !store ~f:
-      (fun (Pack {names; parse; update; allow_inline}) ->
+    List.find_map !store
+      ~f:(fun (Pack {names; parse; update; allow_inline}) ->
         if List.exists names ~f:(String.equal name) then
           if inline && (not allow_inline) then
             Some (Error (`Misplaced (name, value)))
@@ -189,7 +199,7 @@ end = struct
         else None )
     |> Option.value ~default:(Error (`Unknown (name, value)))
 
-  let get {r} = !r
+  let get {get} = get ()
 end
 
 let info =
@@ -332,8 +342,8 @@ module Formatting = struct
     in
     let env = Arg.env_var "OCAMLFORMAT_LEADING_NESTED_MATCH_PARENS" in
     let names = ["leading-nested-match-parens"] in
-    C.flag ~names ~env ~doc ~allow_inline:false ~update:(fun conf x ->
-        {conf with leading_nested_match_parens= x} )
+    C.flag ~default:false ~names ~env ~doc ~allow_inline:false
+      ~update:(fun conf x -> {conf with leading_nested_match_parens= x} )
 
   let let_and =
     let doc =
@@ -362,8 +372,8 @@ module Formatting = struct
       ; ("long", `Long) ]
     in
     let env = Arg.env_var "OCAMLFORMAT_LET_OPEN" in
-    C.choice ~names:["let-open"] ~all ~doc ~env ~allow_inline:true ~update:
-      (fun conf x -> {conf with let_open= x} )
+    C.choice ~names:["let-open"] ~all ~doc ~env ~allow_inline:true
+      ~update:(fun conf x -> {conf with let_open= x} )
 
   let margin =
     let docv = "COLS" in
@@ -383,8 +393,8 @@ module Formatting = struct
     in
     let env = Arg.env_var "OCAMLFORMAT_OCP_INDENT_COMPAT" in
     let names = ["ocp-indent-compat"] in
-    C.flag ~names ~env ~doc ~allow_inline:true ~update:(fun conf x ->
-        {conf with ocp_indent_compat= x} )
+    C.flag ~default:false ~names ~env ~doc ~allow_inline:true
+      ~update:(fun conf x -> {conf with ocp_indent_compat= x} )
 
   let parens_tuple =
     let doc =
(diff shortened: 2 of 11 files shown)


jberdine:
I think it makes sense to have as an option. How well it works will be very dependent on the style of the code in question I think.

Can you elaborate on what you are unhappy about in the current state of this patch?

hhugo:
fix #289 (@jberdine, another n fixing n+6 ;))

osener:
In #289 I suggested yet another "fit-or-vertical" option, but I think it would be also good to accumulate things similar to this under `sparse true` instead.

jberdine:
I agree that there is value in combining various related options into umbrella 'modes', but at the moment `sparse` is somewhat specific (and probably misnamed).

hhugo:
I've added two options. `fun-decl` and `fun-call`.
TODO: implement the same logic for `Pcl_fun` and method in classes.

test/passing/issue289.ml
```diff
@@ -0,0 +1,79 @@
+let foo =
+  let open Gql in
+  [ field "id" ~doc:"Toy ID." ~args:[] ~typ:(non_null guid)
+      ~resolve:(function _ctx -> x.id )
+  ; field "id" ~doc:"Toy ID." ~args:[] ~typppp ~resolve:(function _ctx ->
+        x.id )
+  ; field "id" ~doc:"Toy ID." ~args:[] ~typ:(non_null guid)
+      ~resolve:(function
+      | A -> x.id
```
hhugo: This needs indentation

src/Fmt_ast.ml
```diff
@@ -1420,9 +1425,11 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
           ( wrap_if parens "(" ")"
               (hovbox 0
                  ( hovbox 2
-                     ( hovbox 2
+                     ( ( match c.conf.fun_call with
+                       | `Sparse -> hvbox 2
+                       | `Compact -> hovbox 2 )
```
jberdine: Maybe use `@@` for these applications where the operator is a complex expression

src/Fmt_ast.ml
```diff
@@ -3656,8 +3673,11 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
             $ fmt_extension_suffix c ext
             $ fmt_if_k (Option.is_some in_) (fmt_attributes c ~key:"@" atrs)
             $ fmt_if (first && Poly.(rec_flag = Recursive)) "@ rec"
-            $ fmt " " $ fmt_pattern c xpat
-            $ fmt_fun_args c ~pro:(fmt "@ ") xargs
+            $ fmt " "
+            $ ( match c.conf.fun_decl with
+              | `Compact -> hovbox 0
+              | `Sparse -> hvbox 0 )
```
jberdine: I wonder if it would be better to have a helper function to factor out these matches. Perhaps defined near `groups` to put all these choices together

src/Fmt_ast.ml
```diff
@@ -1425,15 +1428,13 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
...
-                       | `Sparse -> hvbox 2
-                       | `Compact -> hovbox 2 )
-                         ( fmt_args_grouped e0 e1N $ fmt "@ "
-                         $ fmt_label lbl ":"
-                         $ fmt_cmts
-                           @@ hvbox 0
-                                ( fmt "(fun " $ fmt_fun_args c xargs
-                                $ fmt "@ ->" ) )
+                     ( hvbox_if_sparse c.conf.fun_call 2
+                       @@ ( fmt_args_grouped e0 e1N $ fmt "@ "
```
jberdine: Sorry, I guess it's excessive to have a helper and use `@@`

hhugo:
Sorry for the noise, I've messed up quite a bit with rebases.

test/passing/issue289.ml
hhugo: I'm not sure how to fix this one. @jberdine ?

src/Fmt_ast.ml
```diff
@@ -1420,9 +1428,9 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
           ( wrap_if parens "(" ")"
               (hovbox 0
                  ( hovbox 2
-                     ( hovbox 2
+                     ( hvbox_if_sparse c.conf.fun_call 2
                          ( fmt_args_grouped e0 e1N $ fmt "@ "
-                         $ fmt_label lbl ":@,"
+                         $ fmt_label lbl ":"
```
jberdine: Does reverting this change resolve the issue your comment points to?

```diff
-  ; field "id" ~doc:"Toy ID." ~args:[] ~typ:(non_null guid)
-      ~resolve:(function
+  ; field "id" ~doc:"Toy ID." ~args:[] ~typ:(non_null guid) ~resolve:
+      (function
       | A -> x.id
       | B -> c )
```

hhugo: I prefer the version without the break

jberdine: To me that looks as expected. Maybe you still want more indentation though.

src/Fmt_ast.ml
```diff
@@ -1420,9 +1428,9 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
           ( wrap_if parens "(" ")"
               (hovbox 0
                  ( hovbox 2
-                     ( hovbox 2
+                     ( hvbox_if_sparse c.conf.fun_call 2
```
jberdine: Try changing 2 to 4 here or the line above I think.

hhugo: The only thing I came up with is https://github.com/ocaml-ppx/ocamlformat/pull/283/commits/a55ffe1823154a698820d6a3bd9207872541b0f2

src/Fmt_ast.ml
jberdine: I wonder if you would rather disable the special case when the last arg of an apply is a function? Perhaps try commenting out lines 1448-1490, maybe even from 1421, and see if that is preferred. Basically the main point of those special cases is to reduct the indentation when there is a lambda as the last arg of an apply. This is useful for e.g. heavily CPSed code, but doesn't make a huge difference most of the time. The idea is that if it is known to be the last arg, then the indentation of the cases can be less since there isn't a possible confusion between the last case and the next arg.

hhugo: I don't want to give up the special case.
We rely heavily on it when using `List.map` &co

src/Ast.ml
```diff
@@ -140,8 +140,8 @@ let has_trailing_attributes_exp {pexp_desc; pexp_attributes} =
       false
   | _ ->
       List.exists pexp_attributes ~f:(function
-        | {Location.txt= "ocaml.doc" | "ocaml.text"}, _ -> false
-        | _ -> true )
```
jberdine: So that I understand what is desired, in this case the red formatting is preferred, right?

src/Cmts.ml
```diff
@@ -481,9 +481,9 @@ let split_asterisk_prefixed (txt, {Location.loc_start}) =
   let pat =
     String.Search_pattern.create
       (String.init len ~f:(function
-        | 0 -> '\n'
-        | n when n < len - 1 -> ' '
-        | _ -> '*' ))
```
jberdine: And here the red indentation is preferred?

src/Fmt_ast.ml
```diff
@@ -726,19 +729,20 @@ and fmt_core_type c ?(box= true) ?(in_type_declaration= false) ?pro
...
-                   let fmt_cmts = Cmts.fmt c.cmts lab_loc.loc in
-                   fmt_cmts
-                   @@ hvbox 4
-                        ( hvbox 2
-                            ( Cmts.fmt c.cmts lab_loc.loc @@ str lab_loc.txt
-                            $ fmt ":@ "
-                            $ fmt_core_type c (sub_typ ~ctx typ) )
-                        $ fmt_docstring c ~pro:(fmt "@;<2 0>") doc
-                        $ fmt_attributes c ~pre:(fmt " ") ~key:"@" atrs )
-               | Oinherit typ -> fmt_core_type c (sub_typ ~ctx typ) )
```
jberdine: Here too, the red indentation is preferred?

src/Migrate_ast.ml
```diff
@@ -19,8 +19,8 @@ module Parse = struct
   let interface = Parse.interface Versions.ocaml_407
 
   let use_file lexbuf =
-    List.filter (Parse.use_file Versions.ocaml_407 lexbuf) ~f:
-      (fun (p: Parsetree.toplevel_phrase) ->
```
jberdine: In cases like this the indentation is good?

test/passing/issue289.ml
```diff
@@ -0,0 +1,97 @@
...
+  ; field "id" ~doc:"Toy ID." ~args:[] ~typppp ~resolve:(function _ctx ->
+        x.id )
+  ; field
+      "id"
+      ~doc:"Toy ID."
+      ~args:[]
+      ~typ:(non_null guid)
+      ~resolve:(function
+        | A -> x.id
+        | B -> c )
```
jberdine: Indenting less above would be bad, right?

test/passing/issue289.ml
```diff
@@ -0,0 +1,97 @@
...
+  ; field "id" ~doc:"Toy ID." ~args:[] ~resolve:(function
+        | A -> x.id
+        | B -> c )
```
jberdine: But indenting less here would be good, right?

src/Ast.ml
hhugo: Yes

src/Cmts.ml
hhugo: yes

src/Fmt_ast.ml

src/Migrate_ast.ml
hhugo: yes

test/passing/issue289.ml
hhugo: yes

test/passing/issue289.ml
hhugo: yes

hhugo:
I'm happy with the current state.

hhugo:
rebased and squased

src/Fmt_ast.ml
```diff
@@ -1473,14 +1483,16 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
...
-                  $ fmt "(function"
-                  $ fmt_attributes c ~pre:(fmt " ") ~key:"@"
-                      eN.pexp_attributes )
-              $ fmt "@ " $ fmt_cases c ctx'' cs $ fits_breaks ")" " )" )
+                  (hvbox_if_sparse c.conf.fun_call 2
+                     ( fmt_args_grouped e0 e1N $ fmt "@ " $ fmt_label lbl ":"
+                     $ fmt "(function"
+                     $ fmt_attributes c ~pre:(fmt " ") ~key:"@"
+                         eN.pexp_attributes ))
+              $ fmt "@;<1 0>" $ fmt_cases c ctx'' cs $ fits_breaks ")" " )"
```
jberdine: s/`@;<1 0>`/`@ `/ ?

src/Fmt_ast.ml
```diff
@@ -1473,14 +1483,16 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
                is_simple c.conf (fun _ -> 0) (sub_exp ~ctx eI) ) ->
         let e1N = List.rev rev_e1N in
         let ctx'' = Exp eN in
-        hvbox 2
+        hvbox 4
```
jberdine: How would you feel about making this change only if at least one of the new options is enabled? The relative trade-off just does not seem like an overall win to me. It seems that there is much more code that is indented better with 2 than 4 than vice versa.

src/Fmt_ast.ml

src/Fmt_ast.ml
hhugo: seems reasonable

hhugo:
happy to make this a single option. suggestions for a name ?

hhugo:
I've added some support for defining flags with default value been true

src/Conf.ml
```diff
@@ -419,21 +429,32 @@ module Formatting = struct
...
 
 (* Flags that can be modified in the config file that don't affect
    formatting *)
 
-let no_comment_check =
+let comment_check =
   let doc =
-    "UNSAFE: Do not perform any checking of comments and documentation \
-     comments."
+    "UNSAFE: Control wether to check comments and documentation comments."
```
jberdine: Maybe explicitly let-bind default before doc to try to make clear that the doc string is for the inverted flag. This gets tricky.

test/passing/issue289.ml
hhugo: indentation is still weird here. Should we add another box ?

jberdine: This is an example of the case where less indentation is a bit worse, that I don't know how to address. If a box is added around `~resolve:(function...)`, then `~resolve:(function` won't be able to be placed on the previous line unless the whole function expression fits. Since this can only happen on the last arg, I think there is no potential for confusion wrt confusing the body of the function with code that follows, as the following code will be indented less. I think that this case looks a bit unexpected visually, but is not actually confusing.

src/Conf.ml
```diff
@@ -154,17 +155,25 @@ end = struct
...
 
-  let flag ~names ~update ~env ~doc ~allow_inline =
+  let flag ~default ~names ~update ~env ~doc ~allow_inline =
     let open Cmdliner in
-    let term = Arg.(value & flag & info names ~doc ~env) in
+    let invert_flag = default in
+    let names_for_cmdline =
+      if invert_flag then
+        List.filter_map names ~f:(fun n ->
+            if String.length n = 1 then None else Some ("no-" ^ n) )
```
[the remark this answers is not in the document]
hhugo: I would rather wait that we improve & normalise the documentation.

hhugo:
Are you happy if I merge after rebase&squash ?

jberdine:
Something is up with `let rec`:
```
let
    rec sigma_partial_meet' tenv (sigma_acc : Prop.sigma)
        (sigma1_in : Prop.sigma) (sigma2_in : Prop.sigma) : Prop.sigma =
  ()
```

jberdine:
Testing locally, with defaults:
```
let to_loc_trace
      ?(desc_of_source= fun source ->
                          let callsite = Source.call_site source in
                          Format.asprintf "return from %a" Typ.Procname.pp
                            (CallSite.pname callsite))
    ?(source_should_nest= fun _ -> true)
    ?(desc_of_sink= fun sink ->
                      let callsite = Sink.call_site sink in
                      Format.asprintf "call to %a" Typ.Procname.pp
                        (CallSite.pname callsite))
    ?(sink_should_nest= fun _ -> true) (passthroughs, sources, sinks) =
  ()
```
The first arg is indented more (than the others) only with this patch. (This shows that there is a missing break hint before the default values, but that is a pre-existing issue.)

Another case with extra indentation of the first arg (which is also already handled poorly wrt breaking single args):
```
let translate_captured
      { Clang_ast_t.lci_captured_var
      ; lci_init_captured_vardecl
      ; lci_capture_this
      ; lci_capture_kind } ((trans_results_acc, captured_vars_acc) as acc) =
  ()
```
I am not sure whether these changes are just due to the existing poor breaking interacting with the changes, or if there is something unexpected in this patch itself.

hhugo:
Your first example is fixed. together with the break hint for optional arguments.

The second example is unchanged with that feature.

osener:
Works great, thank you @hhugo!

PR #284, opened 2018-08-04
Discussion about whitespaces

hhugo:
One the recurrent feedback I receive is about white-spaces.
- extra white-space before the semi-colon
  - in sequence `e1 ; e2`
- missing white-space before colon in
  - constructor `type t = A: t`
  - record field  `type t = {f: int}`
  - object type `type t = <f: int>`
  - pattern constraint `fun (p: int) -> e`
- missing white-space before equal in
  - record construction `{field= e}`

I'd like to understand the reasons behind the current behaviours. Should this be documented ?
maybe in an faq ?

Assuming that the current behaviour is indented, do you have any suggestion for option name(s) to control that behaviour.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -20,7 +20,9 @@ type t =
   ; escape_chars: [`Decimal | `Hexadecimal | `Preserve]
   ; escape_strings: [`Decimal | `Hexadecimal | `Preserve]
   ; extension_sugar: [`Preserve | `Always]
+  ; field_space: [`Tight | `Loose]
   ; if_then_else: [`Compact | `Keyword_first]
+  ; indicate_nested_or_patterns: bool
   ; infix_precedence: [`Indent | `Parens]
   ; leading_nested_match_parens: bool
   ; let_and: [`Compact | `Sparse]
@@ -30,6 +32,7 @@ type t =
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
   ; quiet: bool
+  ; sequence_style: [`Separator | `Terminator]
   ; type_decl: [`Compact | `Sparse]
   ; wrap_comments: bool
   ; wrap_fun_args: bool }
@@ -52,7 +55,7 @@ end = struct
 
   (** existential package of a Term and a setter for a ref to receive the
       parsed value *)
-  type arg = Arg: 'a Term.t * ('a -> unit) -> arg
+  type arg = Arg : 'a Term.t * ('a -> unit) -> arg
 
   (** convert a list of arg packages to a term for the tuple of all the arg
       terms, and apply it to a function that sets all the receiver refs *)
@@ -133,7 +136,7 @@ end = struct
     -> allow_inline:bool
     -> 'a t
 
-  type pack = Pack: 'a t -> pack
+  type pack = Pack : 'a t -> pack
 
   let store = ref []
 
@@ -312,6 +315,18 @@ module Formatting = struct
     C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
         {conf with extension_sugar= x} )
 
+  let field_space =
+    let doc =
+      "Whether or not to use a space between a field name and the rhs. \
+       This option affects records and objects. Can be set in a config \
+       file with a `field-space = {tight,loose}` line."
+    in
+    let env = Arg.env_var "OCAMLFORMAT_FIELD_SPACE" in
+    let names = ["field-space"] in
+    let all = [("tight", `Tight); ("loose", `Loose)] in
+    C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
+        {conf with field_space= x} )
+
   let if_then_else =
     let doc =
       "If-then-else formatting. Can be set in a config file with an \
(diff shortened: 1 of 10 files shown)


jberdine:
Yes, the current behavior is intended, and these are some of the things that seem to really grab people's individual preferences one way or the other.

These aren't major points I think, but the reasoning behind the choices is along the following lines:

- For sequence expressions, the reasoning is that they are separators, not terminators, and formatting them like terminators is a setup for newcomers who then add a semi after the last expression in the sequence and get confused by the parse or type error. The different formatting to the e.g. C practice seems to be enough of a hint that something is a bit different to help with this a bit. Also, some people think of `;` as analogous to `in`, and `in` also as space both before and after it.

- For not adding a space before colon, one point is that it saves some space, but mainly it is to avoid visually spacing out e.g. record fields from their types further than the type and the name of the next field. Consider `{f: int; g: bool}` vs `{f : int; g : bool}`, in the second one, `g` is visually closer to `int` than to `bool`, and (especially for people familiar with backwards C-style declarations) can look more tightly bound to the preceding type. It is similar for `=` in records.

- For colons in pattern constraints, the current behavior is intended, but I have been thinking that it was a mistake. It was made without considering that the patterns might not be atomic identifiers. Also, in this case, in contrast to the others, `:` does mean `has type` and is logically a binary relation (ignoring typing context, etc.).

src/Fmt_ast.ml
```diff
@@ -1856,8 +1861,15 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
       hvbox 0
         (hvbox_if parens 2
            ( wrap_fits_breaks_if parens "(" ")"
-               (list (sugar_sequence c width xexp) " ;@;<1000 0>"
-                  (fun grp -> list grp " ;@ " (fmt_expression c) ))
+               (list
+                  (sugar_sequence c width xexp)
+                  ( (if c.conf.space_for_sequence then " " else "")
+                  ^^ ";@;<1000 0>" )
```
jberdine: I'd prefer to move `;@;<1000 0>` into the two branches to avoid `^^`

src/Fmt_ast.ml
```diff
@@ -2643,7 +2656,10 @@ and fmt_constructor_arguments c ctx pre args =
           (list lds "@,; " (fmt_label_declaration c ctx))
 
 and fmt_constructor_arguments_result c ctx args res =
-  let pre : _ format = if Option.is_none res then " of@ " else ":@ " in
+  let pre : _ format =
+    if Option.is_none res then " of@ "
+    else (if c.conf.space_before_colon then " " else "") ^^ ":@ "
```
jberdine: Here could we also avoid `^^`

hhugo:
Any suggestion regarding option names ?

src/Conf.ml
```diff
@@ -29,6 +29,9 @@ type t =
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
   ; quiet: bool
+  ; space_before_colon: bool
+  ; space_for_constraint: bool
+  ; space_for_sequence: bool
```
jberdine: maybe `--sequence-style={separator,terminator}`

src/Conf.ml
```diff
@@ -29,6 +29,9 @@ type t =
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
   ; quiet: bool
+  ; space_before_colon: bool
+  ; space_for_constraint: bool
```
jberdine: subsumed by #293 right?

src/Fmt_ast.ml
```diff
@@ -2681,7 +2693,11 @@ and fmt_constructor_arguments c ctx pre args =
           (list lds "@,; " (fmt_label_declaration c ctx))
 
 and fmt_constructor_arguments_result c ctx args res =
-  let pre : _ format = if Option.is_none res then " of@ " else ":@ " in
+  let pre : _ format =
+    if Option.is_none res then " of@ "
+    else if c.conf.space_before_colon then " :@ "
+    else ":@ "
```
jberdine: I think this case is like for constraints: a mistake in the current version, so better to always have the space.

src/Conf.ml
```diff
@@ -29,6 +29,9 @@ type t =
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
   ; quiet: bool
+  ; space_before_colon: bool
```
jberdine: maybe `--field-space={tight,loose}`

src/Conf.ml
hhugo: yes

src/Fmt_ast.ml

src/Conf.ml

src/Conf.ml

hhugo:
I've added another one to the list, indentation in or pattern inside match cases. Name suggestions ?

src/Conf.ml
```diff
@@ -231,6 +234,14 @@ module Formatting = struct
     C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
         {conf with break_string_literals= x} )
 
+  let case_style =
+    let doc = "TODO." in
+    let env = Arg.env_var "OCAMLFORMAT_CASE_STYLE" in
+    let names = ["case-style"] in
```
jberdine: `--no-indicate-nested-or-patterns` ? Note that `--sparse` already does this, but also includes other formatting changes so that (bodies and when guards of) nested or-patterns are not hidden.

hhugo:
squashed and rebased

PR #285, opened 2018-08-04
Improve: configure option for positioning ;;

hhugo:
Also, try to keep adjacent str_item/sig_item and keep at most one blank line between them


src/Conf.ml
```diff
@@ -147,6 +147,19 @@ let escape_strings =
           default
       & info ["escape-strings"] ~doc ~env)
 
+let semi_semi =
+  let doc =
+    ";; position. Can be set in a config file with an `semi-semi \
```
jberdine: s/an/a/

src/Fmt_ast.ml
```diff
@@ -2766,20 +2766,27 @@ and fmt_module_type c ({ast= mty} as xmty) =
...
+               if
+                 dist <= 1
+                 && List.is_empty
+                      (Source.tokens_at c.source
+                         { psig_loc with
+                           loc_start= psig_loc.loc_end
+                         ; loc_end= itm.psig_loc.loc_start } ~filter:
+                         (fun _ -> true ))
+               then fmt "@\n"
+               else fmt "\n@\n"
```
jberdine: Could we have a comment or descriptively-named function indicating what this test is looking for?

hhugo:
The issue with `;; expr` syntax is that it doesn't play well with the toplevel. The toplevel simply ignore everything after `;;`. So for consistency, at JS, we always put `;;` at the end of the previous item.  

The changes about grouping are kind of unrelated and could got in anther PR if you push for it.
We follow the same logic as for consecutive comments, preserving (looking at the source) at most one empty line between item.

jberdine:
I see, the issue with the toplevel is significant.

With this patch I find it a bit odd to have a `;;` immediately following the last of a sequence of structure items when the next one is an eval. Is that the style actually used? I guess separating all structure items with `;;` is out of fashion these days, but what about:
```
let the_last_of_a_sequence_of_str_items = ()

;;
a_pstr_eval
```

Regarding the grouping changes, it is only that I am become a bit uncomfortable about keeping a clear idea of what information about the source locations and concrete syntax is being used to affect formatting. The concern is that any information that is used should be preserved, or else it becomes very fragile when it comes to stabilizing formatting to a fixed-point. The checking done with `Normalize` was previously almost enough to check for this, but now that locations and source text are used more and more, and `Normalize` does not help to check that, we could be heading toward an unstable state. I'm not sure how to achieve it, but it would be good to make the code more explicit about what location and source information is being used and how it is being preserved by the formatting code, so that it is harder to make seemingly innocuous but subtly wrong changes later.

jberdine:
Ok, I think I see, so the rule is to add `;;` at the beginning of the line after each structure item (or just `Pstr_value`?) if either it does not fit on a line or if the next item is a `Pstr_eval`?

It might not be too complicated to change the implementation for `Pstr_value` to emit `;;` when the next item is a `Pstr_eval` regardless of whether the current item breaks.

hhugo:
closing this one, #292 was merged

Issue #272, opened 2018-08-03
Add a way to change ocamlformat configuration using attributes. 


hhugo:
```ocaml
[@@@ocamlformat "extension-point: preserve"]

let x =
  if b then e else " " [@ocamlformat "escape-strings: preserve"]
[@@ocamlformat "if-then-else: keyword-first"]
```

In particular, it would be very useful in conjunction with #271 when formatting code using metaquot.

jberdine:
This could also be an implementation strategy for resolving #263 .

Considering the current way that the `Conf.t` value is threaded through essentially all the formatting code, I think that this should not be difficult to implement.

jberdine:
implemented in #286

Issue #263, opened 2018-08-02
We should be able to locally disable ocamlformat


hhugo:
ocamlformat currently has support for `[@@@ocamlformat.disable]` at toplevel.
`[@@@ocamlformat.disable]` has the effect of disabling ocamlformat for the whole file.

It would be nice to be able to disable ocamlformat locally.
````
[@@@ocamlformat.disable]
let this
    is
       not
           indented
[@@@ocamlformat.enable]

module M = struct
   let f = function 
      | A           |  B -> e
      | CCCC   |  C -> e
   [@@ocamlformat.disable]
end
```

PR #286, opened 2018-08-04
Improve: configure ocamlformat using attributes

hhugo:


hhugo:
fix #272

hhugo:
Also, one should restrict the set of available config in that mode.

hhugo:
The PR also contain a POC for https://github.com/ocaml-ppx/ocamlformat/issues/263

src/Fmt_ast.ml
```diff
@@ -34,6 +34,39 @@ let protect =
...
+                , [] ) } ] ->
+        List.fold ~init:c (String.split ~on:';' str) ~f:(fun c s ->
+            match String.lsplit2 ~on:'=' s with
+            | Some (name, value) -> (
+              try
+                Conf.update c ~name:(String.strip name)
+                  ~value:(String.strip value)
+              with e -> error aname (Exn.to_string e) )
+            | None ->
+                error aname "malformat option, should be [name=value]+" )
```
jberdine: s/malformat/malformed/ ?

jberdine: Would it make sense to share this parsing code with the code that processes .ocamlformat files?

hhugo: the syntax is currently different, are you suggesting to have the same syntax ?

jberdine: Yes, I think it would be easier for users if `--name=value` on the command line, `name=value` in .ocamlformat, and `name=value` as an attribute all meant the same and accepted as close to the same set of options as makes sense. The current config file format was a very quick hack and not really designed.

hhugo: I've refactored the code a bit.
The parsing logic is now shared between cmd-line, config file and attribute.
I also disallow some options in attributes (e.g margin)

hhugo: I've used the same format ocp-indent uses. That is, one line per option, key value separated by`=` (tolerate space for backward compatibility reason). Everything after a `#` is ignored

hhugo:
@jberdine, could you take a look at the refactoring around config

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+    let all =
+      [ ("decimal", `Decimal)
+      ; ("hexadecimal", `Hexadecimal)
+      ; ("preserve", `Preserve) ]
+    in
+    let default = `Preserve in
+    C.choice ~name ~all ~default ~env ~doc ~allow_inline:true ~update:
+      (fun conf x -> {conf with escape_strings= x} )
+
+  let break_infix =
```
jberdine: At one point the options were sorted alphabetically by name excluding any `no_` prefix. Maybe order them since this patch churns them all anyhow. And similarly for the order of the fields in the `t` record type above.

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+    let all = [("compact", `Compact); ("sparse", `Sparse)] in
+    let default = `Compact in
+    C.choice ~name ~all ~default ~env ~doc ~allow_inline:true ~update:
+      (fun conf x -> {conf with type_decl= x} )
+
+  let margin =
+    let docv = "COLS" in
+    let doc =
+      "Format code to fit within $(docv) columns. Can be set in a config \
+       file with a `margin = COLS` line."
```
jberdine: This should mention that it it not supported in attributes.

src/Conf.ml
```diff
@@ -360,172 +486,90 @@ let validate () =
...
+      update ~config ~name:(String.strip var) ~value:(String.strip value)
+        ~inline
+  | [s] -> (
+    match String.split ~on:' ' s with
+    | [] | [""] -> impossible "previous match"
+    | [var; value] ->
+        (* tolerate space separated [var value] to compatibility with older
+           config file format *)
+        update ~config ~name:(String.strip var) ~value:(String.strip value)
+          ~inline
```
jberdine: I'm tempted to issue a deprecation warning to stderr in this case, and remove this fallback in the next version.

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+
+  val int :
+       names:string list
+    -> update:(config -> int -> config)
+    -> default:int
+    -> env:Term.env_info
+    -> doc:string
+    -> docv:string
+    -> allow_inline:bool
+    -> int t
```
jberdine: Maybe dedup these with:
```
  type 'a option_decl =
       update:(config -> 'a -> config)
    -> default:'a
    -> env:Term.env_info
    -> doc:string
    -> allow_inline:bool
    -> 'a t

  val choice : name:string -> all:(string * 'a) list -> 'a option_decl

  val flag : names:string list -> bool option_decl

  val int : names:string list -> docv:string -> int option_decl
```

src/Conf.mli
```diff
@@ -47,3 +47,14 @@ val action : action
 
 val debug : bool
 (** Generate debugging output if true. *)
+
+val parse_config :
+     t
+  -> inline:bool
```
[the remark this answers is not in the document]
hhugo: done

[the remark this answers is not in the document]
hhugo: rather doing it

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+module Formatting = struct
+  let break_string_literals =
+    let doc =
+      "Break string literals. $(b,never) mode formats string literals as \
+       they are parsed, in particular, with escape sequences expanded. \
+       $(b,newlines) mode breaks lines at newlines. $(b,wrap) mode wraps \
+       string literals at the margin. Quoted strings such as \
+       $(i,{id|...|id}) are preserved. Can be set in a config file with a \
+       $(b,break-string-literals) = \
+       $(i,{)$(b,never)$(i,,)$(b,newlines)$(i,,)$(b,wrap)$(i,}) line."
```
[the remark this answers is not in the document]
hhugo: I will open an issue.

src/Conf.mli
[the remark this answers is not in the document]
hhugo: done

src/Conf.ml
```diff
@@ -9,6 +9,24 @@
...
+  ; doc_comments: [`Before | `After]
+  ; parens_tuple: [`Always | `Multi_line_only]
+  ; if_then_else: [`Compact | `Keyword_first]
+  ; infix_precedence: [`Indent | `Parens]
+  ; break_infix: [`Wrap | `Fit_or_vertical]
+  ; type_decl: [`Compact | `Sparse]
+  ; ocp_indent_compat: bool
+  ; quiet: bool
+  ; no_comment_check: bool }
+
```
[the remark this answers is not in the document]
hhugo: moved down

src/Conf.mli
```diff
@@ -47,3 +47,14 @@ val action : action
 
 val debug : bool
 (** Generate debugging output if true. *)
+
+val parse_config :
```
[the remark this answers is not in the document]
hhugo: renamed `parse_line`

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+    -> default:int
+    -> env:Term.env_info
+    -> doc:string
+    -> docv:string
+    -> allow_inline:bool
+    -> int t
+
+  val get : 'a t -> 'a
+
+  val update_config :
```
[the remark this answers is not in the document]
hhugo: `_config` dropped

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+  type config = t
+
+  type 'a t =
+    { names: string list
+    ; parse: string -> 'a
+    ; update: config -> 'a -> config
+    ; allow_inline: bool
+    ; r: 'a ref }
+
+  type pack = Pack: 'a t -> pack [@@ocaml.boxed]
```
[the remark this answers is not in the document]
hhugo: I just wanted the compiler to be silent, didn't really think about it.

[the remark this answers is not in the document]
hhugo: The compiler was warning that I should add an annotation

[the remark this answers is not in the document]
jberdine: Interesting, if I remove it, still no warning. You running 4.07 with the checked-in dune files?

src/Conf.ml
```diff
@@ -60,170 +78,356 @@ end
...
+      with
+      | Some v -> v
+      | None -> user_error (Printf.sprintf "Unknown %s value: %S" name s) []
+    in
+    let r = mk ~default term in
+    let opt = {names= [name]; parse; update; r; allow_inline} in
+    store := Pack opt :: !store ;
+    opt
+
+  let flag ~names ~update ~default ~env ~doc ~allow_inline =
```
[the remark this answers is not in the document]

src/Conf.ml
hhugo: sorted

src/Conf.ml

src/Conf.ml

src/Conf.ml

src/Conf.ml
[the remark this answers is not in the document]
hhugo: It's not needed anymore. I was initially using an inline record

src/Conf.mli
```diff
@@ -11,28 +11,28 @@
 
 (** Configuration options *)
 
-type t = private
```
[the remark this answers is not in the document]
hhugo: don't need it

PR #287, opened 2018-08-04
locally disable ocamlformat

hhugo:


hhugo:
fix #263

src/Fmt_ast.ml
```diff
@@ -3453,110 +3562,116 @@ and fmt_structure_item c ~sep ~last:last_item ?ext ?(use_file= false)
...
-                  ({pexp_desc= Pexp_pack _}, {ptyp_desc= Ptyp_package _})
-              , _ )
-             |Pexp_constraint _, Ppat_constraint _ ->
-                (None, xbody)
-            | Pexp_constraint (exp, typ), _ ->
-                ( Some
-                    (fmt "@ " $ fmt_core_type c ~pro:":" (sub_typ ~ctx typ))
-                , sub_exp ~ctx exp )
-            | _ -> (None, xbody)
+  maybe_disabled c pvb_attributes pvb_loc (fun () ->
```
jberdine: This will format better using `maybe_disabled c pvb_attributes pvb_loc @@ (fun () ->`

src/Fmt_ast.ml
```diff
@@ -34,6 +34,54 @@ let protect =
...
+      error aname
+        (Format.sprintf "unknown suffix %S"
+           (String.chop_prefix_exn ~prefix:"ocamlformat." txt))
+  | _ -> c
+
+let scope c l = {c with conf= List.fold ~init:c.conf l ~f:scope}
+
+let maybe_disabled c attrs loc f =
+  if List.exists attrs ~f:is_disabled then
+    str (Source.string_at c.source loc)
```
jberdine: This will need to remove the comments attached to the locations in the skipped part of the parsetree, right?

src/Fmt_ast.ml
hhugo: indeed

src/Fmt_ast.ml
hhugo: now done

jberdine:
I wonder if the `ocamlformat.disable` attribute should instead be a config option. `[@@ocamlformat "disable"]` seems as usable as `[@@ocamlformat.disable]`, and then `disable` could be put in config files to exclude entire subdirectories, and would make for a more uniform implementation.

hhugo:
There are places where this approach doesn't work.
Sometime the location does not include the attribute resulting in the attribute being dropped.
I didn't have time to look further.

Also, we should probably call Cmts.fmt around ignored block.

src/Conf.ml
```diff
@@ -164,6 +168,14 @@ end = struct
     store := Pack opt :: !store ;
     opt
 
+  let bool_no_cmdline ~name ~update =
```
jberdine: Would it be bad to allow `--disable` on the command line, so that we can document it's intended use as attributes and in config files in the same place with the other options?

src/Fmt_ast.ml
```diff
@@ -2736,6 +2788,7 @@ and fmt_module_type c ({ast= mty} as xmty) =
   let ctx = Mty mty in
   let {pmty_desc; pmty_loc; pmty_attributes} = mty in
   let c = update_config c pmty_attributes in
+  (*maybe_disabled c pmty_loc @@ fun () ->*)
```
jberdine: ?

src/Fmt_ast.ml
```diff
@@ -3133,6 +3201,7 @@ and fmt_module_expr c ({ast= m} as xmod) =
   let ctx = Mod m in
   let {pmod_desc; pmod_loc; pmod_attributes} = m in
   let c = update_config c pmod_attributes in
+  (*maybe_disabled c pmod_loc @@ fun () -> *)
```
jberdine: ?

src/Fmt_ast.ml
```diff
@@ -625,12 +647,14 @@ and fmt_core_type c ?(box= true) ?(in_type_declaration= false) ?pro
   protect (Typ typ)
   @@
   let {ptyp_desc; ptyp_attributes; ptyp_loc} = typ in
+  let c = update_config c ptyp_attributes in
+  maybe_disabled c ptyp_loc
+  @@ fun () ->
```
jberdine: Almost all uses follow this pattern. I wonder if it makes sense to refactor to include`maybe_disabled` into  `update_config` and then write:
```
update_config c attributes
@@ fun c ->
```
instead of:
```
let c = update_config c attributes in
maybe_disabled c loc
@@ fun () ->
```
That might also make it clearer where the config can change but there are no corresponding attributes.

src/Fmt_ast.ml
hhugo: This should return a `block`. I did have time to look at this.

src/Fmt_ast.ml
hhugo: This should return a `block`. I did have time to look at this.

jberdine: Ah yes, ugh, ok.

src/Conf.ml

src/Fmt_ast.ml
hhugo: that make sense

hhugo:
I think I fixed the broken cases  where attributes were not included in locations.
Still need some code cleanup and some more testing.

src/Fmt_ast.ml
```diff
@@ -68,8 +68,59 @@ let update_config c ({txt; loc}, payload) =
...
+  ; esp= Fn.const ()
+  ; epi= None }
+
+let update_config_maybe_disabled c (loc : Location.t) l f =
+  let c = update_config c l in
+  maybe_disabled c loc l f
+
+let update_config_maybe_disabled_block c loc l f =
+  let c = update_config c l in
+  maybe_disabled' c loc l f (fun fmt -> {empty with bdy= fmt})
```
hhugo: this needs testing

src/Fmt_ast.ml
```diff
@@ -90,8 +91,36 @@ let maybe_disabled c (loc : Location.t) f =
...
+let empty =
+  { opn= Fn.const ()
+  ; pro= None
+  ; psp= Fn.const ()
+  ; bdy= Fn.const ()
+  ; cls= Fn.const ()
+  ; esp= Fn.const ()
+  ; epi= None }
+
+let update_config_maybe_disabled c (loc : Location.t) l f =
```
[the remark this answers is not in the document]
hhugo: removed

src/Fmt_ast.ml
```diff
@@ -70,8 +70,9 @@ let update_config c ({txt; loc}, payload) =
 
 let update_config c l = List.fold ~init:c l ~f:update_config
 
-let maybe_disabled c (loc : Location.t) f =
+let maybe_disabled' c (loc : Location.t) l f g =
```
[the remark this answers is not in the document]

src/Source.ml
```diff
@@ -57,6 +57,73 @@ let tokens_at t ?(filter= fun _ -> true) (l : Location.t) :
...
+              if !count = 0 then true else ( Int.decr count ; false )
+          | LBRACKET | LBRACKETBAR | LBRACKETLESS | LBRACKETGREATER
+           |LBRACKETPERCENT | LBRACKETPERCENTPERCENT | LBRACKETAT
+           |LBRACKETATAT | LBRACKETATATAT ->
+              Int.incr count ; false
+          | _ -> false)
+        loc
+    in
+    match l with
+    | None -> assert false
```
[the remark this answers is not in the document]
hhugo: done

test/passing/skip.ml
```diff
@@ -69,3 +69,14 @@ module type S = sig
   val x :  a : t        -> b: t
                -> c : t        -> unit
 end
+
+let x = fun[@ocamlformat "disable"] fc ->
```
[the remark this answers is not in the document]
hhugo: updated

hhugo:
I've done some clean up.

src/Fmt_ast.ml
```diff
@@ -68,8 +68,60 @@ let update_config c ({txt; loc}, payload) =
...
+    let l = String.split ~on:'\n' s in
+    let l =
+      List.mapi l ~f:(fun i s ->
+          if i = 0 then s
+          else
+            drop_while s ~f:(fun i c ->
+                Char.is_whitespace c && i < indent_of_first_line ) )
+    in
+    k (list l "@\n" (fun s -> str s)) )
+  else f c
```
jberdine: Probably easier to read with this short case first

src/Fmt_ast.ml
```diff
@@ -68,8 +68,60 @@ let update_config c ({txt; loc}, payload) =
       !Location.warning_printer loc Caml.Format.err_formatter w ;
       c
 
-let update_config c l =
-  {c with conf= List.fold ~init:c.conf l ~f:update_config}
+let update_config c l = List.fold ~init:c l ~f:update_config
```
jberdine: The shadowing would probably be clearer if the `update_config` above was defined inside the `update_config` here.

src/Fmt_ast.ml
```diff
@@ -68,8 +68,60 @@ let update_config c ({txt; loc}, payload) =
...
+  ; epi: Fmt.t option }
+
+let empty =
+  { opn= Fn.const ()
+  ; pro= None
+  ; psp= Fn.const ()
+  ; bdy= Fn.const ()
+  ; cls= Fn.const ()
+  ; esp= Fn.const ()
+  ; epi= None }
```
jberdine: `block` and `empty` are a bit of an interruption here, maybe better to move them to the top after `type c`

src/Fmt_ast.ml
```diff
@@ -1745,8 +1780,7 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?ext
             (parens || (not (List.is_empty pexp_attributes)))
             "(" ")"
             ( hvbox 2
-                ( fmt_module c keyword name xargs (Some xbody) true xmty
-                    (List.append pmod_attributes pmod.pmod_attributes)
```
jberdine: Maybe I have lost the plot, but why are these attributes no longer needed?

src/Source.ml
```diff
@@ -52,6 +57,74 @@ let tokens_at t ?(filter = fun _ -> true) (l : Location.t) :
...
+      { loc with
+        loc_end= {loc.loc_end with pos_cnum= last_loc.loc_end.pos_cnum} }
+    in
+    let count = ref 0 in
+    let l =
+      find_after t
+        (function
+          | RBRACKET ->
+              if !count = 0 then true else ( Int.decr count ; false )
+          (* It is not clear that a LBRACKET* will ever happen in practive,
```
jberdine: s/that a/that an/  and s/practive/practice/

src/Source.ml
```diff
@@ -52,6 +57,74 @@ let tokens_at t ?(filter = fun _ -> true) (l : Location.t) :
...
+             we're just been defensive here. *)
```
jberdine: s/been/being/

src/Conf.ml
```diff
@@ -632,6 +639,8 @@ let parse_line config ~from s =
            Please use `%s = %s`\n"
           filename lnum name value ;
         update ~config ~from ~name ~value
+    (* special case for disable/enable *)
+    | ["enable"], _ -> update ~config ~from ~name:"disable" ~value:"false"
```
jberdine: This is currently not documented. This PR will necessarily conflict with #302, but maybe add a note to remind us to document the `enable` attribute.

src/Fmt_ast.ml
hhugo: First, `pmod_attributes` and `pmod.pmod_attributes` are the same value.
Second, attributes are already formatted by `fmt_module -> fmt_module_expr c`.
Note that, the change is part of a in a "Fix" commit on its own.

jberdine: Ah I see, ok good, thanks.

hhugo:
rebased & squashed

rbjorklin:
This seems really useful, maybe provide an example of how to use it in the README?

PR #288, opened 2018-08-05
Build: remove uses of `/bin/bash`

smondet:


smondet:
`/usr/bin/env bash` is more portable than `/bin/bash` (which is very often absent: NixOS, OpenWRT, *BSD, ...).

Best is `/bin/sh` I guess (it's POSIX-friendly), but I don't think it works with the `[[ .. ]]` constructs.

I ran into the problem on NixOS.

jberdine:
Oh, sorry, I misread, yes this is good.

PR #293, opened 2018-08-07
Improve: ws before colon for constraint

hhugo:
following comments in #284

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2275,10 +2275,11 @@ and fmt_class_expr c ?eol ?(box= true) ({ast= exp} as xexp) =
            $ hvbox 0 (fmt_class_expr c (sub_cl ~ctx body)) ))
       $ fmt_atrs
   | Pcl_constraint (e, t) ->
-      wrap_fits_breaks "(" ")"
-        ( fmt_class_expr c (sub_cl ~ctx e)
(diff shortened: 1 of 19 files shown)


src/Fmt_ast.ml
```diff
@@ -1518,7 +1517,7 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
   | Pexp_constraint (e, t) ->
       wrap_fits_breaks "(" ")"
         ( fmt_expression c (sub_exp ~ctx e)
-        $ fmt "@ : "
+        $ fmt "@;<1 2>: "
```
jberdine: I suspect that the whole body of the Pexp_constraint case should be wrapped in a box a la Ppat_constraint above, and `@;<1 2>` could go back to `@ `

src/Fmt_ast.ml
```diff
@@ -2277,7 +2277,7 @@ and fmt_class_expr c ?eol ?(box= true) ({ast= exp} as xexp) =
   | Pcl_constraint (e, t) ->
       wrap_fits_breaks "(" ")"
         ( fmt_class_expr c (sub_cl ~ctx e)
-        $ fmt "@ : "
+        $ fmt "@;<1 2>: "
```
jberdine: add box as above?

PR #294, opened 2018-08-07
Improve: preserve extension point formating

### LINKED ISSUE #271: Option to preserve syntax of extension points.
I sometime want to be able to preserve the syntactic difference betrtween 
`[%expr let x in 2]`
and 
`let%expr x in 2`

This is especially important for code using [metaquot](https://github.com/ocaml-ppx/ppx_tools/blob/master/ppx_metaquot.ml)

hhugo:
fix #271

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1903,7 +1903,9 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
                             | Pexp_letmodule _ | Pexp_object _ )
                         ; pexp_attributes= [] } as e1 )
                     , _ ) } as str ) ] )
-    when List.is_empty pexp_attributes ->
+    when List.is_empty pexp_attributes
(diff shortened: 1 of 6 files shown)


src/Fmt_ast.ml
```diff
@@ -1903,7 +1903,11 @@ and fmt_expression c ?(box= true) ?epi ?eol ?parens ?ext ({ast= exp} as xexp)
                             | Pexp_letmodule _ | Pexp_object _ )
                         ; pexp_attributes= [] } as e1 )
                     , _ ) } as str ) ] )
-    when List.is_empty pexp_attributes ->
+    when List.is_empty pexp_attributes
+         && ( Poly.(c.conf.extension_sugar = `Always)
+            || Source_code_position.ascending ext.loc.loc_start
+                 e1.pexp_loc.loc_start
+               > 0 ) ->
```
jberdine: LGTM, though I would always like these things that are sensitive to the concrete syntax to be helpers in `Source` or similar.

Issue #299, opened 2018-08-09
Improve documentation of formatting options.


hhugo:
- [x] factorize the documentation of options regarding the syntax in config files
 (e.g. `Can be set in a 
       config file with an 'wrap-fun-args = {true,false}' line.`)
- [x] review existing options and ensure all values, and meaning of booleans, are documented
- [ ] improve support for boolean flags, changing the default value should update the documentation
- [ ] support short forms for inverted boolean flags, e.g. by changing capitalization
- [x] consider using a single env variable to configure ocamlformat
(e.g. `OCAMLFORMAT=opt1=1,opt2=2 ocamlformat`) 
- [x] divide the 'formatting' options and 'operational' options into separate sections

gpetiot:
Can we consider everything on this list is done? (if I understand correctly the 4th point is the "no-" prefix)

jberdine:
I'm not sure we have reviewed all the existing options documentation, but perhaps. The point about improving the docs for boolean flags is to make the documentation change when the default value is changed, which is not done I think. For inverted short forms, the thought was to use e.g. `-G` for the inverted version of `-g`, similarly to how `--no-goo` is the inverted long form of `--goo`.

hhugo:
We can close, https://github.com/ocaml-ppx/ocamlformat/pull/1062 is merged.

PR #302, opened 2018-08-10
Improve: doc formatting options

gpetiot:
This addresses the first point of issue #299: factorize the documentation of options regarding the syntax in config files.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -100,7 +100,7 @@ module C : sig
     -> allow_inline:bool
     -> 'a t
 
-  val choice : all:(string * 'a) list -> 'a option_decl
+  val choice : all:(string * 'a * string) list -> 'a option_decl
 
   val flag : default:bool -> bool option_decl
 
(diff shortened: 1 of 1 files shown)


src/Conf.ml
```diff
@@ -136,9 +136,37 @@ end = struct
...
+    in
+    let config_file =
+      Format.sprintf
+        "Can be set in a config file with %s '$(b,%s) = %s' line." article
+        name values
+    in
+    let in_attributes =
+      if allow_inline then "" else " Cannot be set in attributes."
+    in
+    Format.sprintf "@,%s@,%s" config_file in_attributes
```
jberdine: I suspect that it would be easier for users if we had a separate section of the man page before the documentation of the options that stated that unless otherwise noted, any option `--option=value` can be set in the config file with an `option=value` line, or with the appropriate attribute. And then for the individual options, we ought to make sure to document all the possible values, e.g. by using `snd Cmdliner_base.enum` (as done for cmdliner's default `--help` option), and also to note the options that can _not_ be set in config files or attributes.

jberdine:
If you could rebase and update the options added to Conf in the meantime (`disable`, I'm not sure if there are any others), then this looks ready to go.

gpetiot:
It seems there is nothing to change for `disable`, anyway I rebased and squashed the commits.

jberdine:
Thanks!

PR #303, opened 2018-08-10
Improve: add option to omit open lines between one-line module items

gpetiot:
Fix issue #133 
Side note: the test is OK with `ocamlformat test/passing/let_decl_compact.ml --let-decl=compact` but not when running `make test` (with the option in the .opts or with inlined in the .ml). I don't see what I am missing.


gpetiot:
Yes it determines if a let is a one-liner based on the source locations. Blank lines between 2 let one-liners are removed (but blank lines between a let and an open or include are preserved). The PR you linked does not seem to fix the issue #133 (I still have the problem on master).

src/Fmt_ast.ml
```diff
@@ -3514,15 +3514,29 @@ and fmt_structure c ctx itms =
...
-          match itm.pstr_desc with
-          | Pstr_include {pincl_mod= me} | Pstr_module {pmb_expr= me} ->
+          match (c.conf.let_decl, itm.pstr_desc) with
+          | _, Pstr_include {pincl_mod= me} | _, Pstr_module {pmb_expr= me} ->
               is_simple_mod me
-          | Pstr_open _ -> true
+          | _, Pstr_open _ -> true
+          | `Compact, Pstr_value _ ->
+              itm.pstr_loc.loc_start.pos_lnum
+              = itm.pstr_loc.loc_end.pos_lnum
```
jberdine: In `Ast.is_simple` the width of formatting the expression without breaking is considered when determining whether an expression "is simple". How would you feel about using a similar approach here instead of using the input source locations? My concern is that if the input has a let binding on one line, but it is too wide to fit within the margin, then the surrounding spacing will be "wrong".

Another point that is not really specific to this change is that for any input file, we should aim for ocamlformat to have the robustness property that formatting the input with margin X and then with margin Y and then again with X should produce the same output as formatting just once with margin X.

Uses of the source locations such as done in this patch make me hesitant because they will lead to this property failing, with the result that formatting is less easily predictable and understandable.

src/Fmt_ast.ml
```diff
@@ -3564,15 +3564,36 @@ and fmt_structure c ctx itms =
...
-          | Pstr_open _ -> true
+          | _, Pstr_open _ -> true
+          | `Compact, Pstr_value (_, bindings) ->
+              let width xe =
+                String.length
+                  (Format.asprintf "%t"
+                     (Cmts.preserve (fun () -> fmt_expression c xe)))
+              in
+              let f acc x =
+                acc && is_simple c.conf width (sub_exp ~ctx x.pvb_expr)
```
jberdine: I'm not sure that it is best to use the same `is_simple` heuristic here as is used for function args, intuitively I think it makes sense for them to be different. Perhaps try something simple like width <= 3/4 margin.

src/Fmt_ast.ml
```diff
@@ -3564,15 +3564,36 @@ and fmt_structure c ctx itms =
...
+              in
+              List.fold bindings ~init:true ~f
+          | _ -> false
+        in
+        let distinct itmI itmJ =
```
jberdine: To me 'distinct' means 'not equal', maybe rename to something longer like 'break_between_open_or_include'?

src/Conf.ml
```diff
@@ -383,6 +384,19 @@ module Formatting = struct
     C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
         {conf with let_and= x} )
 
+  let let_decl =
```
jberdine: I think 'binding' is the more established term than 'declaration'

src/Fmt_ast.ml
```diff
@@ -3564,15 +3564,36 @@ and fmt_structure c ctx itms =
...
+          | `Compact, Pstr_value (_, bindings) ->
+              let width xe =
+                String.length
+                  (Format.asprintf "%t"
+                     (Cmts.preserve (fun () -> fmt_expression c xe)))
+              in
+              let f acc x =
+                acc && is_simple c.conf width (sub_exp ~ctx x.pvb_expr)
+              in
+              List.fold bindings ~init:true ~f
```
jberdine: List.for_all ?

test/passing/let_binding_compact.ml
```diff
@@ -0,0 +1,21 @@
+let z = this one is pretty looooooooooooooooooooooooooooooooooong
+let z = so is this oooooooooooooooooooooooooooooooooooooooooooone
+
+let f x = x + 1
```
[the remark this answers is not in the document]
gpetiot: No it is not intended (but close enough at first). But it is fixed when doing what you suggested (instead of calling `is_simple` at all, just check `(width (sub_exp ~ctx x.pvb_expr)) <= c.margin`), I will revert the changes concerning `is_simple`.

test/passing/let_binding_compact.ml.opts
```diff
@@ -0,0 +1 @@
+--let-binding=compact
```
jberdine: better to change to an attribute

src/Fmt_ast.ml
```diff
@@ -3567,15 +3567,37 @@ and fmt_structure c ctx itms =
...
+          | _, Pstr_open _ -> true
+          | `Compact, Pstr_value (_, bindings) ->
+              let width xe =
+                String.length
+                  (Format.asprintf "%t"
+                     (Cmts.preserve (fun () -> fmt_expression c xe)))
+              in
+              let factor = 4. /. 3. in
+              let f x =
+                is_simple ~factor c.conf width (sub_exp ~ctx x.pvb_expr)
```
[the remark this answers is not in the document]
jberdine: In Compact mode, I do not understand the conceptual reason why `Pstr_value` is treated specially. Wouldn't it make sense to format sequences of one-liners without intervening open lines no matter what form of structure item they are? Perhaps only group items of the same form together, with exceptions to combine `Pstr_type` with `Pstr_typext` and `Pstr_module` with `Pstr_recmodule`. That is, the `is_simple_mod` case above could be done with the width instead in Compact mode, and `break_between_open_or_include` below could be generalized/simplified.

test/passing/let_binding_compact.ml
```diff
@@ -0,0 +1,21 @@
...
+  | `xxxxxxxxxxxxxxxxxx ->
+      yyyyyyyyyyyyyyyyyyyyyyyy
+        zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
+        zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
+        zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
+
+let x = 1
+let y = 2
+let z = this one is pretty looooooooooooooooooooooooooooooooooong
+let z = so is this oooooooooooooooooooooooooooooooooooooooooooone
```
jberdine: It would be good to add to this test so that it covers the interaction between different forms of structure items.

src/Fmt_ast.ml
```diff
@@ -3635,15 +3635,47 @@ and fmt_structure c ctx itms =
...
+                let last = false in
+                String.length
+                  (Format.asprintf "%t"
+                     (Cmts.preserve (fun () -> fmt_structure_item ~last c i)))
+              in
+              width (sub_str ~ctx itm) <= c.conf.margin
           | _ -> false
         in
+        let group itmI itmJ =
+          match (c.conf.let_binding, itmI.pstr_desc, itmJ.pstr_desc) with
```
gpetiot: I modified this function (formerly `break_between_open_or_include`), is this what you had in mind ? (not much shorter)

jberdine: Yes, basically. Some of the repetition of Compact could be avoided with a nested or-pattern, but not a big deal.

Regarding the actual grouping, before this patch simple `open` and `module` items are formatted without an open line between, which might be good to preserve. I am not sure what the best grouping is though. In practice my feeling is that `open`, `include`, and `module` (at least for cases like `module A = B`) are often conceptually good to group together. But it may take some experimentation.

src/Conf.ml
```diff
@@ -454,6 +455,20 @@ module Formatting = struct
     C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
         {conf with let_and= x} )
 
+  let let_binding =
```
jberdine: Oh, and at this point, this name is too specific, maybe `structure_item_grouping` instead?

gpetiot:
There are no more errors when running `OCAMLFORMAT_STRUCTURE_ITEM_GROUPING=compact make -C test test`, but there are still some (minor) formatting differences due to docstrings.

src/Fmt_ast.ml
```diff
@@ -3639,14 +3639,7 @@ and fmt_structure c ctx itms =
...
           | _, Pstr_open _ -> true
-          | `Compact, _ ->
-              let width i =
-                let last = false in
-                String.length
-                  (Format.asprintf "%t"
-                     (Cmts.preserve (fun () -> fmt_structure_item ~last c i)))
-              in
-              width (sub_str ~ctx itm) <= c.conf.margin
+          | `Compact, _ -> Location.width itm.pstr_loc <= c.conf.margin
```
gpetiot: `fmt_structure_item` lost comments even though `Cmts.preserve` is used, I don't understand why. `Location.width` looks like a good replacement, and the short-multi-liners / long-one-liners tests all pass.

src/Fmt_ast.ml
```diff
@@ -3639,14 +3639,7 @@ and fmt_structure c ctx itms =
           | _, Pstr_include {pincl_mod= me} | _, Pstr_module {pmb_expr= me} ->
               is_simple_mod me
           | _, Pstr_open _ -> true
-          | `Compact, _ ->
-              let width i =
-                let last = false in
-                String.length
-                  (Format.asprintf "%t"
-                     (Cmts.preserve (fun () -> fmt_structure_item ~last c i)))
```
jberdine: The argument function to preserve does not get called until asprintf runs and passes it the formatting stream. At that time preserve has already reset the remove flag, and so preserve doesn't work. Sorry, this is an existing but unknown bug in preserve. I'm looking into a fix.

jberdine:
@gpetiot I pushed a commit onto the tip of your branch, with it, and rebased onto #303, my local testing works and the output looks pretty good I think.

jberdine:
So you're right that using the formatted width still doesn't work correctly. See the added test for nested structures in my latest commit. It is an unfortunate performance hit to have to reformat the entire file again whenever changing an item to or from a one-liner. I expect that this feature could be implemented with careful nesting of boxes and breaks, but that approach is almost surely more involved.

hhugo:
Could we have a bit more tests. In particar, adjacent module alias and open

src/Conf.ml
```diff
@@ -514,6 +515,20 @@ module Formatting = struct
...
+  let structure_item_grouping =
+    let doc = "Line breaks between structure items." in
+    let env = Arg.env_var "OCAMLFORMAT_STRUCTURE_ITEM_GROUPING" in
+    let names = ["structure-item-grouping"] in
+    let all =
+      [ ( "sparse"
+        , `Sparse
+        , "$(b,sparse) will always break a line between two structure items."
+        )
+      ; ("compact", `Compact, "$(b,compact) will not.") ]
```
jberdine: Maybe 'compact will not leave open lines between one-liners of similar sorts.'

src/Fmt_ast.ml
```diff
@@ -3710,16 +3710,48 @@ and fmt_structure c ctx itms =
...
-          match itm.pstr_desc with
-          | Pstr_include {pincl_mod= me} | Pstr_module {pmb_expr= me} ->
-              is_simple_mod me
-          | Pstr_open _ -> true
-          | _ -> false
+        let is_simple (itm, c) =
+          match c.conf.structure_item_grouping with
+          | `Compact -> (
+            match itm.pstr_desc with
+            | Pstr_module {pmb_expr= me} -> is_simple_mod me
```
jberdine: Why add back this case? Is there an example of poor output?

src/Fmt_ast.ml
```diff
@@ -3710,16 +3710,47 @@ and fmt_structure c ctx itms =
...
-          match itm.pstr_desc with
-          | Pstr_include {pincl_mod= me} | Pstr_module {pmb_expr= me} ->
-              is_simple_mod me
-          | Pstr_open _ -> true
-          | _ -> false
+        let is_simple (itm, c) =
+          match c.conf.structure_item_grouping with
+          | `Compact ->
+              Location.width itm.pstr_loc <= c.conf.margin
+              && Location.is_single_line itm.pstr_loc
```
jberdine: Using `is_simple_mod` here is probably too specific, it seems better to check that the location is narrow and on one line to catch all cases where there is a force-break in `itm`.

test/passing/structure_item_grouping.ml.opts
```diff
@@ -0,0 +1 @@
+--max-iter=3
```
jberdine: This is needed since the implementation based on checking locations requires reformatting the whole file again when something is changed to or from a one-liner.

gpetiot: Thanks! Did not think of this.

src/Fmt_ast.ml
gpetiot: It seems good to me, thank you.

gpetiot:
I think it is better to keep the formatting of .mli files consistent with .ml files, so my last commit applies this option on signatures (I don't know if it implies the name of the option should change).

src/Conf.ml
```diff
@@ -514,6 +515,23 @@ module Formatting = struct
     C.choice ~names ~all ~env ~doc ~allow_inline:true ~update:(fun conf x ->
         {conf with sequence_style= x} )
 
+  let structure_item_grouping =
+    let doc = "Line breaks between structure items." in
```
jberdine: Better to rename, perhaps `module-item-spacing` (thinking again, 'grouping' is probably an implementation detail). And maybe reword doc to `Spacing between items of structures and signatures.`

src/Conf.ml
```diff
@@ -514,6 +515,23 @@ module Formatting = struct
...
+    let env = Arg.env_var "OCAMLFORMAT_STRUCTURE_ITEM_GROUPING" in
+    let names = ["structure-item-grouping"] in
+    let all =
+      [ ( "sparse"
+        , `Sparse
+        , "$(b,sparse) will always break a line between two structure items."
```
jberdine: maybe reword to match other case: 'will always leave an open line between two items of a structure or signature`

src/Fmt_ast.ml
```diff
@@ -3710,16 +3740,47 @@ and fmt_structure c ctx itms =
           | Pmod_ident _ -> true
           | _ -> false
```
jberdine: For readability, it would be good to refactor `has_doc` to a separate function, probably in `Ast`, and generalized from structure_item to Ast.t, with the implementation for at least the signature_item added.

src/Fmt_ast.ml
```diff
@@ -3710,16 +3740,47 @@ and fmt_structure c ctx itms =
...
+              Location.width itm.pstr_loc <= c.conf.margin
+              && Location.is_single_line itm.pstr_loc
+          | `Sparse -> (
+            match itm.pstr_desc with
+            | Pstr_include {pincl_mod= me} | Pstr_module {pmb_expr= me} ->
+                is_simple_mod me
+            | Pstr_open _ -> true
+            | _ -> false )
+        in
+        let allow_adjacent (itmI, cI) (itmJ, cJ) =
```
jberdine: To make it easier to keep structures and signatures in sync, it would help to refactor this and the analogous function for signatures above to one that operates on Ast.t. This probably should go near the beginning of this module, after the sugar_ functions.

src/Ast.ml
```diff
@@ -116,6 +116,194 @@ let rec is_sugared_list exp =
...
+        Location.width itm.pstr_loc <= c.Conf.margin
+        && Location.is_single_line itm.pstr_loc
+    | `Sparse -> (
+      match itm.pstr_desc with
+      | Pstr_include {pincl_mod= me} | Pstr_module {pmb_expr= me} ->
+          is_simple_mod me
+      | Pstr_open _ -> true
+      | _ -> false )
+
+  let allow_adjacent (itmI, cI) (itmJ, cJ) =
```
jberdine: It is not a big deal, but note that for my comment about making it easier to keep structures and signatures in sync, the idea was to have _one_ function that handled them both, to make it clearer that conceptually they are one thing.

Also, note that while I don't have any issue with this module style, the rest of the existing code that needs to operate over different types of phrase is all expressed in terms of Ast.t.

src/Ast.ml
```diff
@@ -116,6 +116,194 @@ let rec is_sugared_list exp =
...
+    | _ -> false
+
+  let is_simple (itm, c) =
+    match c.Conf.module_item_spacing with
+    | `Compact ->
+        Location.width itm.pstr_loc <= c.Conf.margin
+        && Location.is_single_line itm.pstr_loc
+    | `Sparse -> (
+      match itm.pstr_desc with
+      | Pstr_include {pincl_mod= me} | Pstr_module {pmb_expr= me} ->
```
jberdine: maybe clearer to move definition of `is_simple_mod` here to emphasize that it is just a helper for this case

test/passing/module_item_spacing.ml.ref
```diff
@@ -0,0 +1,69 @@
...
+module B = BB
+open AA
+module C = CC
+
+module M =
+  X
+    (Y)
+    (struct
+      let x = k
+    end)
```
samoht: is there a way to turn this into a one-liner?

jberdine: This breaks because structures are force-broken. This could be made configurable, with a 2-line change in the `Pmod_structure` case of `fmt_module_expr` to put `@<1000 N>` instead of `@<1 N>` only when the config option is set to sparse.

jberdine: I basically have the implementation already, I can send a PR later today.

test/passing/module_item_spacing.ml
```diff
@@ -0,0 +1,71 @@
...
+let x = 1
+let z =
+  this
+    one
+    (is short)
+let y = 2
+let w = this one is toooooooooooooooooooooooooo (looooooooooooooooooooooooog but is (originally a one-liner))
+let k = z
+
+
```
hhugo: Do you expect this double empty line ?

hhugo: Could be due to `(fits_breaks ~force_fit_if:last_item "" "\n"))` line 3777 or so

jberdine:
These have .ref files, which look ok right?

jberdine:
Thanks!

PR #307, opened 2018-08-11
Fix: missing parens around labeled record pattern arg

jberdine:


src/Ast.ml
```diff
@@ -1329,11 +1329,14 @@ end = struct
       , inner ) -> (
       match inner with
       | Ppat_construct ({txt= Lident "::"}, _) -> true
-      | Ppat_construct _ | Ppat_variant _ -> false
+      | Ppat_construct _ | Ppat_record _ | Ppat_variant _ -> false
```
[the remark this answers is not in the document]
jberdine: Without this, then the other changes in this patch cause every record in a sugared list to be needlessly parenthesized. Added a test to check that extra parens are removed.

src/Ast.ml
```diff
@@ -1329,11 +1329,12 @@ end = struct
...
       | _ -> true )
     | ( Pat {ppat_desc= Ppat_construct _}
       , Ppat_construct ({txt= Lident "::"}, _) ) ->
         true
+    | Exp {pexp_desc= Pexp_fun (Optional _, _, _, _)}, Ppat_record _ -> true
```
jberdine: Entertainingly, this is only needed for Optional, but not Labelled args.

test/passing/record.ml
```diff
@@ -27,3 +27,5 @@ let _ = {(x : x) with e1; e2}
 let _ = {(x :> x) with e1; e2}
 
 let _ = {(x#x) with e1; e2}
+
+let f ?l:({f; g}) = e
```
hhugo: can you add a test with a non optional label

jberdine: done

PR #314, opened 2018-08-12
Improve: pexp_open as final arg of infix op

jberdine:


src/Source.ml
```diff
@@ -64,6 +64,16 @@ let contains_IN_token_between t ~(from : Location.t) ~(upto : Location.t) =
   Source_code_position.ascending from.loc_start upto.loc_start < 0
   && (not (List.is_empty (tokens_between t ~from ~upto ~filter)))
 
+let is_long_pexp_open (conf : Conf.t) source {Parsetree.pexp_desc} =
+  match pexp_desc with
+  | Pexp_open (_, {loc}, {pexp_loc}) -> (
+    match conf.let_open with
+    | `Auto | `Long -> true
```
hhugo: Shouldn't `Auto do the same thing as `Preserve  ?

jberdine: This function is probably poorly named. It really means "if this breaks, will its body be in a non-indented form". So for the Auto case, when it breaks it will use the `let open` form, whose body is not indented. Perhaps it would be clearer to move the branch on `conf.let_open` to the use site in `Fmt_ast`.

Issue #301, opened 2018-08-09
Functor arguments are wrapped incorrectly in some cases


osener:
I stumbled upon a case where functor arguments don't get wrapped at the ideal spots. Tested with [`master`](https://github.com/ocaml-ppx/ocamlformat/tree/b2e0d0f3fe0f39a814c4abe59d21ad4909f9b029).

```ocaml
(* This declaration looks odd *)
type request_token =
  Sociaml_oauth_client.Client.Make(Sociaml_oauth_client.Posix.Clock)(Sociaml_oauth_client.
                                                                     Posix.
                                                                     MAC_SHA1)(Sociaml_oauth_client.
                                                                               Posix.
                                                                               Random).
  request_token

(* Whereas this one works well *)
module OauthClient =
  Sociaml_oauth_client.Client.Make
    (Sociaml_oauth_client.Posix.Clock)
    (Sociaml_oauth_client.Posix.MAC_SHA1)
    (Sociaml_oauth_client.Posix.Random)
```

PR #315, opened 2018-08-14
Improve wrapping of functor args in type declaration (issue #301)

gpetiot:


test/passing/wrapping_functor_args.ml
```diff
@@ -0,0 +1,12 @@
+(* This declaration looks odd *)
+type request_token =
+  Sociaml_oauth_client.Client.Make(Sociaml_oauth_client.Posix.Clock)
+    (Sociaml_oauth_client.Posix.MAC_SHA1)
+    (Sociaml_oauth_client.Posix.Random).request_token
```
jberdine: This is clearly better than before, but I think `.request_token` ought to be on the next line since it selects from the functor application, not the last arg. Agree?

gpetiot: I agree, but if I put a cut here it breaks some tests, I will try to fix it.

test/passing/module_attributes.ml
```diff
@@ -14,5 +14,5 @@ include [%ext] [@warning "item"] [@@warning "structure"]
 
 include (val M) [@warning "item"] [@@warning "structure"]
 
-include ( val Aaaaaaaaaaaaaaaa.Bbbbbbbbbbbbbbbb.Cccccccccccccccc.
-              Dddddddddddddddd ) [@warning "item"] [@@warning "structure"]
+include ( val Aaaaaaaaaaaaaaaa.Bbbbbbbbbbbbbbbb.Cccccccccccccccc
+              .Dddddddddddddddd ) [@warning "item"] [@@warning "structure"]
```
jberdine: A separate issue for later, but it would probably be better to break before the attributes unless its a one-liner

test/passing/source.ml.ref
```diff
@@ -8186,8 +8186,8 @@ module USERCODE (X : TYPEVIEW) = struct
   module type F = functor (C : CORE with type V.usert = X.combined) -> BARECODE
```
jberdine: A separate issue for later, but it looks like we're missing a break hint after the `->`

hhugo: Before or after

jberdine: I assumed after, but I don't have large body of code using such explicit functors to get a good feel for what works best.

PR #320, opened 2018-08-15
Improve: no parens for trailing 'not'

hhugo:
refines https://github.com/ocaml-ppx/ocamlformat/pull/275


hhugo:
I'd like to have feedback from @mbarbin @mshinwell

hhugo:
The merge broke master, `make fmt` is needed. Will push a fix in an hour or so if not already fixed

PR #321, opened 2018-08-15
batch of fixes

hhugo:


hhugo:
These were discovered formatting the bucklescript codebase.

src/Ast.ml
```diff
@@ -1628,6 +1628,10 @@ end = struct
       | Pexp_apply (({pexp_desc= Pexp_new _; _} as exp2), _)
         when exp2 == exp ->
           false
+      | Pexp_apply
+          ({pexp_desc= Pexp_ident {txt= Lident txt; _}}, (_, first) :: _)
+        when String.is_prefix ~prefix:"#" txt && (not (first == exp)) -> (
+        match exp.pexp_desc with Pexp_ident _ -> false | _ -> true )
```
jberdine: Ok, so the hash infix ops have a different precedence than the others. I think it would be better to fix this by adding `&& Poly.(prec_ast ctx < Some Apply)` to the `when` guard in `is_right_infix_arg` below.

src/Fmt_ast.ml
```diff
@@ -132,18 +132,27 @@ let rec sugar_arrow_typ c ({ast= typ} as xtyp) =
...
+  let {pcty_desc; pcty_loc; pcty_attributes} = typ in
   match pcty_desc with
   | Pcty_arrow (l, t1, t2) ->
       Cmts.relocate c.cmts ~src:pcty_loc ~before:t1.ptyp_loc
         ~after:t2.pcty_loc ;
-      (l, `core_type (sub_typ ~ctx t1))
-      :: sugar_class_arrow_typ c (sub_cty ~ctx t2)
+      let rest =
+        match pcty_attributes with
+        | [] -> sugar_class_arrow_typ c (sub_cty ~ctx t2)
```
jberdine: In the other cases, it is the attributes of the argument passed to `sugar_` that are checked for emptiness. This case is different, intentionally?

src/Fmt_ast.ml
```diff
@@ -132,18 +132,27 @@ let rec sugar_arrow_typ c ({ast= typ} as xtyp) =
   | Ptyp_arrow (l, t1, t2) ->
       Cmts.relocate c.cmts ~src:ptyp_loc ~before:t1.ptyp_loc
         ~after:t2.ptyp_loc ;
-      (l, sub_typ ~ctx t1) :: sugar_arrow_typ c (sub_typ ~ctx t2)
+      let rest =
+        match t2.ptyp_attributes with
+        | [] -> sugar_arrow_typ c (sub_typ ~ctx t2)
```
jberdine: Would it make sense to move the check for empty attributes into `sugar_` instead?

src/Fmt_ast.ml
hhugo: it is not, thanks

src/Fmt_ast.ml
hhugo: I'm not sure what you mean by this. I think we should rewrite theses functions differently to make it clear that they should return a list of size at least 2

jberdine: I was thinking just for `sugar_` to check if the attributes are empty and if not, don't "sugar" and instead return the arg unchanged. That seems to be a common pattern in these changes: either the attributes are empty and the arg is sugared, or the attributes are non-empty and the arg is left unchanged. But maybe the size condition is better.

src/Ast.ml
```diff
@@ -1675,7 +1675,8 @@ end = struct
             match ctx_desc with
             | Pexp_apply
                 ({pexp_desc= Pexp_ident {txt= Lident i}}, _ :: (_, e2) :: _)
-              when e2 == exp && is_infix_id i ->
+              when e2 == exp && is_infix_id i
+                   && Poly.(prec_ast ctx < Some Apply) ->
```
jberdine: Thinking about this again, I'm not sure my suggestion isn't risky. It would be safer to check that `prec_ast ctx` is `Some p` where `p < Apply`.

hhugo:
Will refactor later. thanks

PR #324, opened 2018-08-17
Improve: attributes on let bindings

jberdine:


jberdine:
@hhugo, this basically just reverts your 
https://github.com/ocaml-ppx/ocamlformat/commit/06a5fc43d978122ceb533ef49f278dd588637aff and updates tests. I'm seeing regressions like the test added to attributes.ml. I don't understand the original fix, as the regtests don't seem to reveal the original problem. Help?

hhugo:
This looks fine to me. I can't remember the reason for 06a5fc4 but based on the commit name, I must have mis-interpreted an error.

PR #325, opened 2018-08-17
Improve: let-module construct 

hhugo:


test/passing/skip.ml
```diff
@@ -93,14 +93,16 @@ let _ = (x [@ocamlformat "disable"]  [@test? _ when e [@test 2]]) 3
 let _ =
   let module X = struct
     let x = 4
-  end in
+  end
+  in
```
jberdine: Do you consider this an improvement? Or is this change to recover a bit from the effects of `disable` in the cases below?

hhugo: It looks more consistent with other `let-in` where `in` fall on its own line in case of a break.

hhugo: To be clear, the change about `let module M = struct end in` was not the target of that PR. Maybe I should restrict the scope and special case to remove this side-effect.

jberdine: Ok, then I am not sure of the intended change, can you point to an example?

hhugo: well, the regtest `module.ml` that was removed from failing and that I forgot add into passing :)

test/passing/source.ml.ref
```diff
@@ -3464,7 +3472,8 @@ type ('k, 'd, 'm) map =
 let add (type k d m) (m : (k, d, m) map) x y s =
   let module M = ( val m
                      : MapT
-                     with type key = k and type data = d and type map = m ) in
+                     with type key = k and type data = d and type map = m )
+  in
```
hhugo: this is what this PR targets

test/passing/skip.ml
jberdine: Ok, let me look

hhugo:
I've added a special case for Pmod_structure

jberdine:
What do you think of #328 instead?

hhugo:
lgtm

PR #326, opened 2018-08-17
Improve: sugar for nestest module_with

hhugo:


src/Fmt_ast.ml
```diff
@@ -2968,19 +2980,21 @@ and fmt_module_type c ({ast= mty} as xmty) =
...
       { empty with
         bdy=
           hvbox 0
             (wrap_if parens "(" ")"
                ( opn $ Option.call ~f:pro $ psp $ bdy $ cls $ esp
                $ Option.call ~f:epi
-               $ list_fl wcs (fun ~first ~last:_ wc ->
-                     fmt_or first "@ with" "@;<1 1>and"
-                     $ fmt_with_constraint c ctx wc ) ))
+               $ list_fl (List.rev wcs)
```
jberdine: The other `sugar_` functions don't return reversed results, so it might be clearer to move this `List.rev` into `sugar_mod_with`

PR #328, opened 2018-08-17
Improve: let-module-in break before `in`

jberdine:


jberdine:
An alternative to #325

PR #329, opened 2018-08-17
Fix: Cmts.preserve

jberdine:


jberdine:
@gpetiot, try to rebase #303 on this and see if you still get dropped comments.

src/Fmt_ast.ml
```diff
@@ -1186,10 +1186,7 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?ext
   let fmt_cmts = Cmts.fmt c.cmts ?eol pexp_loc in
   let fmt_atrs = fmt_attributes c ~pre:(fmt " ") ~key:"@" pexp_attributes in
   let parens = match parens with Some b -> b | None -> parenze_exp xexp in
-  let width xe =
-    String.length
-      (Format.asprintf "%t" (Cmts.preserve (fun () -> fmt_expression c xe)))
-  in
+  let width xe = String.length (Cmts.preserve (fmt_expression c) xe) in
```
jberdine: Here, the intent is to make it hard/impossible to mistakenly pass a fully-applied formatting function to preserve. @hhugo @gpetiot do you think this is enough?

gpetiot: For my PR it is not enough.
If I replace `Location.width itm.pstr_loc <= c.conf.margin` with:
```
              let width i =
                let last = false in
                String.length (Cmts.preserve (fmt_structure_item ~last c) i)
              in
              width (sub_str ~ctx itm) <= c.conf.margin
```
I have a slightly worsed formatting:
```
let w =
  this one is toooooooooooooooooooooooooo
    (looooooooooooooooooooooooog but is (originally a one - liner))
```
is considered a one-liner and is packed with the preceding and following lines.
And I still have the "formatting changed ast" errors on `OCAMLFORMAT_STRUCTURE_ITEM_GROUPING=compact make -C test test`.

src/Fmt_ast.ml
```diff
@@ -631,20 +631,18 @@ let doc_atrs atrs =
...
-  opt doc (fun (({txt; loc} as doc), floating) ->
+  opt doc (fun ({txt; loc}, floating) ->
       let epi =
         match epi with
         | Some _ -> epi
         | None when floating -> Some (fmt "@,")
         | None -> None
       in
-      fmt_if_k
-        (not (Cmts.doc_is_dup c.cmts doc))
```
jberdine: @gpetiot this stale code is what was causing the lost docstrings in #303

src/Cmts.ml
```diff
@@ -402,29 +393,34 @@ let init_use_file =
...
-  update_multi t.cmts_before src before ~f:(fun src_cmts dst_cmts ->
-      List.append src_cmts dst_cmts ) ;
-  update_multi t.cmts_after src after ~f:(fun src_cmts dst_cmts ->
-      List.append dst_cmts src_cmts )
-
 let remove = ref true
 
-let preserve f x =
+let preserve fmt_x x =
+  let buf = Buffer.create 128 in
```
jberdine: @gpetiot the previous version was buggy in that it always used the same buffer, which for nested expressions was just wrong

src/Fmt_ast.ml
hhugo: I wonder if we should fix it instead . I expect this line to be useful together with [--no-check-comments]

jberdine: Perhaps, I'm not sure. It depends on the intended use of `--no-check-comments` I think. To me, it is basically just a debug option to see what the output is, but it never to be actually used. For only that, I'd rather have simpler code with less global state. But maybe you have other uses of `--no-check-comments`?

hhugo: Well, I have been using that option the format the whole compiler code base (https://github.com/hhugo/ocaml/tree/auto) as an experiment/demo.
Thinking more about it, I think this dedup, if any, should be done using an ast_mapper.

jberdine: Yes, as an ast_mapper it would be more predictable wrt which of the duplicates is actually printed, rather than relying on whichever one happened to be hit first by the evaluation of what ought to be a formatting stream * state monad.

PR #330, opened 2018-08-17
Improve: comments before match

jberdine:


hhugo:
Was this change made to prevent comments on the left of the 'match' to be formatted on the right ?
Would someone be sad if we were to preserve comments position relative to the `match` keywork ?

jberdine:
The motivation is to keep cases like:
```
(* comment about entire pattern matching *)
match e with
| ...
```
from being converted to:
```
match
  (* comment about entire pattern matching *)
  e
with
| ...
```
The trouble is really that `match` does not have a location. Or perhaps a different solution would be to attach the comment to the following location only if there aren't keyword tokens in between.

jberdine:
In particular, the reverse conversion of the above example is not itself desired, so I think preserving position wrt `match` would be good.

PR #333, opened 2018-08-19
Improve: breaking before arrows in pattern match cases

jberdine:


src/Ast.ml
```diff
@@ -462,12 +462,14 @@ end = struct
           | Pcf_inherit (_, _, _) -> false
           | Pcf_val (_, _, Cfk_virtual t) -> typ == t
           | Pcf_val
-              (_, _, Cfk_concrete (_, {pexp_desc= Pexp_constraint (_, t)})) ->
+              (_, _, Cfk_concrete (_, {pexp_desc= Pexp_constraint (_, t)}))
+            ->
```
jberdine: these used to over-shoot the margin

PR #334, opened 2018-08-19
Alternative formatting for string literals

hhugo:
This new formatting aims at preserving the visual aspect of new lines and leading whitespaces.
For examples,
```
          "File %S, line %d:\n  \
           Warning: Using deprecated ocamlformat config syntax.\n\
```
would become 
```
          "File %S, line %d:\n
          \  Warning: Using deprecated ocamlformat config syntax.\n"
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -548,34 +548,20 @@ let fmt_constant c ~loc ?epi const =
       let fmt_lines lines =
         hvbox 1
           ( str "\""
-          $ list_pn lines (fun ?prev curr ?next ->
-                let drop = function ' ' | '\t' -> true | _ -> false in
-                let line =
-                  if Option.is_none prev then curr
-                  else String.lstrip ~drop curr
-                in
-                fmt_line line
+          $ list_pn lines (fun ?prev:_ curr ?next ->
+                fmt_line curr
                 $ opt next (fun next ->
-                      let spc =
-                        match
-                          String.lfindi next ~f:(fun _ c -> not (drop c))
-                        with
-                        | Some 0 -> ""
-                        | Some i ->
-                            escape_string c (String.sub next ~pos:0 ~len:i)
-                        | None -> escape_string c next
-                      in
-                      fmt "\\n"
-                      $ fmt_if_k
-                          (not (String.is_empty next))
-                          (str spc $ pre_break 0 "\\" 0) ) )
+                      if String.is_empty next then fmt "\\n"
+                      else if Char.equal next.[0] ' ' then
+                        fmt "\\n" $ pre_break 0 "\\" (-1) $ if_newline "\\"
+                      else fmt "\\n" $ pre_break 0 "\\" 0 ) )
           $ str "\"" $ Option.call ~f:epi )
       in
       let s =
         match (c.conf.break_string_literals, c.conf.escape_strings) with
         | `Never, `Preserve -> Source.string_literal c.source `Preserve loc
         | (`Newlines | `Wrap), `Preserve ->
-            Source.string_literal c.source `Normalize_nl loc
+            Source.string_literal c.source `Normalize loc
         | _ -> s
       in
       match c.conf.break_string_literals with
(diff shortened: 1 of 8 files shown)


jberdine:
I haven't looked at this in detail yet, but would it make sense to work this into the options for `--break-string-literals`?

The idea here is to make the strings in the source look less like the output, but make it more explicit what leading space is in the literal, right?

hhugo:
The idea is to have the string look **more** like the output by having leading ws at the beginning.

jberdine:
Ok, yes, I see now.
I don't think this would need to be under an option, at least until someone complains. Paging @jvillard for comment. It would be nice to remove the existing fiddly code that maintains the "leading" whitespace.
Re hitting the left margin:
```
;;
"File %S, line %d:\
\nWarning: Using deprecated ocamlformat config syntax.\
\nPlease use `%s = %s`\
\n"
```
Seems rather rare and not so bad so I'm disinclined to fret over it.

hhugo:
Before, this become the default, I'd like to have ocp-indent support.

AltGr:
Just a note, few seem to be aware of it, but OCaml supports a format that seems to me better than the two options above:
```
          "File %S, line %d:\n\
          \  Warning: Using deprecated ocamlformat config syntax.\n\
```

(`ocp-indent` supports this too and will indent as shown here)

hhugo:
@AltGr, your example will trigger `illegal backslash escape in string` warning though.
Maybe we should fix the lexer to not raise in that case.

AltGr:
Are you sure? It is a perfectly valid escape and I have never had this warning using it. Just tested with the 4.06.0 toplevel and it's fine.
A warning would be a bug since the escape is documented here: http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#sec80

jberdine:
We would have to ensure that the leading `\` is only emitted in case there is some leading whitespace, otherwise we would get illegal escapes. Is that perhaps what you ran into @hhugo?

hhugo:
Ack, I probably got confused during my testing.

hhugo:
In fact, the issue is with tabs. "\<tab>" will trigger the warning.

test/passing/string.ml.ref
```diff
@@ -9,10 +9,16 @@ let f = function
...
-   cccccccccccccccccc\n	\
-   "
+  "aaaaaaaaaaaaaaaaaaaaaaaaa\n\
+   \tbbbbbbbbbbbbbbbbbbbbbbbbbb\n\
+   \tcccccccccccccccccc\n\
+   \t"
+
+let _ =
+  "aaaaaaaaaaaaaaaaaaaaaaaaa\n\
+   \ bbbbbbbbbbbbbbbbbbbbbbbbbb\n\
```
hhugo: @jberdine, I was unable to have the indentation correct. (one to the left)

jberdine: Did you also try changing the `0` to `-1` in the `pre_break` in `fmt_line` above `fmt_lines`?

hhugo: The issue was that in preserve mode, the string contains `\ `  and the code was checking for leading ws only.

jberdine:
Fails on:
```
;;
"@\n \
 xxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxx xxxxxxxx xxxxxxxxxxx"
```

src/Fmt_ast.ml
```diff
@@ -548,34 +548,20 @@ let fmt_constant c ~loc ?epi const =
...
-                        | Some i ->
-                            escape_string c (String.sub next ~pos:0 ~len:i)
-                        | None -> escape_string c next
-                      in
-                      fmt "\\n"
-                      $ fmt_if_k
-                          (not (String.is_empty next))
-                          (str spc $ pre_break 0 "\\" 0) ) )
+                      if String.is_empty next then fmt "\\n"
+                      else if String.is_prefix next ~prefix:" " then
```
jberdine: maybe just use `next.[0]`

Issue #309, opened 2018-08-11
Command line args should override config file settings


jberdine:
Currently config files are processed after command line arguments, resulting in the command line args being overridden by any setting set in any config file between the file being formatted and the file system root. This is backward, precedence of command line args should be higher than config files.

gpetiot:
I tried to fix this without modifying the whole structure (i.e. processing cmdline before .ocamlformat/attributes) but there is no way to know whether an option has been set on the command line if its value is still equal to the default value, so we cannot decide to only update options (from .ocamlformat/attributes) when the value has not been previously modified by the cmdline.
I am not sure what is the right approach to process the cmdline last, at the moment the processing of arguments (`parse info validate` if I understand) is not parameterized by a conf, so perhaps it is simpler to pass an `~env` to `Term.eval`  that would have been modified by parsing the options in .ocamlformat/attributes (without overriding the existing env). But it still looks overcomplicated, is there a simpler solution that I am not seeing ?

jberdine:
I don't see a simple way. :-( Note that there is another complication: where the .ocamlformat files are searched for depends on the input filename, which is only known by parsing the command line. That might suggest to parse the command line, find the file, read the config files, and parse the command line again. But the call to `parse` sets the backing refs whether the options are specified or not. The approach of using an env for the last parsing of the command line sounds plausible.

hhugo:
We don't want cmdliner to deal with default values. That mean we need cmdliner to return optional values.

One thing I can imagine is,
- keep cmdliner to do validation, error messages, ..
- have cmdliner return unstructured values - `string option` (for example, `Some "max-iters=5"`)
- start with the default config and override config in a specific order with a single code path

```
default
|> update_using_dot_ocamlformat (* split file on newline + parse_n_update  *)
|> update_using_env_variable (* OCAMLFORMAT env variable split on ; or , + parse_n_update *)
|> update_using_cmdline (* fold over the list of string we got from cmdliner to parse_n_update *) 
```

Note that we would drop the environment variable handling from cmdliner and just parse and process a single variable (`OCAMLFORMAT`).

samoht:
A possible way to do this is to have a toplevel `default` config value, which is built by reading the config files (first the global one, then the local one and finally some custom default values). Then the default values for cmdliner will be `default.foo`, `default.bar`, etc.

jberdine:
This approach sounds feasible. Expanding a little:
1. parse command line with cmdliner to find the source files
[...]

PR #337, opened 2018-08-19
Improve error message when max-iters is set to 1

hhugo:


src/Translation_unit.ml
```diff
@@ -236,6 +238,18 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
       | Warning50 l ->
           List.iter l ~f:(fun (l, w) -> !Location.warning_printer l fmt w)
       | exn -> Format.eprintf "%s\n%!" (Exn.to_string exn) )
+  | Not_stable i when i <= 1 ->
+      Format.eprintf "%s: %S needs to be formatted. ([max-iters = 1])\n%!"
```
jberdine: maybe reword: s/needs to be/was not already/

src/Translation_unit.ml
```diff
@@ -102,7 +102,11 @@ let dump xunit dir base suf ext ast =
     xunit.printast (Caml.Format.formatter_of_out_channel oc) ast ;
     Out_channel.close oc )
 
-type result = Ok | Invalid_source of exn | Ocamlformat_bug of exn
+type result =
+  | Ok
+  | Invalid_source of exn
+  | Not_stable of int
```
jberdine: maybe `Unstable`?

hhugo:
I've added explicit `--max-iters` whenever required. The default value it set by environment variable. To use a `.ocamlformat` file, we would first need to address https://github.com/ocaml-ppx/ocamlformat/issues/309.

PR #344, opened 2018-08-21
Improve: new option to control spacing around let bindings

hhugo:


hhugo:
I think I misunderstood the reason for this. If people want to keep this behavor, we should probably encode it in using the `--module-item-spacing` option.

jberdine:
Or it could be modified slightly to add your `;;` after multi-line let-bindings.

hhugo:
Not really happy with names and documentation

hhugo:
waiting for https://github.com/ocaml-ppx/ocamlformat/pull/367 to be merged.

hhugo:
Rebased.
Regarding merging the option with module-item-spacing, I think there are a bit orthogonal. Janestreet would use module-item-spacing=compact with let-binding-spacing=double-semicolon

PR #345, opened 2018-08-21
Improve: update emacs mode to use replace-buffer-contents

hhugo:
`replace-buffer-contents` was included in emacs 26.1 but was unfortunately broken.
It is now fixed in the emacs 26 branch.

The code in the PR tries to detect broken implementations.

### DIFF
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -179,6 +179,14 @@ function."
 	  ok))
     nil))
 
+(defun ocamlformat--patch-buffer (outputfile)
+  (let ((patchbuf (get-buffer-create "*OCamlFormat patch*")))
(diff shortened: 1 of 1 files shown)


jberdine:
Can you describe the benefit of this? I mean, is it worth the complication and risk of the detection code? Is the fix expected to ship before emacs 28?

hhugo:
I would hope the fix to be shipped with `26.2` but I have not evidence of that.

> 4.12 New command replace-buffer-contents.
> This command replaces the contents of the accessible portion of the current buffer with the contents of the accessible portion of a different buffer while keeping point, mark, markers, and text properties as intact as possible.

In practice, it provides a much better user experience as the cursor (most) always stays on the same token.

PR #346, opened 2018-08-21
Improve: add option to not force-break structs

jberdine:


jberdine:
@samoht this is the change to allow one-line structs

samoht:
looks good!

Issue #305, opened 2018-08-10
Switch from ocaml-migrate-parsetree to ppxlib


jberdine:
Instead of depending directly on ocaml-migrate-parsetree, use [ppxlib](https://github.com/ocaml-ppx/ppxlib)'s `Ast`.

Not a high priority currently, but if/when we want to combine ocamlformat with another ppx rewriter, they will need to agree on the Ast version, which ppxlib supports.

hhugo:
@jberdine, what would be a use-case for combining ocamlformat and another ppx ?

emillon:
Duplicate of #1460

hhugo:
Should this be closed ?

PR #347, opened 2018-08-22
Improve: Switch from ocaml-migrate-parsetree to ppxlib

gpetiot:
Fix issue #305

### DIFF
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -14,8 +14,8 @@ depends: [
   "base-unix"
   "cmdliner"
   "dune" {build}
-  "ocaml-migrate-parsetree" {>= "1.0.10"}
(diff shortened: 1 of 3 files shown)


jberdine:
I don't think that this is enough. `Migrate_ast` still makes direct use of `ocaml-migrate-parsetree` rather than going through `ppxlib`.

gpetiot:
Ok, but I think I will give up on this one, there is definitely not enough documentation on ppxlib_ast.

jberdine:
Ok, I agree that `ppxlib` looks sparsely documented currently. This is also not currently pressing, and won't be I think as long as ocaml-migrate-parsetree is separately maintained and as long as we don't want to combine other ppx rewriting passes into one run with ocamlformat.

PR #348, opened 2018-08-22
Improve: divide the 'formatting' options and 'operational' options in…

gpetiot:
…to separate sections

Last point of issue #299.
There is the following constraint with Cmdliner: all customized sections are placed (and sorted alphabetically) before the default OPTIONS section.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -146,30 +146,32 @@ end = struct
 
   let store = ref []
 
-  let in_attributes cond =
-    if cond then "" else "Cannot be set in attributes."
+  let in_attributes ~section cond =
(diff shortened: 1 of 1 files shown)


src/Conf.ml
```diff
@@ -99,8 +99,11 @@ module C : sig
     -> env:Term.env_info
     -> doc:string
     -> allow_inline:bool
+    -> kind:[`Formatting | `Operational]
```
jberdine: maybe s/kind/section/

gpetiot:
> And not directly related to this PR, but is it easy and desirable to remove the section that redundantly lists all the environment variables?

I looked up the cmdliner's API and it does not seem to be feasible.

PR #350, opened 2018-08-22
Improve: simplify setting option defaults, slight --help improvement

jberdine:


jberdine:
@gpetiot, does this look ok to you?

gpetiot:
It looks good to me :+1:

PR #351, opened 2018-08-22
Build: move ocamlformat_support package from branch to directory

jberdine:


avsm:
This is perfect for vendoring, thank you!

hhugo:
Do we need to keep the `ocamlformat_support.opam` file ?

jberdine:
Probably not, if we don't point dune at it either. I'll include this while fixing my breakage of master.

PR #352, opened 2018-08-24
[WIP] Options

gpetiot:
In progress, will fix issue #309


gpetiot:
I think this one is ready to be reviewed.

src/Conf.ml
```diff
@@ -210,27 +214,31 @@ end = struct
...
     in
     let parse = Bool.of_string in
-    let r = mk ~default term in
-    let get () = if invert_flag then not !r else !r in
-    let opt = {names; parse; update; get; allow_inline} in
+    let r = mk ~default:None term in
+    let get () =
+      match !r with
+      | Some x -> Some (if invert_flag then not x else x)
+      | None -> None
```
jberdine: simplify with Option.map ?

src/Conf.ml
```diff
@@ -247,6 +255,18 @@ end = struct
...
 
   let get {get} = get ()
+
+  let default {default} = default
+
+  let update_field {update} = update
+
+  let update_by_name conf (name, value) =
+    let on_pack (Pack ({names; parse} as opt)) =
+      if List.exists names ~f:(String.equal name) then
```
jberdine: List.mem would probably be clearer

test/passing/error4.ml.opts
```diff
@@ -1 +1 @@
---no-comment-check
\ No newline at end of file
+--no-comment-check=true
```
jberdine: Is it possible to not have to change this? Being required to specify true or false is an uncommon interface.

test/passing/cmdline_override.ml.ref
```diff
@@ -0,0 +1,4 @@
+[@@@ocamlformat "module-item-spacing=sparse"]
```
jberdine: Shouldn't attributes have higher priority than anything (but deeper attributes)?

src/Conf.ml
```diff
@@ -806,36 +837,39 @@ let to_absolute file =
...
+  |> update Formatting.let_open
+  |> update Formatting.margin |> update max_iters
+  |> update Formatting.module_item_spacing
+  |> update Formatting.ocp_indent_compat
+  |> update Formatting.parens_tuple
+  |> update quiet
+  |> update Formatting.sequence_style
+  |> update Formatting.type_decl
+  |> update Formatting.wrap_comments
+  |> update Formatting.wrap_fun_args
```
jberdine: It would be nice if this could be refactored into something such as an update for the entire record, so that when a field is added to the type, we will get a compilation error here.

src/Conf.ml
```diff
@@ -711,6 +702,15 @@ let no_version_check =
   let default = false in
   mk ~default Arg.(value & flag & info ["no-version-check"] ~doc ~docs)
 
+let options =
+  let doc = "Aggregate options." in
```
jberdine: It would be good to specify the format of the env var value.

src/Conf.ml
hhugo: Folding over the `store` defined above should do it.

src/Conf.ml
```diff
@@ -210,27 +214,31 @@ end = struct
     let doc = generated_flag_doc ~allow_inline ~doc ~section in
     let docs = section_name section in
     let term =
-      Arg.(value & flag & info names_for_cmdline ~doc ~docs ~env)
+      Arg.(value & opt (some bool) None & info names_for_cmdline ~doc ~docs)
```
hhugo: You can keep the `flag` here and interpret `false` as `None` and `true` as `Some true`

src/Conf.ml
jberdine: Even better

src/Conf.ml
```diff
@@ -111,7 +110,11 @@ module C : sig
 
   val int : default:int -> docv:string -> int option_decl
 
-  val get : 'a t -> 'a
+  val default : 'a t -> 'a
+
+  val update_by_name : config -> string * string -> config
```
hhugo: We should probably use the `update` function defined bellow and properly handle errors with some helpful error message.

src/Conf.ml
```diff
@@ -132,11 +135,11 @@ end = struct
     ; parse: string -> 'a
     ; update: config -> 'a -> config
     ; allow_inline: bool
-    ; get: unit -> 'a }
+    ; get: unit -> 'a option
```
hhugo: Can we rename this field with `cmdline_get` or any other more explicit name.

src/Conf.ml
```diff
@@ -711,6 +696,19 @@ let no_version_check =
...
+    "Aggregate options. Options are specified as a list of comma-separated \
+     pairs of the following form: \
+     $(i,option)$(b,=)$(i,VAL)$(b,,)...$(b,,)$(i,option)$(b,=)$(i,VAL)"
+  in
+  let env = Arg.env_var "OCAMLFORMAT" in
+  let default = [] in
+  let assoc = Arg.(pair ~sep:'=' string string) in
+  let list_assoc = Arg.(list ~sep:',' assoc) in
+  mk ~default
+    Arg.(value & opt list_assoc default & info ["options"] ~doc ~docs ~env)
```
hhugo: ocp-indent uses `--config` and `-c`, should we stick to the same names ?

src/Conf.ml
```diff
@@ -711,6 +686,20 @@ let no_version_check =
   let default = false in
   mk ~default Arg.(value & flag & info ["no-version-check"] ~doc ~docs)
 
+let config =
+  let doc =
+    "Aggregate options. Options are specified as a list of comma-separated \
```
jberdine: I would have written "a comma-separated list", but don't know a grammatical justification for it

src/Conf.ml
```diff
@@ -806,36 +826,26 @@ let to_absolute file =
...
+    match C.update ~config ~name ~value ~inline:false with
+    | Ok c -> (c, errors)
+    | Error e -> (config, e :: errors)
+  in
+  let conf, errors = List.fold_left !config ~init:(conf, []) ~f in
+  try
+    match List.rev errors with
+    | [] -> conf
+    | l ->
+        user_error "malformed OCMALFORMAT environment variable"
```
jberdine: misspelled

src/Conf.ml
```diff
@@ -246,7 +244,13 @@ end = struct
...
     |> Option.value ~default:(Error (`Unknown (name, value)))
 
-  let get {get} = get ()
+  let default {default} = default
+
+  let update_using_cmdline conf =
+    let on_pack conf (Pack {cmdline_get; update}) =
+      match cmdline_get () with None -> conf | Some x -> update conf x
+    in
+    List.fold_left !store ~init:conf ~f:on_pack
```
jberdine: nit: just `List.fold` instead?

test/passing/cmdline_override.ml.opts
```diff
@@ -0,0 +1 @@
+--config=module-item-spacing=compact --module-item-spacing=sparse
```
jberdine: Perhaps add a test with the order of options reversed to check that they are interpreted left-to-right

hhugo:
Could you please rebase onto master when you have a chance ?

test/passing/cmdline_override.ml.opts
jberdine: Ok. I am just conscious of the potential for confusion. Maybe just make sure the doc of --config is clear that it is overridden by any other command line options, as I expect that the env var and --config being "the same" will only be anticipated by people who have implemented something with cmdliner.

ocamlformat-help.txt
```diff
@@ -9,140 +9,125 @@ DESCRIPTION
...
-       --break-cases={fit|nested|all} (absent=fit or OCAMLFORMAT_BREAK_CASES
-       env)
+       section can be set in many ways, its value is determined in the
+       following order (of increasing priority): it can be set in an
+       .ocamlformat configuration file with an 'option = VAL' line, or using
+       the OCAMLFORMAT environment variable: OCAMLFORMAT=option=VAL, ...
+       ,option=VAL, or as an optional parameter on the command line, or with
+       an [@@@ocamlformat "option=VAL"] attribute in the processed file.
+
+       --break-cases={fit|nested|all}
```
jberdine: We should document that the first choice is the default.

ocamlformat-help.txt
```diff
@@ -9,140 +9,125 @@ DESCRIPTION
...
-       section can be set in an .ocamlformat configuration file with an
-       'option = VAL' line, or with an [@@@ocamlformat "option=VAL"]
-       attribute.
-
-       --break-cases={fit|nested|all} (absent=fit or OCAMLFORMAT_BREAK_CASES
-       env)
+       section can be set in many ways, its value is determined in the
+       following order (of increasing priority): it can be set in an
+       .ocamlformat configuration file with an 'option = VAL' line, or using
+       the OCAMLFORMAT environment variable: OCAMLFORMAT=option=VAL, ...
```
jberdine: nit: remove spaces around `...`

ocamlformat-help.txt
```diff
@@ -168,7 +157,7 @@ OPTIONS
        --intf
            Parse file with unrecognized extension as an interface.
 
-       -n N, --max-iters=N (absent=10 or OCAMLFORMAT_MAX_ITERS env)
+       -n N, --max-iters=N
```
jberdine: Is there a good way to keep the doc for the default?

ocamlformat-help.txt
```diff
@@ -9,140 +9,125 @@ DESCRIPTION
...
-       attribute.
-
-       --break-cases={fit|nested|all} (absent=fit or OCAMLFORMAT_BREAK_CASES
-       env)
+       section can be set in many ways, its value is determined in the
+       following order (of increasing priority): it can be set in an
+       .ocamlformat configuration file with an 'option = VAL' line, or using
+       the OCAMLFORMAT environment variable: OCAMLFORMAT=option=VAL, ...
+       ,option=VAL, or as an optional parameter on the command line, or with
+       an [@@@ocamlformat "option=VAL"] attribute in the processed file.
```
jberdine: Should we want to mention `[@@ocamlformat...]` attributes on expressions?

ocamlformat-help.txt
```diff
@@ -9,140 +9,125 @@ DESCRIPTION
...
        Unless otherwise noted, any option --option=VAL detailed in this
-       section can be set in an .ocamlformat configuration file with an
-       'option = VAL' line, or with an [@@@ocamlformat "option=VAL"]
-       attribute.
-
-       --break-cases={fit|nested|all} (absent=fit or OCAMLFORMAT_BREAK_CASES
-       env)
+       section can be set in many ways, its value is determined in the
+       following order (of increasing priority): it can be set in an
+       .ocamlformat configuration file with an 'option = VAL' line, or using
```
jberdine: Should we mention that .ocamlformat files in current and all ancestor directories of the input file are used, with more distant ancestors having lower priority?

ocamlformat-help.txt
gpetiot: I slightly modified the generated doc for int options and choice options (I think it does not make much sense for flags because they are not set by default).

ocamlformat-help.txt
```diff
@@ -151,7 +148,11 @@ OPTIONS (CODE FORMATTING STYLE)
            aligned are not wrapped either. 
 
 OPTIONS
-       -g, --debug (absent OCAMLFORMAT_DEBUG env)
+       -c VAL, --config=VAL (absent OCAMLFORMAT env)
+           Aggregate options. Options are specified as a comma-separated list
+           of pairs: option=VAL,...,option=VAL
```
jberdine: others end with a `.`

src/Conf.ml
```diff
@@ -152,11 +153,17 @@ end = struct
     else "Cannot be set in attributes."
 
   let generated_choice_doc ~allow_inline ~all ~doc ~section =
+    let open Format in
     let each_opt =
-      let f acc (_, _, d) = Format.sprintf "%s@,%s" acc d in
+      let f acc (_, _, d) = sprintf "%s@,%s" acc d in
       List.fold all ~init:"" ~f
```
jberdine: Re-reading this, I see that if you want, you could use pp_print_list as done in generated_choice_docv.

src/Conf.ml
```diff
@@ -269,9 +282,22 @@ let info =
...
+         variable: \
+         $(b,OCAMLFORMAT=)$(i,option)$(b,=)$(i,VAL)$(b,,)...$(b,,)$(i,option)$(b,=)$(i,VAL), \
+         or as an optional parameter on the command line, or with a global \
+         $(b,[@@@ocamlformat \")$(i,option)$(b,=)$(i,VAL)$(b,\"]) \
+         attribute in the processed file, or with an $(b,[@@ocamlformat \
+         \")$(i,option)$(b,=)$(i,VAL)$(b,\"]) attribute on expression in \
+         the processed file."
+    ; `P
+        "(*) .ocamlformat files in current and all ancestor directories of \
+         the input file are used, with more distant ancestors having lower \
```
jberdine: Actually, maybe s/of the/for each/ would be clearer in the case that there are multiple input files.

hhugo:
Lgtm.

jberdine:
@gpetiot, squash and merge when you're ready.

PR #353, opened 2018-08-25
Improve: Support reading file from stdin

bkase:
Hi! Great project! I have a somewhat strange OCaml dev environment, and ocamlformat not supporting stdin input files was causing me problems.

I made small changes to support this new feature:
1. SRC files list can be empty (if so, go to read from stdin mode)
2. If so, you must pass a name (so we know if it's Impl, Intf, etc)
3. In the stdin case, I make a tmp file and mark it to be cleaned up after we're done

I was able to reuse the existing action variants, but I had to change the type of the `file` field in the record slightly. Let me know if you think this should be done some other way! Thanks!

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -889,12 +889,16 @@ let update_using_env conf =
                 , Sexp.List [Sexp.Atom name; Sexp.Atom value] ) ))
   with Sys_error _ -> conf
 
-type 'a input =
-  {kind: 'a; name: string; file: [`Tmp | `Input] * string; conf: t}
+type 'a input = {kind: 'a; name: string; file: string; conf: t}
 
 type action =
   | In_out of [`Impl | `Intf | `Use_file] input * string option
   | Inplace of [`Impl | `Intf | `Use_file] input list
+  | Stdin of
+      { kind: [`Impl | `Intf | `Use_file]
+      ; name: string
+      ; conf: t
+      ; output_file: string option }
 
 let kind_of fname =
   match Filename.extension fname with
@@ -923,19 +927,11 @@ let action =
           | None -> impossible "checked by validate"
           | Some name -> name
         in
-        let file, oc =
-          Filename.open_temp_file "ocamlformat" (Filename.basename name)
-        in
-        In_channel.iter_lines stdin ~f:(fun s ->
-            Out_channel.output_string oc s ;
-            Out_channel.newline oc ) ;
-        Out_channel.close oc ;
-        In_out
-          ( { kind= kind_of name
-            ; name
-            ; file= (`Tmp, file)
-            ; conf= read_config ~filename:name config }
-          , !output )
+        Stdin
+          { kind= kind_of name
+          ; name
+          ; conf= read_config ~filename:name config
+          ; output_file= !output }
     | [input_file] ->
         let name = Option.value !name ~default:input_file in
         In_out
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -668,8 +668,8 @@ let inplace =
 let inputs =
   let docv = "SRC" in
   let doc =
-    "Input files. At least one is required, and exactly one without \
-     --inplace."
+    "Input files. If none are passed, will read from stdin. Exactly one is \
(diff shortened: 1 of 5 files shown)


src/Conf.ml
```diff
@@ -660,8 +660,8 @@ let inplace =
 let inputs =
   let docv = "SRC" in
   let doc =
-    "Input files. At least one is required, and exactly one without \
-     --inplace."
+    "Input files. If none are passed, will read from stdin. Exactly one is \
+     needed for --inplace."
```
jberdine: Question regarding user interface: would it be better to accept `-` as a single filename denoting `stdin` instead of an empty list of input files? I don't really know, but both approach seem to be common.

src/Conf.ml
```diff
@@ -856,16 +857,31 @@ let action =
       (List.map !inputs ~f:(fun file ->
            { kind= kind_of file
            ; name= file
-           ; file
+           ; file= (`Input, file)
            ; conf= read_config ~filename:file config } ))
   else
     match !inputs with
+    | [] ->
+        let name = Option.value_exn !name in
```
jberdine: In general I'd prefer to document the impossible case with something like `impossible "checked by validate"` as below.

src/ocamlformat.ml
```diff
@@ -84,10 +84,17 @@ match Conf.action with
...
-  match
-    In_channel.with_file input_file ~f:(fun ic ->
-        Translation_unit.parse_print (xunit_of_kind kind) conf ~input_name
-          ~input_file ic output_file )
-  with
-  | Ok -> Caml.exit 0
-  | Unstable _ | Ocamlformat_bug _ | Invalid_source _ -> Caml.exit 1 )
+    let after, input_file =
+      match input_file with
+      | `Tmp, tmp_file -> ((fun () -> Caml.Sys.remove tmp_file), tmp_file)
```
jberdine: note that `Unix.unlink` is used elsewhere to remove files

hhugo:
```diff
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -663,8 +663,14 @@ let inputs =
     "Input files. At least one is required, and exactly one without \
      --inplace. If - is passed, will read from stdin."
   in
+  let file_or_dash =
+    let parse, print = Arg.non_dir_file in
+    let parse = function "-" -> `Ok "-" | s -> parse s in
+    (parse, print)
+  in
   let default = [] in
-  mk ~default Arg.(value & pos_all file default & info [] ~doc ~docv ~docs)
+  mk ~default
+    Arg.(value & pos_all file_or_dash default & info [] ~doc ~docv ~docs)
 ```

File: src/Conf.ml [hunk 4]
```diff
@@ -856,16 +866,35 @@ let action =
...
+          | None -> impossible "checked by validate"
+          | Some name -> name
+        in
+        let file, oc =
+          Filename.open_temp_file "ocamlformat" (Filename.basename name)
+        in
+        In_channel.iter_lines stdin ~f:(fun s ->
+            Out_channel.output_string oc s ;
+            Out_channel.newline oc ) ;
+        Out_channel.close oc ;
```
jberdine: I thought I made a comment before, but it seems I fumbled it, so sorry for the piecemeal feedback. But could this processing be delayed from here when config options are being read to later in ocamlformat.ml when the actions are being processed? It would be a better separation if actual processing such as copying files wasn't done in the middle of setting up the config. I think it would also probably be cleaner to add an alternative to the action type for processing stdin, and in that way avoid the distinction between Input and Tmp.

hhugo:
src/ocamlformat_reason.ml needs to be updated

bkase:
ok fixed, seems strange that there's a bunch of duplicated code in that ocamlformat_reason file, but that should be addressed by another PR

jberdine:
I had something slightly easier, I think, in mind. What do you think of the commit I pushed to the end of your branch?

bkase:
Ah yes this is much better, thanks!

bkase:
Look good?

jberdine:
Excellent, thanks!

PR #355, opened 2018-08-25
Fix build without an external ocamlformat_support

avsm:
The new internal library name is `format_`, so use it in the library build instead of referencing `ocamlformat_support`.

With this changeset, ocamlformat successfully builds in an [embedded dune monorepo](http://www.dra27.uk/blog/platform/2018/08/15/dune-vendoring.html) with no external dependencies!


ocamlformat.opam
```diff
@@ -7,7 +7,7 @@ dev-repo: "https://github.com/ocaml-ppx/ocamlformat.git"
 license: "MIT"
 build: [
   ["tools/gen_version.sh" "src/Version.ml" version] {pinned}
-  ["dune" "build" "-p" "ocamlformat_support,ocamlformat" "-j" jobs]
+  ["dune" "build" "-p" "ocamlformat" "-j" jobs]
```
jberdine: s/"ocamlformat"/name/ now that you've unhacked it

PR #356, opened 2018-08-26
Improve: install man-page and track content of ocamlformat --help

hhugo:


samoht:
opam is generating and installing its man page from --help here: https://github.com/ocaml/opam/blob/master/doc/Makefile#L30-L41

Makefile
```diff
@@ -22,7 +22,7 @@ setup: dune-workspace
 
 .PHONY: exe
 exe: setup
-	dune build _build/dev/src/ocamlformat.exe _build/release/src/ocamlformat.exe
+	dune build _build/dev/src/ocamlformat.exe _build/release/src/ocamlformat.exe _build/release/src/ocamlformat-help.txt
```
jberdine: I'm not sure this should be added to the `exe` target. What about adding a `man` or `doc` or something target, and adding that target to `default`?

src/dune
```diff
@@ -20,6 +20,17 @@ let preprocess =
  %s
  (libraries cmdliner ocamlformat_support import ocaml-migrate-parsetree unix))
 
+(rule
+  (targets ocamlformat.1)
+  (deps ocamlformat.exe)
+  (action (with-stdout-to %%{targets} (run ./ocamlformat.exe --help=groff))))
+
+(rule
+  (targets ocamlformat-help.txt)
```
jberdine: Will dune flip its shit if this is `../ocamlformat-help.txt` so that the docs aren't in the source dir?

hhugo: `../ocamlformat-help.txt` will probably not work. but we can move that rule to the parent directory.

hhugo:
@jberdine, what do you think about merging this one first before the other PRs touching the Conf.ml ?

hhugo:
I do not know. @diml ? is it possible ?

Issue #361, opened 2018-08-27
Follow XDG for global config files


samoht:
[XDG](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) is becoming the standard to store configuration files (and it's followed by dune :p). The specs says the global config file should be something like: `./config/<toolname>/file`

samoht:
Basically XDG states that the proliferation of global dot files are annoying, and should all be moved to a `~/.config/` subdir instead. So yes, with XDG `~/.ocamlformat` will have to move to `~/.config/ocamlformat/config` or ~/.config/ocaml/ocamlformat.config` or something like that.

jberdine:
I agree with the basic motivation behind XDG. The part I'm unsure of in the context of ocamlformat is that its configuration is intimately tied source code much more than individual users, at least in the use cases I know of. In particular, the ocamlformat configuration needs to be kept in sync with the source code via source control. I was actually thinking before about how to ensure that no configuration is used that is outside the source control repo.

hhugo:
I kind of agree with @jberdine. Not reading configuration outside the "project" seems right. I even wonder if ocamlformat should simply be disabled if no `.ocamlformat` is found in the project.

samoht:
>  I even wonder if ocamlformat should simply be disabled if no .ocamlformat is found in the project.

I think that's too strict. Most project will just use the "default" formating style, which IMHO needs to be a bit redesigned. The goal of ocamlformat should be to allow the different styles of the community to converge to an "acceptable" default, not to have every project using a different style.

hhugo:
Note, that project can still use whatever default by having an empty `.ocamlformat` in their project.
I agree with the goal of convergence, but in practice, I do not use ocamlformat at home because all code bases I touch (and most often don't own) have different style and I don't want Emacs to reformat entire files on save.

jberdine:
(Note that unless it has changed, adding an empty `.ocamlformat` file just passes the config from the parent through, rather than sets options to defaults.)

I agree that the eventual goal is for the community to converge on a (few) common styles, maybe even one, with a "default" that people who are not overly picky about syntax can get along with. I think the intent is for the config files to move to just specifying one of the umbrella styles, perhaps with a few exceptions. But I don't understand how this is connected to the question of whether to read formatting configuration from within the source tree. It seems that, as something that if changed, changes the source, the formatter config needs to be part of the source tree.

[...]

PR #365, opened 2018-08-28
Improve: Follow XDG for global config files

gpetiot:
Should fix issue #361  @samoht


gpetiot:
Strange failure of the CI, does not seem to be caused by this commit though (no $XDG_CONFIG_HOME/ocamlformat/.ocamlformat by default ?).

samoht:
Maybe `XDG_CONFIG_HOME` is not defined in the environment of the CI?

Also, I think XDG states that the config file should be in a subdir, e.g. `~/.config/ocamlformat/<filename>`.

src/Conf.ml
```diff
@@ -893,9 +897,21 @@ let kind_of fname =
...
+let update_using_xdg =
+  try
+    let filename =
+      Filename.concat
+        (Caml.Sys.getenv "XDG_CONFIG_HOME")
+        "ocamlformat/.ocamlformat"
+    in
+    read_config ~filename ~parents:false
+  with
+  | Caml.Not_found | Not_found_s _ -> Fn.id
```
jberdine: It looks like this code is intentionally reading the env only once at the beginning. Could you make this clearer using `Staged`.

src/Conf.ml
```diff
@@ -296,8 +296,10 @@ let info =
          the processed file."
     ; `P
         "(*) .ocamlformat files in current and all ancestor directories \
-         for each input file are used, with more distant ancestors having \
-         lower priority." ]
+         for each input file are used, as well as the global .ocamlformat \
+         file defined in $(b,\\$XDG_CONFIG_HOME/ocamlformat). The global \
```
jberdine: filename out of sync with code

gpetiot: I suppose the global .ocamlformat file is placed in the directory $XDG_CONFIG_HOME/ocamlformat so the doc looks ok to me. Do you want to rename it ?

jberdine: Ok, I thought the doc was giving the filename rather than the directory name. Either is fine, but my impression is that files under $XDG_CONFIG_HOME are often named differently than when in $HOME, in particular without a leading `.`. I don't have any strong feeling, just think we should follow convention.

src/Conf.ml
```diff
@@ -893,9 +897,20 @@ let kind_of fname =
...
   | _ -> !kind
 
+let update_using_xdg =
+  match Caml.Sys.getenv_opt "XDG_CONFIG_HOME" with
+  | Some xdg_config_home ->
+      let filename =
+        Filename.concat xdg_config_home "ocamlformat/.ocamlformat"
+      in
+      Staged.stage (fun conf -> read_config ~filename ~parents:false conf)
+  | None -> Staged.stage (fun conf -> conf)
```
jberdine: @gpetiot how about like this?

gpetiot: Better of course. Did not know `Staged`, guess I did not use it the best way :)

jberdine:
This looks good to me, I only have the question of whether the conventional thing to do is to read `$XDG_CONFIG_HOME/ocamlformat/.ocamlformat` of something else such as `$XDG_CONFIG_HOME/ocamlformat/config`?

gpetiot:
I did not see any name convention in the XDG spec doc. It is said:

> There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable $XDG_CONFIG_HOME.

> $XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored.

I don't see how it can hurt to keep the name `.ocamlformat` in this directory.

Issue #362, opened 2018-08-27
Format only a fragment on the file


samoht:
`ocp-indent` has an option to fragment only a portion of a file:

```
-l RANGE, --lines=RANGE (absent=-)
           Only re-indent the lines in RANGE (eg. 10-12), adapting to the
           current indentation of surrounding lines. Lines start at 1.
```

I think it's useful to be integrated with some editors (/cc @AltGr and @emillon)

jberdine:
I think this would clearly be useful, but I don't know how to implement it internally. One option might be to reformat the whole file, compute a patch file using diff, filter the hunks by line number, and then apply the resulting patch.

samoht:
Can't you just extract these lines, parse/format as usual (e.g. the fragment should be valid OCaml) and print the new fragments with the same tabulation as the first extracted line?

jberdine:
I don't know how usable it would be, but perhaps something like that could work:
1. split the input file into 3 files by line numbers: prefix, range, suffix
2. format range to a tmp file range' (if this fails, fail the whole format run)
3. prefix every line of range' with the same leading whitespace as the first line of range
4. concat prefix, range', suffix files to obtain the output

gpetiot:
Is there a not-so-difficult way to link the locations between two iterations of ocamlformat ?
Supposing we only want to format lines 9--13, supposing that means only formatting the items in lines 10--12 (it would be great to adjust the value of `--lines` to 10--12 on the way). Supposing the formatting of items in lines 10--12 results in lines 10--14 (that's where I would like to link 10--12 to 10--14), so we update the value of `--lines` to 10--14, and we should reach a fixpoint.

jberdine:
There is nothing internal to ocamlformat that tracks line numbers. Since 
lines are broken largely by Format, it doesn't seem straightforward to 
add. Something could probably be done by diffing the input and output 
files. Maybe some of the implementation of patdiff could be repurposed 
to do so.

gpetiot:
See discussion in #366

PR #366, opened 2018-08-28
[WIP] Format only a fragment

gpetiot:
Should fix issue #362


hhugo:
Here, trying to reformat just the two inner let bindings does not work
```
let f =
let a = aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaa in
let b = aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaa in
a + b
```
I would expect the current approach to only be able to select fragment at the level of toplevel item.

test/passing/fragment.ml.ref
```diff
@@ -0,0 +1,22 @@
...
+    aaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaa
+    aaaaaaaaaaaaaaaaaaa
+
+let a =
+  aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaa
+    aaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaa
+    aaaaaaaaaaaaaaaaaaa
+
+let a = aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaa
+
```
samoht: why is there an extra newline here?

jberdine:
The basic implementation approach in terms of `disabled` attributes can work only if the range of lines specifies a complete sub-forest of the parsetree, right? Is that functional enough? Should that be checked to be true of the given range, and fail otherwise?

gpetiot:
If modified the disable condition so that the formatting is disabled if the construct starts after the end of RANGE, or ends before the start of RANGE.
On the previous example when running with `--lines=2,2` we get:
```
let f =
  let a =
    aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaa
      aaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaaaa
      aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaa
  in
  let b = aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaaaaaaaaaaaaaa in
  a + b
```

test/passing/fragment.ml.ref
gpetiot: Still can't explain this one.

hhugo:
Remember that the line range will be outdated after the first iteration. Also, note that the current implementation of `disable` does not preserve newline between toplevel items.

jberdine:
For an implementation based on `disable`, I think it is necessary to understand whether the desired behavior is to reformat:
1. as little as possible, as long as it includes the indicated range
2. as much as possible, as long as it is contained in the indicated range

@samoht, does one of these make more sense than the other in the use case you have in mind?

jberdine:
Ok, thanks, let's move this PR in that direction then.

It seems that what is needed is to identify the deepest parsetree node whose location contains the given range, enable reformatting on it and its descendants, and consider everything else to be disabled. The `Loc_tree` in `Cmts` used to place comments in the parsetree could be useful for finding the deepest enclosing node.

The current implementation of disabled is geared toward the case where formatting almost everything is enabled, and only a few phrases have formatting disabled. I'm not sure how well this will fit with this PR, which has the opposite expectations. It would be optimal for the disabled attributes to work equally for both, but I suspect it might be complicated to get spacing right in the corner cases. I'm just mentioning in case it is too fiddly: perhaps another implementation strategy would be simpler for this feature.

hhugo:
Before we spend too much time on this. Can we have a description of the expected use case..

samoht:
@hhugo ocp-indent uses this to reformat only a part of the file; it's used by the vim bindings here: https://github.com/OCamlPro/ocp-indent/blob/rewind/tools/ocp-indent.vim#L39 There's probably other ways to do that. 

An other use-case which just came up is to reformat merge diffs, so that when you have conflicts you would only reformat the part of the file where the conflict happened. @emillion just told me it used to have some tooling to handle this case with ocp-indent using the `--lines` argument.

emillon:
Yes, the idea was to parse `git` output to get the line numbers of the current branch vs the merge base, and then pass `--lines` to `ocp-indent`.

hhugo:
Why is this better than reformatting the entire file?

emillon:
That was part of a CI pipeline. If you reformat the entire file, any PR would require reformatting all files touched.

hhugo:
@emillon, sorry, I don't see the issue with reformatting all files touched. Can you give some details ?

emillon:
Sure, I'm happy to share some details. The use case was to gradually switch to `ocp-indent` while keeping code review meaningful. We had several options:

- reindenting all at once would work, but that would also mean losing most of the git history
- reindenting all touched files in each PR. it makes it difficult to do code review, because all code changes are intertwined with formatting the whole file. Reformatting in a different commit would mean breaking `git bisect`, so it's difficult to make this work.

In the end we went with indenting only touched hunks. That way we could keep git history, have reviewable changes, and keep `git bisect` working. The downside was that we had inconsistent formatting within a file, but for a legacy code base it was acceptable.

(TBH I'm not sure I would have picked the same solution with `ocamlformat`. For a formatter to work well, editor integration is necessary, and that probably means indenting at the file granularity.)

gpetiot:
After discussing with @emillon it seems this feature is not really necessary, what do we do about this issue and PR?

hhugo:
I'm happy to close this until someone comes with a real usecase.

Wilfred:
FWIW I'd really like to be able to format parts of a file. It's a common feature of other formatters, including ocp-format. When I'm working with code in an editor, it's really useful to only reformat the bits I'm working on.

I agree that ideally projects would use ocamlformat everywhere, but I don't always have that luxury. 

I was looking at building Emacs integration for this, but it's much easier if the 'format the following part of the file' logic lives entirely in the formatter.

Would you be open to reopening this?

Issue #363, opened 2018-08-28
extend module-item-spacing to toplevel let-and construct 


hhugo:

hhugo:
@gpetiot, would you be able to look at this one ?

PR #367, opened 2018-08-28
Improve: obey module-item-spacing for let-and bindings

gpetiot:
Should fix issue #363

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -335,6 +335,18 @@ module Signature_item : Module_item with type t = signature_item = struct
     || not (allow_adjacent (i1, c1) (i2, c2))
 end
 
+module Expression : Module_item with type t = expression = struct
+  type t = expression
+
+  let is_simple (i, c) =
+    Poly.(c.Conf.module_item_spacing = `Compact)
+    && Location.width i.pexp_loc <= c.Conf.margin
+    && Location.is_single_line i.pexp_loc
+
+  let break_between (i1, c1) (i2, c2) =
+    (not (is_simple (i1, c1))) || not (is_simple (i2, c2))
+end
+
 let may_force_break (c : Conf.t) s =
   let contains_internal_newline s =
     match String.index s '\n' with
@@ -472,6 +484,7 @@ let break_between (i1, c1) (i2, c2) =
   match (i1, i2) with
   | Str i1, Str i2 -> Structure_item.break_between (i1, c1) (i2, c2)
   | Sig i1, Sig i2 -> Signature_item.break_between (i1, c1) (i2, c2)
+  | Exp i1, Exp i2 -> Expression.break_between (i1, c1) (i2, c2)
   | _ -> assert false
 
 (** Precedence levels of Ast terms. *)
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -52,7 +52,7 @@ let protect =
         first := false ) ;
       raise exc
 
-let update_config c l =
+let update_config ?(quiet = false) c l =
   let update_one c ({txt; loc}, payload) =
     let result =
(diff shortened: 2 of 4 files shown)


hhugo:
Can you add some tests with doc comments between let-and constructs
```
let x = x

(** comment *)
and y = y
``` 

```
let x = x

(** floating comment *)

and y = y
```

hhugo:
Also some test with multi-lines
```
let x = x
and y = something
   long
```

```
let y = something
   long
and x = x
```

gpetiot:
I added more tests, there are some `and` that are replaced by `let`, I don't know if I messed up with the first/last variables.

src/Ast.ml
```diff
@@ -335,6 +335,113 @@ module Signature_item : Module_item with type t = signature_item = struct
...
+    | Ppat_extension _ -> false
+    | Ppat_open (_, p) -> pattern_has_doc p
+
+  let is_simple (i, c) =
+    Poly.(c.Conf.module_item_spacing = `Compact)
+    && Location.width i.pexp_loc <= c.Conf.margin
+    && Location.is_single_line i.pexp_loc
+
+  let break_between (i1, c1) (i2, c2) =
+    expr_has_doc i1 || expr_has_doc i2
```
jberdine: Do you have examples where checking for docs here is necessary? It is not obvious to me that checking `pvb_attributes` is not enough.

gpetiot:
I added some commits to take attributes into account, it looks good but it required to print warnings only once to not have extra regressions

gpetiot:
Warnings are printed when we iterate on exp attributes for update_config.

hhugo:
Actually, ignore what I said. I wonder if the proper fix should just be to not call update_config multiple time though.

jberdine:
@hhugo, to avoid calling update_config multiple times, do you have in mind to modify fmt_value_binding to assume the config it receives has already been updated with attributes, or something else?

src/Fmt_ast.ml
```diff
@@ -3682,16 +3682,33 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
...
+      let grps =
+        List.group bindings ~break:(fun (itmI, cI) (itmJ, cJ) ->
+            (not (List.is_empty itmI.pvb_attributes))
+            || (not (List.is_empty itmJ.pvb_attributes))
+            || Ast.break_between
+                 (Exp itmI.pvb_expr, cI.conf)
+                 (Exp itmJ.pvb_expr, cJ.conf) )
+      in
+      let fmt_grp ~first:first_grp ~last:_ bindings =
+        list_fl bindings (fun ~first ~last:_ (binding, _) ->
```
hhugo: should we make use of the config here ?

hhugo:
@jberdine, either that or an optional argument in `update_config` to mutes the warning.

jberdine:
This fails:
```
$ cat test.ml
(** doc *)

type t
$ ocamlformat --break-struct=natural test.ml
ocamlformat.exe: Cannot process "test.ml".
  BUG: formatting changed ast.
[Exit 1]
```

gpetiot:
Are you sure this is due to this PR ?
I will try to fix it as well as other issues in https://github.com/ocaml-ppx/ocamlformat/pull/380

jberdine:
Yes, this fails with f516634 but not with a1bbe9d

gpetiot:
Exactly this commit belongs to this other PR.

jberdine:
Ah, ok, thanks, I must be confused about which commits are where. I'll look around.

jberdine:
Ok, sorry, I was confused, you're right that this one is ok. BTW, I'm not sure why but the title of the commit didn't end up being synced with the title of the PR.

PR #372, opened 2018-08-29
Improve: add support for bigarray sugar index operation

### LINKED ISSUE #371: Use sugar syntax for bigarray index operator.
```
 buf.{i}
 buf.{i,..,k}
 buf.{i,..} <- e
```

hhugo:
Fix #371

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -80,6 +80,53 @@ let index_op_string = (".", '[', ']')
 
 let index_op_array = (".", '(', ')')
 
+let index_op_bigarray = (".", '{', '}')
+
+let all_args_unlabeled args =
+  List.fold_right args ~init:(Some []) ~f:(fun arg acc ->
+      match (acc, arg) with
+      | Some args, (Asttypes.Nolabel, e) -> Some (e :: args)
+      | _ -> None )
+
+let index_op_get_sugar (ident : Longident.t) args =
+  match all_args_unlabeled args with
+  | None -> None
+  | Some args -> (
+    match (ident, args) with
+    | Ldot (Lident "String", "get"), [_] -> Some (index_op_string, args)
+    | Ldot (Lident "Array", "get"), [_] -> Some (index_op_array, args)
+    | Ldot (Ldot (Lident "Bigarray", "Array1"), "get"), [_] ->
+        Some (index_op_bigarray, args)
+    | Ldot (Ldot (Lident "Bigarray", "Array2"), "get"), [_; _] ->
+        Some (index_op_bigarray, args)
+    | Ldot (Ldot (Lident "Bigarray", "Array3"), "get"), [_; _; _] ->
+        Some (index_op_bigarray, args)
+    | ( Ldot (Ldot (Lident "Bigarray", "Genarray"), "get")
+      , [{pexp_desc= Pexp_array l}] )
+      when List.length l > 3 ->
+        Some (index_op_bigarray, l)
+    | _ -> None )
+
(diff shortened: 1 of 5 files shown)


src/Fmt_ast.ml
```diff
@@ -1380,25 +1380,19 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?ext
...
-      fmt_index_op c ctx ~parens index_op_array s i
-  | Pexp_apply
-      ( { pexp_desc= Pexp_ident {txt= Ldot (Lident "String", "get")}
-        ; pexp_attributes= [] }
-      , [(Nolabel, s); (Nolabel, i)] ) ->
-      fmt_index_op c ctx ~parens index_op_string s i
+      ( {pexp_desc= Pexp_ident {txt}; pexp_attributes= []}
+      , (Nolabel, s) :: indices )
+    when Option.is_some (index_op_get_sugar txt indices) ->
+      let op, indices = Option.value_exn (index_op_get_sugar txt indices) in
```
jberdine: It would be nice to avoid the repetition between the when clause and the body, but I don't know a composable way to write it.

src/Ast.ml
```diff
@@ -80,6 +80,66 @@ let index_op_string = (".", '[', ']')
...
+let index_op_get_sugar (ident : Longident.t) args =
+  if
+    List.exists args ~f:(function Asttypes.Nolabel, _ -> false | _ -> true)
+  then None
+  else
+    let args =
+      List.map args ~f:(function
+        | Asttypes.Nolabel, e -> e
+        | _ -> impossible "previous match" )
+    in
```
jberdine: I think I would have written this:
```ocaml
  match all_args_unlabeled args with
  | None -> None
  | Some args ->
```
where
```ocaml
let all_args_unlabeled args =
  List.fold_right args ~init:(Some []) ~f:(fun arg acc ->
      match (acc, arg) with
      | Some args, (Asttypes.Nolabel, e) -> Some (e :: args)
      | _ -> None )
```

src/Ast.ml
```diff
@@ -106,24 +106,12 @@ let index_op_get_sugar (ident : Longident.t) args =
...
-        | _ -> impossible "previous match" )
-    in
-    let e, args =
-      match List.rev args with
-      | e :: args -> (e, List.rev args)
-      | _ -> impossible "previous match"
-    in
+  match all_args_unlabeled args with
+  | None | Some [] -> None
+  | Some (e :: args) -> (
```
hhugo: `e` should be the last element of the list. Explaining the double List.rev above.
We could probably enumerate all three cases that we case about. from `[a1;e]`, `[a1;a2;e]` and `[a1;a2;a3;e]`

jberdine: Oops, my mistake. Enumerating the cases seems good to me. Probably doesn't matter, but likely faster too.

src/Ast.ml
```diff
@@ -80,6 +80,53 @@ let index_op_string = (".", '[', ']')
...
+        Some (index_op_bigarray, args)
+    | ( Ldot (Ldot (Lident "Bigarray", "Genarray"), "get")
+      , [{pexp_desc= Pexp_array l}] )
+      when List.length l > 3 ->
+        Some (index_op_bigarray, l)
+    | _ -> None )
+
+let index_op_set_sugar (ident : Longident.t) args =
+  match all_args_unlabeled args with
+  | None | Some [] -> None
```
jberdine: I guess `Some []` is now redundant

Issue #368, opened 2018-08-28
Compact array


dinosaure:
`ocamlformat` put a break line for each elements of an `array` like this:

```ocaml
  let length =
    [|  0;  1;  2;  3;  4;  5;  6;  7;  8;  8;  9;  9; 10; 10; 11; 11; 12; 12;
       12; 12; 13; 13; 13; 13; 14; 14; 14; 14; 15; 15; 15; 15; 16; 16; 16; 16;
       16; 16; 16; 16; 17; 17; 17; 17; 17; 17; 17; 17; 18; 18; 18; 18; 18; 18;
       18; 18; 19; 19; 19; 19; 19; 19; 19; 19; 20; 20; 20; 20; 20; 20; 20; 20;
       20; 20; 20; 20; 20; 20; 20; 20; 21; 21; 21; 21; 21; 21; 21; 21; 21; 21;
       21; 21; 21; 21; 21; 21; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22;
       22; 22; 22; 22; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23;
       23; 23; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24;
       24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 24; 25; 25;
       25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25;
       25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 25; 26; 26; 26; 26; 26; 26;
       26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26; 26;
       26; 26; 26; 26; 26; 26; 26; 26; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27;
       27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27; 27;
       27; 27; 27; 28 |]
```

A good solution is to keep a compact definition of a big array like this.

jberdine:
It would not be difficult to get e.g.:
```
let length =
[...]

PR #375, opened 2018-08-30
Improve: try to fit simple list/array elements on a single line

gpetiot:
Will fix issue #368.


gpetiot:
I did not see anything awful with `make -C test test test_diff` but it looks like the code of infer is well behind the current state of ocamlformat so the diff is huge and it is easy to miss something.

gpetiot:
It does not change anything, maybe I missed something. Anyway I added an option `--compact-expressions` for this feature (the doc will be added soon).

jberdine:
The dirs under test/code are just normal git repos, so you could use normal git commands to see if they are up to date. An additional technique is to run the test with the merge base of your feature branch, then stage the changes with test_stage, then rerun the test with your feature branch and check the diff with test_diff.

src/Conf.ml
```diff
@@ -376,6 +377,13 @@ module Formatting = struct
     C.choice ~names ~all ~doc ~section (fun conf x ->
         {conf with break_struct= Poly.(x = `Force)} )
 
+  let compact_expressions =
+    let default = false in
+    let doc = "Style for lists and arrays." in
```
jberdine: This could be a bit more descriptive.

src/Conf.ml
```diff
@@ -376,6 +377,13 @@ module Formatting = struct
...
+    let names = ["compact-expressions"] in
```
jberdine: "expressions" is probably overly general, maybe "literals" instead?

src/Fmt_ast.ml
```diff
@@ -90,6 +90,16 @@ let update_config c l =
...
   List.fold ~init:c l ~f:update_one
 
+let fmt_expressions c width to_exp exprs fmt fmt_expr =
+  let grps =
+    List.group exprs ~break:(fun x1 x2 ->
+        (not (is_simple c.conf width (to_exp x1)))
+        || not (is_simple c.conf width (to_exp x2)) )
+  in
+  let fmt_grp exprs = hovbox (-2) (list exprs "@,; " fmt_expr) in
+  if c.conf.compact_expressions then list grps fmt fmt_grp
```
jberdine: Can the computation of grps be sunk into the then branch?

src/Conf.ml
gpetiot: I don't know if "literals" is suited. I was thinking of "compact-linear-data-structures", it designates lists+arrays and not much more, but it is much longer than "compact-lists-arrays" and less explicit.

src/Conf.ml
```diff
@@ -376,6 +377,13 @@ module Formatting = struct
...
+    C.flag ~default ~names ~doc ~section (fun conf compact_expressions ->
+        {conf with compact_expressions} )
```
jberdine: What about handling this like `break-struct`, where the representation is a bool but the option accepts named alternates? For this option, I think using the same choices as `break-infix` would work well. That would also make it easier to give better docs.

src/Conf.ml
jberdine: Yes, "literal" is wrong. Perhaps "break-collection-expressions"?

Also, we could imagine applying this setting to tuples as well.

gpetiot:
I added a fix for default-to-true flags, maybe it would have been better to have it in another PR, but it's so small that I think it is not so bad to put it here.

src/Conf.ml
```diff
@@ -221,18 +222,19 @@ end = struct
...
     in
     let doc = generated_flag_doc ~allow_inline ~doc ~section in
     let docs = section_name section in
-    let term = Arg.(value & flag & info names_for_cmdline ~doc ~docs) in
+    let term = Arg.(value & flag & info names ~doc ~docs) in
     let parse = Bool.of_string in
     let r = mk ~default term in
-    let cmdline_get () = if !r then Some (not invert_flag) else None in
+    let cmdline_get () = if !r then Some true else None in
+    let update c x = update c (if invert_flag then not x else x) in
```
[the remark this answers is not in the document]
gpetiot: Mmh sorry, I thought it was intended to have `no-` options in attributes as well when I tried to write `[@@@ocamlformat "no-break-collection-expressions"]` in commit 8df73c5. I will revert it.

src/Fmt_ast.ml
```diff
@@ -1554,9 +1566,11 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?ext
         ( wrap_fits_breaks "[|" "|]" (Cmts.fmt_within c.cmts pexp_loc)
         $ fmt_atrs )
   | Pexp_array e1N ->
+      let e1N = List.map e1N ~f:(sub_exp ~ctx) in
       hvbox 0
         ( wrap_fits_breaks "[|" "|]"
-            (list e1N "@;<0 1>; " (sub_exp ~ctx >> fmt_expression c))
+            (fmt_expressions c width Fn.id e1N "@;<0 1>; "
```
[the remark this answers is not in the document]
gpetiot: I don't get the same output though, so I will keep it that way for the moment.

src/Fmt_ast.ml
```diff
@@ -90,6 +90,18 @@ let update_config c l =
   in
   List.fold ~init:c l ~f:update_one
 
+let fmt_expressions c width to_exp to_xt exprs fmt fmt_expr =
```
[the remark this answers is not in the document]
gpetiot: Well I don't see what you have in mind for this function, as the types are not the same for lists and for arrays, maybe is was not necessary to factorize it in the first place then?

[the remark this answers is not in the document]
jberdine: Ok, I see. Maybe just replace to_exp and to_xt with one sub_exp function?

PR #381, opened 2018-09-01
Fix: prec of string/array sugar

hhugo:


src/Ast.ml
```diff
@@ -1149,19 +1149,24 @@ end = struct
...
             , e1 :: indices )
           when Option.is_some (index_op_get_sugar txt indices) ->
-            let _, indices =
+            let _, indices' =
               Option.value_exn (index_op_get_sugar txt indices)
             in
-            assert (e0 == exp || snd_f e1 || List.exists indices ~f)
+            assert (
+              e0 == exp || snd_f e1
+              || List.exists indices ~f:snd_f
```
jberdine: Is it worrying that there are expressions left to be formatted which don't survive sugaring? I mean, I would expect that once the indices are sugared, we only try to format expressions that are in the sugared form, but not subexps that exist only in the desugard form. But this check seems to indicate that we do try to format such subexps. Any hints?

hhugo: `x.{1,2,3,4}` is a sugar for `Bigarray.Genarray.get x [| 1; 2; 3; 4 |]`.
Currently, with this example, the context for `4` is `Genarray.get` but `4` is not a "direct" argument of `Genarray.get`.
To make this clearer, we could either be more specific here and change the context used in fmt_ast.ml

jberdine: I see, thanks. I'm not sure what the best way is, but the different cases of e.g. check_exp don't currently all work the same. The older ones look through the intermediate nodes of sugared exps, e.g. the args of the `::` constructor look past the intervening tuple. The index op cases check the sugared form. I guess the first way has the disadvantage of spreading the sugaring logic across more of the code.

Regarding whether to change the context in Fmt_ast, the ctx-ast pair should be defined to be the one which is needed/most useful for the precedence-based parenthesization. In this case it might not matter much, since `e` in `{...e...}` will likely need parens the same times that it does in `[|...e...|]`, but generically the precedence code is expecting the cxt not to be the array that isn't printed.

hhugo: From your previous answer, it's not clear to me whether more changes need to happen.

jberdine: `let _ = f ba.{1, 2, 3, 4}` seems to be a minimal repro for exposed_right_exp midbehaving

hhugo: I've made the changes mentioned above

PR #384, opened 2018-09-01
Fix: while%ext/for%ext

hhugo:


test/passing/extensions.ml.ref
```diff
@@ -81,3 +81,21 @@ let _ =
...
+      done )
+
+let _ =
+  f
+    ( for%ext i = 0 to 1 do
+        ()
+      done )
+    ( while%ext true do
+        ()
+      done )
```
jberdine: :-) where do you find such code...

hhugo: I think it was something like
```ocaml
dont_wait_for begin
   while%lwt cond do
     expr
   done
end
```

I've started to run ocamlformat against all packages available in the opam-repository

PR #386, opened 2018-09-01
Fix: consecutive prefix operator

hhugo:


src/Ast.mli
```diff
@@ -150,6 +150,10 @@ val exposed_right_exp : cls -> expression -> bool
 (** [exposed_right_exp cls exp] holds if there is a right-most subexpression
     of [exp] which is of class [cls] and is not parenthesized. *)
 
+val exposed_left_exp : expression -> bool
+(** [exposed_right_exp cls exp] holds if there is a right-most subexpression
+    of [exp] which is of class [cls] and is not parenthesized. *)
```
jberdine: comment stale

src/Ast.ml
```diff
@@ -1703,6 +1705,12 @@ end = struct
     register_reset (fun () -> Hashtbl.clear memo) ;
     memo
 
+  let rec exposed_left_exp e =
+    match e.pexp_desc with
+    | Pexp_apply (op, _) when is_prefix op -> true
```
hhugo: Do we need a recursive call on `op` here ?

jberdine: It looks like it, though maybe there is some way that such parsetree values are never created. Recursing can't hurt I think.

Issue #357, opened 2018-08-27
Add presets config similar to ocp-indent.


hhugo:
Here is what they do in ocp-indent: https://github.com/OCamlPro/ocp-indent/blob/4830ebf5d1c1b8f47bc152ff13d2c6aa1aad705b/src/indentConfig.ml#L51

I'll like to be able to define a preset config for janestreet, currently hardcoded here:  https://github.com/janestreet/ocamlformat/blob/jane/src/Conf.ml#L813

jberdine:
I think that there are two things going on here:
1. packaging sets of config settings into single styles
2. defining those aggregate styles in source code

To me, 1. is clearly desirable. But I'm not sure about 2. Is it seen as a feature that e.g. to change the definition of the style, using a new version would be needed?

hhugo:
I d like to collect feedback from @AltGr about the setup in ocp-indent

samoht:
As far as I know, most of the people use the default profile for ocp-indent (with some level of tweaks via local `.ocp-config` from power users). Janestreet has its own `ocp-indent` profile which is set in stone since the few initial iterations. And the "beginner" profile is used by `ocamltop`, but I am not sure that was such in a good idea.

IMHO, like in ocp-indent having sensible good default is the priority, and a few profiles corresponding to the code style of large industry users (as long as they don't change too often, and they usually don't :-) )

jberdine:
Ok, it sounds like it makes sense to at least start out by hardcoding the definitions of the styles. If it becomes an issue, supporting defining styles in config files can be done later.

PR #390, opened 2018-09-02
Preset profiles (default & janestreet)

gpetiot:
Should take care of issue #357

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -105,7 +105,8 @@ module C : sig
 
   val section_name : [`Formatting | `Operational] -> string
 
-  val choice : all:(string * 'a * string) list -> 'a option_decl
+  val choice :
+    ?has_default:bool -> all:(string * 'a * string) list -> 'a option_decl
 
   val flag : default:bool -> bool option_decl
 
@@ -153,12 +154,14 @@ end = struct
     if cond || Poly.(section = `Operational) then ""
     else "Cannot be set in attributes."
 
-  let generated_choice_doc ~allow_inline ~all ~doc ~section =
+  let generated_choice_doc ~allow_inline ~all ~doc ~section ~has_default =
     let open Format in
     let default =
-      asprintf "The default value is $(b,%a)."
-        (fun fs (v, _, _) -> fprintf fs "%s" v)
-        (List.hd_exn all)
+      if has_default then
+        asprintf "The default value is $(b,%a)."
+          (fun fs (v, _, _) -> fprintf fs "%s" v)
+          (List.hd_exn all)
+      else ""
     in
     asprintf "%s %a %s %s" doc
       (pp_print_list
@@ -187,11 +190,12 @@ end = struct
     | `Formatting -> Cmdliner.Manpage.s_options ^ " (CODE FORMATTING STYLE)"
     | `Operational -> Cmdliner.Manpage.s_options
 
-  let choice ~all ~names ~doc ~section
+  let choice ?(has_default = true) ~all ~names ~doc ~section
       ?(allow_inline = Poly.(section = `Formatting)) update =
-    let open Cmdliner in
     let _, default, _ = List.hd_exn all in
-    let doc = generated_choice_doc ~allow_inline ~all ~doc ~section in
+    let doc =
+      generated_choice_doc ~allow_inline ~all ~doc ~section ~has_default
+    in
     let docv = generated_choice_docv ~all in
     let opt_names = List.map all ~f:(fun (x, y, _) -> (x, y)) in
     let docs = section_name section in
@@ -760,23 +764,6 @@ let config =
     Arg.(
       value & opt list_assoc default & info ["c"; "config"] ~doc ~docs ~env)
 
-let validate () =
-  if List.is_empty !inputs then
-    `Error (false, "Must specify at least one input file, or `-` for stdin")
-  else if
-    List.equal ~equal:String.equal !inputs ["-"] && Option.is_none !name
-  then `Error (false, "Must specify name when reading from stdin")
-  else if !inplace && Option.is_some !name then
-    `Error (false, "Cannot specify --name with --inplace")
-  else if !inplace && Option.is_some !output then
-    `Error (false, "Cannot specify --output with --inplace")
-  else if (not !inplace) && List.length !inputs > 1 then
-    `Error (false, "Must specify exactly one input file without --inplace")
-  else `Ok ()
-
-;;
-parse info validate
-
 let default =
   { break_cases= C.default Formatting.break_cases
   ; break_collection_expressions=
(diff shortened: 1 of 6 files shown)


src/Conf.ml
```diff
@@ -789,6 +772,70 @@ let default =
...
+  ; ocp_indent_compat= true
+  ; parens_tuple= `Multi_line_only
+  ; sequence_style= `Terminator
+  ; type_decl= `Sparse
+  ; wrap_fun_args= false
+  ; module_item_spacing= `Compact }
+
+let preset_profiles =
+  [ ( "janestreet"
+    , `JaneStreet
```
hhugo: Is this variant really useful ?

src/Conf.ml
```diff
@@ -916,10 +963,15 @@ let update_using_xdg =
...
 
 let build_config ~filename =
-  default
-  |> Staged.unstage update_using_xdg
-  |> read_config ~filename ~parents:true
-  |> update_using_env |> C.update_using_cmdline
+  match !profile with
+  | Some profile ->
+      let f (_, x, _, p) = if Poly.(x = profile) then Some p else None in
+      List.find_map_exn preset_profiles ~f
```
hhugo: I'm don't think we want to disallow customisation here.
I would also expect a "default" preset, similar to ocp-indent.

src/Conf.ml
```diff
@@ -789,6 +772,67 @@ let default =
...
+  ; wrap_fun_args= false
+  ; module_item_spacing= `Compact }
+
+let preset_profiles =
+  [ ( "janestreet"
+    , "$(b,janestreet) is the profile used at JaneStreet."
+    , janestreet_profile ) ]
+
+let profile =
+  let open Format in
```
jberdine: can the scope of this open be reduced, I guess to just the body of doc

src/Conf.ml
jberdine: I suppose that we will soon also have 'compact' and 'sparse' profiles, at which point the variant may be more useful.

src/Conf.ml
jberdine: To my understanding, we may encounter a profile setting via any of the existing ways of configuration. Semantically, processing a profile option should set all the config options according to the profile, overriding everything in any lower priority config settings. And then processing of the rest of the config options should proceed, enabling customization of a profile one option at a time.

src/Conf.ml
jberdine: And of course there should also be a 'default' profile.

src/Conf.ml
jberdine: For cases where there are profile and other options on the command line, I think it makes sense to take the profile as lower priority than the other options.

src/Conf.ml
jberdine: can the scope of this open be reduced, I guess to just the body of doc

src/Conf.ml
```diff
@@ -741,6 +742,39 @@ let config =
...
+  ; sequence_style= `Terminator
+  ; type_decl= `Sparse
+  ; wrap_fun_args= false
+  ; module_item_spacing= `Compact }
+
+let profile =
+  let doc = "Preset profiles." in
+  let names = ["p"; "profile"] in
+  let all =
+    [ ("default", Fn.id, "$(b,default) uses the default values of options.")
```
jberdine: The default profile should set all value to the default values, rather than be a no-op.

src/Conf.ml
```diff
@@ -741,6 +742,39 @@ let config =
...
+    ; ( "janestreet"
+      , apply_janestreet_profile
+      , "$(b,janestreet) is the profile used at JaneStreet." ) ]
+  in
+  C.choice ~names ~all ~doc ~section (fun conf f -> f conf)
```
jberdine: Among the options on the command line, the profile needs to be the lowest priority. Is this aspect currently being handled?

src/Conf.ml
```diff
@@ -741,6 +742,39 @@ let config =
     Arg.(
       value & opt list_assoc default & info ["c"; "config"] ~doc ~docs ~env)
 
+let apply_janestreet_profile c =
+  { c with
```
jberdine: I'm not sure that it makes sense to pass in a config to a profile, since they are intended to set all the values.

src/Conf.ml
hhugo: Similar for the janestreet profile.

src/Conf.ml
hhugo: I would expect that options are processed in order.

src/Conf.ml
```diff
@@ -783,12 +767,66 @@ let default =
...
+  ; sequence_style= `Terminator
+  ; type_decl= `Sparse
+  ; wrap_fun_args= false
+  ; module_item_spacing= `Compact }
+
+let profile =
+  let doc = "Preset profiles." in
+  let names = ["p"; "profile"] in
+  let all =
+    [ ("none", Fn.id, "$(b,none) when no profile is selected.")
```
hhugo: What's the purpose of this one ?

gpetiot: I think we need a profile that does nothing that will be the default value of the option, so that we do not override the options that are already set (since the "default" profile sets each option now) even when no profile is set.

hhugo: Sorry, I don't follow. As far as I understand, `default` is the default value for that option because this is what we start with while build the config.

jberdine: @gpetiot do you think that we need a do-nothing profile, or could we just not specify one in such use cases?

jberdine:
Could you add some tests to check that the overriding and priority handling is correct? At the very least we need to check that specifying an option on the command line later than a profile overrides the profile, and that the profile overrides the config file.

src/Conf.ml
```diff
@@ -800,15 +800,14 @@ let profile =
...
+    [ ( "default"
+      , default
       , "$(b,default) sets each option to its default value." )
     ; ( "janestreet"
-      , (fun _ -> janestreet_profile)
+      , janestreet_profile
       , "$(b,janestreet) is the profile used at JaneStreet." ) ]
   in
-  C.choice ~names ~all ~doc ~section (fun conf f -> f conf)
+  C.choice ~names ~all ~doc ~section (fun _conf f -> f)
```
jberdine: I think that maybe I now understand why you had the `none` profile: not to actually be specified, but to be the default value as far as cmdliner is concerned. Is it currently the case that with no profile specified, processing the command line will set options using the default profile even if there are lower priority files, etc.?

gpetiot: It is what I thought at first, but the choice options are actually default to None so it it is not specified it is None and not Some default, it should be OK.

jberdine: Ok, good. The only niggle is that the doc of the profile option reads like leaving it out is the same as specifying `--profile=default`, which isn't right.

gpetiot: I added the possibility to add a "IGNORE" hidden value for choice options, so that this option is not shown as the default. I'm not fully satisfied with this solution but I don't have any better idea.

src/Conf.ml
```diff
@@ -34,6 +34,7 @@ type t =
   ; module_item_spacing: [`Compact | `Sparse]
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
+  ; profile: t -> t
```
jberdine: Wouldn't `t option` instead of `t -> t` be clearer and still expressive enough.

src/Conf.ml
```diff
@@ -189,8 +189,14 @@ end = struct
   let choice ~all ~names ~doc ~section
       ?(allow_inline = Poly.(section = `Formatting)) update =
     let open Cmdliner in
-    let _, default, _ = List.hd_exn all in
-    let doc = generated_choice_doc ~allow_inline ~all ~doc ~section in
+    let default_name, default, _ = List.hd_exn all in
+    let opt_default, all =
+      if String.equal default_name "IGNORE" then (None, List.tl_exn all)
```
jberdine: Have you considered adding an optional boolean arg to `choice` that could be passed in the profile case and branched on here?

src/Conf.ml
```diff
@@ -783,12 +772,66 @@ let default =
...
+  ; margin= 90
+  ; ocp_indent_compat= true
+  ; parens_tuple= `Multi_line_only
+  ; sequence_style= `Terminator
+  ; type_decl= `Sparse
+  ; wrap_fun_args= false
+  ; module_item_spacing= `Compact }
+
+let _profile =
+  let doc = "Preset profiles." in
```
jberdine: I think that it would help to make this more explicit with something like: `Preset profiles which set $(i,all) options, overriding lower priority configuration.`

src/Conf.ml
```diff
@@ -34,6 +34,7 @@ type t =
   ; module_item_spacing: [`Compact | `Sparse]
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
+  ; profile: t option
```
hhugo: Do we use do any thing with this field?

gpetiot: No we don't, I will remove it.

src/Conf.ml
```diff
@@ -186,15 +189,17 @@ end = struct
...
     let _, default, _ = List.hd_exn all in
-    let doc = generated_choice_doc ~allow_inline ~all ~doc ~section in
+    let doc =
+      generated_choice_doc ~allow_inline ~all ~doc ~section ~has_default
+    in
     let docv = generated_choice_docv ~all in
     let opt_names = List.map all ~f:(fun (x, y, _) -> (x, y)) in
     let docs = section_name section in
     let term =
+      let open Cmdliner in
```
jberdine: nit: this and the next line being in different styles looks a little odd, maybe use `Cmdliner.Arg.(` instead

jberdine:
Do I understand correctly that currently if the env var is set, then in the case where there is an .ocamlformat file including a profile setting, the env var will be used, while if there is no config file but instead a command line --profile setting, then the env var will be ignored?

gpetiot:
> Do I understand correctly that currently if the env var is set, then in the case where there is an .ocamlformat file including a profile setting, the env var will be used

The priority of the env is higher than the priority of the an .ocamlformat file so it can override the options set by the profile yes.

> [...] if there is no config file but instead a command line --profile setting, then the env var will be ignored?

Yes because the command line has the higher priority so the profile set on command line overrides everything else.

PR #391, opened 2018-09-03
Improve: disable ocamlformat if no dot ocamlformat in the project

hhugo:
The PR adds a way to not reformat files that live in a "project" that doesn't contain a `.ocamlformat` file.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -771,6 +771,11 @@ let no_version_check =
   let default = false in
   mk ~default Arg.(value & flag & info ["no-version-check"] ~doc ~docs)
 
+let inside_project_only =
+  let doc = "TODO" in
+  let default = false in
+  mk ~default Arg.(value & flag & info ["inside-project-only"] ~doc ~docs)
+
 let config =
   let doc =
     "Aggregate options. Options are specified as a comma-separated list of \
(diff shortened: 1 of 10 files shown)


hhugo:
@samoht, @gpetiot, @jberdine, alternative suggestion welcome.

jberdine:
IIUC, there is a similar situation with dune, which has resulted in a way of specifying subdirs to ignore. This would result in putting a config file in the dir containing the project root dir to be ignored, and adding that dir's name into the config. I'm not sure if that is better, but think that the space was explored at some length for dune.

hhugo:
Just to be clear, the issue that this PR is trying to solve is about not formatting files that are not meant to be formatted. In particular, I don't want files from the ocaml compiler to be formatted when I edit them in emacs.

jberdine:
I see. I guess perhaps I should enable format-on-save myself since it seems most people use it that way.

For that particular use case, you prefer to call ocamlformat and have it detect that it's disabled, rather than e.g. configuring emacs to not call ocamlformat in the first place? I'm thinking of for instance enabling format-on-save in tuareg-mode-hook only if the file name matches some user-specific whitelist.

This PR basically says to look for an .ocamlformat file in the ancestor directories of the input file, but stop looking when reaching a "project root" dir, right? I'm not sure the implementation is the simplest, but the basic approach seems good to me. We should document it in the man page where it discusses where .ocamlformat files are searched for.

Perhaps this stop at project root mode should be the default?

hhugo:
I'm now curious about how you use ocamlformat if not with format-on-save. Some logic could be implemented in the emacs mode, but similar logic would then need to be implemented in other editor mode a well

samoht:
`M-q` only formats the current top-level expression, right? How do you do this without `--lines`? But maybe the expression is computed by tuareg mode already.

jberdine:
My binding in merlin-mode-map shadows tuareg's, so it formats the whole buffer.

hhugo:
I've reworked the PR.
The behaviour is now triggered by `--inside-project-only` (better name welcome).
This flags is now passed to ocamlformat in the emacs mode.
I've also added a `--print-config` flag that will print config updates together with their origin (see new tests)

hhugo:
I've reverted the change to the emacs mode for now so that the behaviour is preserved (to be discussed in another PR)

src/Conf.ml
```diff
@@ -726,6 +763,14 @@ let inputs =
   mk ~default
     Arg.(value & pos_all file_or_dash default & info [] ~doc ~docv ~docs)
 
+let inside_project_only =
+  let doc =
+    "Do not read .ocamlformat config files outside the project. If no \
+     config file is found, disable ocamlformat."
```
jberdine: What is considered to be inside a project should be documented explicitly. Maybe: `The project root of an input file is taken to be the nearest ancestor directory that contains .git, .hg, or dune-project.`

src/Conf.ml
```diff
@@ -253,30 +265,55 @@ end = struct
...
+      {names; parse; update; cmdline_get; allow_inline; default; to_string}
+    in
     store := Pack opt :: !store ;
     opt
 
-  let update ~config ~name ~value ~inline =
+  let log_update ~from ~name ~value =
+    match from with
+    | `Attribute -> ()
+    | (`File _ | `Config | `Commandline) as from ->
```
jberdine: nit: `as from` redundant

src/Conf.ml
```diff
@@ -726,6 +763,14 @@ let inputs =
   mk ~default
     Arg.(value & pos_all file_or_dash default & info [] ~doc ~docv ~docs)
 
+let inside_project_only =
+  let doc =
+    "Do not read .ocamlformat config files outside the project. If no \
+     config file is found, disable ocamlformat."
+  in
+  let default = false in
+  mk ~default Arg.(value & flag & info ["inside-project-only"] ~doc ~docs)
```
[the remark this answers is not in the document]
jberdine: Is this an option that could productively use a separate env var?

src/Conf.ml
```diff
@@ -993,20 +1055,50 @@ let kind_of fname =
...
+    collect_files ~dir:(Filename.dirname (to_absolute filename)) []
+  in
+  let files =
+    match (xdg_config, !inside_project_only) with
+    | None, _ | Some _, true -> files
+    | Some f, false -> f :: files
+  in
+  let verbose = !print_config in
+  if verbose then
+    Option.iter project_root ~f:(Format.eprintf "project-root=%s\n%!") ;
```
jberdine: call `to_relative`?

src/Conf.ml
hhugo: it is not if you want the next match to be exhaustive.

jberdine: Indeed, I missed that.

src/Conf.ml
hhugo: I don't think so.
It would be harder to read `project-root=../../../../`

src/Conf.ml
```diff
@@ -253,30 +265,55 @@ end = struct
...
           if inline && not allow_inline then
             Some (Error (`Misplaced (name, value)))
           else
-            try Some (Ok (update config (parse value))) with _ ->
-              Some (Error (`Bad_value (name, value)))
+            try
+              let config = update config (parse value) in
+              if verbose then log_update ~from ~name ~value ;
+              Some (Ok config)
+            with _ -> Some (Error (`Bad_value (name, value)))
```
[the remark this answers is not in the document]
hhugo: I've reduced the scope

src/Conf.ml
hhugo: updated

src/Conf.ml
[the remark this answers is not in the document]
hhugo: I'm not sure about the env variable. I like your point about command line flags being more discoverable.

src/Conf.ml
```diff
@@ -253,30 +265,55 @@ end = struct
...
         if List.exists names ~f:(String.equal name) then
           if inline && not allow_inline then
             Some (Error (`Misplaced (name, value)))
           else
-            try Some (Ok (update config (parse value))) with _ ->
-              Some (Error (`Bad_value (name, value)))
+            match parse value with
+            | packed_value ->
+                if verbose then log_update ~from ~name ~value ;
+                Some (Ok (update config packed_value))
```
jberdine: @hhugo I think it's ok to also push `update` out of the handler

hhugo:
I've implemented you suggestion `parse`, it now returns a result.

src/Conf.ml
```diff
@@ -764,6 +831,11 @@ let output =
       & opt (some string) default
       & info ["o"; "output"] ~doc ~docs ~docv)
 
+let print_config =
+  let doc = "Print config" in
```
jberdine: nit: `.` at end.

PR #397, opened 2018-09-05
[WIP] [JS] Improve: various whitespaces

gpetiot:
From the JS fork.


hhugo:
It would be better to split the change, from the reformatting

hhugo:
Also, I suspect that this will need to go under an option

hhugo:
Also, I suspect that this will need to go under an option

hhugo:
You might even need to split it in multiple pr, depending on how many changes are included in that commit

gpetiot:
I don't know if it is a good thing to split it, it may impact many files but the changes are not that big.
Of course we can add an option, although I like this new style better (would like to make it the new default).

jberdine:
Yes, agree, but even more, it would be better to add the options and then perhaps change the settings with the resulting format change in a separate PR.

gpetiot:
Alright, I will split this PR and add new options.

Issue #17, opened 2017-10-21
Give informative error message when comment dropped


jberdine:
Bugs in the formatter can lead to a comment in the original source not being generated in the output. Often such a bug can be worked around by moving the comment slightly. Currently, dropping a comment is detected and an exception is raised. It would be better to issue an standard compiler error message identifying the dropped comment.

PR #401, opened 2018-09-06
Improve: Better error with location when comment dropped

gpetiot:
Should improve the situation for issue #17 
Though raising an error only gives information about the first dropped comment, maybe it is better to use warnings and have a message about all dropped comments?

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -563,7 +563,9 @@ let remaining_comments t =
     Hashtbl.to_alist t
     |> List.concat_map ~f:(fun (ast_loc, cmts) ->
            List.map cmts ~f:(fun (cmt_txt, cmt_loc) ->
-               ( before_after
+               ( cmt_loc
+               , cmt_txt
+               , before_after
                , let open Sexp in
                  List
                    [ List [Atom "ast_loc"; Location.sexp_of_t ast_loc]
--- a/src/Cmts.mli
+++ b/src/Cmts.mli
@@ -119,7 +119,7 @@ val has_within : t -> Location.t -> bool
 val has_after : t -> Location.t -> bool
 (** [has_after t loc] holds if [t] contains some comment after [loc]. *)
 
-val remaining_comments : t -> (string * Sexp.t) list
+val remaining_comments : t -> (Location.t * string * string * Sexp.t) list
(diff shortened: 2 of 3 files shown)


jberdine:
Can you explain the implementation strategy a bit? In particular, I'm not familiar with the mechanics of Location.register_error_of_exn. How does using it compare to constructing the message from the exception directly in quiet_exn?

gpetiot:
Well I am not familiar with this function (found it just before). For me it is just a way to produce standard errors (with locations) from custom exceptions, so I am not sure it is the right tool for the job.

hhugo:
You can probably use `Location.print_loc` to generate some "standard" error message.

hhugo:
I gave it a quick try and push the result on your branch. Probably need some more work.
Here is the result.
```
ocamlformat$ _build/dev/src/ocamlformat.exe test/passing/source.ml 
ocamlformat.exe: Cannot process "test/passing/source.ml".
  BUG: comments dropped.
File "test/passing/source.ml", line 1022, characters 4-33:
"type t is abstract here"
File "test/passing/source.ml", line 1377, characters 2-76:
"Could also use [get_case] for proj, but direct definition is shorter"
File "test/passing/source.ml", line 1317, characters 2-76:
"Could also use [get_case] for proj, but direct definition is shorter"
File "test/passing/source.ml", line 1084, characters 4-33:
"type t is abstract here"
File "test/passing/source.ml", line 1329, characters 2-67:
"Coherence of sum_inj and sum_cases is checked by the typing"
File "test/passing/source.ml", line 1024, characters 21-50:
"in this branch: t = int"
File "test/passing/source.ml", line 1353, characters 9-62:
"One can also write the type annotation directly"
File "test/passing/source.ml", line 1087, characters 26-42:
"t = string"
File "test/passing/source.ml", line 1093, characters 8-45:
"t = ('a, 'b) for some 'a and 'b"
File "test/passing/source.ml", line 1090, characters 8-37:
"t = 'a list for some 'a"
File "test/passing/source.ml", line 1025, characters 26-42:
"t = string"
File "test/passing/source.ml", line 1086, characters 21-50:
"in this branch: t = int"
File "test/passing/source.ml", line 1028, characters 8-37:
"t = 'a list for some 'a"
```

Issue #393, opened 2018-09-04
Isolate configuration files


emillon:
Hello,

I am implementing an integration with dune (see ocaml/dune#1201).

If I understand it correctly, at the moment, the configuration is read in cascade from the closest `.ocamlformat` all the way up until a global `.ocamlformat` file.

In the context of dune, this can cause reproducibility problems, because in a dune project with no `.ocamlformat` file (or with a partial one), running `ocamlformat` on a file will read each developer's global configuration file, and will cause different results.

I'm not sure about the best technical solution, but a `--root` configuration option that never reads files above a particular directory seems enough. This option would also disable reading the global configuration file and from the environment as well.

What do you think about this?

Thanks!

hhugo:
related to https://github.com/ocaml-ppx/ocamlformat/pull/391

jberdine:
Yes, the current behavior of ocamlformat is the way it is just because it was a quick hack and I didn't know a good way to identify the root. Adding a command line option would be very clear, and work well in cases where it is called from something that knows the project root, but maybe would require too much interaction for the use case in #391. What do you think @hhugo, would just a `--root` option work there?

emillon:
To recap I think that there are three use cases:
- `ocamlformat` is invoked from `dune`. `dune` knows the root dir, so `--root` is enough, but detecting `dune-project` or `.git` might work as well (there are probably cases where it will be less precise).
- `ocamlformat` is invoked from an editor. The current directory might be a good root in the case of TUI editors, but for GUI editors the current directory will be `$HOME` or something, so autodetection is probably required
- `ocamlformat` is invoked by hand. This situation is pretty important since it might be the first interaction of a user with this program, and it will potentially happen outside a dune or git directory. Once again, autodetection should be good enough, but there might be surprising cases when a new user tries to configure ocamlformat and the configuration gets ignored.

jberdine:
- It seems clear that there should be a `--root` option for use when the caller knows how to set it.
- I expect that for editors, `CWD` is very often `$HOME` or somewhere below the actual project root, seems too fragile to use to me.
[...]

PR #402, opened 2018-09-06
Improve: --root option to isolate configuration files

gpetiot:
Should fix issue #393


src/Conf.ml
```diff
@@ -899,7 +910,13 @@ let parse_line config ~from s =
 
 let rec read_conf_files conf ~dir ~parents =
   let dir' = Filename.dirname dir in
-  if (not (String.equal dir dir')) && Caml.Sys.file_exists dir then
+  let inside_root =
+    Option.value_map !root ~default:true
+      ~f:(String.is_substring ~substring:dir)
```
jberdine: I would be much more comfortable with converting dir and root to absolute paths (better would be to use Filename.realpath) and checking if calling dirname on dir zero or more times results in root, or using Filename.parts on both and checking if the root list is a prefix of the dir list. Basically, using filenames to do anything but open files is very fragile, and it would be better to try not to.

gpetiot: There is this warning `Warning! this library assumes we are in a POSIX compliant OS. It will not work properly under windows.` in the Core.Filename documentation, so I think it is better to stick with the vanilla Filename lib.

jberdine: Well, treating paths as raw strings is significantly worse. Generally my position is that the consequence of windows failing to implement the established APIs should not be that everything is worse. It would be better to implement something robust, which someone who wants to work on windows can use as a specification for what windows-specific support is needed.

Also, there may be other libraries for filenames than Core, maybe even which have already resolved windows deficiencies.

jberdine: Could you have a look at using [fpath](https://github.com/dbuenzli/fpath) for this, in particular `normalize`. And unless realpath is used somewhere, the docs need to say that input files and project root need to be specified using the same symbolic or hard links.

gpetiot:
The rebase is done.

src/Conf.ml
```diff
@@ -1009,8 +1020,10 @@ let parse_line config ~verbose ~from s =
   | _ -> Error (`Malformed s)
 
 let is_project_root dir =
-  List.exists project_root_witness ~f:(fun name ->
-      Caml.Sys.file_exists (Filename.concat dir name) )
+  Option.value_map !root ~f:(String.equal dir)
```
hhugo: This is eagerly computing the default value even if root is provided

src/Conf.ml
```diff
@@ -1009,8 +1020,11 @@ let parse_line config ~verbose ~from s =
   | _ -> Error (`Malformed s)
 
 let is_project_root dir =
-  List.exists project_root_witness ~f:(fun name ->
-      Caml.Sys.file_exists (Filename.concat dir name) )
+  match !root with
+  | Some root -> String.equal dir root
```
hhugo: This won't work with
- relative path
- root with a trailing separator
- path containing `./` and `../`

Note that the logic around `.ocamlformat` files lookups does not work in presence of `../` already

src/Conf.ml
```diff
@@ -1009,8 +1020,11 @@ let parse_line config ~verbose ~from s =
   | _ -> Error (`Malformed s)
 
 let is_project_root dir =
-  List.exists project_root_witness ~f:(fun name ->
-      Caml.Sys.file_exists (Filename.concat dir name) )
+  match !root with
+  | Some root -> Fpath.(equal (v dir |> normalize) (v root |> normalize))
```
hhugo: normalisation of the root should happen only once.
Also, how does that work when the root is not absolute ?

gpetiot: Now it is normalized only once and it should be absolute (as well as the dir it is compared to).

src/Conf.ml
```diff
@@ -1009,8 +1026,11 @@ let parse_line config ~verbose ~from s =
   | _ -> Error (`Malformed s)
 
 let is_project_root dir =
-  List.exists project_root_witness ~f:(fun name ->
-      Caml.Sys.file_exists (Filename.concat dir name) )
+  match root with
+  | Some root -> Fpath.(equal (v dir |> normalize) root)
```
jberdine: It would probably be better to normalize the dir pathname when first entering collect_files (or earlier) instead of repeatedly here.

src/Conf.ml
```diff
@@ -836,6 +836,17 @@ let print_config =
...
+let root =
+  let docv = "DIR" in
+  let doc =
+    "Root of the project. If specified, only take into account \
+     .ocamlformat configuration files inside $(docv) and its \
+     subdirectories."
+  in
+  let default = None in
+  mk ~default
+    Arg.(value & opt (some dir) default & info ["root"] ~doc ~docs ~docv)
```
jberdine: My feeling is that if root is set, it should be interpreted as if disable-outside-project is also set. Or are there other use cases I'm overlooking? Perhaps it would be best to rename disable-outside-project to something like disable-outside-detected-project?

gpetiot: Seems like a good idea to me.

src/Conf.ml
```diff
@@ -957,6 +968,12 @@ let validate () =
 ;;
 parse info validate
 
+let to_absolute file =
+  Filename.(if is_relative file then concat (Unix.getcwd ()) file else file)
```
jberdine: nit: It would probably be better to switch all filename manipulation to Fpath.

src/Conf.ml
```diff
@@ -1101,9 +1118,11 @@ let xdg_config =
   | None -> None
 
 let build_config ~filename =
-  let files, project_root =
-    collect_files ~dir:(Filename.dirname (to_absolute filename)) []
+  let dir =
+    filename |> to_absolute |> Filename.dirname |> Fpath.v
+    |> Fpath.normalize |> Fpath.to_string
```
jberdine: nit: It might be clearer to use Fpath.segs and express the search up the file system in terms of the list of strings rather than repeatedly performing filename manipulations.

src/dune
```diff
@@ -18,7 +18,7 @@ let preprocess =
  (names ocamlformat ocamlformat_reason)
  (flags (:standard -open Import))
  %s
- (libraries cmdliner format_ import ocaml-migrate-parsetree unix))
+ (libraries cmdliner format_ import ocaml-migrate-parsetree unix fpath))
```
jberdine: libs were alphabetized

ocamlformat.opam
```diff
@@ -16,5 +16,6 @@ depends: [
   "dune" {build}
   "ocaml-migrate-parsetree" {>= "1.0.10"}
   "stdio"
+  "fpath"
```
jberdine: deps were alphabetized

gpetiot:
My last commit is here to feed the discussion, I don't think the implementation is better now that more things are handled by fpath.

src/Conf.ml
jberdine: I don't understand how this change disrupts the CWD handling in test.sh, do you?

jberdine:
From your last commit I guess you prefer repeated calls to dirname instead of segs, right?

gpetiot:
We would have to rebuild the path from the segs anyway, I don't see how this is better than what is currently done.

gpetiot:
I have the feeling that for the moment we should only use Fpath for path normalization then convert it to a string and keep using the original code, so I should probably just revert my last commit.

test/passing/verbose2.ml.ref
```diff
@@ -1,6 +1,10 @@
-project-root={CWD}
-margin=77 ({CWD}/.ocamlformat:1)
-wrap-comments=true ({CWD}/.ocamlformat:2)
-doc-comments=before (Environment variable or --config)
-doc-comments=before (command line)
-
+project-root=/home/gpe/ocamlformat/
+margin=77 (/home/gpe/ocamlformat/.ocamlformat:1)
```
hhugo: This test will need to be fixed.

jberdine:
I might be confused since I don't understand why the paths would need to be rebuilt from the seg list, but the advantage of converting both the root and input to a list of segments is that traversing a list, removing one element at each iteration, gives a much more straightforward termination condition than a loop that repeatedly calls dirname. All the possibly complicated corner cases that might lead to divergence of the dirname-loop are outsourced to the library when using the list of segs.

src/Conf.ml
```diff
@@ -957,6 +968,17 @@ let validate () =
...
+let to_absolute file =
+  Fpath.(if is_rel file then append (Unix.getcwd () |> v) file else file)
+
+let path_exists p = Fpath.to_string p |> Caml.Sys.file_exists
+
+let dirname p = Fpath.split_base p |> fst
+
+let root =
+  Option.map !root
+    ~f:Fpath.(fun x -> v x |> to_absolute |> normalize |> dirname)
```
hhugo: why do we take dirname here ?

test/passing/verbose2.ml.ref
```diff
@@ -1,6 +1,6 @@
-project-root={CWD}
-margin=77 ({CWD}/.ocamlformat:1)
-wrap-comments=true ({CWD}/.ocamlformat:2)
+project-root=../../ocamlformat
```
gpetiot: The previous value was wrong (the test is executed from the `test` directory, not its parent). I think the excessive `../ocamlformat` is due to `relativize` and using `normalize` before or after does not fix this.

src/Cmts.ml
```diff
@@ -41,7 +41,7 @@ end) : sig
 
   val children : t -> Itv.t -> Itv.t list
 
-  val dump : t -> Fmt.t
+  val dump : t -> Formatting.t
```
gpetiot: The renaming is due to a name collision (Bos requires the Fmt package)

gpetiot:
Every comment has been addressed, I think this one is ready to be merged now.

src/Cmts.ml
jberdine: I'm sorry, but no, this does not seem to be worth it. Is there something I'm overlooking? Bos should wrap its internal dependencies instead of polluting the global namespace.

src/Conf.ml
```diff
@@ -1009,25 +1054,41 @@ let parse_line config ~verbose ~from s =
...
+        | Ok true -> filename :: acc
+        | _ -> acc
+      in
+      if is_project_root dir && disable_outside_detected_project then
+        (acc, Some dir)
+      else
+        let dir_exists =
+          match Bos.OS.Dir.exists dir with Ok r -> r | Error _ -> false
+        in
+        if dir_exists then collect_files ~segs:upper_segs acc
```
jberdine: I don't understand the situation where`dir_exists` is false, can you explain when that might happen and what the desired behavior is?

src/Cmts.ml
gpetiot: The problem is that the library named `fmt` (https://github.com/dbuenzli/fmt) is a dependency of `bos`

jberdine: I know, but that doesn't change things wrt this renaming. There are just a few small functions using bos, if it means not using bos and using the standard versions of them instead until bos is fixed to not pollute the global namespace with its internal deps, then so be it.

src/Conf.ml
gpetiot: This case should not exist, I simplified the function.

PR #408, opened 2018-09-08
Add: compact and sparse profiles

jberdine:
This PR adds `compact` and `sparse` profiles, whose settings are up for discussion. Based on feedback so far, it seems clear that there are at least two distinct sets of preferences people have, which possibly fall roughly into wanting "mostly compact" versus "more sparse" code. The hope is that most users will be able to select one of these two profiles and want to change few if any other options. What does everyone think at a high-level, and also about the particular settings?


gpetiot:
It looks good to me. I agree that this kind of profiles would be useful for beginners.

PR #420, opened 2018-09-10
Fix: print error detail in debug mode only

hhugo:


hhugo:
Testded with
```shell
$ echo "let create (type a) (type b) t i w p =let module T = (val (t : (a, b) t)) in T.create i w p" | \
~/l/ocamlformat/_build/dev/src/ocamlformat.exe --name a.ml --quiet -
ocamlformat.exe: Cannot process "/tmp/ocamlformate818e2a.ml".
```

PR #424, opened 2018-09-11
Fix: aliases inside constructor declaration

### LINKED ISSUE #411: Failing to format aliases inside constructor declaration
```ocaml
exception Recursion_error of (Lv6Id.long as 'id) * (string list as 'stack)
```

gpetiot:
Should fix #411


gpetiot:
I am not happy with this patch, so I am open to suggestions.

hhugo:
related #256

Issue #398, opened 2018-09-05
-o /dev/null fails


emillon:
Hi,

The following happens when I'm trying to output to `/dev/null`:

```
 % touch empty.ml
 % ocamlformat empty.ml -o /dev/null
ocamlformat: Cannot process "x.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: unhandled exception.
```

Hooking up `strace` it seems that it's trying to open many files in `/dev`.

While I don't really need `-o /dev/null` to work, I'm a bit surprised by this behaviour. Is there a reason why the tempfiles are not written in `/tmp` (or `$TMPDIR`)?

Thanks!

jberdine:
The temp files are created in the same directory as the final output file because commonly the tmp directory is on another file system and so renaming the final temp file to the output fails, or else there is a spurious expensive copy to do.

Would outputting to stdout and redirecting it with `ocamlformat foo.ml -o - > /dev/null` work for your use case?

emillon:
I don't need specifically `-o /dev/null` to work, but the error message prompted me to report it, I did so :)

I think it'd be enough to be a bit more specific in the error message, like "cannot write temporary file /dev/x28398.ml" for example?

PR #425, opened 2018-09-11
Improve: explicit error message in case of 'permission denied' error

gpetiot:
A nice message for #398

### DIFF
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -277,6 +277,9 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
         exe input_name ;
       Format.eprintf
         "  BUG: formatting did not stabilize after %i iterations.\n%!" i
+  | Ocamlformat_bug (Sys_error msg)
+       when String.is_substring msg ~substring:"Permission denied" ->
+     Format.eprintf "%s: %s.\n%!" exe msg
   | Ocamlformat_bug exn when quiet_exn exn -> ()
   | Ocamlformat_bug exn -> (
       Format.eprintf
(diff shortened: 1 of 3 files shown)


src/Translation_unit.ml
```diff
@@ -272,6 +272,9 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
         exe input_file ;
       Format.eprintf
         "  BUG: formatting did not stabilize after %i iterations.\n%!" i
+  | Ocamlformat_bug (Sys_error msg)
```
hhugo: It feels wrong to use the `Ocamlformat_bug` exception if this not considered to be an ocamlformat bug.

src/ocamlformat.ml
```diff
@@ -75,7 +75,7 @@ match Conf.action with
       List.for_all results ~f:(fun result ->
           match (result : Translation_unit.result) with
           | Ok -> true
-          | Unstable _ | Ocamlformat_bug _ | Invalid_source _ -> false )
```
hhugo: Can we keep the exhaustive match ?

src/Translation_unit.ml
```diff
@@ -155,6 +156,9 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
         Location.input_name := tmp ;
         In_channel.with_file tmp ~f:(parse xunit.parse conf)
       with
+      | exception Sys_error msg
+        when String.is_substring msg ~substring:"Permission denied" ->
```
hhugo: Is "Permission denied" really a special case, I wonder.

PR #426, opened 2018-09-13
[JS] Improve: option 'single-case' = compact/sparse

gpetiot:
From the JS fork (https://github.com/janestreet/ocamlformat/commit/69b4094f9f67b9c64596e94767c7834b547d9cfe)


test/passing/js_source.ml.ref
```diff
@@ -1231,7 +1231,8 @@ let rec devariantize : type t e. e ty_env -> (t, e) ty -> variant -> t =
        | TCarg (sel, t), Some a -> ops.sum_inj (sel, devariantize e t a)
        | TCnoarg sel, None -> ops.sum_inj (sel, Noarg)
        | _ -> raise VariantMismatch
-     with Not_found -> raise VariantMismatch)
+     with
+    | Not_found -> raise VariantMismatch)
```
jberdine: misaligned?

src/Conf.ml
```diff
@@ -853,6 +854,23 @@ module Formatting = struct
       (fun conf x -> {conf with sequence_style= x})
       (fun conf -> conf.sequence_style)
 
+  let single_case =
+    let doc = "Style of single cases." in
+    let names = ["single_case"] in
+    let all =
+      [ ( "compact"
+        , `Compact
+        , "$(b,compact) will try to format single cases in a single line."
```
jberdine: s/single cases in/a single case on/

src/Conf.ml
```diff
@@ -853,6 +854,23 @@ module Formatting = struct
       (fun conf x -> {conf with sequence_style= x})
       (fun conf -> conf.sequence_style)
 
+  let single_case =
+    let doc = "Style of single cases." in
```
jberdine: s/of single cases/of pattern matching expressions with only a single case/

test/passing/js_source.ml.ref
gpetiot: I fixed this alignment but now it is aligned with the `match` instead of the `(` for some other cases when it would have been broken anyway, I prefer this new formatting but added a condition on the hvbox to have a smaller diff.

gpetiot: The last commit has been reverted.

src/Conf.ml
```diff
@@ -853,6 +854,25 @@ module Formatting = struct
       (fun conf x -> {conf with sequence_style= x})
       (fun conf -> conf.sequence_style)
 
+  let single_case =
+    let doc =
+      "Style of pattern matching expressions with only a single case."
+    in
+    let names = ["single_case"] in
```
jberdine: '_' vs '-'

PR #427, opened 2018-09-13
Fix: boxes in let bindings

gpetiot:
From JS fork (https://github.com/janestreet/ocamlformat/commit/684bd4f54e1c5a69b0a933417c359aeb7140501f)

### DIFF
--- a/test/passing/fun_decl.ml
+++ b/test/passing/fun_decl.ml
@@ -20,11 +20,11 @@ let func
   body
 
 let rec func
-       aaaaaaaaaaa
-       bbbbbbbbbbbbbb
-       ccccccccccccccccccc
-       ddddddddddddddddddddd
-       eeeeeeeeeeeeeee =
+   aaaaaaaaaaa
+   bbbbbbbbbbbbbb
+   ccccccccccccccccccc
+   ddddddddddddddddddddd
+   eeeeeeeeeeeeeee =
   body
 
 let to_loc_trace
--- a/test/passing/fun_decl.ml
+++ b/test/passing/fun_decl.ml
@@ -12,47 +12,47 @@ class t =
   end
 
 let func
-    aaaaaaaaaaa
-    bbbbbbbbbbbbbb
-    ccccccccccccccccccc
-    ddddddddddddddddddddd
-    eeeeeeeeeeeeeee =
+   aaaaaaaaaaa
+   bbbbbbbbbbbbbb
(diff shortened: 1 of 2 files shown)


test/passing/fun_decl.ml
```diff
@@ -12,47 +12,47 @@ class t =
...
-    aaaaaaaaaaa
-    bbbbbbbbbbbbbb
-    ccccccccccccccccccc
-    ddddddddddddddddddddd
-    eeeeeeeeeeeeeee =
+       aaaaaaaaaaa
+       bbbbbbbbbbbbbb
+       ccccccccccccccccccc
+       ddddddddddddddddddddd
+       eeeeeeeeeeeeeee =
```
jberdine: Is this really intended?

gpetiot:
Is this good to be merged?

test/passing/fun_decl.ml
```diff
@@ -12,47 +12,47 @@ class t =
...
-    aaaaaaaaaaa
-    bbbbbbbbbbbbbb
-    ccccccccccccccccccc
-    ddddddddddddddddddddd
-    eeeeeeeeeeeeeee =
+   aaaaaaaaaaa
+   bbbbbbbbbbbbbb
+   ccccccccccccccccccc
+   ddddddddddddddddddddd
+   eeeeeeeeeeeeeee =
```
jberdine: I don't understand these changes, but am assuming that they are intended. Is there some discussion that motivates indenting the arguments an odd number of spaces, i.e. in between two natural "tab stops", and furthermore indents the arguments less than the function name?

These changes are nicely guarded, so I don't object to merging this, but just for the record I don't get it. :-)

jberdine: So @hhugo, approve this PR is it's really what you want.

gpetiot: I don't get the indentation either but it comes directly from the JaneStreet fork so I assume they have a use for it. But maybe the indentation could be parameterized with another option like `indent-fun-args=N` ? (with N=3 for JaneStreet)

gpetiot:
Is this one good to go or do you think we need an option to parameterize the indent?

jberdine:
Also for this one, I think @hhugo is the one who needs to make the call on this one.

gpetiot:
This style does not seem to match JS style afterall, maybe just close this PR?

Issue #290, opened 2018-08-05
Consider making ocamlformat read configuration file from ocp-indent


hhugo:
note that once #272 is merge, ocamlformat will use the same syntax as ocp-indent for its configuration file.
\cc @altgr @gpetiot

gpetiot:
Closed with #445

PR #428, opened 2018-09-13
Improve: option 'cases-exp-indent' to adjust indent

gpetiot:
From JS fork (https://github.com/janestreet/ocamlformat/commit/9225436db1aa09009a4cce0f4581753b99f1b74f)

### DIFF
--- a/ocamlformat-help.txt
+++ b/ocamlformat-help.txt
@@ -57,8 +57,9 @@ OPTIONS (CODE FORMATTING STYLE)
            at the margin. The default value is force.
 
        --cases-exp-indent=COLS
-           Indentation of cases expressions (COLS columns). The default value
-           is 4. Cannot be set in attributes.
+           Indentation of cases expressions (COLS columns), except for nested
+           `match` or `try` expressions. The default value is 4. Cannot be
+           set in attributes.
 
        --disable
            Disable ocamlformat. This is used in attributes to locally disable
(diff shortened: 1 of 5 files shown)


hhugo:
Is there an ocp-indent config option that we could reuse?

hhugo:
I don't like the `ocpi_` prefix, we might as well be more explicit `-ocp-indent-match-clause`.
Also it might be a good time to look at https://github.com/ocaml-ppx/ocamlformat/issues/290

gpetiot:
Agreed, I will start looking into this one real soon.

gpetiot:
I think using a prefix allows to differentiate the ocamlformat name (and best name in this case) from the ocp-indent name. Having two different option names without a prefix would be confusing for the user I guess. Do you think it is better to only use the ocp-indent option name?

jberdine:
I'm not sure I understand the intended use cases, so take this with a 
grain of salt. But I think I would have added unprefixed ocamlformat 
options that made sense on their own, and then reading an ocp-indent 
config file would set the corresponding ocamlformat options.

gpetiot:
So options should only be set on the command line using their ocamlformat name, but reading ocp-indent options from ocp-indent configuration files should set the corresponding ocamlformat option ?

gpetiot:
I removed the ocp-indent alternative option name, ocp-indent options are handled by #445

ocamlformat-help.txt
```diff
@@ -56,6 +56,10 @@ OPTIONS (CODE FORMATTING STYLE)
            unconditionally. natural will break struct-end phrases naturally
            at the margin. The default value is force.
 
+       --cases-exp-indent=COLS
+           Indentation of cases expressions (COLS columns). The default value
```
hhugo: There is an exception to this. It is always 2 for nested match/try

PR #429, opened 2018-09-13
Improve: option 'braces-space' to have less whitespaces

gpetiot:
From JS fork (https://github.com/janestreet/ocamlformat/commit/ed1e4b4ff892b6d189f93bdf64da653fb701bded and https://github.com/janestreet/ocamlformat/commit/787d61252e91e61a2059ea3b2036f81f8c1dd7a2) : 1st part, other changes will be in other PRs.

Replaces this PR: https://github.com/ocaml-ppx/ocamlformat/pull/397


jberdine:
In general when adding formatting options, I'd prefer if the option was added with the default set to the current behavior, and then the default can be changed separately. That approach seems clearer to me, and easier for testing and migration. Are there drawbacks I'm overlooking?

gpetiot:
This is what I did, the default is `Loose`, it is `Tight` only for the JS profile (I supposed this is expected since this commit originates from the JS fork), and thus the only modified test is js_source.ml.

jberdine:
I'm sorry, you're right, my confusion, not sure where I was looking to make that conclusion.

gpetiot:
Is this good do go?

src/Conf.ml
```diff
@@ -379,6 +380,19 @@ module Formatting = struct
     C.choice ~names ~all ~doc ~section (fun conf x ->
         {conf with break_cases= x} )
 
+  let braces_space =
+    let doc = "Whether or not to use a space between insides braces." in
```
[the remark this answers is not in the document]
gpetiot: It was an unfortunate copy-paste.

src/Fmt_ast.ml
```diff
@@ -764,8 +764,12 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
         $ fmt ".@ "
         $ fmt_core_type c ~box:false (sub_typ ~ctx t) )
   | Ptyp_tuple typs ->
+      let wrap_if =
+        if Poly.(c.conf.braces_space = `Loose) then wrap_fits_breaks_if
+        else wrap_if_str
+      in
```
jberdine: There is an unfortunate amount of duplication at the moment. Since the objective of this PR seems to be essentially to defeat `wrap_fits_breaks`, it seems to me that it would be clearer and easier to adjust it directly instead. What about passing `c` to `wrap_fits_breaks` instead? I think this boils down to whether there are cases there even with this added option set, you still want the current behavior of `wrap_fits_breaks`? I don't conceptually understand what is different about the cases that add inner spaces with this PR compared to those that have been changed.

src/Conf.ml
```diff
@@ -12,7 +12,8 @@
 (** Configuration options *)
 
 type t =
-  { break_cases: [`Fit | `Nested | `All]
+  { braces_space: [`Tight | `Loose]
```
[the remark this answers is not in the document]
gpetiot: I agree this should be renamed to something like `parens-space`, but I don't see how `indicate-multiline-delimiters` would fit here, all we do is removing some spaces when the option is set, we do not remove the parens.

src/Fmt_ast.ml
gpetiot: I factorized the code, but restricted the option for parens, because the formatting is not looking good when there are semicolons involved for example.

src/Fmt.mli
```diff
@@ -117,12 +117,18 @@ val wrap : s -> s -> t -> t
 (** [wrap prologue epilogue body] formats [prologue] then [body] then
     [epilogue]. *)
 
+val wrap_str : string -> string -> t -> t
+(** As [wrap], but prologue and epilogue are strings. *)
+
```
jberdine: No longer needed.

src/Fmt.mli
```diff
@@ -117,12 +117,18 @@ val wrap : s -> s -> t -> t
...
 val wrap_k : t -> t -> t -> t
 (** As [wrap], but prologue and epilogue may be arbitrary format thunks. *)
 
 val wrap_if : bool -> s -> s -> t -> t
 (** As [wrap], but prologue and epilogue are only formatted conditionally. *)
 
+val wrap_if_str : bool -> string -> string -> t -> t
+(** As [wrap_if], but prologue and epilogue are strings. *)
+
```
jberdine: No longer needed.

src/Fmt.ml
```diff
@@ -127,8 +127,12 @@ let wrap_k x = wrap_if_k true x
 
 let wrap_if cnd pre suf = wrap_if_k cnd (fmt pre) (fmt suf)
 
+and wrap_if_str cnd pre suf = wrap_if_k cnd (str pre) (str suf)
+
 and wrap pre suf = wrap_k (fmt pre) (fmt suf)
 
+and wrap_str pre suf = wrap_k (str pre) (str suf)
+
```
jberdine: No longer needed.

src/Fmt.ml
```diff
@@ -127,8 +127,12 @@ let wrap_k x = wrap_if_k true x
 
 let wrap_if cnd pre suf = wrap_if_k cnd (fmt pre) (fmt suf)
 
+and wrap_if_str cnd pre suf = wrap_if_k cnd (str pre) (str suf)
+
```
jberdine: No longer needed externally, and only one use below: better to remove and inline.

test/passing/js_source.ml.ref
```diff
@@ -2058,26 +2058,29 @@ let rec tc : type n e. n nat -> e ctx -> term -> e checked =
...
     | Cerror _ as e -> e
-    | Cok (body', et) -> Cok (Abs (n, body'), Ar (t, et)) )
+    | Cok (body', et) -> Cok (Abs (n, body'), Ar (t, et)))
   | C m -> Cok (Const m, I)
 ;;
 
 let ctx0 =
   Ccons
-    (Zero, "0", I, Ccons (Suc, "S", Ar (I, I), Ccons (Add, "+", Ar (I, Ar (I, I)), Cnil)))
+    (Zero
```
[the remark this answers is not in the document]
gpetiot: Fixed in #453

PR #430, opened 2018-09-13
Improve: option 'parens-ite'

gpetiot:
From JS fork (https://github.com/janestreet/ocamlformat/commit/0f0289d6fc36177655fb7d21c41c7d55bc700364)
PR open for discussion.
For the moment the option is not used, to read options from Ast.parenze_exp would require to rewrite a big part of the code, is that really what we want?
We could also adopt this new behavior as default.

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -14,15 +14,17 @@
 open Migrate_ast
 open Parsetree
 
-let init, register_reset, leading_nested_match_parens =
+let init, register_reset, leading_nested_match_parens, parens_ite =
   let l = ref [] in
   let leading_nested_match_parens = ref false in
+  let parens_ite = ref false in
   let register f = l := f :: !l in
   let init conf =
     leading_nested_match_parens := conf.Conf.leading_nested_match_parens ;
+    parens_ite := conf.Conf.parens_ite ;
     List.iter !l ~f:(fun f -> f ())
   in
-  (init, register, leading_nested_match_parens)
+  (init, register, leading_nested_match_parens, parens_ite)
 
 (** Predicates recognizing special symbol identifiers. *)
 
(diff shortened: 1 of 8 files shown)


hhugo:
Look at how we do for leading_nested_match_parens

gpetiot:
Thank you, it seems alright now.

src/Ast.ml
```diff
@@ -14,15 +14,19 @@
...
 open Parsetree
 
-let init, register_reset, leading_nested_match_parens =
+let init, register_reset, leading_nested_match_parens, parens_ite =
   let l = ref [] in
-  let leading_nested_match_parens = ref false in
+  let leading_nested_match_parens =
+    ref Conf.(default_profile.leading_nested_match_parens)
+  in
+  let parens_ite = ref Conf.(default_profile.parens_ite) in
```
hhugo: Why do we rely on the default_profile here ? I would have expected the init value of the reference to be meaningless

gpetiot: You are right, I will revert this

src/Ast.ml
```diff
@@ -1920,6 +1922,13 @@ end = struct
             else exposed_right_exp Non_apply exp
         | _ -> exposed_right_exp Non_apply exp )
     in
+    let rec ifthenelse pexp_desc =
+      match pexp_desc with
+      | Pexp_extension (_, PStr [{pstr_desc= Pstr_eval (e, _)}]) ->
+          ifthenelse e.pexp_desc
```
hhugo: this should be probably use `Source.extension_using_sugar`

src/Conf.ml
```diff
@@ -656,6 +657,12 @@ module Formatting = struct
     C.flag ~default:false ~names ~doc ~section (fun conf x ->
         {conf with ocp_indent_compat= x} )
 
+  let parens_ite =
+    let doc = "Parens if-then-else." in
```
hhugo: The documentation should be more specific.

src/Ast.ml
gpetiot: I don't know how this function is supposed to interact with this match, do we only have to consider cases when `extension_using_sugar` stands? or only the opposite? Am I missing something?

gpetiot:
Is this good to go now?

jberdine:
I think @hhugo is the one who needs to make the call on this one.

PR #431, opened 2018-09-13
Improve: parens around any constrained

gpetiot:
From JS fork (https://github.com/janestreet/ocamlformat/commit/cde92c06d2a33c534a7b69d07704355f818ed54c)


jberdine:
Is there any discussion around the motivation behind this change / style? I'm not managing to think of a case where clarity is improved.

test/passing/let_binding.ml
```diff
@@ -1,34 +1,34 @@
...
+let _ : int = x3
 
 let x : int = x4
 
 let _ =
   let (x : int) = x in
   let x : int = x in
-  let (_ : int) = x in
-  let (_ : int) = x in
+  let _ : int = x in
```
jberdine: I think tests like this should be added to a file where the added option is set.

Also, it is cases like this that make me think that this option is confusing, as it seems to format in a way that obscures the underlying parsetree. It does not actually, but only because `_` is not technically an identifier, although just about every programmer thinks of it as an identifier.

gpetiot:
No bug with `test_branch.sh` and the diff looks good.

test/passing/let_binding.ml
jberdine: There should also be tests involving extension points:
```
  let (x : int) = x in
  let x : int = x in
  let (_ : int) = x in
  let _ : int = x in
  let%ext (x : int) = x in
  let%ext x : int = x in
  let%ext (_ : int) = x in
  let%ext _ : int = x in
```

jberdine:
It would be good to include some version of mbarbin's explanation:
```
(* The following does not type check:

   {[
     let%bind _ : unit = return ()
   ]}

   because it is a syntactic suggar for:

   {[
     let%bind _ = ( return () : unit )
   ]}

   Thus, we decided to adopt a consistent approach, and have the paren even when there is
   no extension point and are not strictly necessary.
*)
let (_ : unit) = ()

let f () =
  let%bind (_ : unit) = return () in
  return ()
;;
```
in the docstring.

Also, even without this option set, it seems that there is currently a bug where formatting `let%ext (_ : t) = ` will incorrectly drop the parens.

jberdine:
I'm sorry for the back-and-forth on this, but I have been looking at the parsetrees generated for the test cases with an identifier vs with an any pattern, and with and without parens, and have come away thinking that @hhugo was right originally and we should parenthesize `let _ : t =`, without an option. The parsetree for `let _ : t =` is the same as for `let (_ : t) =` and very similar to `let (x : t) =` but that for `let x : t =` is significantly different as it duplicates the constraint on the body. That duplication affects type inference in some cases. From this perspective, it is more accurate to format the any case with parens even if not strictly necessary.

So if could we take only the first commit of this PR, and add the test cases in the comment above, it'd be good to go.

hhugo:
Yes. Sorry for not coming back to this thread. You explanation sounds very familiar. :)

test/passing/let_binding.ml
```diff
@@ -110,3 +110,14 @@ let _ =
...
+
+let _ =
+  let (x : int) = x in
+  let x : int = x in
+  let (_ : int) = x in
+  let (_ : int) = x in
+  let%ext (x : int) = x in
+  let%ext x : int = x in
+  let%ext (_ : int) = x in
+  let%ext (_ : int) = x in
```
jberdine: I guess we need a .ref file for some of these to be meaningful.

gpetiot: Actually those tests can be removed since they were already present above in the file, and some cases are also tested in extensions.ml.

test/passing/let_binding.ml
```diff
@@ -1,34 +1,34 @@
...
 
 let x : int = x4
 
 let _ =
   let (x : int) = x in
   let x : int = x in
-  let _ : int = x in
-  let _ : int = x in
+  let (_ : int) = x in
+  let (_ : int) = x in
```
jberdine:
```suggestion
  let _ : int = x in
```

PR #441, opened 2018-09-17
Fix: parens around expressions with attributes

### LINKED ISSUE #439: Expressions with attributes are sometimes not parenthesized correctly
echo "let () = f (object end [@imp Show])" | _build/dev/src/ocamlformat.exe --name a.ml -

gpetiot:
Should fix #439

### DIFF
--- a/test/passing/js_source.ml.ref
+++ b/test/passing/js_source.ml.ref
@@ -103,7 +103,7 @@ let () =
      done) [@foo]];
   [%foo (assert true [@foo])];
   [%foo (lazy x) [@foo]];
-  [%foo (object end) [@foo]];
+  [%foo (object end [@foo])];
   [%foo (3 [@foo])];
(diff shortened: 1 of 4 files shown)


hhugo:
#439 is not specific to objects, we have the same issue with many other expression I believe.

test/passing/js_source.ml.ref
```diff
@@ -103,7 +103,7 @@ let () =
       done ) [@foo]];
   [%foo (assert true [@foo])];
   [%foo (lazy x) [@foo]];
-  [%foo (object end) [@foo]];
+  [%foo (object end [@foo])];
```
jberdine: Question: Am I correct that the output before still parses correctly, or is the sanity checking not tight enough?

gpetiot:
I think I got them all this time.

PR #443, opened 2018-09-20
Fix break struct natural

### LINKED ISSUE #376: Issue with "break-struct=natural"
Using `break-struct=natural`, the following code:

```ocaml
module Hash = struct
  include Hash
  module type S = S.HASH
end
```

is reformated as follows:

```ocaml
module Hash = struct include Hash

                     module type S = S.HASH end
```

which looks a bit strange to me.

gpetiot:
Should fix #376

### DIFF
--- a/src/import/Import.ml
+++ b/src/import/Import.ml
@@ -18,8 +18,8 @@ include (
 
       include
         module type of Base
-        (* [Filename], [Format], [Scanf] are all deprecated in [Base], erase
-           them and use the ones from the stdlib. *)
+          (* [Filename], [Format], [Scanf] are all deprecated in [Base],
+             erase them and use the ones from the stdlib. *)
         with module Filename := Base.Filename
          and module Format := Base.Format
          and module Scanf := Base.Scanf
--- a/test/passing/break_struct.ml.ref
+++ b/test/passing/break_struct.ml.ref
@@ -2,10 +2,19 @@
 
 module M = X (Y) (struct let x = k end)
 
-module Hash = struct include Hash module type S = S.HASH end
+module Hash = struct
+  include Hash
+
+  module type S = S.HASH
+end
 
 module Hash = struct
-  include Hash include Hash module type S = S.HASH module type S = S.HASH
+  include Hash
+  include Hash
+
+  module type S = S.HASH
+
+  module type S = S.HASH
 end
 
 module Hash = struct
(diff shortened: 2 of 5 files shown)


gpetiot:
@samoht is this the output you expected?

test/passing/break_struct.ml.ref
```diff
@@ -1,3 +1,35 @@
 [@@@ocamlformat "break-struct=natural"]
 
 module M = X (Y) (struct let x = k end)
+
+module Hash = struct include Hash module type S = S.HASH end
+
+module Hash = struct
+  include Hash include Hash module type S = S.HASH module type S = S.HASH
```
samoht: why does it not break here?

test/passing/break_struct.ml.ref
```diff
@@ -1,3 +1,35 @@
...
+end
+
+module Hash = struct
+  include Hash
+  include Hash
+  include Hash
+
+  module type S = S.HASH
+
```
samoht: why does it adds a newline between the `module type S` sentences?

jberdine:
Probably related to @samoht's nits, below is a case that regresses with this PR:
```
module Vector = struct
  include Vector

  let pp sep pp_elt fs v = List.pp sep pp_elt fs (to_list v)
end
```
should not format to
```
module Vector = struct
  include Vector let pp sep pp_elt fs v = List.pp sep pp_elt fs (to_list v)
end
```

test/passing/break_struct.ml.ref
gpetiot: This is due to the grouping (similar simple items are grouped together) that is done for structure items. Otherwise, everything else seems to be formatted alright

gpetiot:
The new output should be better, what do you all think?

src/import/Import.ml
```diff
@@ -18,8 +18,8 @@ include (
 
       include
         module type of Base
-        (* [Filename], [Format], [Scanf] are all deprecated in [Base], erase
-           them and use the ones from the stdlib. *)
+          (* [Filename], [Format], [Scanf] are all deprecated in [Base],
+             erase them and use the ones from the stdlib. *)
```
gpetiot: The new indentation looks more natural to me with the new boxing

jberdine: I'm not sure I agree, but it's not a big deal, and the rest of the effect of this patch is a significant improvement.

PR #444, opened 2018-09-20
Fix: outdent before arrow

### LINKED ISSUE #370: Bad outdent before -> in corner case
Sometimes there is a bad interaction between nested match expressions and breaking cases before the `->`:
```ocaml
let _ =
  let f x y =
    match x with
    | None -> false
    | Some loooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
  -> ( match y with Some _ -> true | None -> false )
  in
  ()
```

gpetiot:
Should fix #370

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2668,7 +2668,7 @@ and fmt_cases c ctx cs =
             (if first then if_newline "| " else fmt "| ")
         in
         hovbox 4
-          ( open_hovbox 2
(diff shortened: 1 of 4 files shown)


jberdine:
Thanks for looking at this!

This change does not resolve the issue though. If the example in #370 is changed by adding a few more `o`s, then we again have a bad outdent even with this change.

gpetiot:
I am not really satisfied with how the different `break-cases` cases are handled given the number of issues about the formatting of cases. Should we redesign the whole function? (maybe not in this PR)

PR #445, opened 2018-09-20
Read ocp indent config files

gpetiot:
Shoud fix issue #290

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1174,7 +1174,10 @@ let parse_line config ~verbose ~from s =
     | ["apprentice"], _ ->
         update_many ~config ~from ocp_indent_apprentice_profile
     | ["JaneStreet"], _ ->
-        update_many ~config ~from ocp_indent_janestreet_profile
+        Result.( >>= )
+          (update ~config ~from ~name:"profile" ~value:"janestreet")
+          (fun config ->
+            update_many ~config ~from ocp_indent_janestreet_profile )
     | [name], _ -> update ~config ~from ~name ~value:"true"
     | _ -> Error (`Malformed s) )
   | _ -> Error (`Malformed s)
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -875,7 +875,7 @@ let ocp_indent_options =
   ; ( "match_clause"
     , Some
         ( "cases-exp-indent"
-        , "$(b,match-clause) is equivalent to $(b,cases-exp-indent)."
+        , "$(b,match-clause) sets $(b,cases-exp-indent) to the same value."
         , Fn.id ) )
   ; ("ppx_stritem_ext", None)
   ; ("max_indent", None)
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1138,11 +1138,7 @@ let parse_line config ~verbose ~from s =
       List.Assoc.find_exn ocp_indent_options ~equal:String.equal name
     in
     match opt with
-    | None ->
-        if not config.quiet then
-          Format.eprintf "Warning: ocp-indent option %S is unsupported.\n"
-            name ;
-        Ok config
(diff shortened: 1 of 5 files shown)


test/passing/ocp_indent_options.ml.ref
```diff
@@ -0,0 +1,32 @@
+Warning: ocp-indent option "base" is unsupported.
```
hhugo: Do we really want to print a warning here ? I think having 24 lines of warning for every single file is way too noisy

src/Conf.ml
```diff
@@ -1004,6 +1089,11 @@ let parse_line config ~verbose ~from s =
         update ~config ~from ~name ~value
     (* special case for disable/enable *)
     | ["enable"], _ -> update ~config ~from ~name:"disable" ~value:"false"
+    | ["normal"], _ -> update_many ~config ~from ocp_indent_normal_profile
+    | ["apprentice"], _ ->
+        update_many ~config ~from ocp_indent_apprentice_profile
+    | ["JaneStreet"], _ ->
```
hhugo: Should that enable `profile=janestreet` as well ?

test/passing/ocp_indent_options.ml.ref
gpetiot: I think it is important to warn the user when an option is not supported (Is silently ignoring an option really what we want ?), and the messages are not printed when `quiet` is set.

hhugo: Quiet is not the default though.
In the current situation, someone using --ocp-indent-config can either use the quiet mode and never see useful information or always have warnings related to ocp-indent (we're unlikely to support all ocp-indent options anytime soon). I think it's fine to warn if the option is not recognized. I think it's also fine to say in the documentation that some option have no effect.

src/Conf.ml
```diff
@@ -831,6 +831,41 @@ let output =
...
+      in
+      if List.is_empty l then ""
+      else
+        asprintf " Options %a are unsupported."
+          (pp_print_list
+             ~pp_sep:(fun fs () -> fprintf fs ",@ ")
+             (fun fs s -> fprintf fs "$(b,%s)" s))
+          l
+    in
+    asprintf "Read .ocp-indent configuration files.%s" unsupported
```
jberdine: My feeling is that it would be more helpful to document the association between ocp-indent options and ocamlformat options, instead of listing the unsupported ocp-indent options.

gpetiot:
Is this one good to go?

src/Conf.ml
```diff
@@ -842,6 +842,46 @@ let output =
...
+  ; ("align_ops", None)
+  ; ("align_params", None) ]
+
+let ocp_indent_config =
+  let doc =
+    let open Format in
+    let supported =
+      let l =
+        List.filter_map ocp_indent_options ~f:(fun (_, o) ->
+            Option.value_map o ~default:None ~f:(fun (_, doc, _) -> Some doc)
```
jberdine: nit: Could be slightly simpler using `Option.map`

src/Conf.ml
```diff
@@ -842,6 +842,46 @@ let output =
...
 
+let ocp_indent_options =
+  [ ("base", None)
+  ; ("type", None)
+  ; ("in", None)
+  ; ("with", None)
+  ; ( "match_clause"
+    , Some
+        ( "cases-exp-indent"
+        , "$(b,match-clause) is equivalent to $(b,cases-exp-indent)."
```
jberdine: Later, I expect that we will get into the situation where an ocp-indent option `foo` is used to set the ocamlformat option `bar`, but they are not exactly "equivalent". So I wonder whether we should phrase this as `$(b,match-clause) sets $(b,cases-exp-indent) to the same value.` instead?

PR #448, opened 2018-09-23
Don't execute shell scripts directly in build

dra27:
Windows Dune doesn't (yet) understand `#!` scripts. It may well do around the same as opam 2.1 gains the same ability... in the meantime, it's now called via sh (the `%{dep:...}` addition is necessary since Dune only adds dependencies on the first argument of `(run`)

While I'm not averse to CLAs - this is a tiny change, if it really requires a CLA, would someone who's already signed one be so kind as to remark this commit as their own?

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -9,7 +9,7 @@ let preprocess =
 (rule
  (targets Version.ml)
  (deps (universe))
- (action (run ../tools/gen_version.sh Version.ml))
+ (action (run sh %%{dep:../tools/gen_version.sh} Version.ml))
  (mode fallback))
 
 (ocamllex (modules Literal_lexer))


gpetiot:
I marked this commit as my own (and manually added the CLA Signed tag)

gpetiot:
I did not find much info on this.
As this concerns a `dune` file, @emillon can I have your opinion on this?

emillon:
The approach works (it will call `sh` manually and register a dependency to the shell script). I'm not sure about the double percent though - `(run)` will not use `system`, so I don't think it's necessary to quote it on windows. It will probably add a literal `%`.

dra27:
@emillon - it's an ocaml-mode file, the escaping is for `printf`'s benefit

emillon:
All good then! Maybe a way to improve this on dune's side is to add a `(sh ...)` action like we have for `(bash)`.

dra27:
Actually, the better option will be for me to carve out https://github.com/ocaml/opam/pull/3348 to a library, and have Dune vendor it!

emillon:
Yes, that would work as well. We have a similar problem with `make` which has a different name depending on systems (and opam has a builtin solution for this as well).

dra27:
True - that's mostly an orthogonal issue, unless you have a Cygwin-symlink `gmake` or something annoying like that!

gpetiot:
By the way @emillon  is this correct to make this change in ocamlformat instead of doing it in dune?
(cc @jberdine )

dra27:
@gpetiot - it's not correct to assume that shell scripts can be executed directly on Windows. If Dune gets that then, like opam, it would be a largely unique feature (I'm not aware of any other native Windows program which tries to do this and, given the amount of code it takes to do it, it's not too surprising!). It would remain correct for ocamlformat's build system to invoke `sh` directly, even if Dune gained the ability to call the script.

Of course, converting the script to OCaml is ultimately the correct move!

(I'm a Dune dev too, btw, if rather less active!)

emillon:
I agree with @dra27. It might be end up as a nice helper in dune, but the "workaround" is the portable way to do it.

gpetiot:
Alright, thank you for your help!

jberdine:
I don't intend to be grumpy, but I'm really not convinced. Looking at this code change, it looks just plain worse. Am I missing something? I mean, to understand that the version proposed is not just a roundabout way to write the existing version, the reader has to understand that windows can't handle seemingly innocuous invocations of scripts in dune files, and also some fiddly business about what dune considers to be a dependency. The solution to the windows problem cannot be to complicate everything.

dra27:
@jberdine - it’s either `%{dep:...}` or including it in the deps field, but that’s code duplication with the usual entailed brittleness.

I’m not particularly clear why `sh script` is so much more complicated than `./script` - it’s a quite common idiom to avoid issues where the x attribute gets lost in a distribution archive too.

dra27:
It is also worth noting that the manual for run states “to execute a _program_”!

dra27:
It would need to be `run ocaml %%{dep:gen_version.ml}`, but otherwise yes. I could more readily get on board with Dune automatically knowing on Windows that `./foo.ml` means to run `ocaml foo.ml`, though.

Note that you can also avoid this by compiling gen_version at that point (more lines of dune, but fewer “tricks”)

jberdine:
Thanks for the info.

Re compiling gen_version, I agree, but the catch is that it needs to be run from dune (when building distributed archives), opam (when pinning), and make (when building during development). I don't know if it is feasible to compile it in particular when opam pinning.

PR #450, opened 2018-09-28
Ocamlformat diff tool

gpetiot:


jberdine:
Is it necessary to create a directory and run ocamlformat on both files from there, or could the `--name` option be given to specify where to get the config from instead?

gpetiot:
> Is it necessary to create a directory and run ocamlformat on both files from there, or could the `--name` option be given to specify where to get the config from instead?

You are right, I simplified the code accordingly. I also pushed a "tmp" commit for the metadata, it is not finished yet (the branch is [WIP] afterall), don't mind the output.
It would be nice if we could only apply this on .ml files and not doing the check in this tool. I looked up `.gitconfig` and `.gitattributes` but I didn't find anything that could be of use.

jberdine:
Have you tried configuring the file extensions along the lines of https://stackoverflow.com/questions/7403102/git-configure-patterns-for-difftool-and-mergetool ?

Another point in favor of doing the check externally is that sometimes ocamlformat is used on other files, such as dune files in ocaml syntax.

gpetiot:
Those stackoverflow threads are very outdated, it rarely works. The only thing that works is to write this in `~/.gitconfig`:
```
[diff]
    external = <path_to_wrapper_script>
```
But we need a wrapper for the diff tool that chooses which tool to call depending on the file extension.

jberdine:
Ugh, ok I see. For the sake of thinking what the best approach might be, it might be helpful to consider what a user with multiple auto-formatters would need to do. It seems that either all `fmt-diff` tools would need to select their own file types, and ignore the rest, where all tools are run in all cases, or there is a wrapper that dispatches on the extension. It seems that either way there is a wrapper involved, so maybe best to just go with that case. I wonder if there is a commonly used existing one somewhere already.

gpetiot:
It seems the git shenanigans work afterall but we have to use `.git/config` (not versioned?) rather than `.gitconfig` (but `.gitattributes` is fine and can be used instead of `.git/info/attributes`.

`.gitattributes` must contain:
```
*.ml diff=fmt_diff
*.mli diff=fmt_diff
*.mlt diff=fmt_diff
```

`.git/config` must contain:
```
[diff "fmt_diff"]
    command = fmt_diff
```

tools/fmt_diff/fmt_diff.ml
```diff
@@ -0,0 +1,118 @@
...
+  | Error e -> Error e
+
+(** [diff_in_ctx ~ctx ~old_file ~new_file] runs the diff of files [new_file]
+    and [new_file] in the context of file [ctx], so that they both inherit
+    the same configuration. *)
+let diff_in_ctx ~ctx ~old_file ~new_file =
+  Bos.OS.Cmd.resolve (Bos.Cmd.v "ocamlformat") >>= fun fmt ->
+  run_ocamlformat fmt ~ctx ~file_in:old_file >>= fun tmp_1 ->
+  run_ocamlformat fmt ~ctx ~file_in:new_file >>= fun tmp_2 ->
+  Bos.OS.Cmd.must_exist (Bos.Cmd.v "diff") >>= fun diff ->
```
jberdine: As someone who has the following in my .gitconfig:
```
[diff]
        tool = patdiff
[difftool "patdiff"]
	cmd = patdiff-git-wrapper $BASE $LOCAL 0 0 $REMOTE 0 0
```
I would be much happier if there was some way to similarly configure `fmt-diff` (other than PATH shadowing hacks).

tools/fmt_diff/dune
```diff
@@ -0,0 +1,5 @@
+(executable
+  (name        fmt_diff)
```
jberdine: I think `name` is not needed when it's the same as `public_name`

tools/fmt_diff/fmt_diff.ml
```diff
@@ -0,0 +1,118 @@
+open Rresult
+
+(** [run_ocamlformat cmd ~ctx ~file_in] runs the ocamlformat command [cmd]
+    on the input file [file_in] with the configuration context of file [ctx]
+    and returns the path to the output in file. *)
+let run_ocamlformat cmd ~ctx ~file_in =
+  Bos.OS.File.tmp "%s" >>= fun path_out ->
+  Bos.Cmd.(cmd % Fpath.to_string file_in
+           % "--name" % Fpath.to_string ctx % "--disable-outside-project"
```
jberdine: old option name

tools/fmt_diff/fmt_diff.ml
```diff
@@ -0,0 +1,118 @@
...
+           % "-o" % Fpath.to_string path_out)
+  |> Bos.OS.Cmd.run_status
+  |> function
+  | Ok (`Exited 0) -> Ok path_out
+  | Ok _ ->
+      R.error_msg ("ocamlformat on " ^ Fpath.to_string file_in ^ " failed!")
+  | Error e -> Error e
+
+(** [diff_in_ctx ~ctx ~old_file ~new_file] runs the diff of files [new_file]
+    and [new_file] in the context of file [ctx], so that they both inherit
```
jberdine: typo: new_file twice

tools/fmt_diff/fmt_diff.ml
```diff
@@ -0,0 +1,118 @@
...
+                   (short_hash old_hex) (short_hash new_hex) old_mode );
+  print_endline (Format.sprintf "--- a/%s" path);
+  print_endline (Format.sprintf "+++ b/%s" path)
+
+let diff path old_file old_hex old_mode new_file new_hex new_mode =
+  let diff =
+    Bos.OS.File.must_exist (Fpath.v old_file) >>= fun old_file ->
+    Bos.OS.File.must_exist (Fpath.v new_file) >>= fun new_file ->
+    let ctx =
+      if Fpath.(is_prefix (v "/tmp/") new_file) then old_file
```
jberdine: Better to use `Filename.get_temp_dir_name` or similar.

tools/fmt_diff/fmt_diff.ml
```diff
@@ -0,0 +1,118 @@
...
+        ; `Ok () )
+  | _ -> `Ok ()
+
+open Cmdliner
+
+let path =
+  let doc = "" in
+  Arg.(required
+       & pos ~rev:true 6 (some string) None
+       & info [] ~docv:"path" ~doc)
```
jberdine: Is it intentional that docv gets passed literal strings and doc uses punning?

tools/fmt_diff/dune
```diff
@@ -0,0 +1,5 @@
...
+  (public_name fmt_diff)
```
jberdine: My impression is that `fmt-diff` would be more idiomatic than `fmt_diff`.

Also, is it self-inconsistent to have `ocamlformat` and `fmt-diff` executables, rather than `ocamlformat-diff`?

tools/fmt_diff/fmt_diff.ml
gpetiot: I would like to avoid ocamlformat-diff specific configuration files if possible, how about an environment variable $ocamlformat_diff_default=patdiff ?

tools/fmt_diff/dune
gpetiot: I have to upgrade `(lang dune <N>)` from 1.0 to 1.1 for that.

tools/fmt_diff/dune
gpetiot: It doesn't seem to be possible to name it `ocamlformat-diff`, the name of the executable has to be the same as one of the module and it is an invalid module name (it's only a warning but still), we could still copy it manually in the `.opam/<VERSION>/bin` but it is not as straightforward as using `dune install` and it overcomplicates the Makefile.

gpetiot: Those exact lines give me:
```
dune build _build/dev/tools/ocamlformat-diff/ocamlformat-diff.exe _build/release/tools/ocamlformat-diff/ocamlformat-diff.exe
Don't know how to build _build/dev/tools/ocamlformat-diff/ocamlformat-diff.exe
Hint: did you mean _build/dev/tools/ocamlformat-diff/ocamlformat_diff.exe?
Makefile:48: recipe for target 'ocamlformat-diff' failed
make: *** [ocamlformat-diff] Error 1
```
I use `dune` 1.2.1 but I didn't see interesting changes in the 1.3 changelog, am I missing something else?

tools/fmt_diff/fmt_diff.ml
gpetiot: By the way the integration with patdiff works fine with my last commit

tools/fmt_diff/dune
mseri: That sounds like a bug. It used to work as suggested by @jberdine and @avsm

mseri: Weirdly enough, `dune build _build/install/default/bin/ocamlformat-diff`seems to be working for me

gpetiot: I get the same output as I did earlier (wihtout the .exe). What version of `dune` are you all using? If it is indeed fixed in dune-1.3 I don't know if it is worth it to upgrade the dependencies for `dune` and `opam` only for the tool name, as the main use for this tool is to be called by `git diff`.

Makefile
```diff
@@ -44,14 +44,17 @@ opt: setup
 reason: setup
 	dune build _build/dev/src/ocamlformat_reason.exe _build/release/src/ocamlformat_reason.exe
 
+ocamlformat-diff:
```
jberdine: no .PHONY intentionally?

README.md
```diff
@@ -87,6 +88,10 @@ The available profiles are:
 - `sparse` sets options for a generally sparse code style
 - `janestreet` is the profile used at JaneStreet
 
+### Diff
+
+[`ocamlformat_diff`](./tools/ocamlformat-diff/README.md) uses OCamlFormat to apply the same formatting to compared OCaml files, so that the formatting differences between the two files are not displayed. It can easily be used with `git diff`.
```
jberdine: s/_/-/

tools/ocamlformat-diff/README.md
```diff
@@ -0,0 +1,42 @@
+# OCamlFormat diff tool
+
+`ocamlformat_diff` is a tool that uses OCamlFormat to apply the same formatting to compared OCaml files, so that the formatting differences between the two files are not displayed.
```
jberdine: s/_/-/ through whole file I guess

tools/ocamlformat-diff/ocamlformat_diff.ml
```diff
@@ -0,0 +1,135 @@
...
+      else (
+        diff_prelude path old_file old_hex old_mode new_file new_hex
+          new_mode ;
+        print_endline str ;
+        `Ok () )
+  | _ -> `Ok ()
+
+open Cmdliner
+
+let default_diff =
```
jberdine: maybe just `--diff` since it's unclear what 'default' is referring to

tools/ocamlformat-diff/README.md
```diff
@@ -0,0 +1,42 @@
...
+if you want to use the default `diff` command.
+
+Or:
+
+```
+[diff "ocamlformat_diff"]
+    command = ocamlformat_diff --default-diff=patdiff
+```
+if you want to use [patdiff](https://github.com/janestreet/patdiff).
+Feel free to use another diff command as you see fit.
```
jberdine: It would probably still be clear to give only the example with the extra diff option, and mention that it can be omitted to use `diff`, since it is easy to imagine how to remove the option.

PR #452, opened 2018-10-06
Fix: missing parens around partially-applied `::`

### LINKED ISSUE #451: Cannot process cons as function
```
$ ocamlformat a
ocamlformat: Cannot process "a".

$ cat a
(::) 5

$ parse_to_ast
[
  structure_item ([1,0+0]..[1,0+6])
    Pstr_eval
    expression ([1,0+0]..[1,0+6])
      Pexp_construct "::" ([1,0+0]..[1,0+4])
      Some
        expression ([1,0+5]..[1,0+6])
          Pexp_constant PConst_int (5,None)
]
```

I don't know if this is really valid syntax, I only found it cause I'm converting Elm to Ocaml automatically, otherwise I don't think I would have tried this. However, it parses in OCaml, and causes a bug in OCamlformat so I'm reporting it.

jberdine:
fix #451


jberdine:
@gpetiot, done.

PR #453, opened 2018-10-08
Improve: keep a space inside tuples parens

gpetiot:
Fix the collateral damage of #429

### DIFF
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -137,15 +137,16 @@ let wrap_if_fits_and cnd pre suf k fs =
   k fs ;
   fits_breaks_if cnd suf "" fs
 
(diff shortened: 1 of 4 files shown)


src/Fmt.ml
```diff
@@ -137,15 +137,16 @@ let wrap_if_fits_and cnd pre suf k fs =
   k fs ;
   fits_breaks_if cnd suf "" fs
 
-let wrap_fits_breaks_if c cnd pre suf k fs =
-  if (not c.Conf.indicate_multiline_delimiters) && String.equal pre "(" then
+let wrap_fits_breaks_if ?(space = false) c cnd pre suf k fs =
```
jberdine: It seems that the default ought to be true.

Issue #264, opened 2018-08-02
Style option: don't start lines with punctuation


dbuenzli:
It [seems](https://github.com/hhugo/ocaml/commit/9f10047b820103d277392d60b58e30cdcb73a5c9#diff-bfbed4ab457cd252e41f57d62147b4e3R18) that `ocamlformat` currently formats long pairs/lists by breaking before seperators which is somehow at odds with typographic conventions. It would be nice to have an option that breaks after separators (`,`, `;`, `->`, etc.)

jberdine:
Yes, there are a few common styles, and ocamlformat is pretty opinionated in this case. This is in line with the second point from the README:

- Whenever possible, the high-level structure of the code should be obvious by looking only at the left margin, in particular, it should not be necessary to visually jump from left to right hunting for critical keywords/tokens/etc.

But an option would be welcome.

dbuenzli:
I don't find the justification particularly compelling. If I'm reading say a multiline tuple definition newlines and indentation indicate the structure, I don't have to hunt to the right looking for the comma to be sure I'm getting it right.

jberdine:
I can see how the argument isn't compelling. As with most questions about syntax, there are very few arguments that are strong compared to individual experiences and preferences. Which things have burned you, and which you've seen burn others, really matters.

dbuenzli:
> As with most questions about syntax, there are very few arguments that are strong compared to individual experiences and preferences.

Of course but to be honest I believe much more in the established rules and conventions of typography regarding legibility, punctuation and spacing than what seems to have emerged as a fashion among programmers without any compelling justification for it.

What I'm trying to say here is that if you don't have any strong reason not to do so it's better to follow established typographic conventions.

gpetiot:
Closed with #461

Issue #406, opened 2018-09-08
Introduce intermediate data structure between parsetree and Format stream


jberdine:
The current implementation traverses the parsetree and directly produces the `Format` stream. This works well when the desired nesting structure of formatting boxes coincides with the nesting structure of the parsetree. When the desired formatting involves different nesting of boxes than the parsetree, the code gets very complicated and non-compositional.

An alternative is to introduce an intermediate data structure which is constructed by traversing the parsetree, and then it is traversed to produce the `Format` stream. This would enable much simpler and more flexible differences between nesting of boxes relative to parsetree structure.

Currently all the various "sugar" functions, as well as `Fmt_ast.block`, are partial steps in this direction already.

gpetiot:
I lowered the priority of this one, as it looks very time consuming and not critical for now.

jberdine:
Yes, I don't know when the right time to do this will be. At some point we will have spent more time (adding up small pieces) fighting with the current implementation than it would have taken to implement this. But it's hard to predict.

PR #461, opened 2018-10-18
Break separators

gpetiot:
Addressing issue #264 
- [x] semi-colons
- [x] commas
- [x] arrows


src/Conf.ml
```diff
@@ -439,6 +440,23 @@ module Formatting = struct
     C.choice ~names ~all ~doc ~section (fun conf x ->
         {conf with break_infix= x} )
 
+  let break_separators =
+    let doc =
+      "Indicates how are broken an expression with a separator when it \
+       does not fit on a single line."
```
jberdine: Perhaps reword: "Break before or after separators such as `;` in list or record expressions."

src/Conf.ml
```diff
@@ -15,6 +15,7 @@ type t =
   { break_cases: [`Fit | `Nested | `All]
   ; break_collection_expressions: [`Wrap | `Fit_or_vertical]
   ; break_infix: [`Wrap | `Fit_or_vertical]
+  ; break_separators: [`Before | `After]
```
jberdine: Your choice, but my impression from reading the code is that it would be simpler if this was a bool treated as `break_struct` is.

gpetiot: I don't know, I'm not a fan of this choice for `break_struct` to begin with, I find the variants to be clearer than the boolean

gpetiot:
> `git co $(git merge-base master break-separators) && make && (make -C test test_unstage test_clean test_pull test test_stage) && git co break-separators && make && (make -C test test test_diff)`
shows some regressions

This command does not seem to correctly test the branch.

I tested `OCAMLFORMAT=break-separators=after make fmt` and the output looks great.

src/Conf.ml
jberdine: Ok, I don't have a very strong position, but let me explain my thinking to make sure we're on the same page. The trade-off is between the more immediately clear semantics of the variants on the one hand versus the less disruptive code to branch on booleans vs variants. In cases where there are multiple points in the core formatting code that branch on an option, I find that it is easier to keep in mind how they fit into the surrounding formatting code when they are booleans. It does mean that you need to have in mind what the booleans mean, but with careful naming this seems to be not a huge issue. I agree that it is more immediately clear what the meaning is with variants, but basically I would rather optimize for the formatting code, esp. when there are multiple occurrences far apart.

jberdine:
`make fmt` only test the ocamlformat code itself, not the code in the external repos. Can you check that you have them setup correctly? The initial setup steps in the Makefile may have bitrotted.

gpetiot:
> `make fmt` only tests the ocamlformat code itself, not the code in the external repos. Can you check that you have them setup correctly? The initial setup steps in the Makefile may have bitrotted.

I cleaned everything and managed to run the command, everything is okay after my last commit (I added a box I shouldn't have to)

test/passing/break_separators.ml
```diff
@@ -0,0 +1,119 @@
...
+  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ->
+  bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb ->
+  ccccccccccccccccccccccccc
+
+[@@@ocamlformat "ocp-indent-compat"]
+
+type t =
+  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ->
+  bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb ->
+  ccccccccccccccccccccccccc
```
jberdine: Are these change to formatting of `->` in types acceptable in the JS profile? @hhugo

test/passing/break_separators.ml
```diff
@@ -0,0 +1,119 @@
+[@@@ocamlformat "break-separators=after"]
+
+type t =
+  { aaaaaaaaaaaaaaaaaaaaaaaaa: aaaa aaaaaaaaaaaaaaaaaaa;
+    bbbbbbbbbbbbbbbbbbbbbbbbb: bbbbbbbbbbbb bbbb;
+    cccccccccccccccccccccc: ccccccc ccccccccccc cccccccc }
```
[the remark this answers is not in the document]
Khady: And one more request, would it be possible to keep the `;` after the last element?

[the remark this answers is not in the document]
jberdine: @samoht that was my preferred style pre-ocamlformat, but it's involved to implement. If #406 was done, we'd be in a better position to do this sort of formatting in many instances. Note that "this sort" is referring to how the parsetree element for `{ ... }` involves negative indentation internally. More specifically, it is complicated because the format box nesting structure needed to produce the output you want does not match the nesting structure in the parsetree.

gpetiot:
> From the diff of this applied to ocamlformat's source:
> 
> ```
>  val doc_atrs :
> -     (string Location.loc * payload) list
> -  -> (string Location.loc * bool) list option
> -     * (string Location.loc * payload) list
> +  (string Location.loc * payload) list ->
> +  (string Location.loc * bool) list option
> +  * (string Location.loc * payload) list
> ```
> 
> To me this is a significant readability regression. I'd like to hear what the proponents of a separator last style think.

It should be more readable with parens now.

jberdine:
I think this needs quite a few more tests. It can affect many combinations of breaking types within records, and positions of docstrings and comments. Can you comment on the diff produced by test_branch.

gpetiot:
I looked at the diff produced by test_branch and it looks good to me.

test/passing/break_separators.ml
```diff
@@ -0,0 +1,174 @@
...
+      aaaaaaaaaaaaaaaaaa;
+      aaaaaaaaaa } =
+  { aaaaaaaaaaaa= aaaaaaaaaaaaaaaaa;
+    bbbbbbbbbbbbb= bbb bb bbbbbb;
+    cccccc= cccc ccccccccccccccccccccccc }
+
+(* this is an array *)
+let length =
+  [| 0;
+    269999999999999999999999999999999999999999999999999;
```
jberdine: Is this non-alignment intended?

test/passing/break_separators.ml
```diff
@@ -0,0 +1,174 @@
...
+      foo ->
+      foo ->
+      foo * foo ->
+      foo * foo ->
+      foo * foo ]) ->
+  (* foooooooooooooooo *)
+  fooooooooooooooooo
+
+type t = {
+  (* fooooooooooooooooo *)
```
jberdine: Is this non-alignment intended?

test/passing/break_separators.ml
```diff
@@ -0,0 +1,174 @@
...
+
+type t =
+  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ->
+  bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb ->
+  ccccccccccccccccccccccccc
+
+type t =
+  (string Location.loc * payload) list ->
+  ((string Location.loc * bool) list option
+   * (string Location.loc * payload) list) ->
```
jberdine: It seems odd to me to have `*` on the left and `->` on the right, but I guess that is intended.

gpetiot: I am not sure it would more more readable if this operator was on the right.

jberdine:
This hasn't been tested enough: `tools/test_branch.sh break-separators '' 'break-separators=after'` shows failures in test/code/ocaml.

jberdine:
Look at the output with this option enabled on Conf.mli. Is the placement of docstrings relative to `;` intended:
```
type t = {
...
  wrap_comments: bool  (** Wrap comments at margin. *);
...
}
```

test/passing/break_separators.ml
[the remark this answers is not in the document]
gpetiot: To have a consistent style (and correct with respect to comments) there is no `;` printed after the last element of a record.

gpetiot:
`break-separators = after` is now the new default (as it was asked for in the Discourse)

gpetiot:
I agree on the 3-value option.
As for the missing semi-colon at the end, I didn't manage to make a difference between the 1-liner and a multi-liner:
```
type t = {a: a; b: b}

type t = {
  fooooooooooo:  foooooooooooooooooooooooooooo;
  foooooooooooooooo: fooooooooooooooooooooooooooooo;
}
```
the ending semi-colon is wanted for the second case but not the first, and since I can't tell the difference (yet) I choose to not print the semi-colon for now.

jberdine:
Ah, I didn't mean whether or not to have an extra semi at the end, but was referring to how in the quoted diff there is a record with two fields on the same line, but the braces on the preceding and following lines.

jberdine:
In order to emit a trailing semi only in the multi-line case, have you tried something like `fits_breaks "" ";"`?

gpetiot:
> In order to emit a trailing semi only in the multi-line case, have you tried something like `fits_breaks "" ";"`?

I managed to have this, but I had to change the box structure for it, so comments are no longer on the same line when possible. I can't have a box grouping the field declaration and the comments, otherwise I can't tell when the box grouping all fields breaks.

gpetiot:
Before rewriting all the code, is this what you suggest: (do we agree on the absence of the last semicolon for `after` ?) (cc @samoht)

```ocaml
(* break-separators = before *)
type t =
  { fooooooooooo: foooooooooooooooooooooooooooo
  ; foooooooooooooooo: fooooooooooooooooooooooooooooo }

(* break-separators = after*)
type t =
  { fooooooooooo: fooooooooooooooooooooooooooo;
    foooooooooooooooo: fooooooooooooooooooooooooooooo }

(* break-separators = after-and-docked *)
type t = {
  fooooooooooo: foooooooooooooooooooooooooooo;
  foooooooooooooooo: fooooooooooooooooooooooooooooo;
}
```

jberdine:
Could you give an example of the change of comment formatting you mentioned?

To confirm, in the after-and-docked case, the rest of the formatting is like before the case, right?

I think that leaving out the final semi in the after case is fine, but it could be added if its easier. The value of the final semi is to be able to add/reorder fields easily, but there is little value in having the  semi in the presence of the closing brace.

gpetiot:
> Could you give an example of the change of comment formatting you mentioned?
> 
> To confirm, in the after-and-docked case, the rest of the formatting is like before the case, right?
> 
> I think that leaving out the final semi in the after case is fine, but it could be added if its easier. The value of the final semi is to be able to add/reorder fields easily, but there is little value in having the semi in the presence of the closing brace.

Here is an example for type `Conf.t`:
```ocaml
type t = {
   let_binding_spacing: [`Compact | `Sparse | `Double_semicolon];
   let_open: [`Preserve | `Auto | `Short | `Long];
-  margin: int;  (** Format code to fit within [margin] columns. *)
+  margin: int;
+      (** Format code to fit within [margin] columns. *)
   max_iters: int;
       (** Fail if output of formatting does not stabilize within [max_iters]
           iterations. *)
}
```

jberdine:
Hmm, that is unfortunate. The only way I'm seeing to handle it is something analogous to how one-line let bindings are detected to decide whether to put an open line between them.

gpetiot:
For now the `after | after_and_docked` difference is only for the type declarations, is it something we want for record expressions as well?

gpetiot:
> I thought that the difference between `before` and `after_if_docked` would be only about record types, and the difference between that and `after` would be the rest of the changes. Can you confirm? My impression is that we would optimally like the docked version for all cases, but it is too ad hoc and complicated to implement, so should be delayed and added piecemeal, as possible. Does that make sense to you?

That's not what I had in mind, I wanted to do this:
```ocaml
(* break-separators = before *)
type t =
  { fooooooooooo: foooooooooooooooooooooooooooo
  ; foooooooooooooooo: fooooooooooooooooooooooooooooo }

(* break-separators = after *)
type t =
  { fooooooooooo: fooooooooooooooooooooooooooo;
    foooooooooooooooo: fooooooooooooooooooooooooooooo }

(* break-separators = after-and-docked *)
type t = {
  fooooooooooo: foooooooooooooooooooooooooooo;
  foooooooooooooooo: fooooooooooooooooooooooooooooo;
}
```
Where the difference between `before` and `after` is only the position of separators (no difference for brackets). And the difference between `after` and `after-and-docked` is the position of brackets. But for now the docked position of brackets if only implemented for type declarations and not record expressions (but it could be in the future). (more examples in `break_separators.ml`)

gpetiot:
> Can I just again note that I'm concerned that with this option set to `after`, the formatting of `->` in type signatures looks like it could be trouble for (my incomplete understanding of) Jane Street's style.

Aren't people at JS using the `before` style?

jberdine:
> Aren't people at JS using the `before` style?

I think so, but my understanding is that it is only due to a lack of 
options rather than a choice.

ocamlformat-help.txt
```diff
@@ -44,6 +44,13 @@ OPTIONS (CODE FORMATTING STYLE)
            fit-or-vertical vertically breaks expressions if they do not fit
            on a single line. The default value is wrap.
 
+       --break-separators={before|after|after-and-docked}
+           Break before or after separators such as `;` in list or record
+           expressions. before breaks the expressions before the separator.
+           after breaks the expressions after the separator. after-and-docked
+           breaks the expressions after the separator and docks the brackets
```
copy: What does "docks" mean in this context?

Issue #359, opened 2018-08-27
Not breaking after bind/map operators


samoht:
Currently, `>>=` and `>|=` seems to always be put on a newline:

```ocaml
f x >>= fun y ->
g y >>= fun () ->
...
```

is reformated as

```ocaml
f x
>>= fun y ->
g y
>>= fun () ->
...
```

Would it be possible to not do that using an option?

jberdine:
Currently `>>=` etc. are formatted using the same code path as for general binary operator applications, which places the operators at the beginning of lines when it is necessary to break.

To achieve this suggested style, a special case for `bind`-like operators could be added to `fmt_expression` in `Fmt_ast`, to be used instead of [this](https://github.com/ocaml-ppx/ocamlformat/blob/990dda59a17dde1151e37ed0f862196ba909eb99/src/Fmt_ast.ml#L1441).

One consideration is how to choose which operators to use this special case for. Just `>>=` and `>>|`? Some configurable set? Any binop applied to a `fun`(`ction`) 2nd arg? The first option seems unfair to any code that uses its own monadesque or cps combinators. Maybe the last is ok. It would be nice not to need much configuration. Thoughts?

gpetiot:
Closed with #463

vphantom:
@gpetiot Is there a specific combination of configuration options to use in order to keep `expr >>= fun arg ->` groups each on one line?  I'm not getting that behavior with ocamlformat v0.9 `--profile=janestreet -m 80`:

```ocaml
let f_test_noppx () =
  somefunction 10000000 20000000
  >>= fun a -> somefunction 20000000 a >>= fun b -> Ok (b + 42)
;;
```

I'm trying not to use `let%bind` in the name of simplicity in a project which strictly uses ocamlformat, but the above isn't very intuitive compared to the widely used structure:

```ocaml
let f_test_noppx () =
  somefunction 10000000 20000000 >>= fun a ->
  somefunction 20000000 a >>= fun b ->
  Ok (b + 42)
;;
```

gpetiot:
You should try `--no-break-infix-before-func`.

vphantom:
Thanks for the hint!  I get the following, which is very readable:

```ocaml
let f_test_noppx () =
  somefunction 10000000 20000000 >>= fun a ->
  somefunction 20000000 a >>= fun b -> Ok (b + 42)
;;
```

[...]

Issue #462, opened 2018-10-19
do not add linebreak before an anonymous function


samoht:
I am quite fond of:

List.map (fun x ->
    xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx
  ) list
```

but ocamlformat keeps breaking that nice pattern into:

```
List.map
  (fun x ->
    xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx )
  list
```

 :-) Would that be easy to support in the `compact` profile?

gpetiot:
This kind of formatting is difficult to achieve because it relies on formatting boxes that do not respect the shape of the ast, it requires special cases, and it will be difficult to maintain I guess. Same problem for #207.

Issue #207, opened 2018-06-14
Linebreaks in function formatting


jfeser:
I have some code which looks like this:
```
let myfunc : int -> int = fun x ->
  let y = x + 1 in
  y
```
It gets reformatted to:
```
let myfunc : int -> int =
 fun x ->
  let y = x + 1 in
  y
```
It'd be nice to avoid the unnecessary line break after the equals. I can take a shot at fixing this but I'd need some pointers on where to look in the code.

jberdine:
This code is formatted by `Fmt_ast.fmt_value_binding` which itself produces the code from `let` to `=`, and then uses `fmt_expression` to produce from `fun` onward. To achieve what you suggest, I think the easiest approach is to add a special case to `fmt_value_binding` which looks at the body of such bindings with type constraints and splits the `Pexp_fun` into the function arguments and body, analogously to how it is currently done for the case where there are no type constraints.

jberdine:
Easier after #406

enaudon:
I too have a bunch of code that looks like @jfeser's example, and would like this feature.  Out of curiosity, why are issues like this and #364 problematic, while `--break-separators=after-and-docked` is not?

gpetiot:
I'm not sure the fix for this one would be radically different but I don't have the time to look at this one right now, I'm working on the bugs reported lately these days.

PR #463, opened 2018-10-19
Improve: not breaking after bind/map operators

gpetiot:
Should fix issue #359


gpetiot:
I don't know if it is better to use the same option for this one and #207 and #462 (could be `--break-before-fun`)

gpetiot:
@samoht does this fix your issue?

samoht:
Looks great!

src/Conf.ml
```diff
@@ -439,6 +440,15 @@ module Formatting = struct
     C.choice ~names ~all ~doc ~section (fun conf x ->
         {conf with break_infix= x} )
 
+  let break_infix_bind =
+    let doc =
+      "Force infix bind expressions to break before the operator if it \
+       does not fit on a single line."
```
jberdine: I'm a little uncomfortable with the name and description of this option, which only mentions 'bind', while the implementation applies to any infix op with a fun(ction) expression as second argument. Either specialize the implementation or generalize the description.

gpetiot:
The issue on Ast.ml is nox fixed.

jberdine:
Could you add some test cases which check the formatting in cases which break across lines. For example in `e |> fun x -> f`, the current behavior seems odd in case `e` breaks across lines and in case a break is needed between `|>` and `fun`. It seems to me that if `e |> fun x ->` does not fit on a single line, then the break-infix-before-func=true behavior should be used.

ocamlformat-help.txt
```diff
@@ -139,6 +139,11 @@ OPTIONS (CODE FORMATTING STYLE)
            lines between one-liners of similar sorts. The default value is
            sparse.
 
+       --no-break-infix-before-func
+           Breaks infix expressions before the operator if it does not fit on
+           a single line and the operator is followed by an anonymou
```
jberdine: s/anonymou/anonymous

jberdine:
@gpetiot, any comment on my previous question regarding the indentation increase?

jberdine:
An unintended change:
```
--- a/asmcomp/asmgen.ml
+++ b/asmcomp/asmgen.ml
@@ -174,10 +174,10 @@ let set_export_info (ulambda, prealloc, structured_constants, export) =
 let end_gen_implementation ?toplevel ~ppf_dump
     (clambda : clambda_and_constants) =
   Emit.begin_assembly () ;
-  ( clambda
-  ++ Profile.record "cmm" (Cmmgen.compunit ~ppf_dump)
-  ++ Profile.record "compile_phrases" (List.iter (compile_phrase ~ppf_dump))
-  ++ fun () -> () ) ;
+  (clambda
+   ++ Profile.record "cmm" (Cmmgen.compunit ~ppf_dump)
+   ++ Profile.record "compile_phrases"
+        (List.iter (compile_phrase ~ppf_dump)) ++ fun () -> ()) ;
   (match toplevel with None -> () | Some f -> compile_genfuns ~ppf_dump f) ;
   (* We add explicit references to external primitive symbols.  This
      is to ensure that the object files that define these symbols,
```

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
+[@@@ocamlformat "break-infix-before-func=true"]
+
+;;
+f x
+>>= fun y ->
+g y
+>>= fun () ->
+f x >>= fun y -> g y >>= fun () -> f x >>= fun y -> g y >>= fun () -> y ()
```
samoht: I am not sure to understand why they are all on the same line here. Is that expected? I though the rule was:
- either everything (all the `>>=` sequence) can stay on one line and we are fine
- or we break after every element of the sequence. And we break more the `break-infix-before-func=true` than without.

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x
+
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
+  |> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
+  eeeeeeeeeeee eeeeeeeeee |> fun x -> x
```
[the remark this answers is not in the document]
samoht: yes I agree

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+    f x >>= fun y ->
+    g y >>= function
+    | x -> ( f x >>= fun y -> g y >>= function _ -> y () ) )
+
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x
+
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
+  |> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
```
samoht: Is the indentation expected? `ocp-indent` will format this as:

```ocaml
eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
```

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
+  eeeeeeeeeeee eeeeeeeeee |> fun x -> x
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
+  eeeeeeeeeeee eeeeeeeeee
+  |> fun xxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxx ->
```
samoht: Same here, `ocp-indent` will format this as:

```
eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
  eeeeeeeeeeee eeeeeeeeee
|> fun xxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxx ->
```

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+    xxxxxxxxxxx xxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxx xxxxxxxxxxx
+
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x
+
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
+  |> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
```
samoht: same

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxx xxxxxxxxxxxxx -> x
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee
+  |> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx -> x
```
samoht: same

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxxxxxx ->
+    xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
```
samoht: `ocp-indent` will format this as:

```
eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxxxxxx ->
xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
```

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+
+;;
+eeeeeeeeeeeee eeeeeeeeee
+  |> fun xxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxx ->
+    xxxxxxxxxxx xxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxx xxxxxxxxxxx
```
samoht: `ocp-indent` will format this as:

```ocaml
eeeeeeeeeeeee eeeeeeeeee
|> fun xxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxxxxxxx xxxxxxxxx ->
xxxxxxxxxxx xxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxx xxxxxxxxxxx
```

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> function x -> x
+
+;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> function
+| xxxxxx, xxxxxxxxxx, xxxxxxxx, xxxxxxxx -> x
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
+  eeeeeeeeeeee eeeeeeeeee |> function
```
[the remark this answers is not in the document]
samoht: agreed

test/passing/infix_bind.ml
```diff
@@ -0,0 +1,115 @@
...
+| x -> x
+
+;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
+  eeeeeeeeeeee eeeeeeeeee |> function
```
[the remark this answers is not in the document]
samoht: agreed

test/passing/infix_bind.ml
jberdine: Expected: yes, due to implementation difficulty of forcing boxes to break in sync. I'm not saying it's optimal, just expected.

test/passing/infix_bind.ml
jberdine: Not expected to me. It would be good if with `break-infix-before-func=false`, when breaks are needed, the formatting is the same as with `break-infix-before-func=true` wrt indentation, etc.

test/passing/infix_bind.ml
gpetiot: I disagree with the formatting of ocp-indent here, is this really the behavior we want?

test/passing/infix_bind.ml
gpetiot: Are you sure having no indentation is readable here?

gpetiot:
@samoht does the new formatting look alright to you?
@jberdine test_branch does not exhibit any bug and the diff looks good to me

test/passing/infix_bind.ml
samoht: there is now way to "come back" one level up, here, so yes I think that makes sense.

samoht: it's the formating that people are getting used to :p All these light sequence combinators are very similar to `;` which doesn't increase the indentation.

jberdine:
@gpetiot, it seems that the indentation increase I asked about [above](https://github.com/ocaml-ppx/ocamlformat/pull/463#pullrequestreview-170244431) is still happening. Is that expected? If so, do you have any comment about whether that is desired or too involved?

Re setting by default, I'd rather have a release with the option before changing the default if possible. Also, this is another case which conflicts with point 2 of the README code style section.

gpetiot:
Agreed to not make it the default choice. Does the indentation look better now?

gpetiot:
No bug with `test_branch.sh` and the diff looks good.

samoht:
The first changes (not break before `|> function`) looks good to me. For the second one, it seems similar to what `ocp-indent` is doing but I doubt it is related to the current feature :-)

gpetiot:
> Is the following change intended:
> 
> ```
> diff --git a/asmcomp/asmgen.ml b/asmcomp/asmgen.ml
> index 7542d4bbb..863085ae8 100644
> --- a/asmcomp/asmgen.ml
> +++ b/asmcomp/asmgen.ml
> @@ -171,10 +171,10 @@ let set_export_info (ulambda, prealloc, structured_constants, export) =
>  let end_gen_implementation ?toplevel ~ppf_dump
>      (clambda : clambda_and_constants) =
>    Emit.begin_assembly () ;
> -  ( clambda
> -  ++ Profile.record "cmm" (Cmmgen.compunit ~ppf_dump)
> -  ++ Profile.record "compile_phrases" (List.iter (compile_phrase ~ppf_dump))
> -  ++ fun () -> () ) ;
> +  (clambda
> +   ++ Profile.record "cmm" (Cmmgen.compunit ~ppf_dump)
> +   ++ Profile.record "compile_phrases" (List.iter (compile_phrase ~ppf_dump))
> +   ++ fun () -> ()) ;
>    (match toplevel with None -> () | Some f -> compile_genfuns ~ppf_dump f) ;
>    (* We add explicit references to external primitive symbols.  This
>       is to ensure that the object files that define these symbols,
> ```

It is not intended but difficult to fix. Without the option, function `fmt_op_args` was used to make a pretty wrapping, but that formatting is difficult to reproduce when breaking the box structure (or maybe I am missing something).

jberdine:
> The first changes (not break before `|> function`) looks good to me.

Ok, thanks. I must just be unaccustomed to this style.

jberdine:
> It is not intended but difficult to fix. Without the option, 
> function `fmt_op_args` was used to make a pretty wrapping, but that 
> formatting is difficult to reproduce when breaking the box structure 
> (or maybe I am missing something).

I see, thanks for the explanation. I think we should look some more for 
a solution to this. It is not a very nice inconsistency.

gpetiot:
> I see, thanks for the explanation. I think we should look some more for a solution to this. It is not a very nice inconsistency.

I fixed this case.

- no changes with test_branch when the new option is unset
- some changes (new box structure) with test_branch when the new option is set, does this new diff look alright to you?

ocamlformat-help.txt
```diff
@@ -146,6 +146,11 @@ OPTIONS (CODE FORMATTING STYLE)
            lines between one-liners of similar sorts. The default value is
            sparse.
 
+       --no-break-infix-before-func
+           Breaks infix expressions before the operator if it does not fit on
+           a single line and the operator is followed by an anonymous
+           function.
```
jberdine: The documentation should be reworded to reflect the default value. Perhaps: "Break infix operators whose right arguments are anonymous functions specially: do not break after the operator so that the first line of the function appears docked at the end of line after the operator"

src/Fmt_ast.ml
```diff
@@ -1281,7 +1281,7 @@ and fmt_index_op c ctx ~parens ?set {txt= s, opn, cls; loc} l is =
        | None -> fmt ""
        | Some e -> fmt "@ <- " $ fmt_expression c (sub_exp ~ctx e) ))
 
-and fmt_expression c ?(box = true) ?epi ?eol ?parens ?ext
+and fmt_expression c ?(box = true) ?epi ?eol ?parens ?ext ?(indent_wrap = 0)
```
jberdine: nit: since indent_wrap is semantically closely related to parens, I'd order it before ext

PR #464, opened 2018-10-22
[WIP] Improve: Break before anonymous func

gpetiot:
Should fix issue #462 and issue #207


gpetiot:
@samoht does this fix your issue?

test/passing/break_before_func.ml
```diff
@@ -0,0 +1,58 @@
...
+;;
+List.map
+  (* this is a loooooooooooooooooooooooooooooong commmmmmmmmmment
+     blablablabla *)
+  (function
+  | A ->
+      xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
+        xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx
+  ) list
+
```
jberdine: There should also be tests similar to each of the above for the case where there are multiple function expression arguments, to make sure that the formatting in that case isn't too crazy.

gpetiot: More examples above in the file

test/passing/break_before_func.ml
```diff
@@ -0,0 +1,195 @@
...
+    xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
+      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx )
+  (fun xxxxxxxxxxxxxxxxxxxxx yyyyyyyyyyyyyyyy zzzzzzzzzzzzzzzzzz
+       aaaaaaaaaaaaaaaa ->
+    xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
+      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx )
+  (fun xxxxxxxxxxxxxxxxxxxxx yyyyyyyyyyyyyyyy zzzzzzzzzzzzzzzzzz
+       aaaaaaaaaaaaaaaa ->
+    xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
+      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx ) list
```
jberdine: Not breaking before `list` here is a bit dangerous

test/passing/break_before_func.ml
```diff
@@ -0,0 +1,195 @@
+[@@@ocamlformat "break-before-func=false"]
+
+;;
+List.fooo (fun x ->
+    xxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxx
+      xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx
+  )
```
jberdine: Just to check, is it intended that this paren gets bumped to its own line, when there are no following arguments? What do you think @samoht ?

test/passing/break_before_func.ml
jberdine: To expand on this point, elsewhere a lot of effort has been put into avoiding cases where something short can appear on the same line after something that spans multiple lines. Experience shows that it is very easy to visually overlook such code, and by default I think we should keep with that approach and break here too.

gpetiot: Should be fixed now

jberdine:
Could you test this branch with test_branch.sh and look at the diff for files in the test/code/infer/sledge directory. The compact profile is used there, with this option enabled. Some of the changes are as I expect from the intent of this PR, but many look a bit off. I can be more specific, but for a first pass I think you will see several adjustments.

gpetiot:
>           Could you test this branch with test_branch.sh and look at the diff for files in the test/code/infer/sledge directory. The compact profile is used there, with this option enabled. Some of the changes are as I expect from the intent of this PR, but many look a bit off. I can be more specific, but for a first pass I think you will see several adjustments.

@jberdine I think I lost enough time trying to run the script on external regtests (and still can't). I think it will be better to have our own version of the `infer` codebase that we use for regression testing and keep it up to date with ocamlformat master, we don't need to have the last version of the codebase anyway, I think the current one is as diverse as it could be. So we don't have to use git shenanigans to compare different versions of ocamlformat.

jberdine:
I agree that most regressions should be uncovered by using infrequently updated versions of the test sources. Perhaps something based on git submodules, or maybe this is something where a duniverse would work, I don't know. Note that keeping a version of the test code formatted with master isn't enough, as we need to test the changes due to options that may not be enabled. I have no particular attachment to the mechanism using the git staging area, it was just the easiest automation of my manual testing workflow. Alternatives welcome!

gpetiot:
This PR messes too much with the boxes, is really this kind of formatting the direction we want to give ocamlformat?

jberdine:
I agree, to get this to reliably work nicely will involve a lot of changes to box structure. It is this sort of reason that I aborted my original attempts to produce this style of output, and why I mentioned on the original issues that I think that this (type of change) is much easier done after #406.

gpetiot:
Alright I will postpone this until #406 is done, and since the code will be very different by then I'm closing this PR.

PR #465, opened 2018-10-24
Improve: print-config prints all options

gpetiot:
The goal is to print the whole configuration (value of each option, even the default ones) and not to require input files.


samoht:
> we talked about this feature, do you really need to have the source of the value on the output?

I think that would be great, especially to help diagnosing where a configuration parameter comes from.

gpetiot:
This should be good to go now.

samoht:
The result looks great! Did you also remove the constraint that we had to pass an (empty) ml file?

gpetiot:
> The result looks great! Did you also remove the constraint that we had to pass an (empty) ml file?

Yes I did

samoht:
Nice, thanks 👍

jberdine:
Could you add a test to check that when there is a dir with a .ocamlformat file and a subdir also with an .ocamlformat file, the shadowing of the shallower file's options by the deeper one's is correctly reported? In my brief local testing, it looks like something is wrong there:
```
$ (cd test/code/infer/; ocamlformat --print-config sledge/src/sledge.ml) > /dev/null |& grep margin
margin=100 (file .ocamlformat:3)
$ grep margin test/code/infer/sledge/.ocamlformat
margin = 77
```

gpetiot:
> Could you add a test to check that when there is a dir with a .ocamlformat file and a subdir also with an .ocamlformat file, the shadowing of the shallower file's options by the deeper one's is correctly reported? In my brief local testing, it looks like something is wrong there: ...

This bug is fixed and I added a test

Issue #470, opened 2018-10-26
gen_version.sh is not portable


emillon:
Hi,

When `/bin/sh` is not bash, some errors are displayed:

```
          sh src/Version.ml [dev]
../tools/gen_version.sh: 7: ../tools/gen_version.sh: [[: not found
../tools/gen_version.sh: 12: ../tools/gen_version.sh: [[: not found
          sh src/Version.ml [release]
../tools/gen_version.sh: 7: ../tools/gen_version.sh: [[: not found
../tools/gen_version.sh: 12: ../tools/gen_version.sh: [[: not found
dune build _build/release/ocamlformat-help.txt
          sh src/Version.ml [release]
../tools/gen_version.sh: 7: ../tools/gen_version.sh: [[: not found
../tools/gen_version.sh: 12: ../tools/gen_version.sh: [[: not found
```

The reason is that it's invoked with `(sh)`, but this is a bash script. In particular, it uses `[[` to do pattern matching.

There are some ways to do that in a portable way using vanilla `sh`, but I'm wondering if there's a particular reason not to use `dune subst` for this?

Thanks

emillon:
(I can help port that to plain `sh` or `dune subst`)

gpetiot:
Would this issue be fixed by PR #471 ?

jberdine:
Since `gen_version.sh` starts with, `#!/usr/bin/env bash`, I don't understand how a non-bash sh gets run?

jberdine:
But I'm not opposed to having that script modified to work with more shells.

emillon:
@gpetiot Yes it seems so. I'm not familiar enough with ocamlformat internals to say if there's something fundamentally different here, but many projects using dune are using this mechanism. In particular, I believe that the different use cases outlined in the PR should be compatible with `["dune" "subst"] {pinned}`.  For a more conservative patch, you can also keep the `Version` module and leave the only substitution there (`let version = %%VERSION%%`).

@jberdine the script is invoked using `(sh tools/gen_version.sh)` (see #448). This means that it will be executed using `sh tools/gen_version.sh`. When doing that, the `#!` line is ignored (also, consider that some systems don't have bash installed). Sorry I missed that while reviewing the PR.

PR #471, opened 2018-10-26
Improve: make gen_version.sh more portable

gpetiot:
Should fix issue #470
It should not have a negative impact on the release procedure.

### DIFF
--- a/tools/gen_version.sh
+++ b/tools/gen_version.sh
@@ -1,5 +1,3 @@
-#!/usr/bin/env bash
-
 ######################################################################
 #                                                                    #
 #                            OCamlFormat                             #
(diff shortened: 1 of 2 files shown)


tools/gen_version.sh
```diff
@@ -15,12 +15,12 @@
...
 FILE="$1"
 
-if [[ ! -z "$2" ]]; then
+if test -n "$2"; then
     # second arg passed when called from ocamlformat.opam
     VERSION="$2"
 else
     # second arg omitted when called from src/dune
-    if [[ ! "%%VERSION%%" == "%%"*"%%" ]]; then
+    if test "%%VERSION%%" != "%%"*"%%"; then
```
jberdine: Did you test this? It does not behave the same as before at least in the case when passed only one arg.

gpetiot: You are right, didn't find a way to make this more portable though

jberdine:
Looking better. Can you describe the testing done to check that the 4 use cases above still work?

gpetiot:
Tests done:
- old gen_version.sh:
  * `tools/gen_version.sh foo` (no watermark); output: `let version = "5e1622b"`
  * `tools/gen_version.sh v42` (no watermark); output: `let version = "v42"`
  * `tools/gen_version.sf foo` (watermarked with 5e1622b); output: `let version = "5e1622b"`
  * `tools/gen_version.sf foo v42` (watermarked with 5e1622b); output: `let version = "v42"`
- new gen_version.sh:
  * `tools/gen_version.sh foo` (no watermark); output: `let version = "5498451"`
  * `tools/gen_version.sh foo v42` (no watermark); output: `let version "v42"`
  * `tools/gen_version.sh foo` (watermarked with 5498451); output: `let version = "5498451"`
  * `tools/gen_version.sh foo v42` (watermarked with 5498451); output: `let version = "v42"`

jberdine:
Ok, thanks, that looks good. I think what I'll do is add this PR to the bottom of the next release branch and try it while preparing the release, and if all goes well then include it in master.

hhugo:
Can we maybe make the script run with bash in master while waiting for this PR to land ?

gpetiot:
Is this PR still relevant now that #592 is merged?

jberdine:
IIRC there were complaints from people who still don't have bash

hhugo:
Bash should be removed from the dune file
https://github.com/ocaml-ppx/ocamlformat/blob/master/src/dune#L15

gpetiot:
Replaced by #664

Issue #341, opened 2018-08-20
Preserve position of comments in `match (*cmt*) expr with`.


hhugo:
Refines https://github.com/ocaml-ppx/ocamlformat/pull/330

gpetiot:
I don't see any problem with this code, is the problem still relevant for you ?

jberdine:
The implementation of #330 has the undesired effect of transforming:
```
match
  (* comment about entire pattern matching *)
  e
with
| ...
```
to
```
(* comment about entire pattern matching *)
match e with
| ...
```
What should happen is that the comment stays on whichever side of `match` it is in the source.

gpetiot:
For the record, on master it is now formatted as:
```
;;
match (* comment about entire pattern matching *)
      e with k -> z
```

gpetiot:
Update: on master, this:
```ocaml
let foo =
 match
  (* comment about entire pattern matching *)
  e
with
| x -> x
```
is still rewritten as:
```ocaml
let foo =
  (* comment about entire pattern matching *)
  match e with x -> x
```

hhugo:
#867 might resolve this issue.

PR #472, opened 2018-10-26
[HELP] Fix: Preserve position of comments in match (*cmt*) expr with

gpetiot:
Should fix issue #341

### DIFF
--- a/test/passing/comments.ml
+++ b/test/passing/comments.ml
@@ -137,10 +137,12 @@ let () =
   ()
 
 let foo =
+  (* foo *)
   match (* comment about entire pattern matching *)
     e with k -> z
 
 let foo =
+  (* foo *)
   match (* comment about entire pattern matching *)
     e [@foo] with k -> z
 
@@ -157,6 +159,7 @@ let foo =
     e [@foo] with k -> z | fooooooooooooooooooooo -> foooooooooooooooooo
 
 let foo =
+  (* foo *)
   match (* foo *) (* fooooooooooooooo *)
     e (* foo *)  [@foo] with k -> z
  | foooooooooooooooooooo -> fooooooooooooooooo
--- a/test/passing/comments.ml
+++ b/test/passing/comments.ml
@@ -136,21 +136,27 @@ let () =
   (* *)
   ()
 
-;;
-match (* comment about entire pattern matching *)
-      e with k -> z
-;;
-match (* comment about entire pattern matching *)
-      e [@foo] with k -> z
-;;
-match (* foo *) (* fooooooooooooooo *)
-  e (* foo *)  [@foo] with k -> z
-;;
-match (* comment about entire pattern matching *)
-      e with k -> z | foooooooooooooooooooooooooo -> foooooooooooooooooooooo
-;;
-match (* comment about entire pattern matching *)
-      e [@foo] with k -> z | fooooooooooooooooooooo -> foooooooooooooooooo
-;;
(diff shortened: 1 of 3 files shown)


test/passing/comments.ml
```diff
@@ -129,3 +129,22 @@ type t = {
 a : int [@default a]
 (* comment *)
 ;  b : flag }
+
+;;
+match (* comment about entire pattern matching *)
+      e with k -> z
```
jberdine: Note that in
```
(* comment about entire pattern matching *)
match e with k -> z
```
the comment comes before `match` and that should be preserved. It is not enough to always put the comments before or after the `match`, the original position needs to be preserved.

gpetiot: Agreed, but for the record, this is not a regression, this is also formatted like this on master.

gpetiot: This gets formatted as:
```
(* comment about entire pattern matching *)

;;
match e with k -> z
```
as it is the case for any expression on master. Is this what you would like to change? It would belong in another PR.

jberdine: The whole aim of #341 is to make sure that a comment before a token, here `match`, are still before the token after being formatted, and vice versa: a comment after a token stays after it. Details about `;;` are probably tangential, and it might be better to use test cases which are bodies of let-bindings.

gpetiot:
I see https://github.com/facebook/infer/commit/6a4d8b72f0d8cbfcff0affe347bb285fb1284e6a has fixed some issues on the infer repo but it's still not 100% OK. Can you ping me when the test_branch script is 100% OK on master?

test/passing/comments.ml
```diff
@@ -130,21 +130,27 @@ a : int [@default a]
 (* comment *)
 ;  b : flag }
 
-;;
-match (* comment about entire pattern matching *)
-      e with k -> z
```
jberdine: Could you add some tests where the comment comes before `match` as well, to check that their relative positions are preserved.

jberdine:
@gpetiot, I don't understand your command about the infer change. That was to fix the situation for normalizing docstrings, but wasn't intended to be related to this PR.

test/passing/comments.ml
```diff
@@ -129,3 +129,31 @@ type t = {
 a : int [@default a]
 (* comment *)
 ;  b : flag }
+
+let foo =
+  (* foo *)
+  match (* comment about entire pattern matching *)
+    e with k -> z
```
jberdine: Sorry for the piecemeal commentary, but what about the case where the comment and match are on the same line?
```
(* foo *) match (* bar *) e with...
```

jberdine:
Rebasing this on master and running `tools/test_branch.sh preserve-match-cmts 'parse-docstrings=false' 'parse-docstrings=false'` still shows the unstable behavior.

gpetiot:
No there are still regressions with this PR.

gpetiot:
I think I need help on this one, I feel I am close but there are weird comments positionning in case of nested pattern matchings that I can't manage to fix.

e.g.:
```
let foo =
  match x with
  | BinaryOperator (_, exp0, exp1) -> (
      (* pointer arithmetic. somewhere in one of the expressions, there should be at least    one pointer type represented as an access path. just use that access path and forget
                about the arithmetic. if you need to model this more precisely, you should be using
                SIL instead *)
          match HilExp.get_access_exprs exp0 with _ -> x )
```

gpetiot:
We need to start again from the beginning on this one.

PR #475, opened 2018-10-31
Fix: disable-outside-detected-project: disable ocamlformat when no .o…

### LINKED ISSUE #473: `--disable-outside-detected-project` and `.ocp-indent` files
I'm trying to find an emacs configuration that works for both projects that use ocamlformat and projects that don't. I tried the newly added `--disable-outside-detected-project` option which seems good for that. However, when a project doesn't have a `.ocamlformat` file but has a `.ocp-indent` file, then it is considered as using ocamlformat and `ocamlformat --disable-outside-detected-project` does reformat files.

I find this behaviour unexpected as many projects that don't use ocamlformat have `.ocp-indent` files. I suggest that when only a `.ocp-indent` file is found, either:
- files are only re-indented but not reformatted
- nothing special is done and files are printed as it

gpetiot:
…camlformat file is found

Should fix #473

### DIFF
--- a/ocamlformat-help.txt
+++ b/ocamlformat-help.txt
@@ -187,7 +187,8 @@ OPTIONS
            Do not read .ocamlformat config files outside the current project.
            The project root of an input file is taken to be the nearest
            ancestor directory that contains a .git or .hg or dune-project
-           file. If no config file is found, formatting is disabled.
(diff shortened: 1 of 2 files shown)


diml:
Just tried it and I confirm that it works as expected, thanks!

PR #479, opened 2018-10-31
Add PR test script and update contributing guidelines with expected usage

jberdine:


jberdine:
@gpetiot please try this script and see if it works for others or if it depends on my config somehow.

Issue #477, opened 2018-10-31
Formatting of lists inside documentation comments


diml:
I'm trying the recently added support for formatting documentation comments and I find the formatting of lists a little bit too compact. For instance, for the following input:

```ocaml
(** list with short lines:

    - x
    - y
    - z

    list with long lines:

    - xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    - yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy
    - zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

    enumerated list with long lines:

    {ol
     {- xxxxxxxxxxxxxxxxxxxxxxx}
     {- yyyyyyyyyyyyyyyyyyyyyyy}
     {- zzzzzzzzzzzzzzzzzzzzzzz}
    }
*)
```

ocamlformat (at 9d23a512b160c0c78a27616b9d3ddb34d8333451) prints:

```ocaml
(** list with short lines:

    {ul {- x} {- y} {- z}} list with long lines:

    {ul {- xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx}
          {- yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy}
          {- zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz}} enumerated list with long
    lines:

    {ol {- xxxxxxxxxxxxxxxxxxxxxxx} {- yyyyyyyyyyyyyyyyyyyyyyy}
          {- zzzzzzzzzzzzzzzzzzzzzzz}} *)
```

It would seem nicer to me if the output was more something like this:

```ocaml
(** list with short lines:

    {ul
     {- x}
     {- y}
     {- z}}

    list with long lines:

    {ul
     {- xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx}
     {- yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy}
     {- zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz}}

    enumerated list with long lines:

    {ol
     {- xxxxxxxxxxxxxxxxxxxxxxx}
     {- yyyyyyyyyyyyyyyyyyyyyyy}
     {- zzzzzzzzzzzzzzzzzzzzzzz}} *)
```

[...]

PR #480, opened 2018-11-01
Improve the formatting of lists in doc comments

diml:
This implements #477.

### DIFF
--- a/src/Fmt_odoc.ml
+++ b/src/Fmt_odoc.ml
@@ -37,10 +37,6 @@ let fmt_ref_kind = function
   | RK_link -> str ":"
   | RK_custom s -> str s $ str ":"
 
-module List_kind = struct
-  type t = List | Enum
-end
-
 let rec fmt_style style txt =
   let s =
     match style with
@@ -62,8 +58,8 @@ and fmt_text_elt = function
   | PreCode s -> hovbox 0 (wrap "{[\n" "@\n]}" (hovbox 0 (verbatim s)))
   | Verbatim s -> hovbox 0 (wrap "{v\n" "@\nv}" (hovbox 0 (verbatim s)))
   | Style (st, txt) -> fmt_style st txt
-  | List l -> fmt_list List_kind.List l
-  | Enum l -> fmt_list List_kind.Enum l
+  | List l -> fmt_list `List l
+  | Enum l -> fmt_list `Enum l
   | Newline -> fmt "\n@\n"
   | Title (i, None, txt) ->
       hovbox 0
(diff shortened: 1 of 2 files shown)


src/Fmt_odoc.ml
```diff
@@ -85,7 +89,29 @@ and fmt_text_elt = function
...
-and fmt_item txt = hovbox 0 (wrap "{- " "}" (fmt_text txt))
+and fmt_list (kind : List_kind.t) l =
+  let light_syntax =
+    let line_start = fmt (match kind with List -> "- " | Enum -> "+ ") in
+    let fmt_item txt = hovbox 2 (line_start $ fmt_text txt) in
+    vbox 0 (list l "@," fmt_item)
+  in
+  let heavy_syntax =
+    let fmt_item txt = hovbox 0 (wrap "{- " "}" (fmt_text txt)) in
+    let start : (_, _, _) format =
```
jberdine: FMI: is it important that this is a 3-arg format, or would `_ format` be good enough?

src/Fmt_odoc.ml
```diff
@@ -85,7 +89,29 @@ and fmt_text_elt = function
...
+      match kind with List -> "{ul@," | Enum -> "{ol@,"
+    in
+    vbox 1 (wrap start "}" (list l "@," fmt_item))
+  in
+  let print_and_parse fmt =
+    let str = Format.asprintf "\n%t" fmt in
+    Octavius.parse (Lexing.from_string str)
+  in
+  if
+    Poly.equal (print_and_parse light_syntax) (print_and_parse heavy_syntax)
```
jberdine: nice

test/passing/doc_comments.mli
```diff
@@ -127,3 +127,41 @@ val x : x
         s |> String.split_lines |> List.map ~f:String.strip
         |> fun s -> list s "@," Fmt.str
     ]} *)
+
+(** Lists:
+
+    list with short lines:
+
```
jberdine: Is this open line necessary, or just desired?

src/Fmt_odoc.ml
```diff
@@ -37,6 +37,10 @@ let fmt_ref_kind = function
   | RK_link -> str ":"
   | RK_custom s -> str s $ str ":"
 
+module List_kind = struct
+  type t = List | Enum
+end
```
jberdine: Nothing significant, but I'm not sure what the reasoning is behind adding a module vs just using a couple polymorphic variant tags?

src/Fmt_odoc.ml
diml: Indeed, I updated the code

src/Fmt_odoc.ml
diml: It's just a habit I have, I prefer to only use polymorphic variants when I need one of the specific features of polymorphic variants. I find that with regular variants you get errors closer to the actual location of the problem and nicer error messages.

I updated the code to use polymorphic variants instead.

diml:
Thanks, I updated the code

PR #481, opened 2018-11-02
Fix: Sanitize docstring check

gpetiot:
- Fixing the Not_found exception
- Factorize the normalize function

It doesn't fix the case when the formatting changes the docstring, e.g. on test/code/infer/infer/src/IR/Exp.ml we get the error: 

```
Error: Docstring (** Checks whether an expression denotes a location by pointer arithmetic.
Currently, catches array - indexing expressions such as a [i] only. *) added.
```

That can be fixed by modifying the `-` into `--`.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -39,6 +39,7 @@ type t =
   ; module_item_spacing: [`Compact | `Sparse]
   ; ocp_indent_compat: bool
   ; parens_tuple: [`Always | `Multi_line_only]
+  ; parse_docstrings: bool
   ; quiet: bool
   ; sequence_style: [`Separator | `Terminator]
   ; type_decl: [`Compact | `Sparse]
@@ -830,6 +831,13 @@ module Formatting = struct
       (fun conf x -> {conf with parens_tuple= x})
       (fun conf -> conf.parens_tuple)
 
+  let parse_docstrings =
+    let doc = "Parse and format docstrings." in
+    let names = ["parse-docstrings"] in
+    C.flag ~default:true ~names ~doc ~section
+      (fun conf x -> {conf with parse_docstrings= x})
+      (fun conf -> conf.parse_docstrings)
+
   let sequence_style =
     let doc = "Style of sequence." in
     let names = ["sequence-style"] in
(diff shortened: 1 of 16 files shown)


gpetiot:
I have added a `parse-docstrings` option so that the formatting can be disabled locally if it is too weird.

gpetiot:
I improved and precised the check and error message, it could be improved when we have only 1 unstable docstring in the before list and in the after list, but it would require a third variant of type `Location.t * string * string`, but the code is already cryptic enough as it is now.
I don't know if I should export the following type `docstring_error` in Normalize.ml(i) to reduce the code duplication everywhere:
```
type docstring_error =
         [ `Moved of Location.t * Location.t * string
         | `Unstable of Location.t * string ]
```

src/Translation_unit.ml
```diff
@@ -316,26 +320,36 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
...
+                         %s *) moved to @{<loc>%a@}.\n\
+                         %!"
+                        Location.print_loc loc_before (String.strip msg)
+                        Location.print_loc loc_after
+                | `Unstable (loc, s) ->
                     Caml.Format.eprintf
-                      "%!@{<loc>%a@}:@,@{<error>Error@}: Docstring (** %s \
-                       *) added.\n\
+                      "%!@{<loc>%a@}:@,@{<error>Error@}: Formatting of (** \
+                       %s *) is unstable, please tighten up this comment \
```
jberdine: Maybe add after `is unstable` something like `(e.g. parses as a list or not depending on the margin)` to give users a better idea of what is meant by 'unstable' and what to look out for.

src/Translation_unit.ml
```diff
@@ -46,7 +48,9 @@ exception Warning50 of (Location.t * Warnings.t) list
 exception
   Internal_error of
     [ `Ast
-    | `Doc_comment of (Location.t * Location.t * string) list
+    | `Doc_comment of [ `Moved of Location.t * Location.t * string
```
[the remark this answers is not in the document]
gpetiot: Issue #484 submitted

src/Conf.ml
```diff
@@ -830,6 +831,13 @@ module Formatting = struct
       (fun conf x -> {conf with parens_tuple= x})
       (fun conf -> conf.parens_tuple)
 
+  let parse_docstrings =
+    let doc = "Parse and format docstrings." in
+    let names = ["parse-docstrings"] in
+    C.flag ~default:true ~names ~doc ~section
```
jberdine: @gpetiot I think that this should default to `false` initially. We are seeing that enabling it can require changing the source docstrings, and furthermore when this happens it is reported as an internal error. This is a reasonably significant potentially-breaking change, so to ease the upgrade path, this should default to `false` for now, despite what we think ought to be the eventual default set of options.

PR #496, opened 2018-11-08
Test: improve test_branch.sh to allow different config for branch

jberdine:


jberdine:
Hmm, it still doesn't work very well. Options that are set in profiles or config files cannot be configured using the env var.

jberdine:
@gpetiot, are we sure we want the priority of the env var to be lower than config files? I'm not aware of use cases for the env var that are much different from what this test script is trying to do, where it would be better if the env var had the highest priority.

PR #497, opened 2018-11-08
fix break-struct for toplevel items (not in a struct)

### LINKED ISSUE #490: Weird module + include formating with break-struct=natural
When my .ocamlformat configuration file contains `break-struct=natural`, I get pretty weird formatting in a file that contains only a module and an include. They end up both on the same line. This behavior doesn't happen when there is more content in the file.

The .ocamlformat file:

```
break-struct=natural
```

Original file:

```ocaml
module Json = Yojson.Basic

include B
```

Results in:

```ocaml
module Json = Yojson.Basic include B
```

But a file with more content:

```ocaml
module Json = Yojson.Basic

include EsMappingShared

module T = struct
  let v = x
  let y = 100
  let z = 200
end

module Y = struct let v = x end
```

Ends up in a normal formatting:

```ocaml
module Json = Yojson.Basic
include EsMappingShared

module T = struct
  let v = x

  let y = 100

  let z = 200
end

module Y = struct let v = x end
```

gpetiot:
Should fix #490

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3616,7 +3616,7 @@ and fmt_module_expr c ({ast= m} as xmod) =
       ; psp=
           fmt_if_k (not empty)
             (fmt_or c.conf.break_struct "@;<1000 2>" "@;<1 2>")
-      ; bdy= within $ fmt_structure ~in_struct:true c ctx sis
+      ; bdy= within $ fmt_structure c ctx sis
       ; cls= close_box
       ; esp=
           fmt_if_k (not empty)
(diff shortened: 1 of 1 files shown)


src/Fmt_ast.ml
```diff
@@ -3719,20 +3719,21 @@ and fmt_structure c ctx itms =
     List.group itms ~break:(fun (itmI, cI) (itmJ, cJ) ->
         Ast.break_between c.cmts (Str itmI, cI.conf) (Str itmJ, cJ.conf) )
   in
+  let break_struct = c.conf.break_struct || not in_struct in
```
jberdine: I think it would be better to look at `ctx` instead of passing an extra arg.

PR #498, opened 2018-11-08
Add a parens-tuple-patterns configuration option

NathanReb:
Fixes #488

## What's going on

The idea is to allow users to control how they want tuple patterns to be parenthesized.

I added a configuration option called `parens-tuple-patterns` that can be `always` or `multi-line-only` just as it's expression counterpart and default to the latter for retro-compatibility reasons.

I also updated the documentation for the `parens-tuple` option to clarify that it only applies to expressions as changing its name would break compatibility.

There's a new test case in `test/passing/parens_tuple_patterns.ml` to check both behaviour and the default one.

## Tests results

While `make test` passes, I had some issues with `tools/test_branch.sh`.
Running it with master gives the following output:
```
11% 81:641=8s code/infer/infer/src/checkers/NullabilitySuggest.ml                                                                                                                                                  ocamlformat.exe: Cannot process "code/infer/infer/src/bufferoverrun/bufferOverrunModels.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: comments changed.
15% 109:613=5s code/infer/infer/src/checkers/Sanitizer.ml                                                                                                                                                          ocamlformat.exe: Cannot process "code/infer/infer/src/checkers/loopInvariant.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: comments changed.
15% 110:612=5s code/infer/infer/src/checkers/SimpleChecker.ml                                                                                                                                                      ocamlformat.exe: Cannot process "code/infer/infer/src/checkers/control.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: comments changed.
19% 141:581=4s code/infer/infer/src/backend/Tasks.ml                                                                                                                                                               ocamlformat.exe: Cannot process "code/infer/infer/src/backend/OndemandCapture.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: comments changed.
19% 144:578=4s code/infer/infer/src/backend/Differential.ml                                                                                                                                                        ocamlformat.exe: Cannot process "code/infer/infer/src/backend/reporting.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: comments changed.
[...]

PR #503, opened 2018-11-14
Improve: formatting of or-patterns in try expressions

### LINKED ISSUE #489: try and match expressions or-pattern cases are handled inconsistently
The or-pattern cases in a `try` expression are formatted in a different way than in a `match` expression.

If you set `break-cases=all`, the following code:

```ocaml
let _ =
  try () with
  | End_of_file
  | Not_found ->
      ()

let _ =
  match 1 with
  | 1
  | _ ->
      true
```

is formatted as:

```ocaml
let _ =
  try () with
  | ( End_of_file
  | Not_found ) ->
      ()

let _ =
  match 1 with
  | 1
  | _ ->
      true
```

while I would expect the `try` expression to remain the same.

NathanReb:
Fixes #489

I think the issue was that or-patterns in Pexp_try context weren't picked as nested or-patterns.

I wrote a test case first and then tried to make it pass. As I'm not really familiar with the codebase yet it mostly felt like randomly swapping boolean values until it worked so I might be missing something.

As for #498 the `tools/test_branch.sh` still fails (even on master). I'll try to find some time to figure this out if I can.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1027,7 +1027,8 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
       let nested =
         match ctx0 with
         | Pat {ppat_desc= Ppat_or _} -> true
-        | Exp {pexp_desc= Pexp_match _ | Pexp_function _} -> true
+        | Exp {pexp_desc= Pexp_match _ | Pexp_try _ | Pexp_function _} ->
+            true
         | _ -> false
       in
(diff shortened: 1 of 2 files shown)


jberdine:
Thanks! I modified the title as we try to use "Fix" for cases where code that does not parse back to the same abstract syntax is generated, or other hard failures, and to use "Improve" for improving output that is correct according to the parser.

jberdine:
I tested locally and this is good to go.

PR #509, opened 2018-11-15
Fix: comments should not be parsed for diff

gpetiot:
Fix the test_branch.sh script.

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -581,7 +581,7 @@ let remaining_comments t =
 
 let diff c x y =
   let norm z =
-    let f (txt, _) = Normalize.docstring c txt in
+    let f (txt, _) = Normalize.docstring ~parse:false c txt in
     Set.of_list
       (module String)
       (List.map ~f (List.dedup_and_sort ~compare:Poly.compare z))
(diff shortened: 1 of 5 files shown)


jberdine:
Interesting. For my understanding, can you explain the reasoning a little, or what you observed that lead to this change?

gpetiot:
Some of these comments can be successfully parsed as ocamldoc comments but the normalization of those comments is somehow incorrect (I didn't dig further, as we are not supposed to parse them anyway). This is a quick fix so that all pending PR's can be unblocked.

jberdine:
Is what happens that, when `Conf.parse_docstrings` is true, the docstrings in the parsed input and reparsed output are traversed in `Normalize`, causing them to be parsed by `Octavius.parse`, and for some reason the results are different? I don't understand why we are not supposed to parse the docstrings when `Conf.parse_docstrings` is true. Maybe I'm misunderstanding from an earlier point.

gpetiot:
This patch is only for vanilla comments not for docstrings.

gpetiot:
> Ah, `Cmts.diff` calls `Normalize.docstring` on non-docstring comments. I think it would be clearer to add a `Normalize.comment` function that doesn't parse, and use it in `Cmts.diff`, instead of an argument to `Normalize.docstring`. The implementation could be basically the same as now, but I think clearer interface. What do you think?

I agree on the redesign, it should be good to go now

gpetiot:
Continued in #511

Issue #482, opened 2018-11-02
Normalize escaping in docstrings


jberdine:
When parsing and formatting docstrings, escape sequences are sometimes expanded, causing the check to fail. For example on the following docstrings from `ocaml/ocamldoc/odoc_info.mli`
```
(** to extend \{foo syntax *)
(** The different forms of references in \@see tags. *)
```

gpetiot:
You only have this error on branch #481 do you?

jberdine:
I have only only tested the docstring parsing with #480 and #481, so you could be right.

gpetiot:
Seems to be fixed by replacing
```
Format_.asprintf "%a@!"     ...
```
by
```
Format_.asprintf "%a%!"     ...
```
The fix will be done in #481.

jberdine:
There are still some issues with escaping in docstrings.  One failure is on code/ocaml/ocamldoc/odoc_man.ml, where the ast changes as follows:
```
------ odoc_man.old.ast
++++++ odoc_man.new.ast
@|-559,33 +559,33 ============================================================
 |                                            expression (_none_[1,0+-1]..[1,0+-1]) ghost
 |                                              Pexp_ident "Buffer.contents" (_none_[1,0+-1]..[1,0+-1]) ghost
 |                                            [
 |                                              <arg>
 |                                              Nolabel
 |                                                expression (_none_[1,0+-1]..[1,0+-1]) ghost
 |                                                  Pexp_ident "b" (_none_[1,0+-1]..[1,0+-1]) ghost
 |                                            ]
 |                          ]
 |                      None
 |                class_field (_none_[1,0+-1]..[1,0+-1]) ghost
 |                    attribute "ocaml.doc"
 |                      [
 |                        structure_item (_none_[1,0+-1]..[1,0+-1]) ghost
 |                          Pstr_eval
 |                          expression (_none_[1,0+-1]..[1,0+-1]) ghost
-|                            Pexp_constant PConst_string("Printf groff string for the @before information.",None)
+|                            Pexp_constant PConst_string("Printf groff string for the\n@before information. ",None)
 |                      ]
 |                  Pcf_method Public
 |                    "str_man_of_before" (_none_[1,0+-1]..[1,0+-1]) ghost
 |                    Concrete Fresh
 |                    expression (_none_[1,0+-1]..[1,0+-1]) ghost
 |                      Pexp_poly
 |                      expression (_none_[1,0+-1]..[1,0+-1]) ghost
[...]

PR #511, opened 2018-11-16
Fix odoc seps

gpetiot:
Follows up on #509 

Fix unnecessary spaces that changes the formatting.

It is still necessary to modify the docstrings in the `infer` codebase (cc @jberdine), for example:
- sometimes `[A]x` is used and we are supposed to have a space in-between, just use a `-` if the space is not wanted
- sometimes there are unexpected `@`-tags that have to be escaped or put in code blocks
- sometimes `{A}` that are parsed as odoc styles where A is a custom style name and an argument is expected so there is a space between A and the (empty) argument


jberdine:
I am not sure of exactly why the checks are tripping on the infer test, but there are at least some `\[` escape sequences within `[` `]` code blocks where the `\` is lost in the output. AFAIU, this is incorrect as the escape is necessary to produce a `[` character in a code block.

gpetiot:
Seems to also fix #482

src/Fmt_odoc.ml
```diff
@@ -12,6 +12,12 @@
 open Octavius.Types
 open Fmt
 
+let str s =
+  String.substr_replace_all s ~pattern:"@" ~with_:"\\@"
+  |> String.substr_replace_all ~pattern:"{" ~with_:"\\{"
+  |> String.substr_replace_all ~pattern:"}" ~with_:"\\}"
```
gpetiot: Maybe it makes sense to also add `[` and `]`

jberdine: It could very well. Might address one of my comments on your infer PR?

jberdine: Also, it might make sense to use `Base.String.Escaping` for this to avoid a potential performance problem.

gpetiot:
I disabled the parsing of docstring by default.

There are still some issues for the escaped characters, for example `\g` is not modified but `\{` is changed into `{` that makes it a nightmare to format latex strings

src/Fmt_odoc.ml
```diff
@@ -90,11 +89,11 @@ and fmt_text_elt = function
       hvbox 0 (wrap "{!modules:" "}" (list l "@," str))
   | Special_ref SRK_index_list -> str "{!indexlist}"
   | Target (s, l) ->
+      let target = Option.value_map s ~default:"" ~f:(fun s -> s ^ ":") in
+      let latex = String.is_empty target || String.equal target "latex:" in
```
jberdine: Is it clear that latex is special here, or should all `Target`s be left unescaped?

gpetiot: I didn't find any info suggesting that latex is a special case, so I don't escape any target anymore

gpetiot:
I know that you already accepted the changes a long time ago but since the code changed I would rather wait for a last approval before merging this one.

gpetiot:
Last commit should fix #513

gpetiot:
I don't understand how test_branch can still fail on master even though the parsing of docstrings is now disabled by default

jberdine:
I debugged this a bit, it seems that there is a problem (I'm not sure where) that causes ocamlformat to misidentify the project root, leading to the tests failing. Running an ocamlformat command from the `test` directory, mimicing what `make test` does:
```
$ ../_build/dev/src/ocamlformat.exe --no-version-check -i code/infer/infer/src/checkers/cost.ml --print-config |& grep parse-docstrings=
parse-docstrings=true (file ../.ocamlformat:3)
```
Note that test/code/infer/.git does not seem to have stopped the config search from picking up .ocamlformat from higher in the file system. If I explicitly add `--root code/infer`, it works.

PR #515, opened 2018-11-22
[WIP] Fix: break-cases is more consistent and wastes less space

gpetiot:
Should improve the situation of #12 
I think that the small regressions on Source.ml are a small price to pay considering the general improvements (I didn't manage to limit those regressions yet).
I didn't try test_branch yet, I would like to know if this new style is accepted first.

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -1125,8 +1125,7 @@ end = struct
             List.exists cases ~f:(function
               | {pc_lhs} when pc_lhs == pat -> true
               | _ -> false ) )
-      | Pexp_for (p, _, _, _, _) | Pexp_fun (_, _, p, _) -> assert (p == pat)
-      )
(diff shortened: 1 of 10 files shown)


jberdine:
One high-level point is that this changes `break-cases = nested` in a way that it no longer achieves what it was created for. The main point is that in this mode, nested or-patterns are not indicated with different indentation of `|`, and so it is intended that there is a break following every `->` so that the bodies do not get hidden. Did you intended to change the output only in `break-cases = all` mode?

gpetiot:
I mainly focused on `break-cases = all` for now and I was thinking that maybe we could switch to a `break-cases = compact / sparse` instead of the current 3-valued option.
The code would also be easier to maintain.

jberdine:
We need at least options that achieve the objectives of the current `fit` and `nested` options, and there have been requests for a mode where cases are broken ignoring the margin. So perhaps two booleans would work instead, but I don't expect that the implementation will treat them independently.

PR #518, opened 2018-11-23
Disable docstring parsing for external tests

gpetiot:
and fix one of the tests from the `ocaml` repo testsuite


jberdine:
My impression was that the detection of project root had some problem. What do you think?

gpetiot:
I checked that the detection of the project root is fine when the option `disable-outside-detected-project` is set, but it disables the formatting if there is no `.ocamlformat` file at the root (and this is the case for `infer`) so I decided to override the options with configuration files instead.

test/.ocamlformat
```diff
@@ -0,0 +1 @@
+parse-docstrings = false
```
jberdine: This seems too specific, the same problem will happen for any other option. Perhaps set profile = default instead?

gpetiot:
It should be good to go now

PR #524, opened 2018-12-06
amend janestreet profile

mbarbin:
Working on the next internal JS release.


gpetiot:
We should merge #525 before.

gpetiot:
I updated the tests now that #525 is merged, do they look good to you?

PR #526, opened 2018-12-07
Improve: remove more spaces inside parenthesized multiline constructs

gpetiot:
Extends `indicate-multiline-delimiters` (reduces the diff on `core_kernel`).


src/Fmt_ast.ml
```diff
@@ -948,8 +948,13 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
       let parens =
         parens || Poly.(c.conf.parens_tuple_patterns = `Always)
       in
+      let wrap =
```
jberdine: It would probably be better to use a longer name that doesn't shadow `Fmt.wrap`.

PR #527, opened 2018-12-07
Improve: space around collection expressions

gpetiot:
Reduces the diff on `core_kernel`


gpetiot:
No errors with `test_branch.sh` and the diff looks good.

ocamlformat-help.txt
```diff
@@ -183,6 +183,10 @@ OPTIONS (CODE FORMATTING STYLE)
            will always break the line before a single case. The default value
            is compact.
 
+       --space-around-collection-expressions
+           Add a space around collection expressions (lists, arrays,
+           records).
```
jberdine: When I looked at the code, I realized that my first interpretation of this doc string was confused (I expected the added space outside the delimiters, not inside). What about something like: "Add a space inside the delimiters of collection expressions..."?

src/Fmt_ast.ml
```diff
@@ -967,8 +967,12 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
       , Some {ppat_desc= Ppat_tuple [x; y]; ppat_attributes= []} ) -> (
     match sugar_list_pat c pat with
     | Some (loc_xpats, nil_loc) ->
+        let wrap =
```
jberdine: Here and below, better not to shadow `Fmt.wrap`

src/Fmt_ast.ml
```diff
@@ -972,12 +984,8 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
       , Some {ppat_desc= Ppat_tuple [x; y]; ppat_attributes= []} ) -> (
     match sugar_list_pat c pat with
     | Some (loc_xpats, nil_loc) ->
-        let wrap =
-          if c.conf.space_around_collection_expressions then wrap "[ " "@ ]"
-          else wrap_fits_breaks c.conf "[" "]"
-        in
         hvbox 0
-          (wrap
+          (wrap_list c.conf
```
jberdine: Normally we pass `c` in full, and only select out what is needed at the use site. Maybe better to do that here as well.

gpetiot:
I applied the changes to the patterns as well.

PR #528, opened 2018-12-12
Improve: remove indented empty lines separating recursive modules

### LINKED ISSUE #514: indented empty lines separating recursive module definition
This is the minimal example I found:
```
module C = struct
  module rec A : sig
    type t
  end =
    A
  
  and B : sig end = B
end
```
note the trailing whitespace

gpetiot:
Should fix #514

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -367,50 +367,6 @@ module Expression : Module_item with type t = expression = struct
     || not (is_simple (i2, c2))
 end
 
-module type Module_getter = sig
-  type t
-
-  val attributes : t -> attributes
-
-  val loc : t -> Location.t
-end
-
-module Make (S : Module_getter) : Module_item with type t = S.t = struct
-  type t = S.t
-
-  let has_doc itm = Option.is_some (fst (doc_atrs (S.attributes itm)))
-
-  let is_simple (itm, c) =
-    Location.width (S.loc itm) <= c.Conf.margin
-    && Location.is_single_line (S.loc itm)
-
-  let break_between cmts (i1, c1) (i2, c2) =
-    Cmts.has_after cmts (S.loc i1)
-    || Cmts.has_before cmts (S.loc i2)
-    || has_doc i1 || has_doc i2
-    || (not (is_simple (i1, c1)))
-    || not (is_simple (i2, c2))
-end
-
-module Module_type_fields = struct
-  type t = module_type
-
-  let attributes x = x.pmty_attributes
-
-  let loc x = x.pmty_loc
-end
-
-module Module_expr_fields = struct
-  type t = module_expr
-
-  let attributes x = x.pmod_attributes
-
-  let loc x = x.pmod_loc
-end
-
-module Module_type = Make (Module_type_fields)
-module Module_expr = Make (Module_expr_fields)
-
 let may_force_break (c : Conf.t) s =
   let contains_internal_newline s =
     match String.index s '\n' with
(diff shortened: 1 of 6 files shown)


gpetiot:
I refactored the code, tell me if it is an overkill or if we can merge.

src/Fmt_ast.ml
```diff
@@ -138,6 +138,35 @@ let update_config_maybe_disabled_block c loc l f =
   maybe_disabled_k c loc l f (fun bdy ->
       {empty with opn= open_vbox 2; bdy; cls= close_box} )
 
+let fmt_recmodule c ctx (items : 'a list)
+    (get_attributes : 'a -> attributes) (get_loc : 'a -> Location.t)
+    (to_ast : 'a -> Ast.t)
```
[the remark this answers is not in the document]
gpetiot: I did this, it looks over-engineered to me though.

gpetiot:
I would like an approval for my last commit before merging.

src/Fmt_ast.ml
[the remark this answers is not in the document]
jberdine: The code in `Fmt_ast` is improved by this last change, I think, but `Ast` itself is indeed quite heavy. Have you tried eliminating these modules such as `Module_expr` and instead define the functions such as `attributes` directly on `Ast.t`, analogous to how e.g. `Ast.dump` is implemented? I suspect that would be much lighter weight, though maybe there is some constraint I'm overlooking.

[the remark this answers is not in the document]
gpetiot: Does it look better now?

src/Fmt_ast.ml
```diff
@@ -3838,10 +3863,11 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
   | Pstr_open open_descr -> fmt_open_description c open_descr
   | Pstr_primitive vd -> fmt_value_description c ctx vd
   | Pstr_recmodule bindings ->
-      hvbox 0
-        (list_fl bindings (fun ~first ~last binding ->
-             fmt_module_binding c ctx ~rec_flag:true ~first binding
-             $ fmt_if (not last) "@,@\n" ))
+      fmt_recmodule c ctx bindings
```
jberdine: How about mapping `(fun x -> Mod x)` over `bindings`, and...

src/Fmt_ast.ml
```diff
@@ -3838,10 +3863,11 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
...
+        (module Ast.Module_binding_fields)
```
jberdine: removing this argument, and...

src/Fmt_ast.ml
```diff
@@ -3838,10 +3863,11 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
...
+        (fun c ctx ~rec_flag ~first b ->
```
jberdine: matching `b` with `Mod` here?

And the similar changes around for the other call site of `fmt_recmodule` above. Doing this would allow to remove `Module_fields_getter` and its implementations I think.

gpetiot: The matching won't be enough to get the original type back, but I simplified the implementation as much as I could.

src/Ast.ml
```diff
@@ -500,11 +500,51 @@ end
...
+  | Typ x -> x.ptyp_attributes
+  | Cty x -> x.pcty_attributes
+  | Pat x -> x.ppat_attributes
+  | Exp x -> x.pexp_attributes
+  | Cl x -> x.pcl_attributes
+  | Mty x -> x.pmty_attributes
+  | Mod x -> x.pmod_attributes
+  | Sig _ -> assert false
+  | Str _ -> assert false
+  | Top -> assert false
```
jberdine: I'm not sure that it's better to crash in these cases instead of returning the empty list. AFAIU e.g. signatures do not currently have attributes in the parsetree, but it would make sense if they did.

src/Ast.ml
```diff
@@ -500,11 +500,51 @@ end
...
+  | Typ x -> x.ptyp_loc
+  | Cty x -> x.pcty_loc
+  | Pat x -> x.ppat_loc
+  | Exp x -> x.pexp_loc
+  | Cl x -> x.pcl_loc
+  | Mty x -> x.pmty_loc
+  | Mod x -> x.pmod_loc
+  | Sig x -> x.psig_loc
+  | Str x -> x.pstr_loc
+  | Top -> assert false
```
jberdine: As above, maybe it would be better to return Location.none instead of crashing.

PR #530, opened 2018-12-14
Improve: option --diff to display the diff without overwriting the file

gpetiot:
Relying on [patdiff](https://github.com/janestreet/patdiff) for now, not sure adding a new dep is a good thing.

Is this what you had in mind @pascutto ?

### DIFF
--- a/ocamlformat-help.txt
+++ b/ocamlformat-help.txt
@@ -206,6 +206,9 @@ OPTIONS
            Aggregate options. Options are specified as a comma-separated list
            of pairs: option=VAL,...,option=VAL.
 
+       --diff
+           Print the diff.
+
        --disable-outside-detected-project
            Do not read .ocamlformat config files outside the current project.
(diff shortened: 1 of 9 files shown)


jberdine:
Could you provide some more context on the use case for this? It has the feel of something that ought to be handled by the shell. In particular, is it different from `ocamlformat test.ml | diff test.ml -`?

gpetiot:
> Could you provide some more context on the use case for this? It has the feel of something that ought to be handled by the shell. In particular, is it different from `ocamlformat test.ml | diff test.ml -`?

The idea was to have exactly this, but with something better like `patdiff`, maybe it's better to not bother modifying the Conf.action type and just have a bash script.

jberdine:
Yes, or send a PR against patdiff to accept stdin. In general I don't think that it is scalable to implement all the workflows within ocamlformat itself. We need to rely on the (user) shell to fit the core functionality into the desired end workflow.

PR #532, opened 2018-12-22
Improve: new default indicate-multiline-delimiters=false

gpetiot:
I think this new behavior is something the community would like.


jberdine:
Maybe I am alone, but looking at the diff with this enabled doesn't make me so happy to be honest. The spaces following `(` that are removed by this option have been relied on when designing other aspects of the formatting. For example, the alignment of tuples is a significant regression I think. I am not sure that all of the changes caused by switching this option are the desired consequences of the original goal. Does anyone have any comments in that regard?

Might it be that people only really are surprised by the spaces before `)`? My impression is that such a change would break much less of the rest of the formatting.

I have not seen any examples showing how the spaces lead to misreading code, or heard any other arguments other than what amounts to a "popularity contest", so it is hard to understand the real issue and motivation in order to address it while minimizing collateral damage. Help?

gpetiot:
So is it better to modify `indicate-multiline-delimiters=false` to only remove the spaces before `)` and leave `(` unchanged?

jberdine:
Perhaps a smaller change, what about adjusting `indicate-multiline-delimiters=false` so that at least tuples and infix ops use the spacing around `(` of `indicate-multiline-delimiters=true` and we can see what it looks like and maybe gather feedback from those who want `indicate-multiline-delimiters=false`?

PR #533, opened 2018-12-24
Preserve shebang

### LINKED ISSUE #523: Shebang triggers syntax error
Code:
```ocaml
#!/usr/bin/env ocaml

(* ... *)
```
Error:
```shell
ocamlformat: ignoring "hello.ml" (syntax error)
File "hello.ml", line 1, characters 0-3:
Error: Syntax error
```
Judging by this StackOverflow post [recommending preprocessing the shebang away with sed](https://stackoverflow.com/a/7616621) these things are not super welcome in OCaml community. Github search [returns a number of examples](https://github.com/search?q=%22%2Fusr%2Fbin%2Fenv%22++language%3AOCaml&type=Code), some of them including the top-level directives like `#load` which trigger the same syntax error. I'm too new to be sure if these examples fall under the TOS listed in the readme _("OCamlFormat requires source code that meets the following conditions...")_ . Preprocessing the thing away with sed would require to stitch it back on again :)

gpetiot:
Should fix #523 

For now, only the first line is tested, maybe something smarter could be done in the future?

### DIFF
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -83,6 +83,7 @@ let parse parse_ast (conf : Conf.t) ic =
   in
   Warnings.parse_options false (W.to_string warnings) ;
   let lexbuf = Lexing.from_channel ic in
+  Lexer.skip_hash_bang lexbuf ;
   Location.init lexbuf !Location.input_name ;
   let warning_printer = !Location.warning_printer in
   let w50 = ref [] in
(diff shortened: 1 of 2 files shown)


src/Translation_unit.ml
```diff
@@ -83,6 +83,7 @@ let parse parse_ast (conf : Conf.t) ic =
   in
   Warnings.parse_options false (W.to_string warnings) ;
   let lexbuf = Lexing.from_channel ic in
+  Lexer.skip_hash_bang lexbuf ;
```
hhugo: We should be able to know how much was skipped looking at the lexbuf. That way, we would be sure we don't drop anything.

gpetiot: It should be addressed by my last commit.

PR #536, opened 2018-12-31
Improve: less linebreaks for break-cases

gpetiot:
To reduce the diff on `core_kernel`, closer to what most users want anyway.


test/passing/break_cases_nested.ml
```diff
@@ -2,8 +2,7 @@
 
 let f x = function
   | C | P (this, test, [is; wide; enough; _to; break], [the; line]) | A | K
-    ->
-      1
+    -> 1
```
jberdine: With the 'nested' option, these breaks need to be preserved. The reasoning behind this is that `break-cases=nested` is intended for use with `indicate-nested-or-patterns=false`, and the breaks are needed to keep from hiding a short body within a sequence of cases that otherwise looks like a nested or-pattern.

Some people will prefer to take that risk, but we need to keep a the safe alternative.

gpetiot: I chose to have a dedicated function for each case, I feel it's easier to maintain than having a generalized function for the 3 cases.

jberdine: I have to look more later, but initially I'm not sure about the split into separate functions. I agree that each one is simpler than the combined one, but there is still quite a bit of duplication of code that handles other things than just break-cases that we will need to keep in sync across the 3 functions.

gpetiot: I factorized the code in a way it still shows where the box structure is different from one case to another.

src/Fmt_ast.ml
```diff
@@ -2854,22 +2840,55 @@ and fmt_cases c ctx cs =
...
+        hovbox 0 (fmt_expression c ?parens:parens_for_exp xrhs)
+        $ fmt_if_k parens_here
+            (fmt_or c.conf.indicate_multiline_delimiters "@ )" "@,)")
+      in
+      match c.conf.break_cases with
+      | `Fit ->
+          let fmt_arrow_close_box =
+            fmt "@;<1 -2>"
+            $ fmt_if_k parens_here (open_hvbox 2)
+            $ fmt "->" $ fmt_if parens_here " ("
```
jberdine: `fmt_arrow_close_box` in this case is not quite the same as before, intended?

test/passing/ocp_indent_options.ml.ref
```diff
@@ -2,7 +2,7 @@ let _ =
   let f x y =
     match x with
     | None -> false
-    | Some loooooooooooooooooooooooooooooooooooooooooooooooooooooooooong -> (
-      match y with Some _ -> true | None -> false )
+    | Some loooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
+    -> ( match y with Some _ -> true | None -> false )
```
jberdine: I think that it's correct to break here (I haven't checked the margin carefully), but the indentation should be 2 spaces more. Seems to be the same as the case from selection.ml.

gpetiot:
The commit 1a149e2 is avoids to format this:
```ocaml
let f = function
  | (`A | `B)[@bar] | `C ->  ()
```
like this:
```ocaml
let f = function
  (| (`A | `B)[@bar]) | `C ->  ()
```
It shouldn't break anything else.

gpetiot:
The purpose of this PR is also to write the formatting in a way it's easier to maintain: the data and the computation are defined in two separate files (for now, only for this case, we can propagate to each and every case later).
New file architecture:
 - Params.ml defines the specific parameters for each option
 - Fmt_ast.ml defines the general formatting, this formatting function is parameterized with the data issued by Params.ml

test/passing/js_source.ml.ref
```diff
@@ -90,8 +90,12 @@ let () =
     (let open M in
     ()) [@foo]];
   [%foo fun [@foo] x -> ()];
-  [%foo function[@foo] x -> ()];
-  [%foo try[@foo] () with _ -> ()];
+  [%foo
+    function[@foo]
+    | x ->  ()];
```
jberdine: extra space after `->`?

test/passing/break_cases_all.ml
```diff
@@ -14,70 +13,71 @@ let f x = function
 let f =
   let g = function
     | H
-      when x y <> k ->
-        2
+      when x y <> k -> 2
```
jberdine: Maybe not for this PR, but is the break before `when` intended?

samoht:
Just a quick remark: `different` doesn't sound very explicit as an option name (e.g. it doesn't really tell what that option does).

gpetiot:
> Just a quick remark: `different` doesn't sound very explicit as an option name (e.g. it doesn't really tell what that option does).

Would `break-cases = group-or` be more explicit?

test/passing/break_cases_all.ml
gpetiot: fixed

gpetiot:
The diff computed by `test_branch.sh` is consistent with the diff on the regression tests.

test/passing/js_source.ml.ref
```diff
@@ -4061,12 +4306,20 @@ let lexpr_ops (ops : ('a, 'a) #ops Lazy.t) =
...
+      | #expr as x -> expr#free x
 
     method subst ~sub =
-      function #lambda as x -> lambda#subst ~sub x | #expr as x -> expr#subst ~sub x
+      function
+      | #lambda as x -> lambda#subst ~sub x
+      | #expr as x -> expr#subst ~sub x
 
-    method eval = function #lambda as x -> lambda#eval x | #expr as x -> expr#eval x
+    method eval =
```
hhugo: `function` could stay on the line before, similar to let bindings.

hhugo: Probably should go in another PR

gpetiot: related to #462

test/passing/break_cases_all.ml
```diff
@@ -4,80 +4,90 @@ let f x = function
...
   match exp.pexp_desc with
   | Pexp_sequence _
   | Pexp_extension
       ( _
       , PStr [{pstr_desc= Pstr_eval ({pexp_desc= Pexp_sequence _}, []); _}]
-      ) ->
-      true
-  | _ ->
-      false
+      ) -> true
```
jberdine: Just to check, is this output where a case body is placed on the same line at the end of a multi-line pattern intended? Technically, it is the same situation as after a nested or-pattern, but maybe not what is intended nevertheless.

test/passing/break_cases_all.ml
```diff
@@ -4,80 +4,90 @@ let f x = function
...
-    ()
-| _ ->
-    ()
+  when fffffffffffffff bbbbbbbbbb yyyyyyyyyy -> ()
+| _ -> ()
 
 let is_sequence exp =
   match exp.pexp_desc with
   | Pexp_sequence _
   | Pexp_extension
```
jberdine: Is `indicate-nested-or-patterns` false for some reason here? If not, this case looks like it is not respecting it.

test/passing/break_cases_all.ml
jberdine: Also, I think a test case is needed where there is a multi-line pattern with a `when` clause, as in that case there needs to be a break before the `when`. There are a number of regressions of this form such as:
```
--- a/infer/src/IR/HilInstr.ml
+++ b/infer/src/IR/HilInstr.ml
@@ -68,8 +68,7 @@ let of_sil ~include_array_indexes ~f_resolve_id (instr : Sil.instr) =
       , Const (Cfun callee_pname)
       , (target_exp, _) :: (Sizeof {typ= cast_typ}, _) :: _
       , loc
-      , _ )
-    when Typ.Procname.equal callee_pname BuiltinDecl.__cast ->
+      , _ ) when Typ.Procname.equal callee_pname BuiltinDecl.__cast ->
       analyze_id_assignment (Var.of_id ret_id) target_exp cast_typ loc
   | Store (lhs_exp, typ, rhs_exp, loc) ->
       let lhs_access_expr =
```
I suspect that the change to avoid breaking before `when` for single-line patterns in `all` mode needs to be more narrowly targeted. But for multi-line patterns, I would argue that in all styles of match cases, there should be a break before`when` clauses on multi-or patterns. Without that, readers are strongly led to associate the `when` with the last pattern, which is wrong and a source of bugs.

jberdine:
Regarding naming, to summarize (my understanding of) the high-level styles, we have:

- fit: just break naturally at the margin

- different: force breaking between top-level cases (ie, not nested or-patterns), otherwise break naturally at the margin

- all: force breaking between all top-level cases as well as nested or-patterns

- nested: like different but also break before each body so indicate-nested-or-patterns is not needed, except allows one-liners

So what about renaming `different` to `toplevel`?

test/passing/break_cases_all.ml
gpetiot: fixed, thanks

test/passing/break_cases_all.ml
gpetiot: My last commit should fix that, running `test_branch` with `break-cases=all` on both looks like the output of this branch is more correct than it is on master.

jberdine: Looking quickly at the diff from `test_branch` with `break-cases=all`, I didn't spot any issues with `when`. But for the diff without setting any options, there are still regressions like the one in the diff above. This is in `nested` mode.

Using `break-cases=toplevel`, I still see situations where a case body follows a multiline pattern without a break after the `->`:
```
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -1263,15 +1250,13 @@ end = struct
       | Pstr_primitive _ | Pstr_type _ | Pstr_typext _ | Pstr_exception _
        |Pstr_module _ | Pstr_recmodule _ | Pstr_modtype _ | Pstr_open _
        |Pstr_class _ | Pstr_class_type _ | Pstr_include _ | Pstr_attribute _
-        ->
-          assert false )
+        -> assert false )
     | Mod {pmod_desc= Pmod_unpack e1} -> (
       match e1 with
       | { pexp_desc=
             Pexp_constraint
               (e, {ptyp_desc= Ptyp_package _; ptyp_attributes= []})
-        ; pexp_attributes= [] } ->
-          assert (e == exp)
+        ; pexp_attributes= [] } -> assert (e == exp)
       | e -> assert (e == exp) )
     | Cl ctx ->
         let rec loop ctx =
```
Is this change intended, and intended to be combined with the case-breaking?

gpetiot: These changes (fitting the rhs expression after the arrow if it fits even if the pattern broke) were intended, I reverted them for `all` and I can revert them for `nested` and `toplevel` as well.

jberdine: My impression is that the choice of whether to tack expressions on to the end of multiline patterns is mostly orthogonal to the style of breaking cases. Also, adding short phrases on the same line after multiline ones is a significant departure from the rest of the formatting. So I'd say that it would be best not to do it in the same PR.

In cases where the only thing at the beginning of the line is "punctuation" like closing parens or `->`, it would be fine. But the implementation of that could be complicated.

jberdine: And can I bump the request for a test case in 
https://github.com/ocaml-ppx/ocamlformat/pull/536#discussion_r252651181 . The situation with `when` seems subtle enough that it will help to have tests for it rather than only seeing such regressions on the full repo test diffs.

src/Fmt_ast.ml
```diff
@@ -2867,54 +2869,31 @@ and fmt_cases c ctx cs =
...
-            | true, false, false -> fmt "@;<1 2>"
-            | true, false, true -> fmt " (@;<1 2>"
-            | true, true, false -> fmt " "
-            | true, true, true -> fmt " (@;<1 4>" )
-          $ hovbox 0
-              ( hovbox 0 (fmt_expression c ?parens:parens_for_exp xrhs)
-              $ fmt_or_k c.conf.indicate_multiline_delimiters
-                  (fmt_if parens_here "@ )")
-                  (fmt_if parens_here "@,)") ) ) )
+      let guard g =
```
jberdine: Don't you think it would be clearer to inline this? There is only one use, right?

src/Params.mli
```diff
@@ -0,0 +1,25 @@
...
+ *                            OCamlFormat                             *
+ *                                                                    *
+ *  Copyright (c) 2019-present, Facebook, Inc.  All rights reserved.  *
+ *                                                                    *
+ *  This source code is licensed under the MIT license found in the   *
+ *  LICENSE file in the root directory of this source tree.           *
+ *                                                                    *
+ **********************************************************************)
+
+module Cases : sig
```
jberdine: I'm still unclear what the reasoning is for this module. Wouldn't it be simpler to have `type cases` with `val get_cases`?

gpetiot:
> Thanks, looks good! A couple minor comments, but not need for another round of review.
> 
> Also, a question that probably should have occurred to me earlier: Do we need the `all` case? I'm not aware of a use case, since as far as I know it is too sparse for anyone. If it simplifies anything, might be worth removing.

Yes removing `all` will simplify the code but I'm afraid someone could be relying on this option. What should be do before removing an option or changing a default value? Making a survey on the discuss or on slack?

jberdine:
Anil made a post on discuss a while ago asking for feedback. I think that a comment in that thread would be a good idea.

To be diligent, despite being pre-1.0, we could release 0.9 with `all`, but emit a warning whenever the config handling code sees it set. The warning could announce plans to remove it, and ask users to send feedback if they would like it to remain.

jberdine:
But I think that everyone who wanted `all` would prefer `toplevel`.

gpetiot:
> But I think that everyone who wanted `all` would prefer `toplevel`.

Maybe we should just rename `toplevel` to `all` then, so we don't break the existing configurations.

jberdine:
Good point. We will have a config-breaking release later, we can open an issue to remember to rename `all` then.

PR #537, opened 2018-12-31
Improve: remove unnecessary parens when open module

### LINKED ISSUE #534: Unnecessary parens around Module.{ record }
For example, here's a diff that ocamlformat produces:

```diff
-  let Cstruct.{ buffer = bigstring; off = offset; len = length } = buffer in
+  let Cstruct.({ buffer = bigstring; off = offset; len = length }) = buffer in
```

gpetiot:
Should fix #534 
No regression with test_branch.sh

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1161,10 +1161,19 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
            (fmt "exception@ " $ fmt_pattern c (sub_pat ~ctx pat)))
   | Ppat_extension ext -> hvbox 2 (fmt_extension c ctx "%" ext)
   | Ppat_open (lid, pat) ->
+      let opn, cls =
+        let can_skip_parens =
+          match pat.ppat_desc with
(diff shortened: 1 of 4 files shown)


src/Fmt_ast.ml
```diff
@@ -1161,10 +1161,19 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
            (fmt "exception@ " $ fmt_pattern c (sub_pat ~ctx pat)))
   | Ppat_extension ext -> hvbox 2 (fmt_extension c ctx "%" ext)
   | Ppat_open (lid, pat) ->
+      let opn, cls =
+        let can_skip_parens =
+          match pat.ppat_desc with
+          | Ppat_array _ | Ppat_record _ -> true
+          | _ -> (
```
jberdine: Shouldn't there be a case for tuples, which consults parens_tuple_patterns? Consider cases such as:
```
[@@@ocamlformat "parens-tuple-patterns=always"]

let M.((x, y)) = ()
```

gpetiot: Indeed it should

Issue #521, opened 2018-11-24
Preserve grouping of module items


copy:
Consider this cut down human-formatted code:

```ocaml
let cmos_rtc_seconds = 0x00
let cmos_rtc_seconds_alarm = 0x01
let cmos_rtc_minutes = 0x02

let log_other = 0x000001
let log_cpu = 0x000002
let log_fpu = 0x000004

let cr0_pe = 1 lsl 0
let cr0_mp = 1 lsl 1
let cr0_em = 1 lsl 2
```

Depending on the configuration, ocamlformat either drops the newlines or adds newlines between every `let` binding, while I'd like to preserve the grouping. I don't consider this a failure of the formatter, as I don't expect the formatter to understand how to group top-level bindings by their semantics, but I do find the human-formatted version much more readable.

jberdine:
Note that if you use `let`-`and` groups of binding you can get the spacing I think you want:
```
let cmos_rtc_seconds = 0x00
and cmos_rtc_seconds_alarm = 0x01
and cmos_rtc_minutes = 0x02

let log_other = 0x000001
and log_cpu = 0x000002
and log_fpu = 0x000004

let cr0_pe = 1 lsl 0
and cr0_mp = 1 lsl 1
and cr0_em = 1 lsl 2
```

copy:
That's very neat and probably works as a work-around for me. It does feel a little bit like misuse of `and` though, would you recommend doing this in real code?

jberdine:
Opinions will vary I expect, but `and` does semantically group bindings together, and the formatting reflects that to some extent. I wouldn't give anyone a hard time during code review over using `and` here. Adding comments would also spread the groups out, which might be more helpful.

gpetiot:
I agree with Josh here `and` can be used for that, but if you have the time to contribute to ocamlformat, you can have a look at `Ast.break_between` it will be easy to add a new option that groups the bindings if they have the same prefix.

gpetiot:
Closed with #538

PR #538, opened 2018-12-31
Improve: module-item-spacing=preserve

gpetiot:
Should address #521


jberdine:
I am really not sure about this one. This change is quite a special case in the implementation, looking at the names in formatted code, and changing behavior based on them. It is very unclear to me that we want to go down that route, and if so, there should be a more compelling use case I think. We could find some suboptimal formatting in any code base, and change ocamlformat to do a bit better in that case, but this is not a tenable path forward over time. This change feels essentially indistinguishable from an instance of that approach. Maybe I have misunderstood?

For the original issue, wouldn't a more widely useful solution be to implement a way to detect and preserve groups of one-liners?

src/Ast.ml
```diff
@@ -273,6 +273,24 @@ module Structure_item : Module_item with type t = structure_item = struct
...
+        , (Pstr_module _ | Pstr_recmodule _ | Pstr_open _ | Pstr_include _)
+        )
+       |Pstr_modtype _, Pstr_modtype _
+       |Pstr_class _, Pstr_class _
+       |Pstr_class_type _, Pstr_class_type _
+       |Pstr_attribute _, Pstr_attribute _
+       |Pstr_extension _, Pstr_extension _ ->
+          true
+      | Pstr_value (_, [{pvb_loc= x}]), Pstr_value (_, [{pvb_loc= y}]) ->
+          not (Source.empty_line_between x y)
```
jberdine: Couldn't you refactor this to avoid duplicating these many cases by combining the Compact and Preserve cases, and adding a check of `module_item_spacing` in the `Pstr_value` case? Maybe use `space_between` defined below to simplify?

src/Source.ml
```diff
@@ -15,6 +15,9 @@ type t = string
 
 let create s = s
 
+let empty_line_between l1 l2 =
+  Location.(l2.loc_end.pos_lnum - l1.loc_start.pos_lnum > 1)
```
jberdine: This doesn't check that the line between is *empty*.

ocamlformat-help.txt
```diff
@@ -133,11 +133,12 @@ OPTIONS (CODE FORMATTING STYLE)
...
-       --module-item-spacing={sparse|compact}
+       --module-item-spacing={sparse|preserve|compact}
            Spacing between items of structures and signatures. sparse will
-           always break a line between two items. compact will not leave open
-           lines between one-liners of similar sorts. The default value is
-           sparse.
+           always break a line between two items. preserve preserves the
+           spacing between items of structures and signatures. compact will
+           not leave open lines between one-liners of similar sorts. The
+           default value is sparse.
```
jberdine: I'm not sure if the 'preserve' terminology is quite right. The implementation of preserve is like compact, but will space things out if they are in the input. But it will add spaces between items as compact would. I expect that people reading 'preserve' will be surprised by that. I'm not thinking of an alternative term at the moment though. Perhaps it's enough to just expand the documentation.

gpetiot:
I kept the `preserve` value for the option but improved the documentation, factorized and checked that the line in-between is indeed empty.

src/Conf.ml
```diff
@@ -836,6 +836,10 @@ module Formatting = struct
       [ ( "sparse"
         , `Sparse
         , "$(b,sparse) will always break a line between two items." )
+      ; ( "preserve"
+        , `Preserve
+        , "$(b,preserve) will not leave open lines between of-liners of \
```
jberdine:
```suggestion
        , "$(b,preserve) will not leave open lines between one-liners of \
```

src/Source.ml
```diff
@@ -23,6 +23,11 @@ let string_between t (l1 : Location.t) (l2 : Location.t) =
     (* can happen e.g. if comment is within a parenthesized expression *)
     None
 
+let empty_line_between t l1 l2 =
+  let is_empty s = String.strip s |> String.is_empty in
```
jberdine: Perhaps premature optimization, but I'd be tempted to avoid copying the string by using `String.lfindi` with an `~f` that checks for non-whitespace chars.

src/Source.ml
```diff
@@ -23,6 +23,11 @@ let string_between t (l1 : Location.t) (l2 : Location.t) =
...
+  Location.(l2.loc_end.pos_lnum - l1.loc_start.pos_lnum > 1)
+  && Option.value_map ~f:is_empty ~default:true (string_between t l1 l2)
```
jberdine: Could use `Option.for_all` here.

test/passing/module_item_spacing.ml
```diff
@@ -108,3 +108,20 @@ let x = 1
 (* floating *)
 
 let y = 2
+
+;;
+[@@@ocamlformat "module-item-spacing=preserve"]
```
jberdine: It would probably be useful to duplicate the code below and check all 3 values.

src/Ast.ml
```diff
@@ -356,9 +360,16 @@ module Expression : Module_item with type t = expression = struct
     Poly.(c.Conf.module_item_spacing = `Compact)
     && Location.is_single_line i.pexp_loc c.Conf.margin
 
-  let break_between cmts (i1, c1) (i2, c2) =
+  let space_between s (i1, c1) (i2, c2) =
```
jberdine: Maybe `vspace_between` would be clearer?

gpetiot:
> ```ocaml
>  module C : sig
>    type config = t
> -
>    type 'a t
> -
>    type parsed_from = [`File of Fpath.t * int | `Attribute]
> -
>    type updated_from = [`Env | `Commandline | `Parsed of parsed_from]
> 
>    type 'a option_decl =
> ```
> 
> Since the current default is sparse, there are open lines between these items in the input, but with preserve they are being removed. AFAIU this isn't intended, right?

The new specific processing of `module-item-spacing=preserve` should prevent this issue.

jberdine:
Am I missing something or does enabling this lead to unexpected removal of open lines such as:
```
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -1890,7 +1890,6 @@ end = struct
       in
       mem_cls_cl cls cl
       || Hashtbl.Poly.find_or_add memo (cls, cl) ~default:exposed_
-
   and mark_parenzed_inner_nested_match exp =
     let exposed_ () =
       let continue subexp =
```

gpetiot:
> Am I missing something or does enabling this lead to unexpected removal of open lines such as: [...]

Fixed with my last commit, but the interaction with the `let-binding-spacing` is unclear and creates weird behaviors

gpetiot:
@jberdine is this good enough to be merged for 0.9 soon or do you want to postpone this later?

src/Source.ml
```diff
@@ -41,6 +41,12 @@ let string_between t (l1 : Location.t) (l2 : Location.t) =
   then None
   else Some (String.sub t ~pos ~len)
 
+let empty_line_between t l1 l2 =
+  let non_whitespace _i c = not (Char.is_whitespace c) in
```
jberdine: nit: FWIW I tend to avoid names like `_i` since the code can still cheat and use it, and by itself `i` isn't so helpful for documentation. But as you choose.

src/Source.ml
```diff
@@ -41,6 +41,12 @@ let string_between t (l1 : Location.t) (l2 : Location.t) =
...
+  let is_empty s = String.lfindi s ~f:non_whitespace |> Option.is_none in
+  Location.(l2.loc_end.pos_lnum - l1.loc_start.pos_lnum > 1)
```
jberdine:
```suggestion
  Location.(l2.loc_end.pos_lnum - l1.loc_start.pos_lnum) > 1
```

gpetiot:
> This is looking good now, thanks for all the iteration!
> 
> A few comments inline, and a test request, but this doesn't need another round of review.
> 
> Does your previous comment
> 
> > the interaction with the let-binding-spacing is unclear and creates weird behaviors
> 
> still apply? If so, it would be good to add a test or two that demonstrate this, and it'd be good to open an issue to track it.

Yes it still applies, I added a test and created issue #704.

Issue #508, opened 2018-11-15
Choose between begin/end and parens


gpetiot:
Cf [Guidelines](https://ocaml.org/learn/tutorials/guidelines.html)

There could be a different option for each kind of block:
- conditional branches
- match or try
- ...

For each option the following values: `preserve`/`begin-end`/`parens`

gpetiot:
Duplicate of #491

PR #539, opened 2019-01-02
[WIP] Improve: option 'exp-grouping'

gpetiot:
Should eventually address #508


jberdine:
I must admit that I'm hesitant about such a change that increases the code complexity and opens the door for many corner case bugs (due to e.g. the parser not needing surrounding whitespace to identify parens, while `begin` and `end` will need surrounding space some times, but just *when* could be involved). That seems like something to hold off doing until people are complaining loudly. Are we sure we want to do this now?

gpetiot:
> I must admit that I'm hesitant about such a change that increases the code complexity and opens the door for many corner case bugs (due to e.g. the parser not needing surrounding whitespace to identify parens, while `begin` and `end` will need surrounding space some times, but just _when_ could be involved). That seems like something to hold off doing until people are complaining loudly. Are we sure we want to do this now?

I agree that this PR increases the code complexity and we don't get anything in return. I'm closing this PR until people start complaining.

PR #541, opened 2019-01-10
[WIP] Improve: use metaquot

gpetiot:
in progress

- downgrading `Migrate_ast` to work with `Ast_406` instead of `Ast_407` because there is no `metaquot_407`
- patterns can't (always) be rewritten with `metaquot` if they contain `as`
- I am simplifying several code chunks as I encounter them

### DIFF
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -31,6 +31,7 @@ depends: [
   "fpath"
   "ocaml-migrate-parsetree" {>= "1.0.10"}
   "octavius"
+  "ppx_tools_versioned"
   "stdio"
 ]
 synopsis: "Auto-formatter for OCaml code"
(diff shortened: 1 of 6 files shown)


jberdine:
Can you discuss the motivation behind this?

I'm uncomfortable with making a change that potentially makes it much harder to upgrade or test newer compilers. For instance, in the current code base, testing with the trunk compiler is pretty easy by using the raw compiler-libs modules instead of the wrappers in migrate-parsetree. This has usually been doable by commenting out a few definitions. With metaquot it seems that it would be more involved, or basically impossible until metaquot is updated.

It is perhaps also a personal point, but I do a lot of searching by constructor name, and this change would defeat that.

gpetiot:
Alright I get your point, I will close this PR and open a new one containing only the code simplifications  unrelated to metaquot.

PR #546, opened 2019-01-14
Improve: module item spacing in sparse mode

jberdine:
- refactor the margin check into Location.is_single_line
- revise some is_simple checks from true to requiring is_single_line


jberdine:
Noticed while reviewing #538.

Issue #522, opened 2018-11-24
Long indentation on comment after pattern match body


copy:
This example looks unnatural to me:

```ocaml
;;
match x with XXXXXXXXXXXXXXXXXXXXXXXXXX -> (* *)
                                           x
```

Related: https://github.com/ocaml-ppx/ocamlformat/issues/341#issuecomment-423521961

jberdine:
Yes, this is suboptimal. In this case the comment was determined to be associated with the following `x`.  If that is logically the right thing to do (it would be useful to know if that is true in you real example), then there should probably be a line break after the `->`. There could be a lot of undesirable collateral damage of such a change done simply though. It will have to be checked.

copy:
> it would be useful to know if that is true in you real example

Yes, in my real code the comment is attached to the following `x` and a line break after the `->` would have been my choice.

jberdine:
Should be fixed now, please re-open if not.

PR #339, opened 2018-08-19
Improve: comments attachment

hhugo:

Issue #306, opened 2018-08-11
Suboptimal comment placement


jberdine:
Cases where the current comment attachment heuristic does not agree with expectation:
- [ ] `let f (* should be after f, but is before x *) ~x = e`
- [ ] `let f ~l:(* should be after f, but is before x *) x = e`
- [ ] `let f ~(x : t (* should be after t, but is before body *)) = e`

Cases where the attachment seems right, but formatting is suboptimal:
- [ ] `let f ~((* should be before x, but is before constraint *) x : t) = e`

Cases that seem to be due to the OCaml frontend producing unexpected `Location`s:
- [ ] `let g ?y:((* 1 *) D y) = e` while `let f ~x:((* 0 *) C x) = d` is ok, `Location` of pattern seems to differ significantly with `?` vs `~`

jberdine:
Note that one potential source of difficulty with comment placement around labeled arguments is that in the parsetree there is no `Location.t` for the label
```
    | Pexp_fun of arg_label * expression option * pattern * expression
```
where
```
  type arg_label =
      Nolabel
    | Labelled of string (*  label:T -> ... *)
    | Optional of string (* ?label:T -> ... *)
```
It might be helpful to consider the token stream when locating comments, so that in e.g. `f (* comment *) ~x` the heuristic can attach the comment after `f` instead of before `x` due to the presence of the token for `~` between the comment and `x`.

hhugo:
Cross referencing https://github.com/ocaml-ppx/ocamlformat/pull/246

jberdine:
Related to #198

PR #548, opened 2019-01-17
Fix: comments attachment

### LINKED ISSUE #543: Unexpected pyramid of increasing indentation for commented arguments
Given this code:

```ocaml
let emit_wrapper_function =
  Hhas_function.make function_attributes name body
    (Hhas_pos.pos_to_span ast_fun.Ast.f_span)
    false (* is_async *)
    false (* is_generator *)
    false (* is_pair_generator *)
    hoisted true (* no_injection *)
    true (* inout_wrapper *)
    is_interceptable false
    (* is_memoize_impl *)
    Rx.NonRx false
```

`ocamlformat` produces:

```ocaml
$ ocamlformat /tmp/pyramid.ml 
let emit_wrapper_function =
  Hhas_function.make function_attributes name body
    (Hhas_pos.pos_to_span ast_fun.Ast.f_span)
    false (* is_async *)
          false (* is_generator *)
                false (* is_pair_generator *)
                      hoisted true (* no_injection *)
                                   true (* inout_wrapper *)
                                        is_interceptable false
    (* is_memoize_impl *)
    Rx.NonRx false
```

gpetiot:
Improves #339 (didn't manage to push commits to this one), also seems to fix #543


gpetiot:
> The output changes due to this PR are improvements in the large majority of cases. There are cases where comments are converted from ending a line to not ending it. Also, the placement heuristic makes more use of source line numbers, and hence line breaks. So the formatting code does not preserve the information about the input that is used to determine the output, which is dangerous. While this makes me hesitate, I am not thinking of a way to change the implementation so that this information is preserved. So I don't have any particular suggestion at the moment, but want to think on it some more.

What are the cases that you absolutely want to preserve?

jberdine:
I wouldn't state it so strongly as "absolutely want to preserve", but an example from the diff is:
```
--- a/infer/src/base/CommandLineOption.ml
+++ b/infer/src/base/CommandLineOption.ml
@@ -1000,8 +1000,8 @@ let wrap_line indent_string wrap_length line0 =
     let word_length =
       let len = String.length word in
       if String.is_prefix ~prefix:"$(b," word || String.is_prefix ~prefix:"$(i," word then
-        len - 4 (* length of formatting tag prefix *)
-        - 1 (* APPROXIMATION: closing parenthesis that will come after the word, or maybe later *)
+        len - 4 (* length of formatting tag prefix *) - 1
+        (* APPROXIMATION: closing parenthesis that will come after the word, or maybe later *)
       else len
     in
     let new_length = line_length + String.length word_sep_str + word_length in
```
The way that `- 1` moves to the end of a line after a comment is suboptimal.

I am trying a couple ideas. Don't hold your breath, but give me a little time.

jberdine:
What I tried is not going work. The basic idea was to in certain cases such as application arguments, wrap an exp and its associated comments in a box. The intent is to keep an exp from e.g. moving to the end of the preceding line without taking its following comment with it. This does not work without significant changes due to the current comment placement. I suspect that the basic approach there needs to be revised. Consider:
```ocaml
let f x =
  if xxxxxxxxxxxxxxxxxxxxxxxxxxxx then
    x - 1 (* comment attached to [x - 1] *)
    - 2 (* comment attached to entire if-then-else exp that should be attached to just [2] *)
  ;
  ()
```
Note that the first comment gets attached to `x - 1`, which I think is reasonably expected. On the other hand, the second comment gets attached to the entire if-then-else expression, which is probably not what the author intended. This happens since the smallest expression location containing the location of the second comment is the first argument of the `;`.

Note that the parser has a number of choices of which expression's location in which to include the second comment's location: `2`, `x - 1 - 2`, `if ... 2`. These choices are the same as those ocamlformat has for where to place the comment. The current comment placement approach is very dependent on / sensitive to this location choice. This is heading into the direction of considering these placement questions in the parser itself, either in order to choose locations enabling robust placement later by ocamlformat, or by e.g. adding the comments as attributes to the parsetree instead of just producing a separate list of them.

Cc @let-def for info re the last paragraph on sensitivity to the parser's choice of locations.

gpetiot:
I will try to preserve cases like this one before merging this PR.

gpetiot:
I tried a few things without success so I think I will keep it like this.

src/Cmts.ml
```diff
@@ -256,9 +256,11 @@ let is_adjacent t (l1 : Location.t) (l2 : Location.t) =
...
       | _ -> false )
 
-(** Heuristic to choose between placing a comment after the previous loc or
-    before the next loc. Places comment after prev loc only if they are
-    "adjacent", and the comment and next loc are not "adjacent". *)
+(** Heuristic to choose between placing a comment after the previous
+    location or before the next one. Places comment after the previous
+    location only if they are "adjacent", and the comment and the next
+    location are not "adjacent" or the comment and the previous location are
+    on the same line (different from the next location) *)
```
gpetiot: @jberdine you said on the parent PR that the documentation needed to be updated, I personally find this comment too long and confusing, are you sure there is a smart way to document this without paraphrasing the code? (can we just remove it?)

jberdine: There is the point that how can people expect to understand what the formatter does if we can't describe the heuristic used. So I think that there is definite value in the description, though probably more in user-facing documentation than in code documentation.

That said, if the best explanation is unusable for user-facing documentation and amounts to paraphrasing the code, then we might as well remove it (except the first sentence).

jberdine:
This may resolve #522

jberdine:
The problems in #306 should be rechecked after this is merged

PR #549, opened 2019-01-17
Fix: missing break before `; _`

jberdine:


jberdine:
Noticed when investigating margin questions.

PR #550, opened 2019-01-18
Improve: calculate length of comment strings using UTF8

jberdine:
When there are UTF8 sequences in comments, they are commonly displayed
using one column per decoded code point. This patch changes the string
length calculation from using the number of bytes in the encoding to
the number of decoded code points.


cfcs:
The length of codepoints is not the number of printed characters though, so this still wrong, albeit less wrong. :-/

I think the correct approach here would be to use `uuseg` to count grapheme clusters.

PR #552, opened 2019-01-18
Revise formatting of compact single case matches

jberdine:
The reasoning for the compact single case match only really applies for `try`, not for `match`, so specialize. Also, the current compact formatting can be improved wrt readability.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -305,7 +305,8 @@ end = struct
     let docs = section_name section in
     let term = Arg.(value & flag & info names_for_cmdline ~doc ~docs) in
     let parse s =
-      try Ok (Bool.of_string s) with _ ->
+      try Ok (Bool.of_string s)
+      with _ ->
         Error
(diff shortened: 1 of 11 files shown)


src/Conf.ml
```diff
@@ -305,7 +305,8 @@ end = struct
     let docs = section_name section in
     let term = Arg.(value & flag & info names_for_cmdline ~doc ~docs) in
     let parse s =
-      try Ok (Bool.of_string s) with _ ->
+      try Ok (Bool.of_string s)
+      with _ ->
```
jberdine: It helps quick understanding of the control flow to have `with` on the left, and doesn't cost indentation of the handler body.

jberdine:
@gpetiot this might avoid troubles with indentation in #536

src/Conf.ml
gpetiot: I agree it looks better

PR #553, opened 2019-01-18
Fix typo in help text

Wilfred:


gpetiot:
Thanks!

jberdine:
Note that ocamlformat-help.txt is generated, so to fix this the source documentation in Conf.ml needs to be changed. I've just done that and force-pushed.

Issue #454, opened 2018-10-08
Remove windows hack for running scripts in build actions


jberdine:
Revert #448 once dune is able to hide window's inability to run scripts.

gpetiot:
Fixed by #664 ?

jberdine:
It's not really fixed. There is still the `%%{dep` stuff needed in the `dune` file since we're not allowed to make `gen_version.ml` a shebang script either.

Julow:
@jberdine The `%{dep:...}` syntax is to tell dune what are the dependencies. The real "hack" is calling the interpreter explicitly: `ocaml %{dep:...}`.
I don't think that is an uncommon way to call a script. What do you think?

PR #559, opened 2019-01-22
Fix call to gen_version script in dune file

Julow:
`sh` is the legacy shell on most platforms and this script use bash features

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -12,7 +12,7 @@ Jbuild_plugin.V1.send
 (rule
  (targets Version.ml)
  (deps (universe))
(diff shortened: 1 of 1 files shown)


jberdine:
Because windows is feeble, apparently the call to the shell is needed. See https://github.com/ocaml-ppx/ocamlformat/issues/454. Also see https://github.com/ocaml-ppx/ocamlformat/pull/471 which changes this script to avoid needing bash, I think.

Julow:
I updated this PR to not rely on the shebang (`#!`) and invoke `bash` directly.
May be closed in favor of https://github.com/ocaml-ppx/ocamlformat/pull/471

PR #560, opened 2019-01-22
Switch the default value of --doc-comments from after to before

### LINKED ISSUE #556: Switch the default value of --doc-comments from after to before
To avoid confusing type comments with constructor comments, type comments are always placed before the type definition:

```ocaml
(** Your comment on [t] here *)
type 'a t = Some of 'a | None (** Your comment on [None] here *)
```

but the default value for doc comments on values is after:

```ocaml
let f x y = x + y
(** Your comment on [f] here *)
```

It would be much nicer for comments to have a consistent position across the different structure item types.

Since the type comment can't be moved, I would like to propose changing the default to `--doc-comments=before`.

mrmr1993:
This fixes #556.


jberdine:
Closing this as we'll change profiles and defaults of many options together in other PRs.

Issue #506, opened 2018-11-15
Position of 'in' keyword for multiline bindings


gpetiot:
Currently if a binding can't fit on one line, the `in` keyword is written alone on a line:
```
    let pair (Arg (trm_x, set_x)) (Arg (trm_y, set_y)) =
      let trm_xy = Term.(const (fun a b -> (a, b)) $ trm_x $ trm_y) in
      let set_xy (a, b) = set_x a ; set_y b in
      Arg (trm_xy, set_xy)
    in
    ...
```

Another option would be to let it at the end of the body:
```
    let pair (Arg (trm_x, set_x)) (Arg (trm_y, set_y)) =
      let trm_xy = Term.(const (fun a b -> (a, b)) $ trm_x $ trm_y) in
      let set_xy (a, b) = set_x a ; set_y b in
      Arg (trm_xy, set_xy) in
    ...
```

jberdine:
I implemented this very early on, but it was problematic because it sometimes caused confusion. In situations where the body of the `let` is long(er), and the indentation of the body varies, especially if it is deep near the end, it can be very difficult to quickly visually identify the point where the indentation reduces to the level of the `let`. People would end up putting the cursor on the `let` and scrolling down to find the end. Placing the `in` on the left resolved these problems. This is a case where formatting manually one would be able to choose which position to use for the `in` based on human understanding, but the formatter cannot be so clever. It has not seemed worth it or a good idea to devise a heuristic to decide when to use one form vs the other.

An option could be added, but it seems superfluous and I'd like to see a compelling argument before anyone spent time implementing it, and adding code complexity to maintain.

PR #562, opened 2019-01-23
Adding option for lonely in-s

raphael-proust:
This PR adds an option (lonely-in) to place the `in` delimiters of `let`-`in` bindings on the last line as the expression that is bound – instead of a line of its own.

E.g.,
```
let v =
  List.fold_left
    (fun acc x -> …)
    empty
    xs in
```

Note that the compact profile sets the lonely-in option for more compactness. I'm not sure what the policy is regarding PRs changing existing profiles. Is it ok?

PS: Most of the work was done by @mathiasbourgoin but I've probably messed it up when tidying it up for the PR.


src/Conf.ml
```diff
@@ -50,7 +50,8 @@ type t =
   ; space_around_collection_expressions: bool
   ; type_decl: [`Compact | `Sparse]
   ; wrap_comments: bool
-  ; wrap_fun_args: bool }
+  ; wrap_fun_args: bool
+  ; end_of_line_in: bool }
```
gpetiot: We would like to keep the options alphabetically sorted in Conf.ml and Conf.mli (in type definitions, profiles, etc.)

gpetiot:
Also referencing #506, as [Josh's comment](https://github.com/ocaml-ppx/ocamlformat/issues/506#issuecomment-456579262) should be taken into account.

raphael-proust:
I'll push more commits to fix the ordering, fix the formatting, and fix the existing tests.

I'm not clear on how to use the tests for an option like this one. Can you point to some more info or give a brief overview?

As for #506:  
The whole source code of https://gitlab.com/tezos/tezos/ is formatted (by hand currently) with end-of-line `in`. When we review MRs by new contributors, we have to tell them about that convention; we want to automate that part of the review.
Moreover, end-of-line `in`s are pretty much equivalent to end-of-line `)`/`]`/`}` – but for binding expressions specifically. It can lead to the same kinds of layout. When it makes impairs reading (because of long expressions with varying indentation), similar techniques as with `)` can be used to ease reading: introduce names for sub-expressions, set editor to highlight matching bind keyword, etc.

jberdine:
Thanks for your PR! One admin point is that generally it's best not to change profiles or defaults at the same time as adding an option. It is much easier to test for unintentional regressions when the option is disabled that way, and also makes the upgrade story easier for users.

src/Fmt_ast.ml
```diff
@@ -1975,7 +1975,10 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?(indent_wrap = 0) ?ext
                       ?ext:(if first then ext else None)
                       ctx binding
                       ~in_:(fun indent ->
-                        fmt_if_k last (break 1 (-indent) $ fmt "in") )
+                          fmt_if_k last
+                            (fmt_or_k c.conf.end_of_line_in
+                               (fmt " in")
```
jberdine: One thing that should be explicitly tested is what happens if just the ` in` would exceed the margin. There is a question of whether it is better to break before the `in`, as if this option was disabled, or to break the line before the `in`. Analogous choices elsewhere in the formatter opt for allowing the `in` to end up on its own line. Usually that is much simpler to implement, and leads to less dramatic formatting changes when code is edited.

test/passing/long_let_in.ml
```diff
@@ -0,0 +1,36 @@
+let () =
+  let short = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let fooo = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let baaar = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let long = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
```
jberdine: Try a case like
```
  let violate_margin = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1111 + 1 in
```
to see that `in` gets pushed past the margin.

test/passing/long_let_in.ml
```diff
@@ -0,0 +1,36 @@
+let () =
```
jberdine: It isn't discoverable, sorry, but we are trying to move away from using the `.opts` files unless necessary. So in cases like this it is easier to work with to add an attribute at the toplevel of the test file. It can also be helpful to add another attribute after the tests with the option set the other way and duplicate (some of) the tests to show the effect.

```suggestion
[@@@ocamlformat "end-of-line-in=true"]

let () =
```

src/Conf.ml
```diff
@@ -1234,7 +1246,8 @@ let default_profile =
       C.default Formatting.space_around_collection_expressions
   ; type_decl= C.default Formatting.type_decl
   ; wrap_comments= C.default Formatting.wrap_comments
-  ; wrap_fun_args= C.default Formatting.wrap_fun_args }
+  ; wrap_fun_args= C.default Formatting.wrap_fun_args
+  ; end_of_line_in= C.default Formatting.end_of_line_in }
```
jberdine: same ordering comment as for the type definition

src/Conf.ml
```diff
@@ -985,6 +986,17 @@ module Formatting = struct
...
       (fun conf wrap_fun_args -> {conf with wrap_fun_args})
       (fun conf -> conf.wrap_fun_args)
+
+  let end_of_line_in =
+    let default = false in
+    let doc =
+      "Place $(i,in)s (from $(i,let)-$(i,in)-s) at the end of the last \
+       line of the expression it delimits instead of on a line of its own."
+    in
+    let names = ["end-of-line-in"] in
```
jberdine: Considering symmetry with the other option names, I think it would be better to use e.g. `no-break-before-in`. What do others think?

test/passing/long_let_in.ml
```diff
@@ -0,0 +1,42 @@
...
+  let fooo = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let baaar = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let long = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let longer = 11 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let longeerer = 11 + 11 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let longest = 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 + 11 in
+  let violate_margin = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1111 + 1 in
+  ignore short; ignore set; ignore seat; ignore long; ignore longer;
+  ignore longerer; ignore longest ;
+  ignore violate_margin
```
gpetiot: Can we remove all these ignore in the tests and put `()` instead? It is a lot of noise in the tests.

src/Conf.ml
raphael-proust: I prefer when options are written as positive in the source code and configuration files (`break-before-in=false` is ok `no-break-before-in=true` is not).

I don't particularly mind about the default (`true` or `false`, leading to `--no-X` or `--X`) for command-line arguments.

src/Conf.ml
raphael-proust: Done

test/passing/long_let_in.ml
raphael-proust: Done

test/passing/long_let_in.ml
raphael-proust: Added the test. I'll fix it so that the output looks within the expected range of acceptable.

I made a quick attempt using `fits_breaks` but I thing I used it wrong.

test/passing/long_let_in.ml
raphael-proust: Done

src/Conf.ml
jberdine: The current behavior should be the default (for at least one release). Avoiding `no-` prefixes in the source would be good. It would be good to check that the docstring is phrased so that it makes sense from the perspective of changing from the default when reading the output of --help.

src/Fmt_ast.ml
raphael-proust: This is implemented and tested in `long_let_in`. Some of it is definitely implementation choices (I don't want to add too many different options (break before in: always? when there is no space? never?)

I tried to have some sense of consistency:

In the most common case (when it's not just `in` falling in the margin), the general visual impression of the code is a series of vertically aligned `let`s. In other words, the way the visual presentation relates to the AST is that a certain horizontal spacing is reserved for `let` keywords of a series of let-bindings and for the final expression in which all the bindings are available.

In the rarer occasion of the `in` keyword falling alone in the margin, I attempted to keep that general impression/relationship. This is achieved by placing the `in` on its own line with the same level of indentation as the expression bound by the `let`.

```
let x =
  …
  in
let y =
  …
  in
x + y
```

This is the most consistent approach I found. It keeps the same visual clues for the same types of AST.

src/Conf.ml
```diff
@@ -642,6 +643,17 @@ module Formatting = struct
       (fun conf x -> {conf with doc_comments= x})
       (fun conf -> conf.doc_comments)
 
+  let end_of_line_in =
```
jberdine: Something like `break-before-in` would still be more consistent with the existing option naming scheme.

gpetiot:
Is this PR still under development? What is missing (besides a rebase) ?

src/Conf.ml
raphael-proust: I renamed the option and rephrased the docstring. The docstring is written for the positive option because of the option name (`break-before-in`). However, because it defaults to `true` the option is presented in the help/man as `--no-break-before-in`.

raphael-proust:
The syntax errors seem to be casued by `let-and-in` constructs. They are formatted as `let-in-and-in` with the current code. I'll fix this now.

raphael-proust:
The error was obvious in hindsight and the fix was simple. I think I addressed all the comments and made all the requested changes. Let me know if there are more changes to do.

raphael-proust:
I made a cleaner branch: https://github.com/raphael-proust/ocamlformat/tree/break-before-in rebased on master

raphael-proust:
@gpetiot  @jberdine let me know if I need to make further changes.

test/passing/long_let_in.ml.ref
```diff
@@ -0,0 +1,59 @@
...
+
+let flat : unit =
+  let short = 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let fooo =
+    1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let baaar =
+    1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 in
+  let long =
+    1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1
+    in
```
hhugo: Having `in` alone and indented looks weird

raphael-proust: I made it unindented.

There is discussion to allow margin breaking for the corner case (along the lines of `break_before_in: [ ``Always | ``Only_to_save_margin | ``Never ]` with `Never` simply ignoring the margin for the purpose of laying the `in`) I might make another PR in the future.

gpetiot:
The output looks better now, can you rebase the branch on master? (your changes of Fmt_ast have to be reported in the fmt_let function)
Thanks.

raphael-proust:
Done. (There was also a small conflict in `Conf.ml` because of changes to another `break*` variable.)

src/Fmt_ast.ml
```diff
@@ -383,7 +383,10 @@ let fmt_let c ctx ~ext ~rec_flag ~bindings ~body ~parens ~attributes
                   ?ext:(if first then ext else None)
                   ctx binding
                   ~in_:(fun indent ->
-                    fmt_if_k last (break 1 (-indent) $ fmt "in") )
+                    fmt_if_k last
+                      (fmt_or_k c.conf.break_before_in
+                         (break 1 (-indent) $ fmt "in")
+                         (fits_breaks " in" "@;<1 -2>in")) )
```
Julow: Could you try using `Format.pp_print_fits_or_breaks` directly (or through a new function in `Fmt`) to avoid hardcoding the indent ?

raphael-proust: Good catch! Thanks!

I fixed it but I went for an easier solution: build the required string using `indent`. I think the two commits should be squashed when merged.

I can add a function to the `Fmt` module if it's more in line with the usage of the project. My preferred solution would be something along the lines of `fits_or: string -> t -> t` which checks if there's enough space for the `string` and otherwise uses `t`. I can't see the necessary helper functions in `Format` though.

Julow: It's better but I'm not fan of the approach.
`fits_breaks` cannot take a format as parameter, because it is based on `Format.pp_print_fits_or_breaks` that only takes strings. The "@;\<n o>" part is actually parsed here: https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Fmt.ml#L103
This is obfuscation to me and I would prefer a direct binding to `pp_print_fits_or_breaks` but I understand this may be out of the scope of this PR.

jberdine:
@raphael-proust: an admin point is that we need the CLA to be signed. Do you expect this to be already taken care of (and the bot that adds the CLA tags is just confused) or is this yet to do?

raphael-proust:
Oh… I forgot to ask my employer about the CLA. I'll ask them and see if I need to do anything on their side.

I'll also have to understand what CLAs mean. And especially how is it useful considering I'm already pushing commits that include the license file.

jberdine:
@raphael-proust IANAL but AFAIU the main point of the CLA is to enable relicensing. For example recently many facebook open source projects that were licensed under "BSD + Facebook patents" were changed to standard MIT. It would have been much hard to do that without CLAs in place, and in many cases it probably would not have happened.

jberdine:
I'm closing this just for workflow. Please reopen when there is progress on the CLA, and this will need to be rebased.

PR #574, opened 2019-01-29
Implement box debugging

Julow:
Print colored brackets around boxes to a temporary *.boxes.ml* file

I needed this to understand the behavior of `fmt_module`, which I still don't understand.

When printed to a terminal, the debug file looks like this: https://imgur.com/hPrcAkn

### DIFF
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -23,6 +23,12 @@ let set_margin n fs =
   Format.pp_set_margin fs n ;
   Format.pp_set_max_indent fs (n - 1)
 
+(** Debug of formatting -------------------------------------------------*)
+
+let pp_color_k color_code k fs =
+  let c = Format.sprintf "\x1B[%dm" in
+  Format.fprintf fs "@<0>%s%t@<0>%s" (c color_code) k (c 0)
+
 (** Break hints and format strings --------------------------------------*)
 
 let break n o fs = Format.pp_print_break fs n o
@@ -150,15 +156,59 @@ let wrap_fits_breaks ?(space = true) conf x =
 
 (** Boxes ---------------------------------------------------------------*)
 
-let open_box n fs = Format.pp_open_box fs n
+let box_debug_enabled = ref false
 
-and open_vbox n fs = Format.pp_open_vbox fs n
+let with_box_debug k fs =
+  let g = !box_debug_enabled in
+  box_debug_enabled := true ;
+  k fs ;
+  box_debug_enabled := g
 
-and open_hvbox n fs = Format.pp_open_hvbox fs n
+let box_stack = ref []
 
-and open_hovbox n fs = Format.pp_open_hovbox fs n
+let box_depth_colors = [|32; 33; 34; 31; 35; 36|]
 
-and close_box fs = Format.pp_close_box fs ()
+let box_depth_color () =
+  let depth = List.length !box_stack in
+  box_depth_colors.(depth % Array.length box_depth_colors)
+
+let debug_box_open open_sym close_sym n fs =
+  if !box_debug_enabled then (
+    let openning =
+      if n = 0 then open_sym else Format.sprintf "%s‹%d›" open_sym n
+    in
+    pp_color_k (box_depth_color ())
+      (fun fs -> Format.fprintf fs "@<0>%s" openning)
+      fs ;
+    box_stack := close_sym :: !box_stack )
+
+let debug_box_close fs =
+  if !box_debug_enabled then
+    match !box_stack with
+    | [] -> pp_color_k 41 (fmt "@<0>]") fs
+    | close_sym :: tl ->
+        box_stack := tl ;
+        pp_color_k (box_depth_color ())
+          (fun fs -> fmt "@<0>%s" fs close_sym)
+          fs
+
+let open_box n fs =
+  debug_box_open "«" "»" n fs ;
+  Format.pp_open_box fs n
+
+and open_vbox n fs =
+  debug_box_open "⟨" "⟩" n fs ;
+  Format.pp_open_vbox fs n
+
+and open_hvbox n fs =
+  debug_box_open "⦑" "⦒" n fs ;
+  Format.pp_open_hvbox fs n
+
+and open_hovbox n fs =
+  debug_box_open "⟪" "⟫" n fs ;
+  Format.pp_open_hovbox fs n
+
+and close_box fs = debug_box_close fs ; Format.pp_close_box fs ()
 
 (** Wrapping boxes ------------------------------------------------------*)
 
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -166,7 +166,8 @@ let with_box_debug k fs =
 
 let box_depth = ref 0
 
-let box_depth_colors = [|32; 33; 34; 31; 35; 36|]
+(* Numeric part of the ANSI escape sequence for colors *)
+let box_depth_colors = [|32; 33; 94; 31; 35; 36|]
 
 let box_depth_color () =
   box_depth_colors.(!box_depth % Array.length box_depth_colors)
@@ -184,6 +185,7 @@ let debug_box_open box_kind n fs =
 let debug_box_close fs =
   if !box_debug_enabled then
     if !box_depth = 0 then
+      (* mismatched close, red background *)
       pp_color_k 41 (fun fs -> Format.fprintf fs "@<0>]") fs
     else (
       Int.decr box_depth ;
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -164,48 +164,41 @@ let with_box_debug k fs =
   k fs ;
   box_debug_enabled := g
 
-let box_stack = ref []
+let box_depth = ref 0
 
 let box_depth_colors = [|32; 33; 34; 31; 35; 36|]
 
 let box_depth_color () =
-  let depth = List.length !box_stack in
-  box_depth_colors.(depth % Array.length box_depth_colors)
+  box_depth_colors.(!box_depth % Array.length box_depth_colors)
 
-let debug_box_open open_sym close_sym n fs =
+let debug_box_open box_kind n fs =
   if !box_debug_enabled then (
     let openning =
-      if n = 0 then open_sym else Format.sprintf "%s‹%d›" open_sym n
+      if n = 0 then box_kind else Format.sprintf "%s<%d" box_kind n
     in
     pp_color_k (box_depth_color ())
-      (fun fs -> Format.fprintf fs "@<0>%s" openning)
+      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>>" openning)
       fs ;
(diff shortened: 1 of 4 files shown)


src/Fmt.ml
```diff
@@ -150,15 +156,42 @@ let wrap_fits_breaks ?(space = true) conf x =
...
+
+let box_depth_colors = [|32; 33; 34; 31; 35; 36|]
+
+let box_depth_color () =
+  let depth = List.length !box_stack in
+  box_depth_colors.(depth % Array.length box_depth_colors)
+
+let debug_box_open open_sym close_sym fs =
+  if !box_debug_enabled then (
+    pp_color (box_depth_color ()) (fun fs -> fmt "@<0>%s" fs open_sym) fs ;
```
jberdine: If you end up passing the format stream around explicitly, it is probably clearer to not use the `fmt` combinators and just use `Format.fprintf` directly.

Julow:
> We have to make sure all boxes in the code are opened/closed using the Fmt API, i.e. no `@[ ... @]`, if there is not many changes we can do that in this PR.

Fixed

src/Fmt.ml
```diff
@@ -150,15 +156,42 @@ let wrap_fits_breaks ?(space = true) conf x =
...
+  if !box_debug_enabled then
+    match !box_stack with
+    | [] -> pp_color 41 (fmt "@<0>]") fs
+    | close_sym :: tl ->
+        box_stack := tl ;
+        pp_color (box_depth_color ())
+          (fun fs -> fmt "@<0>%s" fs close_sym)
+          fs
+
+let open_box n fs = debug_box_open "<" ">" fs ; Format.pp_open_box fs n
```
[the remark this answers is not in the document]
Julow: Colors are enough on my terminal but this is a great idea.
The problem with break hints is that they are almost always inserted in the format string and `Fmt.break` is almost never used.

src/Fmt.ml
Julow: `Format.fprintf` takes the `fs` as first parameter so it would not be much better.

gpetiot:
> @gpetiot There is hundreds of occurrence of `@;` and `@,` and like 3 of `Fmt.break`. I'm not sure it is the right approach

This change has to be made sooner or later, but I agree we can wrap up this PR and do it in another one.

src/Fmt.ml
```diff
@@ -150,15 +156,48 @@ let wrap_fits_breaks ?(space = true) conf x =
...
+          (fun fs -> fmt "@<0>%s" fs close_sym)
+          fs
+
+let open_box n fs =
+  debug_box_open "«" "»" fs ;
+  Format.pp_open_box fs n
 
-and open_vbox n fs = Format.pp_open_vbox fs n
+and open_vbox n fs =
+  debug_box_open "⟨" "⟩" fs ;
```
Julow: Length of lines with unicode characters is wrong ?

src/Fmt.mli
```diff
@@ -146,6 +146,10 @@ val wrap_fits_breaks_if :
 
 (** Boxes ---------------------------------------------------------------*)
 
+val box_debug_enabled : bool ref
+(** If set to true, brackets will be printed to represent boxes. For debug
+    purposes, by default, false *)
```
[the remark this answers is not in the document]
Julow: Done. This is a great change but it does not help the readability of `Translation_unit`

gpetiot:
Regarding the style of the output I would expect something like:
(Of course I don't oppose to having a fancy character instead of the name of the box, but we need the indentation of the box)
```xml
  <hvbox 2
       foo
       <break 1 0>
       <hovbox 3
           foo
           foo
           <break 1 2>
           foo
       >
       <break 1000 0>
       <hvbox 0
           foo
           <break 1 4>
           foo
       >
  >
```
or something more xml/html-like if it's easier to generate pretty pictures from it:
```xml
  <hvbox 2>
       foo
       <break 1 0>
       <hovbox 3
           foo
           foo
           <break 1 2 />
           foo
       </hovbox>
       <break 1000 0 />
       <hvbox 0
           foo
           <break 1 4 />
           foo
       </hvbox>
  </hvbox>
```

src/Fmt.ml
jberdine: `@<0>` should take care of it if I understand your question correctly

src/Fmt_ast.ml
```diff
@@ -3506,25 +3507,24 @@ and fmt_module c ?epi keyword name xargs xbody colon xmty attributes =
...
-                    $ psp $ bdy
-                    $ fmt_if_k (Option.is_some pro) cls
-                    $ esp
-                    $ ( match next with
-                      | Some (_, Some {opn; pro= Some _}) ->
-                          opn $ open_hvbox 0
-                      | _ -> fmt "" )
-                    $ Option.call ~f:epi )
-              $
+              let arg =
```
jberdine: Maybe clearer to write something like:
```
              let maybe_box =
                match arg_mtyp with
                | Some {pro= None} -> hvbox 2
                | _ -> Fn.id
              in
              fmt "@ "
              $ maybe_box
                  ( fmt "(" ... fmt ")" )
```

src/Fmt.ml
Julow: I agree.

src/Fmt.ml
Julow: No. I mean there may be a bug in Ocamlformat:

```
let open_box n fs =
  debug_box_open "«" "»" fs ;
  Format.pp_open_box fs n

let open_box n fs = debug_box_open "<" ">" fs ; Format.pp_open_box fs n
```

jberdine: I see. Currently only comments are interpreted as if unicode as far as length goes. For actual code, I don't know what the right thing is, since AFAIU code is supposed to be ascii or maybe latin-1, with some idea that in the future it will be utf8. I just don't feel that I know what the problems might be, so have avoided unescaped unicode in non-comments.

src/Fmt_ast.ml
Julow: Done.

src/Fmt.ml
Julow: We should extend this to string literals, I think unicode characters in strings is as common as in comments.

src/Fmt_ast.ml
```diff
@@ -3506,25 +3507,24 @@ and fmt_module c ?epi keyword name xargs xbody colon xmty attributes =
...
-              $
-              match arg_mtyp with
-              | Some {pro= None} -> fmt ")@]"
-              | _ -> fmt ")" ) )
+              let maybe_box k =
+                match arg_mtyp with Some {pro= None} -> hvbox 2 k | _ -> k
+              in
+              fmt "@ "
+              $ maybe_box
+                  ( fmt "(" $ fmt_str_loc c name
```
jberdine: minor: could now use `wrap "(" ")"` instead

src/Translation_unit.ml
```diff
@@ -144,24 +144,39 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
...
-      if not Conf.debug then Filename.open_temp_file ~temp_dir:dir base ext
+    dump xunit dir base ".old" ".ast" ast ;
+    let remaining_comments, tmp =
+      let print_and_get_remaining_comments wrap oc =
+        let source = Source.create source_txt in
+        let cmts_t = xunit.init_cmts source conf ast comments in
+        let fs = Format.formatter_of_out_channel oc in
+        postprocess fs ;
+        Fmt.set_margin conf.margin fs ;
+        wrap (xunit.fmt source cmts_t conf ast) fs ;
```
jberdine: `wrap` is not very informative, and is used elsewhere, what about `maybe_debug` or something? And it is probably clearer to just have a named bool arg instead of passing either Fn.id or Fmt.with_box_debug around.

src/Fmt_ast.ml
Julow: Yes

src/Translation_unit.ml
Julow: I agree, I changed to use a boolean parameter.

src/Translation_unit.ml
```diff
@@ -149,25 +149,39 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
...
-        (tmp, oc)
+        let print_to_file ?box_debug base_name =
+          let tmp = Filename.concat dir base_name in
+          Format.eprintf "%s@\n%!" tmp ;
+          let oc = Out_channel.create ~fail_if_exists:false tmp in
+          let cmts = print_and_get_remaining_comments ?box_debug oc in
+          (cmts, tmp)
+        in
+        let tmp = print_to_file (Format.sprintf "%s.%i%s" base i ext) in
+        ignore
```
gpetiot: (not important) A good practice (that is not always enforced in this project) is to check for the return type of the ignored expression.

Julow: This type is quite complicated, is `_ * _` good enough ?

src/Fmt.ml
```diff
@@ -150,15 +156,52 @@ let wrap_fits_breaks ?(space = true) conf x =
...
+
+let with_box_debug k fs =
+  let g = !box_debug_enabled in
+  box_debug_enabled := true ;
+  k fs ;
+  box_debug_enabled := g
+
+let box_depth = ref 0
+
+let box_depth_colors = [|32; 33; 34; 31; 35; 36|]
```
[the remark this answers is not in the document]
Julow: Done

src/Fmt.ml
```diff
@@ -150,15 +156,52 @@ let wrap_fits_breaks ?(space = true) conf x =
...
+    in
+    pp_color_k (box_depth_color ())
+      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>>" openning)
+      fs ;
+    Int.incr box_depth )
+
+let debug_box_close fs =
+  if !box_debug_enabled then
+    if !box_depth = 0 then
+      pp_color_k 41 (fun fs -> Format.fprintf fs "@<0>]") fs
```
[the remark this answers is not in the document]
Julow: Added comment.

src/Translation_unit.ml
gpetiot: I find this syntax prettier:
```ocaml
let ( (_ : (Location.t * string * string * Sexp.t) list lazy_t)
    , (_ : string) ) =
  print_to_file ~box_debug:true
     (Format.sprintf "%s.%i.boxes%s" base i ext)
in
```
But I won't force it, maybe Josh will have an opinion about it

Julow: I don't like this but I won't force either.

jberdine: I tend to prefer to keep all the "mess" related to `ignore` together and subordinate to the actual code, so put `|> (ignore : _ * _ -> _)` after the computation. And unless there is some particular documentation value in specifying the type more, I wouldn't do more than make sure that the computation actually gets executed.

Julow: I did that.

Issue #569, opened 2019-01-25
Issues preventing Janestreet adoption


Julow:
This is some issues that are blocking Janestreet adoption of Ocamlformat

# ~Or patterns on the same line~

Pull request: https://github.com/ocaml-ppx/ocamlformat/pull/536

# ~Non-doc comments in record definitions~

Pull request: https://github.com/ocaml-ppx/ocamlformat/pull/580

# ~Extra space before doc comments in record definitions~

Pull request: https://github.com/ocaml-ppx/ocamlformat/pull/575

# Inconsistent break before module signature when there is constraint

Issue: https://github.com/ocaml-ppx/ocamlformat/issues/572

# ~Missing break before closing bracket of polymorphic variants~

Pull request: https://github.com/ocaml-ppx/ocamlformat/pull/583

# Type annotation on first-class modules

Issue: https://github.com/ocaml-ppx/ocamlformat/issues/629

# Configure strategies for breaking long functions declarations

Issue: https://github.com/ocaml-ppx/ocamlformat/issues/626

# Keep tag-only doc comments on the same line

Issue: https://github.com/ocaml-ppx/ocamlformat/issues/627

# Break before multi-line expression in `let module X =` body

Issue: https://github.com/ocaml-ppx/ocamlformat/issues/628

# Extra break in first-class module syntax

~Pull request: https://github.com/ocaml-ppx/ocamlformat/pull/585~

# ~Extra break in module pack type~

Pull request: https://github.com/ocaml-ppx/ocamlformat/pull/600

jberdine:
It would be good to separate these into individual issues so that we can discuss and track progress.

Julow:
Splitted.

PR #575, opened 2019-01-29
Remove extra space before documentation in record declarations

Julow:
This is one of https://github.com/ocaml-ppx/ocamlformat/issues/569. There is 2 spaces just before documentation comments in:

```ocaml
type t =
  { x : int  (** comment *)
  ; y : int  (** comment for y *)
  }
```

This space may be there on purpose, should I add an option to put it back?

EDIT:
Now this PR is about adding the `doc_comments_padding` option
It's `2` by default and `1` in the Janestreet profile


hhugo:
@jberdine, can you explain the reasoning behind these double spaces?

Julow:
We could add a case to the `doc-comments` option, e.g `After_padded`

jberdine:
The reasoning for two spaces isn't deep, just based on feedback to the effect that the docstrings end up crowding the code itself. This is more apparent when the docstrings are not vertically aligned with each other, as then they are generally quite visually distinct.

An additional case for doc-comments could work, but might be confusing since the docstrings for e.g. record fields are always placed after.

But however it is configured this is not specific to records, the spacing should be the same for e.g. docstrings on constructors.

hhugo:
@mbarbin, do you have strong opinions on this ?

Julow:
I updated this PR for variant and extension constructors and polymorphic variant and object types.
An option like `doc_comments_padding=2` would make sense ?

gpetiot:
> I updated this PR for variant and extension constructors and polymorphic variant and object types.
> An option like `doc_comments_padding=2` would make sense ?

It makes sense to have this option if we can't agree on the appropriate number of spaces.

Julow:
I added the `doc-comments-padding` option. @jberdine I set `1` as the default, do you prefer we keep `2` ?

jberdine:
Yes, in general: keep defaults of new options to current behavior for at least one release.

Julow:
I set the default back to `2` but keep `1` in the Janestreet profile.

jberdine:
Could you rebase and resolve conflicts so I can test?

Julow:
Rebased and cleaned.

gpetiot:
You can test with tools/test_branch.sh that it doesn't break anything, but otherwise looks good.

PR #580, opened 2019-01-30
Break comment record

Julow:
In:
```ocaml
type t =
  { x : int (* comment *)
  ; y : int
  (* comment for y *)
  }
```
The first comment is attached to the type and the second to the label declaration.

This PR move it directly after the type.

(One of https://github.com/ocaml-ppx/ocamlformat/issues/569)

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3069,25 +3069,22 @@ and fmt_label_declaration c ctx lbl_decl ?(last = false) =
   @@ fun c ->
   let doc, atrs = doc_atrs pld_attributes in
   let indent = if Poly.(c.conf.break_separators = `Before) then 2 else 0 in
-  let fmt_cmts =
-    Cmts.fmt c.cmts ~eol:(break_unless_newline 1 indent) pld_loc
-  in
-  fmt_cmts
-  @@ hvbox 4
-       ( hvbox 3
-           ( hvbox 2
-               ( fmt_if Poly.(pld_mutable = Mutable) "mutable "
-               $ fmt_str_loc c pld_name
-               $ fmt_if Poly.(c.conf.field_space = `Loose) " "
-               $ fmt ":@ "
-               $ fmt_core_type c (sub_typ ~ctx pld_type)
-               $ Cmts.fmt_after c.cmts pld_loc
-               $ fmt_if_k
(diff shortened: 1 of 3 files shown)


src/Fmt_ast.ml
```diff
@@ -3081,6 +3081,7 @@ and fmt_label_declaration c ctx lbl_decl ?(last = false) =
                $ fmt_if Poly.(c.conf.field_space = `Loose) " "
                $ fmt ":@ "
                $ fmt_core_type c (sub_typ ~ctx pld_type)
+               $ Cmts.fmt_after c.cmts pld_loc
```
jberdine: We need to be careful of the relative position of comments, `;`, attributes and docstrings. I'm not sure that comments should come before all of the others.

Also, if `fmt_after` is called explicitly, then things should be reworked to avoid `Cmts.fmt` and use `fmt_before` instead.

Also, we don't want to break the line up to the comment just because the comment doesn't fit, so there should be something like a box around "mutable" through `fmt_core_type` (which is what the `hvbox 4` is currently for). For example:
```
--- a/infer/src/biabduction/State.ml
+++ b/infer/src/biabduction/State.ml
@@ -20,8 +20,9 @@ type failure_stats =
   ; (* number of node failures (i.e. at least one instruction failure) *)
     mutable node_ok: int
   ; (* number of node successes (i.e. no instruction failures) *)
-    mutable first_failure: (Location.t * Procdesc.Node.t * int * Errlog.loc_trace * exn) option
-  (* exception at the first failure *) }
+    mutable first_failure:
+      (Location.t * Procdesc.Node.t * int * Errlog.loc_trace * exn) option
+      (* exception at the first failure *) }
```

Something doesn't seem to fit quite right here. Can you explain your reasoning for why formatting the comments with `fmt_after` where you have it solves your problem relative to them being formatted by `Cmts.fmt` after the `hvbox 4` closes?

Julow: In the example in this PR, the first comment is printed by `fmt_core_type` and the second by this function, so I tried to move the comment closer to where it should be printed. I agree this is a bit hacky.

I tried an other approach [on this branch](https://github.com/ocaml-ppx/ocamlformat/compare/master...Julow:break_comment_record_steal_comment_from_type) where I "steal" the comment after the type but now they are not printed where they should.

PR #584, opened 2019-02-01
[WIP] Ocamlformat lib

gpetiot:
The goal is to have an `ocamlformat-api` package that can be used in other projects.

Libraries:
- ocamlformat-base: contains the import file that is auto-included by dune
- ocamlformat-import: contains the Format vendored library
- ocamlformat-api: the main file is Translation_unit
the libraries `cmdliner` and `fpath` are no longer dependencies of ocamlformat-api.

Binaries:
- ocamlformat
- ocamlformat_reason

The external API of ocamlformat-api would be:
- the current Translation_unit.mli
- the Conf.t type
Any idea about the API is welcomed.

Still in progress.
- [ ] I'm not satisfied with having to use refs in Conf to minimize the changes. For now only `debug` still uses a ref in the API.
- [ ] better main file for the API


gpetiot:
I didn't figure out how to fix the CI for this PR, if someone has some time to look at the travis file it would be awesome.

jberdine:
Thanks for working on this! I haven't had a change to look in detail yet, but a couple initial comments:

It would be simpler if we could have just one `libocamlformat` library, plus the executables implemented using it. Have you tried that and run into trouble? Is there some benefit to having 3 libs instead of 1 that I'm overlooking?

I expect that the main library entry point should probably be something along the lines of accepting as arguments:
  1. a string with the filename of the original source file, to be used in place of the `--name` cl arg
  2. a string containing the source concrete syntax (perhaps this will be the contents of the file, or perhaps the filename is only used for finding config and the string is different)
  3. a parsetree and list of comments (which might be the result of parsing the concrete syntax, or might have been transformed e.g. by some ppx, but with the constraint that any Location's in the parsetree must correspond to the right points in the concrete syntax string)
  4. an arg indicating the source kind (impl, intf, or use file) (for the purpose of reparsing and checking)
  5. a file name to be created containing the output (or maybe it would be better to instead return the name of a tmp file that the lib has created containing the output)
  6. maybe a couple args for e.g. controlling debug mode, if necessary (but from the perspective of a lib, maybe having an env var to control that sort of thing is better than needing some way to e.g. tell dune to pass debug to libocamlformat, and maybe we don't need to support setting debug mode for the lib)

In particular, I don't think that the lib client ought to have to do anything about Conf. The lib should do all the lookup in config files, etc. itself. Currently `Conf.action` combines some info (e.g. inplace or not file handling) that only concerns the implementation of the cli, but not the actual formatting, plus some info (e.g. impl vs intf) that is relevant to the formatting. I think that some refactoring here to separate these sorts of info will be needed. I expect it will take some refactoring to be able to eliminate the cmdliner dep from the lib, as the code for command line options and reading configs is currently mixed.

Also, the idea is for the lib api to take args that provide the info it really needs to operate, without making assumptions or constraints about how files are handled, etc. Additional convenience wrappers can always be added.

Does that make sense?

gpetiot:
I will try to put as much things as possible in Api.Conf and only put the option override from cmdline in Cli, there is a bit of work ahead.

hcarty:
It would be useful to split the IO into a separate library. That would allow js_of_ocaml compatibility for projects like sketch.sh.

jberdine:
We need to keep the ability to reparse the produced output and check that it is the same. It's not clear to me that this can be done without writing a file to give to the parser. Do you have any suggestions?

gpetiot:
Waiting for #589, it would be great if we could remove all I/O operations from ocamlformat-api (or whatever we choose to name it).

gpetiot:
Replaced by #589 
(this would be a nightmare to rebase onto #589)

PR #585, opened 2019-02-01
No break module pack

Julow:
(one of https://github.com/ocaml-ppx/ocamlformat/issues/569)

An extra break is added when the expression does not fit on one line:

```
let _ =
  (module M
  : S
    with type long_long_type = t
     and type long_long_long_type = u
     and type very_long_long_type = t )

let _ =
  ( module struct
    type t
  end
  : S )
```

A little indentation is added because I moved the parenthesis outside of the box.

### DIFF
--- a/test/passing/js_source.ml.ref
+++ b/test/passing/js_source.ml.ref
@@ -3199,9 +3199,7 @@ let make_set (type s) cmp =
     type t = s
 
     let compare = cmp
-  end)
-  : Set.S
-    with type elt = s)
+  end) : Set.S with type elt = s)
 ;;
 
 (* No type annotation here *)
@@ -3271,8 +3269,7 @@ let m =
 let m =
   (module struct
     let x = 3
-  end
-  : S)
+  end : S)
 ;;
 
 ;;
@@ -3310,8 +3307,7 @@ end
 let rec (module M : S') =
   (module struct
     let f n = if n <= 0 then 1 else n * M.f (n - 1)
-  end
-  : S')
+  end : S')
 in
 M.f 3
 
@@ -3421,17 +3417,14 @@ module SSMap = struct
 end
 
 let ssmap =
-  (module SSMap
-  : MapT
+  (module SSMap : MapT
     with type key = string and type data = string and type map = SSMap.map)
 ;;
 
 let ssmap =
   (module struct
     include SSMap
-  end
-  : MapT
-    with type key = string and type data = string and type map = SSMap.map)
+  end : MapT with type key = string and type data = string and type map = SSMap.map)
 ;;
 
 let ssmap :
@@ -5837,8 +5830,7 @@ end
 let v =
   (module struct
     let x = 3
-  end
-  : S)
+  end : S)
 ;;
 
 module F () = (val v)
@@ -5863,8 +5855,7 @@ let v =
     type t = int
 
     let x = 3
-  end
-  : S)
+  end : S)
 ;;
 
 module F () = (val v)
--- a/test/passing/module.ml
+++ b/test/passing/module.ml
@@ -40,8 +40,7 @@ module type X = sig
   val n : (module S with type t = t)
 
   val o :
-    (module
-     S
+    (module S
        with type long_long_type = t
         and type long_long_long_type = u
         and type very_long_long_type = t)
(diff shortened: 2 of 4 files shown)


test/passing/module.ml
```diff
@@ -28,3 +28,26 @@ module C = struct
...
+let _ =
+  ( module M : S
+    with type long_long_type = t
+     and type long_long_long_type = u
+     and type very_long_long_type = t )
+
+let _ =
+  ( module struct
+      type t
+  end : S with type t = t )
```
gpetiot: I think `end` should start in the same column as `module`. Also can you check the output when there is enough type constraints to break the line in this case?

Julow: This was a bug, fixed.

test/passing/js_source.ml.ref
```diff
@@ -3189,12 +3189,10 @@ let sort (type s) (module Set : Set.S with type elt = s) l =
 (* No real improvement here? *)
 let make_set (type s) cmp =
   (module Set.Make (struct
-    type t = s
+     type t = s
```
jberdine: Note that in general the output produced for cases where a paren precedes a keyword indents from the paren, not the keyword, so this added indentation is in conflict with that. The choice of where to indent from can be discussed, but I don't think it shouldn't be changed just in this case.

test/passing/module.ml
```diff
@@ -28,3 +28,41 @@ module C = struct
 
   and B : sig end = B
 end
+
+let _ =
+  ( module M : S
+    with type long_long_type = t
```
jberdine: This seems to be just wrong, `with` modifies `S`, not `module M`. Am I missing something?

test/passing/module.ml
```diff
@@ -28,3 +28,41 @@ module C = struct
...
+     and type long_long_long_type = u
+     and type very_long_long_type = t )
+
+let _ =
+  ( module struct
+      type t
+    end : S with type t = t )
```
jberdine: Here a phrase is placed at the end of the last line of the preceding phrase which spans multiple lines. This is something that the rest of the formatting tries hard not to do. Maybe here an exception is ok, since `end` will not ever have anything before it on the same line. But can you just confirm that this is intended?

Julow: This is intended but not required.

test/passing/js_source.ml.ref
Julow: The `end` is also indented to be aligned with the `module` keyword. Should I undo this and open a new PR ?

test/passing/module.ml
```diff
@@ -28,3 +28,41 @@ module C = struct
...
+let _ =
+  ( module struct
+      type t
+    end : S )
+
+let _ =
+  ( module struct
+      type t
+    end
+  : Loooooooooooooooonnnnnnnnnnnnnnnng.Looooooooooooooonnnnnnnnnnnnnng.Mod
```
Julow: Should the `:` be on the same line as `end` and the closing `)` just after the type ?

test/passing/module.ml
Julow: I agree, I added indentation

test/passing/module.ml
jberdine: The `:` is currently in the right place. Due to the length of the name, I think the `)` is in the right place or else it would encroach into the margin.

test/passing/module.ml
```diff
@@ -28,3 +28,65 @@ module C = struct
...
+    (module S
+         with type long_long_type = t
+          and type long_long_long_type = u
+          and type very_long_long_type = t)
+end
+
+module X = (val X.n : S with type t = t)
+
+module Y = ( val X.o : S
+               with type long_long_type = t
```
jberdine: Here there is indentation that depends on width of user-code, the name `Y`. This is something that's considered to be a bug. (Though I think it is possible currently, at least due to interactions with comments.)

test/passing/js_source.ml.ref
```diff
@@ -3414,17 +3410,14 @@ module SSMap = struct
 end
 
 let ssmap =
-  (module SSMap
-  : MapT
+  (module SSMap : MapT
```
jberdine: Here, since `MapT` is modified by `with`, we need a break before `:`.

test/passing/module.ml
jberdine: (Before `:`, not `with`.) I think that for these where the module type expression fits on a line, it could be placed after the `end` as here. But once the module type expression does not fit on a line, we need a break before `:`. So I think that the question is whether it is worth having what seems to be two different layouts. I wouldn't object to having two layouts, but in the past others have been more hesitant.

test/passing/js_source.ml.ref
Julow: I removed that indentation.

test/passing/module.ml
Julow: I'll implement a different layout.

test/passing/module.ml
Julow: Fixed

test/passing/js_source.ml.ref
Julow: This is intended, I'll add an option to control the whole PR

PR #589, opened 2019-02-03
Change API to not rely on the file-system to format sources

hhugo:
Following a comment from @jberdine  https://github.com/ocaml-ppx/ocamlformat/pull/584#issuecomment-459972698

> We need to keep the ability to reparse the produced output and check that it is the same. It's not clear to me that this can be done without writing a file to give to the parser. Do you have any suggestions?

### DIFF
--- a/src/Reason.mli
+++ b/src/Reason.mli
@@ -15,12 +15,14 @@ open Migrate_ast
 open Parsetree
 open Translation_unit
 
-val input_impl : Conf.t -> In_channel.t -> structure with_comments
+type 'a t = {origin_filename: string; ast_and_comment: 'a with_comments}
+
+val input_bin_impl : In_channel.t -> structure t
 (** Reads a serialized structure from an input channel. It is assumed to be
     the output of `refmt --print=binary_reason` where `refmt` has been
     compiled with the same version of `ocaml` as `ocamlformat`. *)
 
-val input_intf : Conf.t -> In_channel.t -> signature with_comments
+val input_bin_intf : In_channel.t -> signature t
 (** Reads a serialized signature from an input channel. It is assumed to be
     the output of `refmt --print=binary_reason` where `refmt` has been
     compiled with the same version of `ocaml` as `ocamlformat`. *)
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -20,8 +20,7 @@ type 'a with_comments =
 
 (** Operations on translation units. *)
 type 'a t =
-  { input: Conf.t -> input_file:string -> In_channel.t -> 'a with_comments
-  ; init_cmts:
+  { init_cmts:
       Source.t -> Conf.t -> 'a -> (string * Location.t) list -> Cmts.t
   ; fmt: Source.t -> Cmts.t -> Conf.t -> 'a -> Fmt.t
   ; parse: Lexing.lexbuf -> 'a
@@ -38,9 +37,6 @@ type 'a t =
       -> Normalize.docstring_error list
   ; normalize: Conf.t -> 'a with_comments -> 'a
   ; printast: Caml.Format.formatter -> 'a -> unit }
-
-(** Existential package of a type of translation unit and its operations. *)
-type x = XUnit : 'a t -> x
 
 exception Warning50 of (Location.t * Warnings.t) list
 
@@ -74,23 +70,20 @@ end = struct
   let enable x = abs x
 
   let to_string x =
-    String.concat ~sep:"" (List.map ~f:(Format.sprintf "%+d") x)
+    String.concat ~sep:"" (List.map ~f:(Printf.sprintf "%+d") x)
 end
 
-let parse parse_ast (conf : Conf.t) ~input_file ic =
+let parse parse_ast (conf : Conf.t) source =
+  let lexbuf = Lexing.from_string source in
   let warnings =
     W.enable 50
     :: (if conf.quiet then List.map ~f:W.disable W.in_lexer else [])
   in
   Warnings.parse_options false (W.to_string warnings) ;
-  let lexbuf = Lexing.from_channel ic in
-  Lexer.skip_hash_bang lexbuf ;
   let hash_bang =
-    let ic = In_channel.create input_file in
+    Lexer.skip_hash_bang lexbuf ;
     let len = lexbuf.lex_last_pos in
-    let buf = Base.Bytes.create len in
-    ignore (In_channel.really_input_exn ic ~buf ~pos:0 ~len) ;
-    Base.Bytes.to_string buf
+    String.sub source ~pos:0 ~len
   in
   Location.init lexbuf !Location.input_name ;
   let warning_printer = !Location.warning_printer in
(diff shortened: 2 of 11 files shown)


hhugo:
This makes `Translation_unit` no longer depend on `Unix`

hhugo:
cc @hcarty

src/Translation_unit.ml
```diff
@@ -114,94 +108,95 @@ let parse parse_ast (conf : Conf.t) ic =
...
 let ellipsis_cmt = ellipsis 50
 
-let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
-    ofile =
+let make_dump xunit input_name dir =
   let dir =
-    match ofile with
-    | Some ofile -> Filename.dirname ofile
-    | None -> Filename.get_temp_dir_name ()
+    match dir with Some dir -> dir | None -> Filename.get_temp_dir_name ()
```
gpetiot: Can be rewritten with `Option.value`

src/Translation_unit.ml
```diff
@@ -114,94 +108,95 @@ let parse parse_ast (conf : Conf.t) ic =
...
+        with_file suf ext (fun oc -> Out_channel.output_string oc fmted)
+      in
+      [("output file", String.sexp_of_t file)]
+    else []
+  in
+  (dump_ast, dump_formatted)
+
+let format xunit (conf : Conf.t) ~input_name ~source ~parsed ~output_file =
+  Location.input_name := input_name ;
+  let dump_ast, dump_formatted =
```
gpetiot: Can we go further and make `dump_ast` and `dump_formatted` parameters of the `format` function?

src/Translation_unit.ml
hhugo: This would force the evaluation of `Filename.get_temp_dir_name`. I guess it's ok in that case,

src/Translation_unit.ml
hhugo: I just did that. The API doesn't look great. I don't really know how to make it better.

src/Translation_unit.ml
hhugo: @gpetiot, how do you feel about taking over that PR ? I don't have much time to spend on this. Feel free to push to the branch directly.

src/Translation_unit.ml
hhugo: @gpetiot, how do you feel about taking over that PR ? I don't have much time to spend on this. Feel free to push to the branch directly.

jberdine:
Can I ask what the plan is wrt these changes, as I don't know what the objective is either in the abstract or in detail? I think that the main thing I don't understand is how we are going to call the parser on the generated output. Can that be done without a file?

hhugo:
@jberdine, I'm a bit confused by your question. In this PR, we do call the parser without a using file.

src/Translation_unit.ml
gpetiot: @hhugo can you allow contributors to push to this branch? Right now I can't, thanks!

hhugo:
@jberdine, did we change something regarding locks on branchs/PRs ?
@gpetiot, this branch lives in ocaml-ppx/ocamlformat. I'm surprised you cannot push to it. Do you know what I need to change ?

jberdine:
@hhugo nothing has changed that I know, and it looks like @gpetiot should have permission based on https://github.com/ocaml-ppx/ocamlformat/settings/branch_protection_rules/3895401

hhugo:
It seems that I cannot force push the branch myself. 
```
git push origin -f hh-refactor 
Counting objects: 9, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (9/9), done.
Writing objects: 100% (9/9), 4.49 KiB | 920.00 KiB/s, done.
Total 9 (delta 7), reused 0 (delta 0)
remote: Resolving deltas: 100% (7/7), completed with 5 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/hh-refactor.
remote: error: Cannot force-push to a protected branch
To github.com:ocaml-ppx/ocamlformat.git
 ! [remote rejected] hh-refactor -> hh-refactor (protected branch hook declined)
error: failed to push some refs to 'git@github.com:ocaml-ppx/ocamlformat.git'
```

jberdine:
@hhugo, sorry for my confused question, now I see. Still have not had a chance to look at this in detail, will try to get to it soon.

jberdine:
I tried a force push with a test branch and it failed for me too. I'll ask our team to see if our automation has broken things.

hhugo:
I've rebased the PR and made few more changes.

gpetiot:
I added an additional layer between ocamlformat.ml / ocamlformat_reason.ml and Translation_unit.ml : the files `ocaml_api.ml` /  `reason_api.ml`. The goal is to hide all the xunit shenanigans to the users of the library and don't use file operations. It can be a first step towards exporting an OCamlformat API (instead of what I tried in #584), what do you think?

hhugo:
you should probably simplify signature for dump_ast and dump_formatted

hhugo:
It's not clear that we need 3 function in Ocaml_api. What about only one ?

src/Translation_unit.mli
```diff
@@ -33,34 +32,34 @@ type 'a t =
...
-  -> string option
-  -> result
-(** [parse_print xunit conf input_name input_file input_channel output_file]
-    parses the contents of [input_channel], using [input_name] for error
-    messages, and referring to the contents of [input_file] to improve
-    comment placement, and writes formatted code into [output_file]. *)
+  -> source:string
+  -> parsed:('a with_comments, exn) Result.t
+  -> unit
+  -> (string, error) Result.t
```
jberdine: The only thing a client can do with the returned error is pass it to `print_error`, right? So what about partially applying `print_error` to the error and returning the `print_error` function instead?

hcarty:
Thank you @hhugo!

src/Translation_unit.mli
hhugo: A client could deconstruct the error manually.

src/Translation_unit.ml
```diff
@@ -54,6 +50,8 @@ exception
 
 let internal_error msg kvs = raise (Internal_error (msg, kvs))
 
+let sprintf = Printf.sprintf
```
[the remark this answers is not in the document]
hhugo: Some intermediate state of this PR had many occurrences. Not worth it anymore

src/Translation_unit.mli
```diff
@@ -33,34 +32,34 @@ type 'a t =
...
+  | Ocamlformat_bug of {input_name: string; exn: exn}
   | User_error of string
 
-val parse :
-     (Lexing.lexbuf -> 'a)
-  -> Conf.t
-  -> input_file:string
-  -> In_channel.t
-  -> 'a with_comments
+val parse : (Lexing.lexbuf -> 'a) -> Conf.t -> string -> 'a with_comments
```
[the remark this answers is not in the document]
hhugo: `format` takes a `('a with_comments, exn) result` so that it can be used by both ocamlformat and ocamlformat_reason. `parse` is used by `ocamlformat.ml` to generate that `('a with_comments, exn) result` value. I don't see a better solution/api at the moment.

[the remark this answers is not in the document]
jberdine: Yes, ok. I guess that if integrated e.g. into the ppx pipeline of dune, we'd not use the special treatment of warnings and hash-bang lines in `parse`, and just pass the parsetree that comes down the pipe to `format`.

src/Translation_unit.mli
jberdine: Yes, true. I'm just wondering whether that is more functionality than a client is going to want, and hence maybe the narrower interface would be preferable. I have no strong view either way, just mentioning it.

src/Reason.mli
```diff
@@ -15,12 +15,12 @@ open Migrate_ast
 open Parsetree
 open Translation_unit
 
-val input_impl : Conf.t -> In_channel.t -> structure with_comments
+val input_bin_impl : In_channel.t -> string * structure with_comments
```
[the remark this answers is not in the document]
hhugo: introduced a new return type.

src/Translation_unit.mli
```diff
@@ -33,34 +32,34 @@ type 'a t =
...
 
-val parse_print :
-     x
+val format :
+     'a t
   -> Conf.t
+  -> ?dump_ast:(suffix:string -> 'a -> unit)
```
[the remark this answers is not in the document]
hhugo: cleaned up

hhugo:
@gpetiot, I've made additional changes. What do you think about the current state ?

hhugo:
I've just rebased on top of the reason fix.

gpetiot:
> @gpetiot, I've made additional changes. What do you think about the current state ?

Is there a reason why you removed the Formattable module for ocaml_api but not for reason_api?

hhugo:
No really

hhugo:
I guess we could remove it as well

src/ocaml_api.mli
```diff
@@ -0,0 +1,20 @@
...
+
+val format :
+     Conf.t
+  -> kind:[`Impl | `Intf | `Use_file]
+  -> ?dump_ast:(suffix:string -> (Formatter.t -> unit) -> unit)
+  -> ?dump_formatted:(suffix:string -> string -> string option)
+  -> input_name:string
+  -> source:string
+  -> unit
+  -> (string, Translation_unit.error) Result.t
```
jberdine: To be a self-contained external interface, we should define the type or errors here, or in some other way avoid referring to internal ocamlformat modules. I would be inclined to try to partially apply `print_error` to the error and just return a `Formatter.t -> unit` here.

hhugo:
I want to step back on this PR a little.
I'm tempted to revert many of the change and just address what the title says and not try to replace (https://github.com/ocaml-ppx/ocamlformat/pull/584)
The changes were already significant. what do people think ?

hhugo:
I went ahead and created https://github.com/ocaml-ppx/ocamlformat/pull/611/

jberdine:
I would appreciate if we could discuss a little what the plan is that this fits into. I don't actually understand the motivation of doing the PR title. It seemed to be part of refactoring to expose a library interface that could e.g. be used in a ppx pipeline rather than from the ocamlformat command line. But I don't see what accessing the file system has to do with that. At some point there was a passing comment on some PR about jsoo, but I don't know how that fits in. Help?

hhugo:
I don't really know what people have planned regarding exposing a library. This PR was not really connected to any of this (even if it turned out to be useful).

The initial goal was just rely less on the file system and simplify the code.
Currently, `Translation_unit` formats into a file and load the file back immediately to check for fix-point and eventually loop. We should not need to touch the disk for that.

This PR also allows us to remove non-trivial code to keep the fs clean. 
```
      match (Conf.action, ofile) with
      | _, None ->
          Out_channel.output_string stdout fmted ;
          Unix.unlink tmp ;
          Ok
      | In_out _, Some ofile -> Unix.rename tmp ofile ; Ok
      | Inplace _, Some ofile when i > 1 -> Unix.rename tmp ofile ; Ok
      | Inplace _, Some _ -> Unix.unlink tmp ; Ok ) 
```
```
            let result =
              print_check ~i:(i + 1) ~conf ~ast:new_.ast
                ~comments:new_.comments ~prefix:new_.prefix
                ~source_txt:fmted ~source_file:tmp
            in
            Unix.unlink tmp ; result
```

Regarding jsoo, I think it was mentioned that online ocaml toplevel could format code in the web browser  directly. Jsoo has some basic support for a speudo filesystems so the current implem could maybe work.

jberdine:
Ok, thanks, that makes things clearer. I am totally in favor of not doing silly things like creating unnecessary temporary files. I'm less sure about changes that seem to be adding function arguments to plumb in the code to write files. I'll look at the other PR now.

hhugo:
probably look at https://github.com/ocaml-ppx/ocamlformat/pull/611

PR #592, opened 2019-02-03
Make gen_version.sh use bash.

### LINKED ISSUE #470: gen_version.sh is not portable

hhugo:
fix #470 now not waiting for #471.

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -12,7 +12,7 @@ Jbuild_plugin.V1.send
 (rule
  (targets Version.ml)
  (deps (universe))
- (action (run sh %%{dep:../tools/gen_version.sh} Version.ml))
+ (action (bash "%%{dep:../tools/gen_version.sh} Version.ml"))
  (mode fallback))
 
 (ocamllex Literal_lexer)
--- a/src/dune
+++ b/src/dune
@@ -12,7 +12,7 @@ Jbuild_plugin.V1.send
 (rule
  (targets Version.ml)
  (deps (universe))
- (action (bash "%%{dep:../tools/gen_version.sh} Version.ml"))
+ (action (run bash %%{dep:../tools/gen_version.sh} Version.ml))
  (mode fallback))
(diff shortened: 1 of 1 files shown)


src/dune
```diff
@@ -12,7 +12,7 @@ Jbuild_plugin.V1.send
 (rule
  (targets Version.ml)
  (deps (universe))
- (action (run sh %%{dep:../tools/gen_version.sh} Version.ml))
+ (action (bash "%%{dep:../tools/gen_version.sh} Version.ml"))
```
jberdine: IIUC, `run` is necessary to prevent dune from using `system`, and unquoted `%%{` arg are necessary for dune to track dependencies correctly.

gpetiot: This assumes bash is installed, can we just merge #471 given the diff is really small and tests have been run?

jberdine:
We already depend on bash.

PR #593, opened 2019-02-03
Improve: split Fmt_ast.ml

gpetiot:
- [x] put sugar functions in Sugar.ml
  - [x] write documentation

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3762,31 +3762,7 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
       || pat_is_extension pat
     then (xpat, [], None, xbody)
     else
-      let sugar_polynewtype pat body =
-        let ctx = Pat pat in
-        match pat.ppat_desc with
-        | Ppat_constraint (pat2, {ptyp_desc= Ptyp_poly (pvars, _typ)}) ->
-            let rec sugar_polynewtype_ xpat pvars0 pvars body =
-              let ctx = Exp body in
-              match (pvars, body.pexp_desc) with
-              | [], Pexp_constraint (exp, typ) ->
-                  Some (xpat, pvars0, sub_typ ~ctx typ, sub_exp ~ctx exp)
-              | ( {txt= pvar; loc= loc1} :: pvars
-                , Pexp_newtype ({txt= nvar; loc= loc2}, exp) )
-                when String.equal pvar nvar ->
-                  Cmts.relocate c.cmts ~src:loc2 ~before:loc1 ~after:loc1 ;
-                  sugar_polynewtype_ xpat pvars0 pvars exp
-              | _ -> None
-            in
-            Cmts.relocate c.cmts ~src:pat.ppat_loc ~before:pat2.ppat_loc
-              ~after:pat2.ppat_loc ;
-            sugar_polynewtype_ (sub_pat ~ctx pat2) pvars pvars body
-        | _ -> None
-      in
-      match sugar_polynewtype pat body with
-      (* format
-       *    let f: 'r 's. 'r 's t = fun (type r) -> fun (type s) -> (e : r s t)
-       * as let f: type r s. r s t = e *)
+      match Sugar.sugar_polynewtype c.cmts pat body with
       | Some (xpat, pvars, xtyp, xbody) ->
           let fmt_cstr =
             fmt "@ : type "
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -170,329 +170,6 @@ let fmt_recmodule c ctx items f ast =
          $ fmt_grp ~first grp
          $ fits_breaks_if ((not break_struct) && not last) "" "\n" ))
 
-let rec sugar_arrow_typ c ({ast= typ} as xtyp) =
-  let ctx = Typ typ in
-  let {ptyp_desc; ptyp_loc} = typ in
-  match ptyp_desc with
-  | Ptyp_arrow (l, t1, t2) ->
-      Cmts.relocate c.cmts ~src:ptyp_loc ~before:t1.ptyp_loc
-        ~after:t2.ptyp_loc ;
-      let rest =
-        match t2.ptyp_attributes with
-        | [] -> sugar_arrow_typ c (sub_typ ~ctx t2)
-        | _ -> [(Nolabel, sub_typ ~ctx t2)]
-      in
-      (l, sub_typ ~ctx t1) :: rest
-  | _ -> [(Nolabel, xtyp)]
-
-let rec sugar_class_arrow_typ c ({ast= typ} as xtyp) =
-  let ctx = Cty typ in
-  let {pcty_desc; pcty_loc} = typ in
-  match pcty_desc with
-  | Pcty_arrow (l, t1, t2) ->
-      Cmts.relocate c.cmts ~src:pcty_loc ~before:t1.ptyp_loc
-        ~after:t2.pcty_loc ;
-      let rest =
-        match t2.pcty_attributes with
-        | [] -> sugar_class_arrow_typ c (sub_cty ~ctx t2)
-        | _ -> [(Nolabel, `class_type (sub_cty ~ctx t2))]
-      in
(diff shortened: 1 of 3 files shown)


src/Fmt_ast.ml
```diff
@@ -1196,7 +873,9 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
           | Ppat_array _ | Ppat_record _ -> true
           | Ppat_tuple _ -> Poly.(c.conf.parens_tuple_patterns = `Always)
           | _ -> (
-            match sugar_list_pat c pat with Some _ -> true | None -> false )
+            match Sugar.sugar_list_pat c.cmts pat with
```
Julow: It would be shorter with `Option.is_some`

src/Fmt_ast.ml
```diff
@@ -1207,8 +886,8 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
 
 and fmt_fun_args c ?(pro = fmt "") args =
   let fmt_fun_arg = function
-    | Val (Nolabel, xpat, None) -> fmt_pattern c xpat
-    | Val
+    | Sugar.Val (Nolabel, xpat, None) -> fmt_pattern c xpat
+    | Sugar.Val
```
Julow: It's not needed to prefix with `Sugar.` every time, only once

src/Fmt_ast.ml
```diff
@@ -2070,7 +1753,9 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?(indent_wrap = 0) ?ext
           | Pexp_array _ | Pexp_record _ -> true
           | Pexp_tuple _ -> Poly.(c.conf.parens_tuple = `Always)
           | _ -> (
-            match sugar_list_exp c e0 with Some _ -> true | None -> false )
+            match Sugar.sugar_list_exp c.cmts e0 with
```
Julow: `Option.is_some` ?

src/Sugar.ml
```diff
@@ -0,0 +1,340 @@
...
+        ~after:els.pexp_loc ;
+      (Some (sub_exp ~ctx cnd), sub_exp ~ctx thn, pexp_attributes)
+      :: sugar_ite cmts (sub_exp ~ctx els)
+  | Pexp_ifthenelse (cnd, thn, None) ->
+      Cmts.relocate cmts ~src:pexp_loc ~before:cnd.pexp_loc
+        ~after:thn.pexp_loc ;
+      [(Some (sub_exp ~ctx cnd), sub_exp ~ctx thn, pexp_attributes)]
+  | _ -> [(None, xexp, pexp_attributes)]
+
+let sugar_sequence conf cmts width xexp =
```
Julow: Other sugar functions may use `Conf.t` in the future ? Otherwise, we should move the `List.group ...` part back to `Fmt_ast.ml` and keep only the `sugar_sequence_` part here.

src/Fmt_ast.ml
gpetiot: It seems to be needed, and since it indicates where the constructor comes from I prefer this rather than opening Sugar.

src/Fmt_ast.ml
```diff
@@ -170,329 +170,11 @@ let fmt_recmodule c ctx items f ast =
...
-        ~after:els.pexp_loc ;
-      (Some (sub_exp ~ctx cnd), sub_exp ~ctx thn, pexp_attributes)
-      :: sugar_ite c (sub_exp ~ctx els)
-  | Pexp_ifthenelse (cnd, thn, None) ->
-      Cmts.relocate c.cmts ~src:pexp_loc ~before:cnd.pexp_loc
-        ~after:thn.pexp_loc ;
-      [(Some (sub_exp ~ctx cnd), sub_exp ~ctx thn, pexp_attributes)]
-  | _ -> [(None, xexp, pexp_attributes)]
-
 let sugar_sequence c width xexp =
```
jberdine: It is odd to leave this small piece and move the rest. Can you explain?

src/Fmt_ast.ml
jberdine: The existing code uses the style where type annotations are added to function arguments instead. See e.g. `fmt_longident`.

src/Sugar.ml
```diff
@@ -0,0 +1,340 @@
...
+ *  LICENSE file in the root directory of this source tree.           *
+ *                                                                    *
+ **********************************************************************)
+
+open Migrate_ast
+open Asttypes
+open Parsetree
+open Ast
+
+let rec sugar_arrow_typ cmts ({ast= typ} as xtyp) =
```
jberdine: Since these are in a `Sugar` module, remove the `sugar_` prefixes on the names.

src/Sugar.mli
```diff
@@ -0,0 +1,169 @@
...
+val sugar_functor_type :
+     Cmts.t
+  -> for_functor_kw:bool
+  -> module_type Ast.xt
+  -> (label loc * module_type Ast.xt option) list * module_type Ast.xt
+(** [sugar_functor_type cmts for_functor_kw m] returns the list of module
+    types applied to the functor of module type [m]. [for_functor_kw]
+    indicates if the keyword [functor] is used. The sugar is different when
+    used with the [functor] keyword. The syntax M(A : A)(B : B) cannot
+    handle [_] as module name. [cmts] are relocated to the functor arguments
```
jberdine: The comment about `_` as a module name is really an implementation-explanation comment and should be moved back.

src/Sugar.mli
```diff
@@ -0,0 +1,169 @@
...
+
+val sugar_functor :
+     Cmts.t
+  -> for_functor_kw:bool
+  -> module_expr Ast.xt
+  -> (label loc * module_type Ast.xt option) list * module_expr Ast.xt
+(** [sugar_functor cmts for_functor_kw m] returns the list of module types
+    applied to the functor of module [m]. [for_functor_kw] indicates if the
+    keyword [functor] is used. The sugar is different when used with the
+    [functor] keyword. The syntax M(A : A)(B : B) cannot handle [_] as
```
jberdine: And here.

src/Fmt_ast.ml
Julow: I meant qualifying the first and not the other is enough for the compiler to known where the constructors come from. Type annotations is good too.

src/Fmt_ast.ml
jberdine: I meant that it is probably better to inline this little snippet into the couple places it is used, as all the other similar calls of `List.group`.

src/Fmt_ast.ml
```diff
@@ -621,6 +298,8 @@ let fmt_variance = function
   | Contravariant -> fmt "-"
   | Invariant -> fmt ""
 
+let fmt_private_flag flag = fmt_if Poly.(flag = Private) "@ private"
```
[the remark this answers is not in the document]
gpetiot: Yes but this function didn't need to be in the big `let rec ... and ... ` recursive definition so I put it outside.

src/Sugar.mli
```diff
@@ -0,0 +1,165 @@
...
+
+    {[
+        if c1 then e1
+        else if c2 then e2
+        else e3
+    ]}
+
+    will return the following list:
+    [(Some c1, e1); (Some c2, e2); (None, e3)] (attributes are voluntarily
+    omitted. [cmts] are relocated to the sub-expressions to have more
```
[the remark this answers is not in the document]
gpetiot: Sorry, I meant they are omitted in this example, it should be `[(Some c1, e1, attr1); (Some c2, e2, attr2); (None, e3, attr3)]` instead but I simplified it by omitting the attributes.

src/Fmt_ast.ml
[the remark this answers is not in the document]
jberdine: Yeah, I know, but that didn't seem to be a big deal (it's clear from the definition of this little function that it can't cause divergence). But at the beginning I was referring to parsetree.mli all the time, which is less true now.

src/Sugar.mli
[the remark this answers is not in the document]
jberdine: Ah, ok. In that case I think it would be clearer to just not mention attributes.

PR #600, opened 2019-02-07
Fix extra break in module pack type

Julow:
Small fix to remove an extra break in:

```
val x :
  (module
   M
     with type looooooooooooooooooooooonnnnnnnnnnnnnnnng_type = t
      and type more_types = u)
```

### DIFF
--- a/test/passing/source.ml.ref
+++ b/test/passing/source.ml.ref
@@ -3502,8 +3502,7 @@ let ssmap =
   )
 
 let ssmap :
-    (module
-     MapT
+    (module MapT
        with type key = string
         and type data = string
         and type map = SSMap.map) =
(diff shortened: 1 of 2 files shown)


gpetiot:
Is the result of test_branch looking good?

Julow:
Yes.

test/passing/source.ml.ref
```diff
@@ -3502,8 +3502,7 @@ let ssmap =
   )
 
 let ssmap :
-    (module
-     MapT
+    (module MapT
```
jberdine: It wouldn't be bad to have an additional test where `MapT` is long enough to break, to keep track of what happens in that case.

Julow: Added.

PR #601, opened 2019-02-07
Fix regression for variants with docstrings

### LINKED ISSUE #598: Regression for variants with docstrings
#486 introduced a regression in the formatting of variants with docstrings:
```
val x :
  [ `X of int
    (** Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed *) ]
```
is formatted
```
val x :
  [ `X of
    int
    (** Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed *) ]
```
after #486.

gpetiot:
Fix #598 
The diff of test_branch looks good:
```ocaml
--- a/infer/src/absint/TransferFunctions.mli
+++ b/infer/src/absint/TransferFunctions.mli
@@ -39,12 +39,10 @@ end
 
 module type DisjunctiveConfig = sig
   val join_policy :
-    [ `JoinAfter of
-      int
+    [ `JoinAfter of int
       (** when the set of disjuncts gets bigger than [n] the underlying domain's join is called to
        collapse them into one state *)
-    | `UnderApproximateAfter of
-      int
+    | `UnderApproximateAfter of int
       (** When the set of disjuncts gets bigger than [n] then just stop adding new states to it,
          drop any further states on the floor. This corresponds to an under-approximation/bounded
          approach. *)
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -636,10 +636,11 @@ and fmt_row_field c ctx = function
       let c = update_config c atrs in
       let doc, atrs = doc_atrs atrs in
       hvbox 0
-        ( fmt_str_loc c ~pre:"`" name
-        $ fmt_if (not (const && List.is_empty typs)) " of@ "
-        $ fmt_if (const && not (List.is_empty typs)) " & "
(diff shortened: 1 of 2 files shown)


gpetiot:
cc @Julow for review

Issue #504, opened 2018-11-14
If then else formatting


NathanReb:
Currently the following code:
```ocaml
let _ =
  if condition then
    let a = 1 in
    let b = 2 in
    a + b
  else if other_condition then
    12
  else
    0
```
is either formatted as:
```ocaml
let _ =
  if condition then
    let a = 1 in
    let b = 2 in
    a + b
  else if other_condition then 12
  else 0
```
or, if `if-then-else=keyword-first`, as:
```ocaml
let _ =
  if condition
  then
    let a = 1 in
    let b = 2 in
    a + b
  else if other_condition
  then 12
  else 0
```
What troubles me the most here is the fact that expressions that wraps are treated differently from expressions that don't.

Internally we always format it as in the original example, the main reasons are:
1. It is more consistent as you don't have expressions on their own line with their own indentation level and expressions "hidden" after a keyword mixed in the same conditional expression.
2. It is more line-diff friendly when expressions get longer.
3. it's (probably subjectively) more legible to separate the conditions and the keywords from the corresponding expressions as they then have different indentation levels.

I would love to have a configuration option in `ocamlformat` that would let me format conditionals that way.
I think it should be independent from the option configuring whether `if-then-else` should always break or be written on a single line if possible (so should `keyword-first` if you ask me) and be part of an option that describe how `if-then-else` should be formatted when they wrap.
Also to clarify my ideal behaviour here, I'd like this formatting to apply to all wrapping conditional expressions ie I'd like this:
```ocaml
let _ =
  if condition_1 then 1
  else if condition_2 then 2
  else 3
```
to be formatted as:
```ocaml
let _ =
  if condition_1 then
    1
  else if condition_2 then
    2
  else
    3
```
mostly on the account of 2. and 3.

Do these concerns seem sensible to you ? Would you be open to splitting `if-then-else` into several options and to have such formatting in `ocamlformat` ?

If so, I'm as I always happy to contribute!

jberdine:
This is related to #358. I think that it is a common desire to synchronize the breaking of the arms of conditional expressions. There are degrees, e.g. if all arms are short, it may be preferably to put them on the same line as their conditions, so it may make sense to leave options for that. But in particular having the ability to produce code in the style of your first example would be good. I expect that your very last example is too sparse for many. One smooth progression could be:
```
if cond1 then arm1 else if cond2 then arm2 else arm3
```
if it all fits on a line, then
```
if cond1 then arm1
[...]

PR #603, opened 2019-02-08
Improve: If-then-else = fit-or-vertical mode

gpetiot:
Should address #504 (cc @NathanReb let us know if it is what you had in mind)

The test_branch diff looks good.


test/passing/ite.ml
```diff
@@ -41,3 +41,64 @@ f
   else () )
 
 let () = if [@test] true then () else if [@other] true then ()
+
+[@@@ocamlformat "if-then-else=fit-or-vertical"]
```
Julow: I think it would be better to have a `ite_fit_or_vertical.ml` symlink to this file and `ite_fit_or_vertical.ml.opts` that add the option. Same for `keyword-first`. That way, future tests will be run with the 3 options.

gpetiot: @jberdine what is your opinion on introducing symbolic links for the tests?

jberdine:
Is the expectation of this that short cases will be formatted as
```
if cond1 then arm1 else if cond2 then arm2 else arm3
```
and then directly go all the way to
```
if cond1 then
  arm1
else if cond2 then
  arm2
else
  arm3
```
without the possibility of
```
if cond1 then arm1
else if cond2 then arm2
else arm3
```
?

If so, we should keep #504 open as this is only a partial solution.

gpetiot:
> Is the expectation of this that short cases will be formatted as
> 
> ```
> if cond1 then arm1 else if cond2 then arm2 else arm3
> ```
> and then directly go all the way to
> 
> ```
> if cond1 then
>   arm1
> else if cond2 then
>   arm2
> else
>   arm3
> ```
> without the possibility of
> 
> ```
> if cond1 then arm1
> else if cond2 then arm2
> else arm3
> ```
> ?
> 
> If so, we should keep #504 open as this is only a partial solution.

Isn't it the `compact` mode?

jberdine:
It is not entirely the same. What I mean above is that if it fits, format as
```
if cond1 then arm1 else if cond2 then arm2 else arm3
```
then if it does not fit, format as
```
if cond1 then arm1
else if cond2 then arm2
else arm3
```
as long as each cond and arm is short enough that *none* of them hits the margin, and once *one* does, us the fully-sparse form
```
if cond1 then
  arm1
else if cond2 then
  arm2
else
  arm3
```

gpetiot:
> It is not entirely the same. What I mean above is that if it fits, format as
> (...)
> as long as each cond and arm is short enough that _none_ of them hits the margin, and once _one_ does, us the fully-sparse form

I'm not sure it's possible to do that

jberdine:
I had not found a way to do it, but fresh eyes on a problem can sometimes see new ways.

As an easier compromise, what would people think about using `fit` if there are <= 2 arms, and otherwise using `fit-or-vertical`. This would allow short non-nested cases to be formatted compactly, without getting into the suboptimal cases of using fit for nested if-then-else expressions.

jberdine:
Also, can you add a test case where the first arm is an if-then-else, such as:
```ocaml
if some condition then
  if some nested condition then some action
  else some other action
else some default action
```
plus variants to test how the indentation of the arms correlates when they break in such cases.

gpetiot:
> Also, can you add a test case where the first arm is an if-then-else, such as:
> 
> ```ocaml
> if some condition then
>   if some nested condition then some action
>   else some other action
> else some default action
> ```
> 
> plus variants to test how the indentation of the arms correlates when they break in such cases.

I added some tests.

jberdine:
Looking at the diff produced by `tools/test_branch.sh if-then-else-sparse '' 'if-then-else=fit-or-vertical'` I see some cases such as:
```
@@ -1558,16 +1568,20 @@ end = struct
     prec_ast ast
     >>| fun prec_ast ->
     let cmp = Poly.compare prec_ctx prec_ast in
-    if cmp < 0 then (* ast higher precedence than context: no parens *)
-      false
-    else if cmp > 0 then (* context higher prec than ast: add parens *)
-      true
+    if cmp < 0 then
+      (* ast higher precedence than context: no parens *)
+    false
+    else if cmp > 0 then
+      (* context higher prec than ast: add parens *)
+    true
     else if
       Poly.(assoc_of_prec prec_ast = which_child && which_child <> Non)
```
I realize that comments are hard, but is the lost indentation of `false` here something you have tried to avoid?

gpetiot:
> Looking at the diff produced by `tools/test_branch.sh if-then-else-sparse '' 'if-then-else=fit-or-vertical'` I see some cases such as:
> 
> ```
> @@ -1558,16 +1568,20 @@ end = struct
>      prec_ast ast
>      >>| fun prec_ast ->
>      let cmp = Poly.compare prec_ctx prec_ast in
> -    if cmp < 0 then (* ast higher precedence than context: no parens *)
> -      false
> -    else if cmp > 0 then (* context higher prec than ast: add parens *)
> -      true
> +    if cmp < 0 then
> +      (* ast higher precedence than context: no parens *)
> +    false
> +    else if cmp > 0 then
> +      (* context higher prec than ast: add parens *)
> +    true
>      else if
>        Poly.(assoc_of_prec prec_ast = which_child && which_child <> Non)
> ```
> 
> I realize that comments are hard, but is the lost indentation of `false` here something you have tried to avoid?

The comments are attached to the expressions, would the following formatting be okay for you?
```ocaml
let foo =
  if cmp < 0 then
    (* ast higher precedence than context: no parens *) false
  else if cmp > 0 then
    (* context higher prec than ast: add parens *) true
  else if Poly.(assoc_of_prec prec_ast = which_child && which_child <> Non)
  then
    foo
```

jberdine:
I'd really like to try to avoid changing comments that end a line in the 
source into ones that have something following them.

gpetiot:
Keeping the comments at the end of the line compels us to use a hovbox containing the condition and the expression, and doing this makes the purpose of this PR impossible, i.e. having this:

```ocaml
let _ =
  if condition then
    let a = 1 in
    let b = 2 in
    a + b
  else if other_condition then
    12
  else
    0
```

instead of this:

```ocaml
let _ =
  if condition then
    let a = 1 in
    let b = 2 in
    a + b
  else if other_condition then 12
  else 0
```

test/passing/ite.ml
Julow: I did that in https://github.com/ocaml-ppx/ocamlformat/pull/619

jberdine:
:-(

I expect that at some point we will want to discuss with Pierre Weis et al about the possibility of adding some functionality to Format to support this sort of situation. Perhaps something like a type of box which either fits or forces all of its immediate sub-boxes to break. Such an extension might fit well into the existing interface of Format. But I'm unsure about how large a change it would involve for the current Format implementation.

test/passing/ite.ml
jberdine: Symlinks seem ok to me.

gpetiot:
As a short term solution, maybe we could add a warning in the `if-then-else = fit-or-vertical` documentation saying that comments may be moved, maybe add the ability to flag an option "unstable".

jberdine:
I'm not sure. If we had a reasonably clear plan that would resolve it soon, it would be a somewhat different situation, but at the moment we don't have any timeframe at all for such a change.

Do we know cases where the output with either the fit or keyword-first options is a hard blocker, where fit-or-vertical would unblock things?

NathanReb:
Sorry for stepping in so late, had a lot on my plate lately.

First of all thank you so much @gpetiot for your help on this, it's much appreciated. :heart: 

We've been wanting to use `ocamlformat` at Cryptosense for a while now and the if-then-else formatting was the last blocker. I'd be very happy with the current state of `fit-or-vertical`. We don't write that many comments except for documentation and as I mentionned in #504 I'm not interested in the intermediate case:
```ocaml
if cond1 then arm1
else if cond2 then arm2
else arm3
```

I don't want to put any kind of pressure on any of you and if you feel like this feature isn't ready to be merged yet and require some more refactoring beforehand I would totally understand.
On the other hand if it does get in I'd be the happiest person on earth and would spend the few days following the `0.9` release adding `ocamlformat` to all of our projects!

gpetiot:
If we need a new type of box to have this behavior, it will take more time so I remove the 0.9 tag on this PR for now.
Any contribution to `Format` (our own vendored version) in this direction is welcomed :)

NathanReb:
Ok!

Thank you for having tried to squeeze this in 0.9 anyway! I will happily contribute as soon as I get more time!

gpetiot:
@jberdine my last commit 0c01892 fixes the alignment of keywords in case of parentheses, I fixed it for fit-or-vertical because it looked weird. Do you want me to fix it for the default option as well or is it intended?

src/Fmt_ast.ml
```diff
@@ -1696,6 +1699,36 @@ and fmt_expression c ?(box = true) ?epi ?eol ?parens ?(indent_wrap = 0) ?ext
...
+                              $ fmt_or parens "@;<1 2>" "@ "
+                              $ fmt "then" )
+                        | None -> fmt "else" )
+                      $ fmt_if parens_bch " (" $ fmt "@;<1 2>"
+                      $ fmt_expression c ~pro ~eol:(fmt "@;<1 2>")
+                          ~box:false ~parens:false xbch
+                      $ fmt_if parens_bch
+                          ( if c.conf.indicate_multiline_delimiters then " )"
+                          else ")" ) )
+                    $ fmt_if (not last) "@ "
```
Julow: This function is extremely big. Also, I see a bit of repetition, would the `Params` thing help ?

gpetiot: It will once we find a common box structure for a the cases. I didn't look at that, maybe it could take some time, I think this factorization step could be done in a future PR, since it will also impact the existing cases and the diff would be less readable than it is now.

gpetiot:
@NathanReb does the output after the commit 0c01892 look better to you or do you prefer the previous output?
Maybe this keyword alignment could be fixed across all options in a separate PR.

test/passing/ite_fit_or_vertical.ml.ref
```diff
@@ -0,0 +1,137 @@
...
+    this is more )
+  else
+    e
+
+;;
+f
+  ( if loooooooooooooooooooooooooooooooooooooooooooooooooooooooooong then
+      ()
+    else
+      () )
```
jberdine: Here, it looks like indentation starts from the point before `if` rather than the point before `(`. Is this correct, and intended? If so, it would be good to move that into a separate option. There are other cases where indentation starts from before leading parens (and I currently consider it a bug when this doesn't happen), and while it's expected for opinions to vary, we should aim to work toward a point where each value of the option is consistently obeyed.

NathanReb:
> @NathanReb does the output after the commit `0c01892` look better to you or do you prefer the previous output?

If you're talking about keywords being aligned together rather than with the opening parens then I do indeed prefer after `0c01892`! That might be subjective though.

test/passing/ite_fit_or_vertical.ml.ref
gpetiot: The indentation starting from the `if` was precisely the point of the last commit, but I can revert it, ship the PR without it (and keep the current indentation behavior), and we can put the changes of this last commit in another PR indeed.

gpetiot:
Alright I reverted the last commit about the indentation, I will submit a separate PR adding an option for this new indentation.
Is this ready to be merged now @jberdine ?

test/passing/ite_fit_or_vertical.ml.ref
```diff
@@ -0,0 +1,137 @@
+let _ = if b then e else ( e1 ; e2 )
```
trefis: I can't help but feel there shouldn't be spaces inside the parentheses above.
It looks to me like the `indicate_multiline_delimiters` option is used incorrectly here. That is, if it's on ocamlformat will always print spaces, even though the parentheses might be on the same line.

PR #604, opened 2019-02-08
Fix ocamlformat_reason

### LINKED ISSUE #596: Add some tests for ocamlformat_reason

hhugo:
I suspect ocamlformat_reason has been broken for a while.
- fix the implementation
- add a test to spot regressions, fix #596

### DIFF
--- a/ocamlformat_reason.opam
+++ b/ocamlformat_reason.opam
@@ -31,7 +31,7 @@ depends: [
   "octavius" {>= "1.2.0"}
   "stdio"
   "uutf"
-  "reason" {= "3.3.7"}
+  "reason" {>= "3.2.0"}
 ]
 
 available: [ ocaml-version >= "4.04.1" & ocaml-version < "4.05" ]
--- a/ocamlformat_reason.opam
+++ b/ocamlformat_reason.opam
@@ -34,8 +34,4 @@ depends: [
   "reason" {= "3.3.7"}
 ]
 
-conflicts: [
-  "lwt" {>="3.1.0"} (* ocamlformat does not need lwt, but reason requires utop requires lwt, so exclude 3.1.0 since it does not build *)
-]
-
 available: [ ocaml-version >= "4.04.1" & ocaml-version < "4.05" ]
--- a/ocamlformat_reason.opam
+++ b/ocamlformat_reason.opam
@@ -21,15 +21,21 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
+  "ocaml" {>= "4.05"}
   "base" {>= "v0.11.0"}
   "base-unix"
   "cmdliner"
-  "dune" {build}
+  "dune" {build & >= "1.1.1"}
   "fpath"
-  "ocaml-migrate-parsetree" {>= "1.0.6"}
-  "octavius"
-  "reason" {= "1.13.4"}
+  "ocaml-migrate-parsetree" {>= "1.0.10"}
+  "octavius" {>= "1.2.0"}
   "stdio"
-  "lwt" {< "3.1.0"} (* ocamlformat does not need lwt, but reason requires utop requires lwt, so exclude 3.1.0 since it does not build *)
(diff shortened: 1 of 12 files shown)


src/Source.ml
```diff
@@ -18,29 +18,30 @@ let create s = s
...
 let merge (l1 : Location.t) ~(sub : Location.t) =
   let base = l1.loc_start.pos_cnum in
   { l1 with
     loc_start= {l1.loc_start with pos_cnum= base + sub.loc_start.pos_cnum}
   ; loc_end= {l1.loc_end with pos_cnum= base + sub.loc_end.pos_cnum} }
 
-let string_from_loc t (l : Location.t) =
-  let pos = l.loc_start.pos_cnum in
-  let len = Location.width l in
-  String.sub t ~pos ~len
```
jberdine: +1

test-reason/test-gen.ml
```diff
@@ -0,0 +1,17 @@
...
+let x = 3
+
+type schoolPerson = Teacher | Director | Student of string
+
+let greeting person =
+  match person with
+  | Teacher -> ( ("Hey Professor!" [@reason.raw_literal "Hey Professor!"]) )
+  | Director -> (
+      ("Hello Director." [@reason.raw_literal "Hello Director."]) )
+  | ((Student ("Richard"[@reason.raw_literal "Richard"]))[@explicit_arity])
```
jberdine: clearly no one is using this :-\

src/Reason.ml
```diff
@@ -13,25 +13,91 @@
...
+      * 'concrete
+      -> ast
+
+let find_magic magic x =
+  let rec loop = function
+    | [] ->
+        let prefix = String.sub magic ~pos:0 ~len:9 in
+        if
+          String.equal prefix
+            (String.sub Ast_402.Config.ast_impl_magic_number ~pos:0 ~len:9)
```
jberdine: 402?

test-reason/test-gen.ml
hhugo: :)

src/Reason.ml
hhugo: the version doesn't matter, we only care about the prefix.

hhugo:
I've updated version constraint

ocamlformat_reason.opam
```diff
@@ -21,15 +21,21 @@ build: [
...
+  "ocaml-migrate-parsetree" {>= "1.0.10"}
+  "octavius" {>= "1.2.0"}
   "stdio"
-  "lwt" {< "3.1.0"} (* ocamlformat does not need lwt, but reason requires utop requires lwt, so exclude 3.1.0 since it does not build *)
+  "uutf"
+  "reason" {= "3.3.7"}
 ]
+
+conflicts: [
+  "lwt" {>="3.1.0"} (* ocamlformat does not need lwt, but reason requires utop requires lwt, so exclude 3.1.0 since it does not build *)
```
hcarty: This should not be required anymore IIRC - the utop wrapper was split into a separate `rtop` package.

src/Reason.ml
```diff
@@ -13,25 +13,92 @@
...
 
-let input_impl = input Config.ast_impl_magic_number
+let input_impl _conf ic :
+    Migrate_ast.Parsetree.structure Translation_unit.with_comments =
+  let _filename, comments, x = input ic in
+  match x with
+  | Impl (bin_version, ast) ->
+      let module Bin_version = (val bin_version) in
+      let to_current =
+        Migrate_parsetree.Versions.(migrate (module Bin_version) ocaml_407)
```
jberdine: Probably better to extend/use `Migrate_ast` so that we only have to remember to change one file when changing ast versions.

ocamlformat_reason.opam
hhugo: ack

src/Reason.ml
hhugo: done

PR #607, opened 2019-02-09
Fix reason wrt strings and chars

hhugo:
Fix the fallback code path when sources are not available.
Strings and chars were not properly escaped.


jberdine:
Sorry, I missed it before, but could `test-reason` be renamed to `test/reason` or something similar under `test`? Because of file header bots and whatnot...

hhugo:
I did the move in https://github.com/ocaml-ppx/ocamlformat/pull/608 assuming it could be merge soon.

PR #608, opened 2019-02-09
Improve reason

hhugo:
Sanitise the ast parsed from reason files before formatting it. In particular
- tune the ast in presence of "explicit_arity" (previously done inside `norm`) 
- remove special reason attributes


jberdine:
I was trying to keep all the reason-specific stuff in `Reason`. Couldn't the special cases in these changes be implemented as a parsetree transformation in `Reason.mapper`? Note that previously this strategy worked for handling e.g. the `explicit_arity` attributes.

hhugo:
Removing special attributes can probably be done in Reason.ml.

Leveraging the content of attributes to recover original string literal escaping is more tricky.
I see that https://github.com/ocaml-ppx/ocamlformat/pull/611 is enough to recover the escaping from the original source. So I guess it's fine to not rely on the special attributes now.

jberdine:
For cases where the concrete source string is used, a better long-term solution is probably to fix a scheme of attributes that represent the needed information. Then when the concrete syntax is available, it would be queried at the beginning to populate the attributes all at once, and then the core formatting logic would not need to refer to the concrete syntax again. The reason support could then translate the reason attributes or other info to the attributes the formatter needs.

Related:
The uses of the concrete syntax for comment placement could also fall into this category, if we changed the way we store comments from hashtables to attributes. This would be a step toward better reason interop, if refmt also changed to add comments as attributes.

hhugo:
Storing comments in attributes will limit where comments can be attached. I don't know if this would be problematic or not

jberdine:
Do you know of places in the parsetree where there is a `Location.t` and no attributes? A while ago I did a search and think the only case at that point was `with_constraint`.

hhugo:
I didn't check. Note that the ast of ocaml 4.08 has received a fair amount of changes.

hhugo:
I've moved all the logic in `Reason.ml`.
This PR is currently based on top of #611 to see how `test-reason` behave when sources are available.
It should be moved on to on master before merging.

hhugo:
Please just review https://github.com/ocaml-ppx/ocamlformat/pull/608/commits/01b42e617de324c93723191707bd63a47c2699d4

src/Reason.ml
```diff
@@ -11,69 +11,183 @@
...
-  in
-  let comments = List.map comments ~f:(fun c -> (c.text, c.location)) in
-  (filename, comments, find_magic magic ast)
-
-let input_impl _conf ic :
-    Migrate_ast.Parsetree.structure Translation_unit.with_comments =
-  let _filename, comments, x = input ic in
+module Binary_reason = struct
+  (** Type of Reason binary serialized data, which must agree with the type
+      used by the implementation of `refmt`. *)
```
jberdine: This probably goes on `reason_data` below?

src/Reason.ml
```diff
@@ -11,69 +11,183 @@
...
+
+  type category = int
+
+  (* | EndOfLine | SingleLine | Regular *)
```
jberdine: docstring?

PR #611, opened 2019-02-10
Do not rely on the file-system to format sources (second attempt)

hhugo:


src/ocamlformat.ml
```diff
@@ -46,80 +44,85 @@ let intf : _ Translation_unit.t =
...
+  | None -> Out_channel.output_string Out_channel.stdout data
+  | Some output_file -> Out_channel.write_all output_file ~data
+
 ;;
 match Conf.action with
 | Inplace inputs ->
-    let results : Translation_unit.result list =
-      List.map inputs
+    let output_file = None in
+    let results =
```
jberdine: maybe rename results to errors

src/Translation_unit.ml
```diff
@@ -406,18 +279,138 @@ let parse_print (XUnit xunit) (conf : Conf.t) ~input_name ~input_file ic
...
+  Location.input_name := input_name ;
+  (* iterate until formatting stabilizes *)
+  let rec print_check ~i ~(conf : Conf.t) (t : _ with_comments) ~source :
+      (string, error) Result.t =
+    dump_ast ~suffix:".old" t.ast ;
+    let format ~box_debug =
+      let buffer = Buffer.create (String.length source) in
+      let source_t = Source.create source in
+      let cmts_t = xunit.init_cmts source_t conf t.ast t.comments in
+      let fs = Format.formatter_of_buffer buffer in
```
jberdine: maybe change this Format to Format_, and the same for the next 2, and then remove the module alias at the top and `Caml.` prefixes on the other occurrences of Format?

src/Translation_unit.ml
```diff
@@ -137,169 +121,39 @@ let ellipsis n msg =
...
+  tmp
+
+let dump_ast ~input_name ?output_file ~suffix fmt =
+  if Conf.debug then
+    let ext = ".ast" in
+    let (_filename : string) =
+      with_file input_name output_file suffix ext (fun oc ->
+          fmt (Caml.Format.formatter_of_out_channel oc) )
+    in
+    ()
```
jberdine: How about `with_file ... |> (ignore : string -> _)` instead of `let ... = ... in ()`?

src/Translation_unit.ml
jberdine: see commit I pushed to your branch

hhugo:
I've pushed a commit to address you feedback.

hhugo:
@jberdine, can I get another small review including the last 3 commits I pushed ?

Issue #610, opened 2019-02-09
Ignored ocamlformat attribute


jberdine:
Formatting
```
[@@@ocamlformat "wrap-comments=true"]

type t =
  | Aaaaaaaaaa (* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. *)
  | Bbbbbbbbbb
```
ignores the attribute while passing `--wrap-comments` does not.

PR #613, opened 2019-02-11
Fix missing break before comment

### LINKED ISSUE #609: Missing break before comment
#548 introduced a regression where breaks are missing before comments. Consider:
```
type t =
  | Aaaaaaaaaa (* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. *)
  | Bbbbbbbbbb
```
With `wrap-comments=true` this formats to:
```
type t =
  | Aaaaaaaaaa (* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
                  do eiusmod tempor incididunt ut labore et dolore magna
                  aliqua. *)
  | Bbbbbbbbbb
```
while it used to format to:
```
type t =
  | Aaaaaaaaaa
  (* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
     tempor incididunt ut labore et dolore magna aliqua. *)
  | Bbbbbbbbbb
```
While the previous indentation of the comment is suboptimal, the current indentation is wrong.

gpetiot:
Fix #609 

The diff of this form:

```ocaml
--- a/stdlib/camlinternalFormatBasics.ml
+++ b/stdlib/camlinternalFormatBasics.ml
@@ -256,11 +256,14 @@ Format_subst_ty constructor).
 type block_type =
   | Pp_hbox (* Horizontal block no line breaking *)
   | Pp_vbox (* Vertical block each break leads to a new line *)
-  | Pp_hvbox (* Horizontal-vertical block: same as vbox, except if this block
+  | Pp_hvbox
+  (* Horizontal-vertical block: same as vbox, except if this block
                  is small enough to fit on a single line *)
-  | Pp_hovbox (* Horizontal or Vertical block: breaks lead to new line
+  | Pp_hovbox
+  (* Horizontal or Vertical block: breaks lead to new line
                  only when necessary to print the content of the block *)
-  | Pp_box (* Horizontal or Indent block: breaks lead to new line
+  | Pp_box
```
is in fact closer to the original source (due to the non-wrapping mode).

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2785,14 +2785,15 @@ and fmt_constructor_declaration c ctx ~first ~last:_ cstr_decl =
   $ Cmts.fmt_before c.cmts pcd_loc
   $ Cmts.fmt_before c.cmts loc
   $ fmt_or_k first (if_newline "| ") (fmt "| ")
-  $ hovbox 2
-      ( hvbox 2
-          ( Cmts.fmt c.cmts loc
-              (wrap_if (is_symbol_id txt) "( " " )" (str txt))
-          $ fmt_constructor_arguments_result c ctx pcd_args pcd_res )
-      $ fmt_attributes c ~pre:(fmt "@;") ~key:"@" atrs
-      $ fmt_docstring_padded c doc )
-  $ Cmts.fmt_after c.cmts ~pro:(fmt " ") ~epi:(fmt "@ ") pcd_loc
+  $ hvbox 0
+      ( hovbox 2
+          ( hvbox 2
+              ( Cmts.fmt c.cmts loc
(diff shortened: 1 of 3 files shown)


jberdine:
Does your comment mean that you think that the break before the comment should be added only in wrap mode?

gpetiot:
> Does your comment mean that you think that the break before the comment should be added only in wrap mode?

I don't know I need to understand the issue #610 about the wrap mode first, but at least we fixed the regression.

gpetiot:
> Does your comment mean that you think that the break before the comment should be added only in wrap mode?

After fixing the bug yes, done in #615.

PR #619, opened 2019-02-12
Add module_annotation option

Julow:
*Work in progress*

It has two values:
- `sparse` the current behavior, default:

```ocaml
let _ =
  ( module struct
    type t
  end
  : S
    with type t = y )
```

- `compact` don't break if possible the module signature:

```ocaml
let _ =
  ( module struct
	type t
  end : S with type t = y )
```


test/passing/js_source.ml.ref
```diff
@@ -3225,8 +3225,7 @@ end
```
Julow: The js profile has `module-annotation=compact`

test/passing/module_annotation_compact.ml
```diff
@@ -0,0 +1 @@
+module.ml
```
Julow: This is a symlink so that future tests are tested with both modes

test/passing/module_attributes.ml
```diff
@@ -6,7 +6,7 @@ end [@warning "item"] [@@warning "structure"]
 
 include M [@warning "item"] [@@warning "structure"]
 
-include (M : S) [@warning "item"] [@@warning "structure"]
+include (M : S ) [@warning "item"] [@@warning "structure"]
```
gpetiot: This looks like a regression

test/passing/source.ml.ref
```diff
@@ -3488,7 +3487,7 @@ module SSMap = struct
 end
 
 let ssmap =
-  (module SSMap
+  ( module SSMap
```
gpetiot: This also looks like a regression (unless it's intended?)

Julow:
This PR is WIP for too long and will conflict badly with #665. I close it and try it again on top of #665.

PR #631, opened 2019-02-15
Improve emacs setup

Khady:
ocamlformat doesn't depend on merlin. So it is better to add the keybinding in tuareg's map.


gpetiot:
Looks good, thanks.

PR #632, opened 2019-02-15
tools/update_tests.sh --all

Julow:
Add an option to update all the tests

### DIFF
--- a/tools/update_tests.sh
+++ b/tools/update_tests.sh
@@ -16,8 +16,6 @@
 # The arguments are the names of the tests to update (without the directory),
 #  e.g.:  update_tests.sh comments.ml comments_in_record.ml
 
-set -e
-
 ROOT=$(git rev-parse --show-toplevel)
 EXE="$ROOT"/_build/dev/src/ocamlformat.exe
 OLD_DIR=$(pwd)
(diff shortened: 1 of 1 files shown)


tools/update_tests.sh
```diff
@@ -16,8 +16,6 @@
 # The arguments are the names of the tests to update (without the directory),
 #  e.g.:  update_tests.sh comments.ml comments_in_record.ml
 
-set -e
```
Julow: Some tests cause errors (`error1.ml`) and it is not needed by the script.

tools/update_tests.sh
```diff
@@ -52,6 +50,11 @@ function update () {
     cd $OLD_DIR
 }
 
+if [[ $1 = "--all" ]]; then
```
Julow: Very basic option handling.

gpetiot: In #471 I've used the `test` keyword instead of double brackets (for portability concerns), maybe we could use it here as well (but not blocking if you choose not to).

Julow: The script starts with `#!/usr/bin/env bash`. Also I use arrays. I'm not sure it is usefull but I can try to make it more "portable".

PR #636, opened 2019-02-26
don't include ocamlformat_diff in ocamlformat

Khady:
Right now, the oamlformat_diff tool is part of the ocamlformat package. It looks like a mistake. And it means than an a fresh opam switch, `opam pin add ocamlformat --dev` will fail.


gpetiot:
Good catch, thanks!

Issue #627, opened 2019-02-12
Keep tag-only doc comments on the same line


Julow:
```
include Module (** @inline *)
```

Disabled if the comment contains text

PR #637, opened 2019-02-26
Fit tag-only doc comments

Julow:
Add the `doc-tagonly-fit` options to fix https://github.com/ocaml-ppx/ocamlformat/issues/627:

```
include Module (** @inline *)
```

Implemented on open, include and module bindings.


test/passing/tag_only.ml
```diff
@@ -0,0 +1,87 @@
...
+
+open Module (** @author A *)
+
+open Module (** @inline *)
+
+include Abc (** @inline *)
+
+include struct
+  type t
+end (** @inline *)
```
Julow: This is not really intended.

gpetiot:
I'm not a big fan of the option's name (random suggestions: `fit-tag-only-doc-comments` if you want to keep a boolean, or `tag-only-doc-comments = sparse / compact` if you want a sparse/compact variant).
Also don't forget to update the sparse / compact preset profiles.
Otherwise it looks good.

Julow:
I agree. What about `doc-comments-tag-only = fit / default` ? So that it has the `doc-comments` prefix and the value `default` means no special treatment

gpetiot:
By the way I think you should check the interaction of this option with `doc_comments`, non-fitting tags should not always go before the corresponding code.

Julow:
> By the way I think you should check the interaction of this option with the `doc_comments` option, non-fitting tags should not always go before the corresponding code.

It seems like `doc_comments=after` in only handled for `val` and `external` statements. This is PR only implements `doc_comments_tag_only=fit` for `include`, `open` and `module` statements.

gpetiot:
Alright looks good now, I will merge it.

Issue #621, opened 2019-02-12
Break between module "with" constraints if expression does not fit


Julow:
Current formatting (with `profile=janestreet`):

```
let _ =
  (module Module.Name
  : Module.Signature
    with type long_long_type = a and type very_long_long_type = b)
```

Expected:

```
let _ =
  (module Module.Name
  : Module.Signature
    with type long_long_type = a
     and type very_long_long_type = b)
```

gpetiot:
Fixed with #639

PR #639, opened 2019-02-27
Improve: break the list of 'with type' module constraints

gpetiot:
Should address #621 
I think we always want this new behavior, hence no new option

### DIFF
--- a/test/passing/js_source.ml.ref
+++ b/test/passing/js_source.ml.ref
@@ -3425,7 +3425,7 @@ let make_set (type s) cmp =
     let compare = cmp
   end)
   : Set.S
-      with type elt = s)
+    with type elt = s)
 ;;
 
 (* No type annotation here *)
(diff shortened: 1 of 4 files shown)


test/passing/js_source.ml.ref
```diff
@@ -3649,20 +3649,25 @@ end
...
+     and type data = string
+     and type map = SSMap.map)
 ;;
 
 let ssmap :
-    (module MapT with type key = string and type data = string and type map = SSMap.map)
-    =
+    (module MapT with type key = string
+                  and type data = string
+                  and type map = SSMap.map) =
```
gpetiot: So far no success breaking before `with` without introducing more regressions elsewhere.

Julow: What kind of regressions ?

gpetiot: You can look at the diff of my last commit 2d68308, there are net improvements (even if we break more lines sometimes) but for some cases I'm not sure if it's really better

gpetiot:
Any thoughts on this? Can I merge it?

PR #642, opened 2019-03-02
Build: update to base v0.12

jberdine:

PR #646, opened 2019-03-05
Fix build on OCaml 4.06.1

Julow:
Unrelated to https://github.com/ocaml-ppx/ocamlformat/pull/642 but since it was merged, build was broken on my machine.

### DIFF
--- a/src/Source.ml
+++ b/src/Source.ml
@@ -92,9 +92,9 @@ let extend_loc_to_include_attributes t (loc : Location.t)
           let loc =
             match payload with
             | PStr [] -> loc
-            | PStr l -> (List.last_exn l).pstr_loc
+            | PStr l -> (List.last_exn l).Parsetree.pstr_loc
             | PSig [] -> loc
-            | PSig l -> (List.last_exn l).psig_loc
+            | PSig l -> (List.last_exn l).Parsetree.psig_loc
             | PTyp c -> c.ptyp_loc
             | PPat (p, None) -> p.ppat_loc
             | PPat (_, Some e) -> e.pexp_loc
--- a/src/Reason.ml
+++ b/src/Reason.ml
@@ -196,7 +196,7 @@ type 'a t =
 let input_bin_impl ic : Migrate_ast.Parsetree.structure t =
   let origin_filename, comments, x = Binary_reason.input ic in
   match x with
-  | Impl (bin_version, ast) ->
+  | Binary_reason.Impl (bin_version, ast) ->
       let module Bin_version = (val bin_version) in
       let to_current =
         Migrate_parsetree.Versions.(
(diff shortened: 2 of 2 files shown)


src/Source.ml
```diff
@@ -92,9 +92,9 @@ let extend_loc_to_include_attributes t (loc : Location.t)
           let loc =
             match payload with
             | PStr [] -> loc
-            | PStr l -> (List.last_exn l).pstr_loc
+            | PStr l -> (List.last_exn l).Parsetree.pstr_loc
```
Julow: This was just a warning.

test/passing/js_source.ml.ref
```diff
@@ -9792,4 +9792,4 @@ module Indexop = struct
   let x, y, z = Def.(h.%["one"], h.%("two"), h.%{"three"})
 end
 
-type t = |
+(* type t = | *)
```
Julow: This change may not be required, we can just assume that we are using 4.07 when running tests ?

gpetiot: So you were using 4.06.1, didn't you already get errors on this construct?

Julow: No.

gpetiot: Do you have an idea of what's happening? Why do we have to comment this now? It causes an error on 4.06.1 I don't understand why you didn't have this error previously.

hhugo:
ocamlformat compiles/installs fine with ocaml 4.06.0

```
$ ocamlc -version
4.06.1
$ make clean
rm -Rf _coverage
find ./ -name 'bisect*.out' -delete
rm -rf _build dune-workspace
$ dune build -p ocamlformat
$
```
note that call `make` will fail because more warning are enabled in that case.

hhugo:
`ocamlformat_reason` doesn't work and this PR fixes this.

src/Source.ml
hhugo: I think it's fine to have warning trigger with older version of ocaml, @jberdine, any opinion ?

test/passing/js_source.ml.ref
hhugo: I would not expect tests to pass with older version of the compiler.

jberdine: Yes, there are too many moving parts to expect tests to pass with old compilers. My position is just one, heavily influenced by a development context that is perhaps uncommon, and perhaps not widely shared, but I don't think it's worth the time to worry about 4.06 when 4.08 is about to be released. I am much less worried about requiring a current compiler to build ocamlformat than I am about generating code that only the current compiler can parse, but that is a different question. For building ocamlformat itself, I'd say just only develop using the current compiler (to save your own sanity by avoiding chasing down the consequences of the differences between the parsers in different versions) and bump the minimum required version.

Julow: I agree for the tests, I removed the commit.

Julow:
I think this is ready to be merged.

PR #648, opened 2019-03-05
Link to man page from readme

### LINKED ISSUE #647: Make ocamlformat documentation easy to find online
It would be good to put up the ocamlformat documentation somewhere online or at least in an easy-to-find location in the repo so that potential users don't have to install ocamlformat to find it. Best-case scenario would be something like this: https://scalameta.org/scalafmt/docs/configuration.html

yawaramin:
Fix #647

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ OCamlFormat is documented in its man page and through its internal help:
 * `ocamlformat --help`
 * `man ocamlformat`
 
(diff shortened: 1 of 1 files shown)


yawaramin:
Hello @jberdine @hhugo for your viewing pleasure :-)

gpetiot:
Thanks

PR #651, opened 2019-03-06
Fix cmt placement infix op

### LINKED ISSUE #617: Comment placement regression in infix op applications
#548 seems to have had some unfortunate effects on the placement of comments in infix op applications. Consider:
```ocaml
let () =
  xxxxxxxxxx
  || (* xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx *)
     xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
Before, this formatted to itself. After, it formats to:
```ocaml
let () =
  xxxxxxxxxx
  (* xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx *)
  || xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
Note that the comment has moved from after the infix op to after it.

gpetiot:
Fix #617 
No regression with test_branch

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -270,20 +270,20 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
         let same_line_as_next l =
           next.loc_start.pos_lnum = l.loc_start.pos_lnum
         in
-        let char_before ~in_ =
+        let sequence_symbol_before =
           let pos_cnum = prev.loc_end.pos_cnum - 1 in
           let loc_end = {prev.loc_end with pos_cnum} in
           let char_loc = {prev with loc_start= loc_end} in
           let str = Source.string_at t.source char_loc in
-          List.mem in_ str ~equal:String.equal
+          let c = str.[0] in
+          Char.equal c ';' || Char.equal c '|' || Char.equal c '&'
         in
-        let symbol_before = char_before ~in_:[";"; "|"] in
         let prev, next =
           if not (same_line_as_prev next) then
             let next, prev =
               List.partition_tf cmtl ~f:(fun (_, l1) ->
                   same_line_as_next l1
-                  || (same_line_as_prev l1 && symbol_before)
+                  || (same_line_as_prev l1 && sequence_symbol_before)
                   || not (same_line_as_prev l1) )
             in
             (prev, next)
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -270,16 +270,28 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
         let same_line_as_next l =
           next.loc_start.pos_lnum = l.loc_start.pos_lnum
         in
+        let is_infix_id i =
+          match (i.[0], i) with
+          | ( ( '$' | '%' | '*' | '+' | '-' | '/' | '<' | '=' | '>' | '|'
+              | '&' | '@' | '^' | '#' )
+            , _ )
+           |( _
+            , ( "!=" | "land" | "lor" | "lxor" | "mod" | "::" | ":=" | "asr"
+              | "lsl" | "lsr" | "or" | "||" ) ) ->
+              true
+          | _ -> false
+        in
         let infix_symbol_before =
           let pos_cnum = prev.loc_end.pos_cnum - 1 in
           let loc_end = {prev.loc_end with pos_cnum} in
-          let char_loc = {prev with loc_start= loc_end} in
-          let str = Source.string_at t.source char_loc in
-          match str.[0] with
-          | '$' | '%' | '*' | '+' | '-' | '/' | '<' | '=' | '>' | '|'
-           |'&' | '@' | '^' | '#' | ';' ->
-              true
-          | _ -> false
+          match Source.position_before t.source loc_end with
+          | Some loc_start ->
+              if loc_start.pos_cnum < prev.loc_end.pos_cnum then
+                let op_loc = {prev with loc_start} in
+                let str = Source.string_at t.source op_loc in
+                String.equal str ";" || is_infix_id str
+              else false
+          | None -> false
         in
         let prev, next =
           if not (same_line_as_prev next) then
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -270,17 +270,6 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
         let same_line_as_next l =
           next.loc_start.pos_lnum = l.loc_start.pos_lnum
         in
-        let is_infix_id i =
-          match (i.[0], i) with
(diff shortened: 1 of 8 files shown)


src/Cmts.ml
```diff
@@ -270,20 +270,20 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
-          let loc_end =
-            {prev.loc_end with pos_cnum= prev.loc_end.pos_cnum - 1}
-          in
+        let char_before ~in_ =
+          let pos_cnum = prev.loc_end.pos_cnum - 1 in
+          let loc_end = {prev.loc_end with pos_cnum} in
           let char_loc = {prev with loc_start= loc_end} in
           let str = Source.string_at t.source char_loc in
-          String.equal str ";"
+          List.mem in_ str ~equal:String.equal
```
Julow: I'm not fan of the `List.mem` approach, I'd prefer pattern matching or multiple `String.equal`

src/Cmts.ml
```diff
@@ -270,20 +270,20 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
         in
+        let symbol_before = char_before ~in_:[";"; "|"] in
```
Julow: Why these characters ? Eg. `&&`:
```
let () =
  xxxxxxxxxx
  (* xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx *)
  && xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

I don't really understand this code but why can't this be done in `Fmt_ast.ml` with more informations ? (eg. `Pexp_apply ... when is_infix_id`)

gpetiot: This is done in Cmts.ml because the function deciding where the comments should be attached to (after the previous entity or before the next one) is located in this file.
Sure we can add more characters, I think maybe we could simply test if it's not an alphanumeric character (and not `)`)

src/Cmts.ml
gpetiot: done

src/Cmts.ml
gpetiot: done

src/Cmts.ml
```diff
@@ -270,20 +270,20 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
-            {prev.loc_end with pos_cnum= prev.loc_end.pos_cnum - 1}
-          in
+        let sequence_symbol_before =
+          let pos_cnum = prev.loc_end.pos_cnum - 1 in
+          let loc_end = {prev.loc_end with pos_cnum} in
           let char_loc = {prev with loc_start= loc_end} in
           let str = Source.string_at t.source char_loc in
-          String.equal str ";"
+          let c = str.[0] in
+          Char.equal c ';' || Char.equal c '|' || Char.equal c '&'
```
Julow: Why not all infix operators ? `Ast.is_infix_id` do this.

gpetiot: There would be a cyclic dependency between Ast and Cmts, I copied 2 lines of pattern matching from `Ast.is_infix_id` instead. Re-organizing all these utility functions into another file is out of the scope of this PR we can do it later.

Julow: It does not handle operators like `land`. I think you should do this while you can otherwise we will forget and won't do it.

src/Cmts.ml
```diff
@@ -270,20 +270,35 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
           next.loc_start.pos_lnum = l.loc_start.pos_lnum
         in
-        let colon_before =
-          let loc_end =
-            {prev.loc_end with pos_cnum= prev.loc_end.pos_cnum - 1}
-          in
-          let char_loc = {prev with loc_start= loc_end} in
-          let str = Source.string_at t.source char_loc in
-          String.equal str ";"
+        let is_infix_id i =
```
gpetiot: Copied this function for now, not sure where to put it, maybe in import/Import.ml? or creating a Utility.ml file?

src/Cmts.ml
gpetiot: done

src/Cmts.ml
Julow: I have nothing against creating a new file. Alternatively, you could move the few functions that handle comments in `Ast.ml` to `Cmts.ml` to reverse the dependency. Both make sense to me.

gpetiot: I parameterized the Ast functions with has_before/has_after functions instead, to remove the dependency to Cmts.ml

src/Cmts.ml
```diff
@@ -270,20 +270,24 @@ let partition_after_prev_or_before_next t ~prev cmts ~next =
...
-          String.equal str ";"
+        let infix_symbol_before =
+          let pos_cnum = prev.loc_end.pos_cnum - 1 in
+          let loc_end = {prev.loc_end with pos_cnum} in
+          match Source.position_before t.source loc_end with
+          | Some loc_start ->
+              if loc_start.pos_cnum < prev.loc_end.pos_cnum then
+                let op_loc = {prev with loc_start} in
+                let str = Source.string_at t.source op_loc in
+                String.equal str ";" || Ast.is_infix_id str
```
Julow: Thanks for the cleanup. It does not work with keyword operators:

```
let () =
  xxxxxxxxxx
  (* xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx *)
  land xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

and or-patterns:

```
let ( xxxxxxxxxx
    (* xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx *)
    | xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx )
    =
  a
```

gpetiot: I fixed the first case, as for the patterns, having the comment before the `|` is consistent with how it's done for other patterns kinds (see in pattern matching).

PR #657, opened 2019-03-08
Improve: check flag to check whether the input files already are formatted

### LINKED ISSUE #655: Add a --check flag for pre-commit hooks
Something I value highly in Prettier et al., is the ability to ensure divergent code doesn't make it into the master branch. Towards that end, most of them provide a feature [like `--check`](https://prettier.io/docs/en/cli.html#check), that doesn't actually *write* reformatted files to disk, but simply *checks* whether reformatting causes any changes, and dumps a list of disagreeing files to stdout if there are any.

If it exists with a failure status-code, this can even be used as a git [`pre-commit` hook](https://git-scm.com/docs/githooks#_pre_commit) to ensure that commits with non-conforming code are rejected — and then can be used with tools like [`lint-staged`](https://github.com/okonet/lint-staged#examples) to slot in alongside similar tools in one's codebase.

gpetiot:
Fix #655

### DIFF
--- a/src/ocamlformat_reason.ml
+++ b/src/ocamlformat_reason.ml
@@ -62,6 +62,7 @@ match Conf.action with
 | In_out ({kind= `Use_file; _}, _) ->
     user_error "Cannot convert Reason code with --use-file" []
 | Inplace _ -> user_error "Cannot convert Reason code with --inplace" []
+| Check _ -> user_error "Cannot check Reason code with --check" []
 | In_out
     ( {kind= (`Impl | `Intf) as kind; file= "-"; name= input_name; conf}
     , output_file ) -> (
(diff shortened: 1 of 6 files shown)


src/ocamlformat_reason.ml
```diff
@@ -95,3 +95,24 @@ match Conf.action with
         to_output_file output_file s ;
         Caml.exit 0
     | Error _ -> Caml.exit 1 )
+| Check inputs ->
```
hhugo: Similar to `--inplace`, `Check` doesn't make sense for ocamlformat_reason - the source is a binary file

jberdine:
Do we need this since we have `-n 1`?

gpetiot:
> Do we need this since we have `-n 1`?

For me the benefit is to be more efficient since we don't print the output.

PR #658, opened 2019-03-09
Fix: disallow '-' with other inputs

hhugo:


src/Conf.ml
```diff
@@ -1390,16 +1390,20 @@ let (_profile : t option C.t) =
     (fun _ -> !selected_profile_ref)
 
 let validate () =
+  let inputs_len = List.length !inputs in
+  let has_stdin = List.exists ~f:(String.equal "-") !inputs in
   if !print_config then `Ok ()
-  else if List.is_empty !inputs then
+  else if inputs_len = 0 then
     `Error (false, "Must specify at least one input file, or `-` for stdin")
-  else if List.equal String.equal !inputs ["-"] && Option.is_none !name then
```
Julow: If `--name` is required for stdin, `--inplace` is not compatible with `--name` and `--inplace` is required for multiple inputs, I think the problem was just this line (`List.equal` instead of `List.exists`). But the new error message avoid confusion.

PR #659, opened 2019-03-09
Restore compat with base.v0.11

hhugo:
This would be useful before v0.9.
Quoting @avsm: 

> The issue is that we need to move all the platform tools at the same time, or else the opam universe splits (i.e. ocamlformat will not be coinstallable with anything else that requires Base v0.11).

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -207,7 +207,7 @@ end = struct
   let empty : t = (empty_start, empty_end)
 
   let invariant (smap, emap) =
-    assert (List.equal Poly.equal (Map.to_alist smap) (Map.to_alist emap))
+    assert (Poly.equal (Map.to_alist smap) (Map.to_alist emap))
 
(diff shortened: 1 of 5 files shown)


hhugo:
The first commit is part of #658

src/Cmts.ml
```diff
@@ -207,7 +207,7 @@ end = struct
   let empty : t = (empty_start, empty_end)
 
   let invariant (smap, emap) =
-    assert (List.equal Poly.equal (Map.to_alist smap) (Map.to_alist emap))
+    assert (Poly.equal (Map.to_alist smap) (Map.to_alist emap))
```
Julow: Why not `Map.equal Poly.equal smap emap` ?

hhugo: `smap` and `emap` don't have the same type.

Julow: I see.

avsm:
Thanks @hhugo, this is perfect

PR #661, opened 2019-03-10
Reduce the use of Poly comparisons

hhugo:


src/Fmt_odoc.ml
```diff
@@ -116,10 +116,10 @@ and fmt_list kind l =
...
     Octavius.parse (Lexing.from_string str)
   in
-  if
-    Poly.equal (print_and_parse light_syntax) (print_and_parse heavy_syntax)
-  then light_syntax
-  else heavy_syntax
+  match (print_and_parse light_syntax, print_and_parse heavy_syntax) with
+  | Ok x, Ok y when Poly.equal x y -> light_syntax
+  | Ok _, Ok _ -> heavy_syntax
+  | Error _, _ | _, Error _ -> heavy_syntax
```
hhugo: @gpetiot, is this correct ?

gpetiot: Yes it is.

PR #662, opened 2019-03-10
Fix: formatting of end line comments

### LINKED ISSUE #660: Bad formatting of end line comments
Hi !
Using the latest version of ocamlformat, and default settings,
```
include Bar (* this is commenting the include line *)
```
was formatted to 
```
include Bar

(* this is commenting the include line *)
```
Wouldn't it be better to move the comment above the code than two lines below ? This seems like a semantic change.

gpetiot:
Fix #660 
Overall the comment placement is closer to the original source.

The new breaks on the `[%%foo:   ...  ]` seem to be more consistent (discriminating on the context to only break for toplevel extensions does not seem to work).

### DIFF
--- a/test/passing/comments.ml.opts
+++ b/test/passing/comments.ml.opts
@@ -1,1 +1,1 @@
---max-iter=4
+--max-iter=3
--- a/test/passing/comments.ml.opts
+++ b/test/passing/comments.ml.opts
@@ -1,1 +1,1 @@
---max-iter=3
+--max-iter=4
--- a/test/passing/js_source.ml.ref
+++ b/test/passing/js_source.ml.ref
@@ -38,20 +38,26 @@ x]
 
 let ([%foo 2 + 1] : [%foo bar.baz]) = [%foo "foo"]
 
-[%%foo module M = [%bar]]
+[%%foo
+module M = [%bar]]
 
 let ([%foo let () = ()] : [%foo type t = t]) = [%foo class c = object end]
 
-[%%foo: 'a list]
+[%%foo:
+'a list]
 
 let ([%foo: [`Foo]] : [%foo: t -> t]) = [%foo: < foo : t > ]
 
-[%%foo? _]
-[%%foo? Some y when y > 0]
+[%%foo?
+_]
+
+[%%foo?
+Some y when y > 0]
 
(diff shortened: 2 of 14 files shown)


test/passing/object.ml
```diff
@@ -30,7 +30,8 @@ let _ =
 
     constraint t = 'a t
 
-    [%%ext salut, "hello"]
+    [%%ext
+    salut, "hello"]
```
Julow: Why this change ? It's not related to comments.

gpetiot: It was due to the added hvbox around the structure_item and the comments after, sometimes the break was expected for the extensions, even if it fitted on one line (but not always).
I finally put the hvbox only if there are comments after, to reduce the diff.

test/passing/comments.ml.opts
```diff
@@ -1 +1 @@
---max-iter=3
\ No newline at end of file
+--max-iter=4
```
Julow: Does this means that this PR make Ocamlformat a little slower ?

test/passing/js_source.ml.ref
```diff
@@ -8809,7 +8799,6 @@ external t : float -> float = "t" [@@unboxed]
 
 let _ = ignore ( + )
 let _ = raise Exit 3
-
 (* comment 9644 of PR#6000 *)
```
Julow: Why this one is not on the same line ?

test/passing/comments.ml.opts
gpetiot: That means we need an extra iteration yes (the reason is not obvious from my changes though)

test/passing/js_source.ml.ref
gpetiot: fixed

gpetiot:
The changes on the `[%%foo .... ]` extensions are due to the new hvbox, but the output is closer to the input code anyway.

test/passing/comments.ml.opts
gpetiot: fixed

gpetiot:
I fixed all the regressions, should be ready to go now.

@pascutto does it look better on the codebase you have been formatting?

PR #664, opened 2019-03-10
Improve: make gen_version an ocaml script

hhugo:
alternative to https://github.com/ocaml-ppx/ocamlformat/pull/471


gpetiot:
Looks good to me.

Julow:
I think we should use `dune subst` instead to avoid maintaining the tool.
It does the same thing, except for the `--tags` option to `git describe`. But I think it is not a problem.

tools/gen_version.ml
```diff
@@ -0,0 +1,42 @@
+(**********************************************************************
+ *                                                                    *
+ *                            OCamlFormat                             *
+ *                                                                    *
+ *  Copyright (c) 2017-present, Facebook, Inc.  All rights reserved.  *
```
jberdine: year

Julow:
In the first case, we just use `"%%VERSION%%"` instead of `Version.version` and for the release it will work fine.
In the case of pinned or dev, the version will be the commit hash instead of the previous tag.

jberdine:
A version of any binary that is just %%VERSION%% is not acceptable. Every build needs the correct version.

Julow:
That is not what I meant, `dune subst` will substitute the watermarks inside of the source files. The problem is that for non release builds, the version will be the commit hash (eg. `c781b4dd` or `c781b4dd-dirty`).

tools/gen_version.ml
hhugo: fixed

jberdine:
@Julow, I'm sorry but I don't understand your suggestion. Would you like to send something more concrete with what you mean? Or probably it is better to see if we can use https://github.com/ocaml/dune/pull/1930.

jberdine:
(But to be clear, the current plan is to include this PR in 0.9, but more improvement is welcome.)

Julow:
@jberdine Sorry for not being clear. I implemented my suggestion in #712.

PR #665, opened 2019-03-11
Improve: fmt_module

gpetiot:
@Julow I feel like it's better to rework `fmt_module` (with a minimal diff in formatting, mainly improving the code) before working on #619 and other module-related issues.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3467,12 +3467,10 @@ and fmt_module_expr c ({ast= m} as xmod) =
       in
       if Option.is_some pro_a then
         { blk_a with
-          opn= opn_a
-        ; pro=
+          pro=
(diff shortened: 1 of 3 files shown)


gpetiot:
I think the diff is already big enough and we can merge this one, I also think the indentation changes are an improvement, do you agree @jberdine @hhugo @Julow ?

jberdine:
This causes many regressions with test_branch due to the indentation of modules not taking leading parens into account. An option could be added to ignore leading parens when determining the indentation, but it should be applied to other syntactic forms too.

jberdine:
Is it expected that this will be quick to fix, or should we revert in order to make the 0.9 release?

gpetiot:
We only have to revert this commit: 4377c44

PR #663, opened 2019-03-10
Improve: new profiles conventional and ocamlformat

gpetiot:
See #656

- New `conventional` profile, with the following changes with respect to the default profile:
   * break-separators: from before to after-and-docked
   * break-cases: from fit to nested
   * break-infix-before-func: from true to false
   * break-sequences: from false to true
   * field-space: from tight to loose
   * sequence-style: from separator to terminator
- New `structure-left-margin` profile, equivalent to the default profile.

Didn't change the default values yet, so that people can try the `conventional` profile for the 0.9 release, and we can discuss about making the beginner values default for the next release.

PR #669, opened 2019-03-14
Fix indent of after comments in record types

gpetiot:
Fix issue reported in https://github.com/ocaml-ppx/ocamlformat/pull/663#issuecomment-472601110
It looks better to me, I don't know why we had a 4-spaces indentation before.


test/passing/comments_in_record.ml.ref
```diff
@@ -39,8 +39,8 @@ type t =
         (* aaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbb
```
Julow: I think the "weird formatting" comment is because the comment for the last field is not indented the same as for the other fields. You are making a regression here.
Also the other comments were not indented (https://github.com/mirage/irmin/pull/582/files#diff-4846077446d291e891a3940001669978R135)

gpetiot:
You are right, I explained in #663  why it won't be fixed.

PR #670, opened 2019-03-14
Fix space-around-collection-expressions for record/variant definitions

Julow:
The option did not work for record definitions when `break-separators=after-and-docked` and was never implemented for polymorphic variants.


src/Fmt_ast.ml
```diff
@@ -634,25 +634,28 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
...
-        $ fits_breaks
-            (if protect_token then " ]" else "]")
-            (if force_break then "@;<1000 0>]" else "@ ]") )
+        ( match (flag, lbls, rfs) with
+        | Closed, None, [Rinherit _] ->
+            fmt "[ | " $ row_fields rfs $ closing
+        | Closed, None, _ ->
+            let opening = if space_around then "[ " else "[" in
+            fits_breaks opening "[ " $ row_fields rfs $ closing
+        | Open, None, _ -> fmt "[> " $ row_fields rfs $ closing
```
Julow: Should we remove those spaces when `space-around-collection-expressions` is `false` ?
eg.
```ocaml
type t = [>`A]
type t = [<`A]
type t = [<`A>`B]
```

Julow: After writing this, I'd say no.

gpetiot:
We should add tests of multi-line variants and also add tests with comments inside, but otherwise looks good.

test/passing/js_source.ml.ref
```diff
@@ -6738,10 +6739,10 @@ end
 
 module C =
   Combine (struct
-      type 'a t = 'a constraint 'a = [> ]
+      type 'a t = 'a constraint 'a = [>  ]
```
hhugo: double space here

test/passing/js_source.ml.ref
```diff
@@ -6729,7 +6730,7 @@ let f x = (x : 'a vlist :> 'b vlist)
 let f (x : 'a vlist) : 'b vlist = x
 
 module type Poly = sig
-  type 'a t = 'a constraint 'a = [> ]
+  type 'a t = 'a constraint 'a = [>  ]
```
hhugo: double space here

test/passing/js_source.ml.ref
Julow: Fixed.

Julow:
I refactored the tests a little and used symlinks to test this options.

PR #679, opened 2019-03-18
Fix space around tuples

gpetiot:
From https://github.com/janestreet/ocamlformat/commit/4feebdd2e42acf7c64c3eb58eda606ca52548b6b

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -353,12 +353,7 @@ let wrap_record c =
   else wrap_fits_breaks c.conf "{" "}"
 
 let wrap_tuple ~parens ~no_parens_if_break c =
-  let wrap_fits_breaks_if _c cnd pre suf k fs =
-    fits_breaks_if cnd pre (pre ^ " ") fs ;
-    k fs ;
-    fits_breaks_if cnd suf ("@ " ^ suf) fs
-  in
-  if parens then wrap_fits_breaks_if c.conf true "(" ")"
+  if parens then wrap_fits_breaks c.conf "(" ")"
   else if no_parens_if_break then Fn.id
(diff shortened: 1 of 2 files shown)


src/Fmt_ast.ml
```diff
@@ -352,6 +352,16 @@ let wrap_record c =
   if c.conf.space_around_collection_expressions then wrap "{ " "@ }"
   else wrap_fits_breaks c.conf "{" "}"
 
+let wrap_tuple ~parens ~no_parens_if_break c =
+  let wrap_fits_breaks_if _c cnd pre suf k fs =
```
hhugo: This should probably move to [Fmt]

hhugo: Also, I wonder why code for tuple is so different from list, array, record  ..

src/Fmt_ast.ml
```diff
@@ -352,6 +352,16 @@ let wrap_record c =
...
+    fits_breaks_if cnd pre (pre ^ " ") fs ;
+    k fs ;
+    fits_breaks_if cnd suf ("@ " ^ suf) fs
+  in
+  if parens then wrap_fits_breaks_if c.conf true "(" ")"
```
Julow: Why did you rewrite `wrap_fits_breaks_if` ? The one from `Fmt` should work fine here.

src/Fmt_ast.ml
gpetiot: I'm not sure this function would fit in Fmt, for now Fmt only contains generic functions that are not designed for any particular AST node.
If we want to extract wrap_tuple, wrap_record and so on, maybe we should create a file Fmt_ast_helper to gather the AST specific functions.

PR #680, opened 2019-03-18
Fix weird space before type with ocp-indent-compat

gpetiot:
From https://github.com/janestreet/ocamlformat/commit/d6241cfd3fa32c486689e1d1666e277ef674c03d

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ install:
   - opam install --with-test ocamlformat
 script:
   - make test
-  - opam install ocamlformat
+  - opam install --with-test ocamlformat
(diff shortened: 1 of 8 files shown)


hhugo:
This only make sense if ocp-indent is applied after ocamlformat. This might be a good time to add a test for that include ocp-indent.

hhugo:
@Julow, I don't understand why you want a new option here ?

Julow:
Using `ocp-indent-compat` seems arbitrary and some people may want to have the `->` before, without the initial padding (I do).

gpetiot:
One of those it-works-locally-but-not-on-the-CI bugs.

Julow:
I think it's because you must pass `--with-test` to `opam install` in `.travis.yaml` in the `install` rule.

gpetiot:
Not necessary anymore since #694 has been merged

PR #681, opened 2019-03-18
Improve split attribute in let binding for expect test with uncaught exn

### LINKED ISSUE #649: Unfortunate interaction with `dune promote` and `ppx_inline_test`
With ppx_inline_test v0.12.0, if an expect test raises an exception, the result after `dune promote` looks like:
```ocaml
    let%expect_test _ =
      assert false ;
      [%expect.unreachable]
    [@@expect.uncaught_exn {|
      (* CR expect_test_collector: This test expectation appears to contain a backtrace.
         This is strongly discouraged as backtraces are fragile.
         Please change this test to not include a backtrace. *)

      "Assert_failure test.ml:5:6"
      Raised at file "test.ml", line 4, characters 6-18
      Called from file "collector/expect_test_collector.ml", line 225, characters 12-19 |}]
```
This formats to:
```ocaml
    let%expect_test[@expect.uncaught_exn
                     {|
      (* CR expect_test_collector: This test expectation appears to contain a backtrace.
         This is strongly discouraged as backtraces are fragile.
         Please change this test to not include a backtrace. *)

      "Assert_failure test.ml:5:6"
      Raised at file "test.ml", line 4, characters 6-18
      Called from file "collector/expect_test_collector.ml", line 225, characters 12-19 |}] _
        =
      assert false ;
      [%expect.unreachable]
```
(note the attribute moved) and then fixing the exception (change `false` to `true` in the example) and re-running the test fails:
```
Uncaught exception:
  
  ("Expect test evaluator bug"
   (exn (Invalid_argument "Negative length: -506"))
   (backtrace
     "Raised at file \"stdlib.ml\", line 34, characters 20-45\
    \nCalled from file \"src/ordered_collection_common.ml\", line 9, characters 7-48\
    \nCalled from file \"src/string.ml\", line 128, characters 4-84\
    \nCalled from file \"matcher/matcher.ml\" (inlined), line 254, characters 32-66\
    \nCalled from file \"matcher/matcher.ml\", line 392, characters 10-49\
    \nCalled from file \"list.ml\", line 117, characters 24-34\
    \nCalled from file \"src/list0.ml\" (inlined), line 22, characters 40-81\
    \nCalled from file \"matcher/matcher.ml\", line 290, characters 4-1023\
    \nCalled from file \"src/exn.ml\", line 66, characters 8-11\
    \nRe-raised at file \"src/exn.ml\", line 69, characters 6-15\
    \nCalled from file \"evaluator/ppx_expect_evaluator.ml\" (inlined), line 174, characters 6-112\
    \nCalled from file \"evaluator/ppx_expect_evaluator.ml\", line 187, characters 8-43\
    \nCalled from file \"evaluator/ppx_expect_evaluator.ml\", line 208, characters 6-83\
    \n")
   (filename src/symbheap/equality_test.ml))
```
It seems that `ppx_inline_test` is relying on an aspect of the concrete syntax it generates that `ocamlformat` does not preserve.

I don't know what the right fix is.

gpetiot:
From https://github.com/janestreet/ocamlformat/commit/c60d7378061fda35c158f399451735465ccef87f

fix #649

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3964,7 +3964,11 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
     match xbody.ast with {pexp_desc= Pexp_fun _} -> 1 | _ -> 2
   in
   let at_attrs, at_at_attrs =
-    match ext with None -> (atrs, []) | Some _ -> ([], atrs)
+    match ext with
+    | None -> (atrs, [])
+    | Some _ ->
+        List.partition_tf atrs ~f:(fun ({txt}, _) ->
+            not (String.equal txt "expect.uncaught_exn") )
   in
   let stmt_loc = Sugar.args_location xargs in
   fmt_docstring c ~epi:(fmt "@\n") doc1
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3967,8 +3967,8 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
     match ext with
     | None -> (atrs, [])
     | Some _ ->
-        List.partition_tf atrs ~f:(fun ({txt}, _) ->
-            not (String.equal txt "expect.uncaught_exn") )
+        List.partition_tf atrs ~f:(fun ({loc}, _) ->
+            Location.compare_start loc pvb_expr.pexp_loc < 1 )
   in
   let stmt_loc = Sugar.args_location xargs in
   fmt_docstring c ~epi:(fmt "@\n") doc1
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3963,6 +3963,9 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
   let indent =
     match xbody.ast with {pexp_desc= Pexp_fun _} -> 1 | _ -> 2
   in
+  let at_attrs, at_at_attrs =
+    match ext with None -> (atrs, []) | Some _ -> ([], atrs)
+  in
   let stmt_loc = Sugar.args_location xargs in
   fmt_docstring c ~epi:(fmt "@\n") doc1
   $ Cmts.fmt_before c pvb_loc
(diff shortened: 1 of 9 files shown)


hhugo:
The reason for this logic is https://github.com/ocaml-ppx/ocamlformat/issues/649
```
    let%expect_test _ =
      assert false ;
      [%expect.unreachable]
    [@@expect.uncaught_exn {|
      (* CR expect_test_collector: This test expectation appears to contain a backtrace.
         This is strongly discouraged as backtraces are fragile.
         Please change this test to not include a backtrace. *)

      "Assert_failure test.ml:5:6"
      Raised at file "test.ml", line 4, characters 6-18
      Called from file "collector/expect_test_collector.ml", line 225, characters 12-19 |}]
```
I'm not particularly happy with the side effects.

hhugo:
Can you add a tests for the thing reported in #649 ?

gpetiot:
Is this ready to go or is it too specific?

hhugo:
It seems to be very specific indeed. Could you maybe check ppx_expect to see if there are any other cases to cover.

hhugo:
Or preserve the position from the source ..

Julow:
That would be better, indeed.

src/Fmt_ast.ml
```diff
@@ -3943,6 +3943,13 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
   let indent =
     match xbody.ast with {pexp_desc= Pexp_fun _} -> 1 | _ -> 2
   in
+  let at_attrs, at_at_attrs =
+    match ext with
+    | None -> (atrs, [])
```
hhugo: Should we preserve the location even when there are not extension ?

hhugo:
I'm curious how this behaves with ocamlformat_reason. Could you add an example for reason ?

hhugo:
The same logic could be applied to all other items if needs be. But let's keep that for later.

gpetiot:
> I'm curious how this behaves with ocamlformat_reason. Could you add an example for reason ?

I don't know the state of ocamlformat_reason but I can't get the right syntax for attributes (it shouldn't delay this PR anyway).

hhugo:
Can you add test with attribute before and after for both [let] and [let in]

test/passing/let_binding.ml
```diff
@@ -121,3 +121,19 @@ let _ =
   let%ext (_ : int) = x in
   let%ext _ : int = x in
   ()
+
+let fooo = fooooooooooo [@@foo];;
+
+let fooo = fooooooooooo [@@foo]
+and fooo = fooooooooooo [@@foo];;
+
+let foooo = fooooooooo [@foo] in
```
hhugo: Should be @@

PR #684, opened 2019-03-19
Improve: disable conf in files and attributes

gpetiot:
From https://github.com/janestreet/ocamlformat/commit/1af58007af7f8627133b14fa47414b9180a2dcc9

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1523,13 +1523,10 @@ let parse_line config ~from s =
           ~from:(`Parsed (`File x))
           ~name ~value ~inline:false
     | name, `Attribute ->
-        if
-          !disable_conf_attrs
-          && (not (String.equal name "disable"))
-          && not config.quiet
-        then (
-          Format.eprintf "Warning: Configuration in attribute %S ignored.\n"
-            s ;
+        if !disable_conf_attrs then (
+          if not config.quiet then
+            Format.eprintf
+              "Warning: Configuration in attribute %S ignored.\n" s ;
           Ok config )
         else
           C.update ~config
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1065,6 +1065,16 @@ let comment_check =
     (fun conf x -> {conf with comment_check= x})
     (fun conf -> conf.comment_check)
 
+let disable_conf_attrs =
+  let doc = "Disable configuration in attributes." in
+  mk ~default:false
+    Arg.(value & flag & info ["disable-conf-attrs"] ~doc ~docs)
+
+let disable_conf_files =
+  let doc = "Disable .ocamlformat configuration files." in
+  mk ~default:false
+    Arg.(value & flag & info ["disable-conf-files"] ~doc ~docs)
+
 let disable_outside_detected_project =
   let witness =
     String.concat ~sep:" or "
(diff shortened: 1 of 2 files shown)


src/Conf.ml
```diff
@@ -1509,13 +1519,21 @@ let parse_line config ~from s =
               , Format.sprintf "expecting %s but got %s" Version.version
                   value ))
     | name, `File x ->
-        C.update ~config
-          ~from:(`Parsed (`File x))
-          ~name ~value ~inline:false
+        if !disable_conf_files then
+          Error (`Malformed "Disabled file configuration of ocamlformat")
```
Julow: This will trigger an error when there is a `.ocamlformat` file in the current directory.

Julow: Also the error message is pretty ugly:

```
  ("invalid format" "Disabled file configuration of ocamlformat")
  ("invalid format" "Disabled file configuration of ocamlformat")
  ("invalid format" "Disabled file configuration of ocamlformat"))
```
(followed by stack trace)

src/Conf.ml
```diff
@@ -1509,13 +1519,21 @@ let parse_line config ~from s =
...
+        else
+          C.update ~config
+            ~from:(`Parsed (`File x))
+            ~name ~value ~inline:false
     | name, `Attribute ->
-        C.update ~config
-          ~from:(`Parsed `Attribute)
-          ~name ~value ~inline:true
+        if !disable_conf_attrs && not (String.equal name "disable") then
+          Error
```
Julow: This error is not shown to the user and attributes are ignored.

gpetiot:
> What is the motivation for these options ?

The motivation is to not have configuration files and attributes silently overriding the preset configuration (here by Janestreet but it could be useful to others)

src/Conf.ml
gpetiot: fixed

src/Conf.ml
gpetiot: fixed

src/Conf.ml
```diff
@@ -1513,9 +1523,18 @@ let parse_line config ~from s =
           ~from:(`Parsed (`File x))
           ~name ~value ~inline:false
     | name, `Attribute ->
-        C.update ~config
-          ~from:(`Parsed `Attribute)
-          ~name ~value ~inline:true
+        if
+          !disable_conf_attrs
+          && (not (String.equal name "disable"))
+          && not config.quiet
```
Julow: If `--quiet` is passed, attributes are not ignored. You meant a nested `if` ?

src/Conf.ml
```diff
@@ -1519,17 +1519,18 @@ let parse_line config ~from s =
...
+          ~from:(`Parsed (`File x))
+          ~name ~value ~inline:false
     | name, `Attribute ->
-        if !disable_conf_attrs && not (String.equal name "disable") then
-          Error
-            (`Malformed
-              "Disabled configuration of ocamlformat in attributes")
+        if
+          !disable_conf_attrs
+          && (not (String.equal name "disable"))
```
hhugo: This is very janestreet related, I don't think we should specialize "disable" just yet.

src/Conf.ml
gpetiot: good catch

Issue #266, opened 2018-08-02
Improve parens around infix operator.


hhugo:
There have been a bit a back and forth on that topic: https://github.com/ocaml-ppx/ocamlformat/commit/ea5ed9ffeae42fe9caaa1005786f3d83da802157, https://github.com/ocaml-ppx/ocamlformat/issues/115, https://github.com/ocaml-ppx/ocamlformat/pull/141

I'll copy paste comments from ea5ed9ffeae42fe9caaa1005786f3d83da802157:
 
> Add discretionary parens on nested binops with different precedence
> 
> Add parens around an arg of a binop that is also a binop of a
> different precedence, unless the line breaks, in which case the
> nesting structure is indicated by indentation.
> 
> Examples:
```
;; f (a || (b && c))

;; f
     ( aaaaaaaaaaaaaaaaaaaaa
     || (bbbbbbbbbbbbbbbbbbbbb && ccccccccccccccccccccc) )

;; f
     ( aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
     || bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
        && ccccccccccccccccccccccccccccccccc )

;; f ((a && b) || c)

;; f
     ( (aaaaaaaaaaaaaaaaaaaaa && bbbbbbbbbbbbbbbbbbbbb)
     || ccccccccccccccccccccc )

;; f
     ( aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
       && bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
     || ccccccccccccccccccccccccccccccccc )
```

hhugo:
cc @mshinwell

mshinwell:
I don't think the rule about using indentation instead of parentheses should ever apply.  It's just too easy to make a mistake here.  Bear in mind this is a long-standing problem, hence the provision of warnings in compilers like GCC, which don't take account of indentation.

(I'm also unclear why there are spaces just inside parentheses in the examples above; that seems highly non-standard.)

jberdine:
See #275 for an option for using parens, and added parens for `not`.

Re `not`, do you think it is special compared to any other application?

This is probably a case where using type information would help, so that only cases which had multiple well-typed interpretations, up to the grammar's ambiguity, would need to be explicitly disambiguated.

jberdine:
[...]

PR #686, opened 2019-03-19
Improve: parens around not expressions

gpetiot:
From https://github.com/janestreet/ocamlformat/commit/7fcd0e265d584ce43abbfc0c2ab3ebb4240afc6f


gpetiot:
Is this one ready to go?

hhugo:
I'm not sure about this one. This was previously discussed in https://github.com/ocaml-ppx/ocamlformat/issues/266. The special case for "not" is kind of weird.
I would wait for feedback from @jberdine

Julow:
I'm not fan of the new option `parens-not`. What do you think of adding a value to `infix-precedence`, eg. `parens-disambiguate`, which could be used to add other special cases ? Or to change the option name to be more general and extensible in the future ?

gpetiot:
From #266 it looks like the current implementation does a good job, @hhugo do you have concrete examples that motivated you to write the original patch in the first place?

hhugo:
Some people were unhappy about it. @mbarbin, any opinion on this ?

mbarbin:
I am looking at jane with master without this PR, and I can't find bad examples at the moment.  I think this is OK to close this PR and we can re-open if we find a compelling example to add to the comment.  I think the rule to parenthesize not expressions seems good on master, we can plan on using it for 0.10 and see whether we get some complementary feedback on it for a future version of ocamlformat.

PR #694, opened 2019-03-20
Fix value binding ocp indent compat

gpetiot:
From https://github.com/janestreet/ocamlformat/commit/b461c6001b0787a5c9aaa4bc712708092d682e81
I had to add some conditions to limit the changes to ocp-indent-compat mode.
But I'm having trouble to see what the original commit is trying to achieve, as the original commit doesn't generate valid OCaml code on js_source.ml
I also didn't report the changes on the `:` position, otherwise we get a regression with test/passing/ocp_indent_compat.ml:
```ocaml
external i : (int -> float[@unboxed]) = "i" "i_nat"
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -551,7 +551,7 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
   @@ fun c ->
   ( match (ptyp_desc, pro) with
   | Ptyp_arrow _, Some pro when c.conf.ocp_indent_compat ->
-      fmt_or pro_space "@;" "@," $ str pro $ fmt " "
+      fmt_if pro_space "@;" $ str pro $ fmt " "
   | _, Some pro -> fmt_if pro_space " " $ str pro $ fmt "@ "
   | _ -> fmt "" )
   $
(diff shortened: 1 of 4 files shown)


src/Fmt_ast.ml
```diff
@@ -2685,6 +2687,11 @@ and fmt_cases c ctx cs =
                   (fmt_or c.conf.indicate_multiline_delimiters "@ )" "@,)")
               ) ) )
 
+and is_arrow_or_poly = function
```
Julow: This does not need to be mutually recursive. Could you move it to the top of the file ?

test/passing/js_source.ml.ref
```diff
@@ -3629,7 +3630,8 @@ let ssmap =
      and type map = SSMap.map)
 ;;
 
-let ssmap :
+let ssmap
+    :
     (module MapT with type key = string
```
Julow: The `:` should be on the same line as the type.

gpetiot:
I think I fixed what was wrong in the original commit, @hhugo does the ocp-indent-compatible output look consistent with your original intent?

hhugo:
This looks good.

PR #697, opened 2019-03-21
Fix ast changed when record ident constrained

### LINKED ISSUE #696: BUG: ast changed.
```
echo "let a () = A { A.a : t; }" | _build/dev/src/ocamlformat.exe - --name a.ml
```
cc @greg

gpetiot:
Fix #696


hhugo:
We might have the same issue with [Pexp_override]

gpetiot:
I can't reproduce the same bug on `let a () = A {<x = (A.a : t)>}`

PR #690, opened 2019-03-20
Fix break before parens no wrap fun args

gpetiot:
Fix #682 
To me a lot of the changes are improvements, very few regressions but I didn't manage to get rid of them.

Issue #682, opened 2019-03-19
weird break before parens with --no-wrap-fun-args 


hhugo:
```
$ echo " let _ = let _ = fffffffffff aaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbbbbbbb ~f:(fun x -> return xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx) in 2 " \
 | _build/dev/src/ocamlformat.exe --name a.ml --no-wrap-fun-args -m 67 -


let _ =
  let _ =
    fffffffffff
      aaaaaaaaaaaaaaaaaaaaaaaaaaaa
      bbbbbbbbbbbbbbbbbbbbbbbbbb
      ~f:(fun x -> return xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    )
  in
  2
```

gpetiot:
The fact that we try to put `~f:(fun x ->` is put on the same line as the function name `fffffffffff` when it's possible messes up the box structure and makes this difficult to fix.

PR #698, opened 2019-03-21
Fix incorrect ocaml code

gpetiot:
#690 broke test_branch by generating invalid OCaml code
Better fix for #682 
If you run test_branch, you have to run it against the commit just before #690 was merged.


hhugo:
What broke ? Should we add regression tests ?

gpetiot:
It was due to `fits_breaks "@ )" "@ )"` the first argument should not contain `@` so we incorrectly printed `@` before a closing parenthesis in the generated OCaml.

PR #700, opened 2019-03-22
Fix extra space ifthenelse

### LINKED ISSUE #695: Extra space in if-then-else
There should be no space after `(` in `then ( if b`

```
$ echo "if fffffffffffffff aaaaa bb then (if b then aaaaaaaaaaaaaaaa ffff) else aaaaaaaaaaaa qqqqqqqqqqq" | /j/office/app/dev-tools/prod/apply-style/1.4-5/ocamlformat --profile janestreet --name a.ml -
;;
if fffffffffffffff aaaaa bb
then ( if b then aaaaaaaaaaaaaaaa ffff)
else aaaaaaaaaaaa qqqqqqqqqqq
```

gpetiot:
Fix #695, no regression with test_branch

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1754,7 +1754,11 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                                       $ fmt "@ " $ fmt_expression c xcnd )
                                   $ fmt "@ then" )
                             | None -> fmt "else" )
-                          $ fmt_if parens_bch " (" $ fmt "@ "
+                          $ fmt_or parens_bch
+                              ( if c.conf.indicate_multiline_delimiters then
+                                " (@ "
+                              else " (@," )
+                              "@ "
                           $ fmt_expression c ~box:false ~parens:false xbch
                           )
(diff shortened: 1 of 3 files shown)


src/Fmt_ast.ml
```diff
@@ -1754,7 +1754,11 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                                       $ fmt "@ " $ fmt_expression c xcnd )
                                   $ fmt "@ then" )
                             | None -> fmt "else" )
-                          $ fmt_if parens_bch " (" $ fmt "@ "
+                          $ fmt_or parens_bch
+                              ( if c.conf.indicate_multiline_delimiters then
+                                " (@ "
+                              else " (@," )
```
hhugo: The indentation of `else` looks weird

gpetiot: Agreed but according to Josh https://github.com/ocaml-ppx/ocamlformat/pull/603#discussion_r262851548 it's the expected behavior until we add a new option for that.

jberdine: Yes, this is the expected form for `if-then-else=compact`

PR #712, opened 2019-03-25
Use dune subst instead of gen_version.sh

### LINKED ISSUE #470: gen_version.sh is not portable

Julow:
Alternative to #664 (alternative to #471 (fix #470)).

`tools/gen_version.sh` is not needed and can be replaced by `dune subst`.

The problem with this approach is that for pinned or dev builds, the `"%%VERSION%%"` will be the commit hash.


jberdine:
Thanks for writing this up. What I don't understand with this approach is how each build is tagged with the correct version. Have I misunderstood?

Julow:
They are not, unless you run `subst`.

jberdine:
I see, thanks. Ok, so we're still with #664.

Julow:
Why is it important to tag local dev builds ?

jberdine:
Depending on the work flow, exes built in this way end up being deployed elsewhere, and used to run tests. It is very easy to get confused about what test results correspond to what version if the exes have garbage versions internally. It would probably be ok if the correct version didn't get set until `dune install`, but until the next version of dune, I don't think that is possible.

Julow:
I understand. Note that `opam install ./ocamlformat.opam` triggers the subst:
```
 > ocamlformat --version
6da8dfd
```

I close in favor of #664.

jberdine:
Yes, but the overhead of `opam install` is much higher than that of `dune install`. Perhaps it is my failure to read the docs, but since opam2 it seems impossible to pin local packages without opam making a copy of the source dir. And even worse, if the project root isn't the git root, then it has to use rsync, which copies all the build artifacts. So e.g. in my normal setup a few dirs down from ocamlformat/test there are a couple 10s of GB of llvm binaries.

PR #711, opened 2019-03-25
Fix doc comments tag only when docstring parsing disabled

gpetiot:
Fix #707, no regression with test_branch

PR #713, opened 2019-03-25
Fix: parse docstrings once

gpetiot:
Follow up on #711

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -360,7 +360,7 @@ let wrap_tuple ~parens ~no_parens_if_break c =
 
 let doc_atrs = Ast.doc_atrs
 
-let _parse_docstring str_cmt =
+let parse_docstring str_cmt =
   match Octavius.parse (Lexing.from_string str_cmt) with
   | Error _ -> Error ()
   | Ok parsed as ok ->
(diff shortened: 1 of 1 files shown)


src/Fmt_ast.ml
```diff
@@ -360,16 +360,14 @@ let wrap_tuple ~parens ~no_parens_if_break c =
...
-  if not c.conf.parse_docstrings then Error ()
-  else
-    match Octavius.parse (Lexing.from_string str_cmt) with
-    | Error _ -> Error ()
-    | Ok parsed as ok ->
-        if Conf.debug then (
-          Octavius.print Caml.Format.str_formatter parsed ;
-          Caml.Format.eprintf "%s%!" (Caml.Format.flush_str_formatter ()) ) ;
-        ok
+let _parse_docstring str_cmt =
```
jberdine: Tangential question: why does this have a leading `_` in it's name, thereby defeating the dead code checks?

Julow: I did that to split `fmt_docstring` into `_parse_` and `_fmt_docstring` to implement `fmt_docstring_around`. I agree the names are not very good. I was hoping to discuss that in the review.

jberdine: Ok, thanks. Maybe `fmt_parsed_docstring` instead of `_fmt_docstring` and just remove the leading `_` from `_parse_docstring`?

PR #715, opened 2019-03-26
Remove: deprecated config file syntax parsing

jberdine:


src/Conf.ml
```diff
@@ -1574,16 +1574,6 @@ let parse_line config ~from s =
       , from )
```
Julow: You could remove the `from` parameter. Also, splitting the string is not usefull because the pattern only match lists of one element, you could use strip instead.

PR #717, opened 2019-03-26
Improve: put the equal first for ocp-indent-compat

### LINKED ISSUE #624: Inconsistent break in type annotations
Using the Janestreet profile,

```
let ssmap :
    (module MapT with type key = string and type data = string and type map = SSMap.map)
    =
  ()
```

Compared to an arrow type:

```
let ssmap
    :  (module MapT with type key = string and type data = string and type map = SSMap.map)
    -> unit =
  ()
```

gpetiot:
Fix #624 
Before the fix:
```ocaml
let ssmap              
    : (module MapT with type key = string and type data = string and type map = SSMap.map)
  =
  ()
;;

let ssmap
    :  (module MapT with type key = string and type data = string and type map = SSMap.map)
    -> unit =
  ()
;;
```

After the fix:
```ocaml
let ssmap              
    : (module MapT with type key = string and type data = string and type map = SSMap.map)
  =
  ()
;;

let ssmap
    :  (module MapT with type key = string and type data = string and type map = SSMap.map)
    -> unit
  =
  ()
;;
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -4023,8 +4023,7 @@ and fmt_value_binding c ~rec_flag ~first ?ext ?in_ ?epi ctx binding =
                 $ wrap_fun_decl_args ~stmt_loc c (fmt_fun_args c xargs) )
             $ Option.call ~f:fmt_cstr )
         $ fmt_or_k c.conf.ocp_indent_compat
-            ( if Option.is_some fmt_cstr then fmt "@ ="
-            else fits_breaks " =" "@;<1000 0>=" )
+            (fits_breaks " =" "@;<1000 0>=")
             (fmt "@;<1 2>=") )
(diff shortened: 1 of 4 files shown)


gpetiot:
> The change looks good.
> 
> Unrelated, but I don't see the connection between `ocp-indent` and this feature. I feel like this and other `ocp-indent-compat` features could have their own option.

I don't really have an opinion on this, maybe @hhugo or @jberdine ?

jberdine:
I don't have a strong opinion, but if all else is equal, fewer options is better. The `ocp-indent-compat` option was originally intended for changes to output which was on its own fine, but caused ocp-indent to indent wildly when post-processing the output of ocamlformat. It is nice to have one option to enable to catch all the known cases of this. If we receive requests to enable only a subset of them, then it would make sense to split some off into separate options that ocp-indent-compat would also set. But I don't know that it makes sense to eagerly do this before there are requests.

Issue #571, opened 2019-01-28
Use the Odoc parser instead of Octavius


Julow:
Octavius is the documentation comments parser. It will be deprecated soon, in favor of [Odoc](https://github.com/ocaml/odoc).

## TODO

- Rewrite the `Fmt_odoc` module
    The types are different, obviously, but very similar

- Vendor odoc sources
    Odoc's `parser` and `model` internal libraries are not released yet on opam

- Whitespaces informations are lost
    In the AST, sentenses words are splitted and consecutive whitespaces, including newlines are represented as `\`Space`. Newlines can be retrieved by comparing start/end locations.
    A bit of work is required to preserve ASCII-art and fancy headers

- Small difficulties
    - Heading labels (`{0:label Heading}`) are not optional in the AST and so always printed
    - Deciding between heavy and light syntax for lists should be done with a better heuristic than print-parse-compare
    - Paths_types (identifiers, references) is very complex
        We only need to convert them to strings, there is (not exported) functions to do that in Odoc's `html/comment.ml`

Julow:
I started working on this a little, [on this branch](https://github.com/Julow/ocamlformat/tree/odoc_instead_of_octavius)

avsm:
cc @jonludam @aantron so they see this integration work ongoing.  In particular, try to structure the odoc vendoring so that it's standalone and in a single directory so that it can be updated easily.  I have a prototype called [duniverse](https://github.com/avsm/duniverse) which could be cleaned up to automate the self-contained vendoring extraction into ocamlformat for this purpose.

We may also need a little bit of Dune help from @rgrinberg in order to parameterise ignoring the vendoring in some cases, but I'm not sure that'll be needed if odoc is purely a binary and not an OCaml library as well.

jberdine:
Re whitespace, ascii-art, etc., I think that we need to be careful and scope down the problem. In general trying to preserve information that is not in the AST is complicated to implement and hard to make robust. I don't think that ocamlformat should attempt to format docstrings more closely to the source or make them more readable, etc. than odoc does. That is, if any particular docstring is not displayed as desired by odoc, then ocamlformat should not attempt to do differently. If such formatting is needed, then handling it will be needed in odoc as well, and special casing it by referring back to the concrete source in ocamlformat will just be duplicate work done in a not very reusable or principled way.

jberdine:
[...]

PR #721, opened 2019-03-27
Improve: parse doc comments with Odoc

Julow:
(issue #571)

Work in progress, but the current state is relatively clean and can be reviewed.

Rewrite `Fmt_odoc` and `Normalize` to use Odoc.
It adds a lot of regressions and fixes a few bugs.

Odoc `1.4.0` exports its internal libraries, it's no longer needed to vendor it. Other small problems remain.


src/Sugar.mli
```diff
@@ -97,9 +97,9 @@ val ite :
     if-then-else expressions, e.g.:
 
     {[
-        if c1 then e1
-        else if c2 then e2
-        else e3
+if c1 then e1
+else if c2 then e2
+else e3
```
jberdine: Is this indentation intended?

src/dune
```diff
@@ -21,7 +21,7 @@ Jbuild_plugin.V1.send
  (names ocamlformat ocamlformat_reason)
  (flags (:standard -open Import))
  %s
- (libraries cmdliner format_ fpath import ocaml-migrate-parsetree octavius unix uutf))
+ (libraries cmdliner format_ fpath import ocaml-migrate-parsetree unix uutf odoc.model odoc.parser))
```
jberdine: alphabetize

test/passing/doc_comments.mli
```diff
@@ -13,23 +13,25 @@ val escape : string -> string
 
 type title = string * int * string * string * string
 
-(** {1 Standard section names} *)
+(** {1:standard-section-names Standard section names} *)
```
jberdine: Is it necessary to add `:standard-section-names`? Seems very redundant.

test/passing/doc_comments.mli
```diff
@@ -62,24 +66,37 @@ val doc_to_plain :
   -> string
 (** [doc_to_plain b ~subst s] using [b], subsitutes in [s] variables by
     their [subst] definition and renders cmdliner directives to plain text.
+
     @raise Invalid_argument in case of illegal syntax. *)
 
 val k : k
 (** this is a comment
+
```
jberdine: Are these extra open lines intended / necessary?

test/passing/doc_comments.mli
```diff
@@ -180,7 +201,7 @@ val x : x
 
 (** [a]'c [b]'s [c]'c *)
 
-(** return true if {%\gamma(lhs) \subseteq \gamma(rhs)%} *)
+(** return true if [\gamma(lhs) \subseteq \gamma(rhs)] *)
```
jberdine: The `{%` `%}` markup ought to be preserved, right?

test/passing/doc_comments.mli
```diff
@@ -189,7 +210,7 @@ val x : x
 
 (** for [Global]s *)
 
-(** generic command: ∀xs.{foot}-{post} *)
+(** generic command: ∀xs.[foot]-[post] *)
```
jberdine: We need to avoid rewrites like this, they are semantically meaningful. If the original is ill-formed somehow, better to fail.

test/passing/doc_comments.mli
Julow: This is not optional in Odoc's ast. Work is needed to remove them.

src/Sugar.mli
Julow: A bit of work is needed to fix this. It is intended for verbatim blocks to avoid pushing ascii tables to the margin.

test/passing/doc_comments.mli
Julow: Not intended. I added the blank line to separate lists and paragraphs. I'll add a special case for tags (and module references, verbatim and code blocks).

test/passing/doc_comments.mli
Julow: In this case (and the next one) Odoc gives a warning and we could refuse to format with an error.

test/passing/doc_comments.mli
jberdine: Ok I see. Are they generated if omitted, so that we could detect if they have the generated name and then not print them?

test/passing/doc_comments.mli
jberdine: Oh. So odoc doesn't support the whole ocamldoc language? `{%` and `%}` are documented markup.

src/Sugar.mli
jberdine: Ok. I don't really know what the best choice ought to be, but expect that people will want to be able to keep things out of the left margin. But maybe not  if that means having inexplicable extra space in the generated docs.

test/passing/doc_comments.mli
Julow: Users may want to explicitly add the label. Reading the original string is possible but tricky. A good solution would be to modify Odoc.

test/passing/doc_comments.mli
Julow: Odoc gives a warning and turn it into a code span:
```
'{%...%}' (raw markup) needs a target language.
Suggestion: try '{%html:...%}'.
```
Warnings are easily handled, I did not code this part yet.
In case of warning, do you think we should print the warning and let the doc comment as is or refuse to format the whole file ?
The first seems more reasonable but without a `-Werror` flag (or maybe when `--check` is set), it could stay unnoticed.

test/passing/doc_comments.mli
Julow: It is:
```
'{abc}': bad markup.
Suggestion: did you mean '{!abc}' or '[abc]'?
```

test/passing/doc_comments.mli
jberdine: This seems like an odoc bug / divergence from spec to me. The `latex` language is the default: http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sss%3Atarget-specific-syntax

test/passing/doc_comments.mli
jberdine: Ok, thanks. I think it would be better to fail in such cases. Or at least warn, or have an option to choose. There are, I expect, lots of docstrings in the wild that are ill-formed, and it isn't only ocamlformat that will be doing the wrong thing with them. So probably better to give users feedback as early as possible. Hopefully we can get line numbers, etc. for an error message.

hhugo:
How far are we to have this PR ready ? I'm wondering if we should try to fix Octavious for 4.08 or if this PR will land in reasonable time ?

Julow:
Sorry for the delay.

I think this PR can land in reasonable time from now. I'm not 100% confident it can before the OCaml release.

It is not ready for merging but should be ready for a new review.

I rebased on master.

Since last reviews:
- Removed unecessary blank lines between some elements
- Don't format comments when Odoc generated warnings
	Because Odoc may have lost informations. In practice Odoc try its best to not to.
- Use Odoc parser's `Ast` instead of `Model.Comment`
	It is a bit simpler and the parser is a lot simpler to call.
	I made a PR to Odoc: https://github.com/ocaml/odoc/pull/351
	This fixes label on headings
- Removed terminating break in list elements with heavy syntax
- Fixed the formatting of references
	This part is ridiculously hard. Odoc could be modified to improve this but it would be a big change.

src/Fmt_odoc.ml
```diff
@@ -9,120 +9,218 @@
...
+open Odoc__parser.Ast
+module Names = Odoc__model.Names
+module Location_ = Odoc__model.Location_
 
-let escape_brackets =
-  String.Escaping.escape ~escapeworthy:['['; ']'] ~escape_char:'\\'
+(** Escape characters if they are not already escaped. [escapeworthy] should
+    return [true] if the character should be escaped, [false] otherwise. *)
+let ensure_escape ?(escape_char = '\\') ~escapeworthy s =
+  let module E = String.Escaping in
```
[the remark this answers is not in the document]
Julow: It's local but I agree. It is used only once anyway.

src/Fmt_odoc.ml
```diff
@@ -9,120 +9,218 @@
...
-  | RK_const -> str "!const:"
-  | RK_link -> str ":"
-  | RK_custom s -> str "!" $ str s $ str ":"
-
-let rec fmt_style style txt =
+let str_verbatim = Fmt.str
+
+let fmt_if_not_empty lst fmt = fmt_if (not (List.is_empty lst)) fmt
+
+let ign_loc f with_loc = f with_loc.Location_.value
```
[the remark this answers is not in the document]
Julow: Changed.

src/Fmt_odoc.ml
```diff
@@ -9,120 +9,218 @@
...
+      ("instance-variable:", dot (p :> t) (InstanceVariableName.to_string s))
+  | `Label (p, s) -> ("label:", dot (p :> t) (LabelName.to_string s))
+  | `Resolved r -> fmt_reference_resolved r
+
+let rec fmt_inline_element : inline_element -> Fmt.t = function
+  | `Space -> fmt "@ "
+  | `Word w ->
+      (* Escape lines starting with '+' or '-' *)
+      let escape =
+        if String.length w > 0 && Char.(w.[0] = '+' || w.[0] = '-') then
```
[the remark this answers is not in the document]
Julow: Yes.

src/Fmt_odoc.ml
```diff
@@ -152,39 +250,85 @@ and fmt_text txt =
...
   in
   hovbox 0 (list_pn txt f)
+    ]} *)
 
-let at () = char '@'
+and fmt_nestable_block_element : nestable_block_element -> t = function
+  | `Paragraph elems -> hovbox 0 (fmt_inline_elements elems)
+  | `Code_block s ->
+      vbox 0 (fmt "{[@;<1 -999>" $ str_verbatim s $ fmt "@ ]}")
+  | `Verbatim s -> vbox 0 (fmt "{v@;<1 -999>" $ str_verbatim s $ fmt "@ v}")
```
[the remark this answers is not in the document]
Julow: The goal is to remove any indentation. What are the alternatives ?

Julow:
Now that my PRs on Odoc have been merged (https://github.com/ocaml/odoc/pull/351, https://github.com/ocaml/odoc/pull/364), this can be implemented more easily.

Rebased on master and fixed a few bugs to to the rebase and changes in Odoc.
Currently, it is needed to pin Odoc to master.

test/passing/doc_comments.mli
Julow: This is now fixed.

src/Fmt_odoc.ml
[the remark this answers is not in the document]
Julow: I changed the code.

src/Sugar.mli
Julow: Whitespaces are preserved in verbatim/code blocks. The user is free to indent it.

ocamlformat.opam
```diff
@@ -29,7 +29,7 @@ depends: [
   "dune" {build & >= "1.1.1"}
   "fpath"
   "ocaml-migrate-parsetree" {>= "1.3.1"}
-  "octavius" {>= "1.2.0"}
+  "odoc" {>= "1.4.0"}
```
avsm:
```suggestion
  "odoc" {>= "1.4.1"}
```

just merged https://github.com/ocaml/opam-repository/pull/14349

ocamlformat_reason.opam
```diff
@@ -28,7 +28,7 @@ depends: [
   "dune" {build & >= "1.1.1"}
   "fpath"
   "ocaml-migrate-parsetree" {>= "1.0.10"}
-  "octavius" {>= "1.2.0"}
+  "odoc" {>= "1.4.0"}
```
avsm:
```suggestion
  "odoc" {>= "1.4.1"}
```

ocamlformat.opam
Julow: Finally available to `opam update`. Upgraded.

Julow:
Rebased on master. The history is a mess, this should be reviewed file by file.

gpetiot:
The diff with test_branch is:
```
diff --git a/src/Sugar.mli b/src/Sugar.mli
index 81514eb..db5f9f6 100644
--- a/src/Sugar.mli
+++ b/src/Sugar.mli
@@ -93,9 +93,9 @@ val ite :
     if-then-else expressions, e.g.:
 
     {[
-        if c1 then e1
-        else if c2 then e2
-        else e3
+if c1 then e1
+else if c2 then e2
+else e3
     ]}
 
     will return the following list:
@@ -141,11 +141,9 @@ val polynewtype :
     pattern [pat] with body [exp]. e.g.:
 
     {[
-        let f: 'r 's. 'r 's t = fun (type r) -> fun (type s) -> (e : r s t)
+let f: 'r 's. 'r 's t = fun (type r) -> fun (type s) -> (e : r s t)
     ]}
 
     Can be rewritten as:
 
-    {[
-        let f: type r s. r s t = e
-    ]} *)
+    {[ let f: type r s. r s t = e ]} *)
```

I still think starting the verbatim code blocks at column 0 looks weird, but it shouldn't delay the merge.

src/Fmt_ast.ml
```diff
@@ -415,14 +415,19 @@ let fmt_label lbl sep =
...
+  let location = loc.Location.loc_start in
+  let parsed = Odoc_parser.parse_comment_raw ~location ~text in
+  match parsed with
+  | {value; warnings= []} -> Ok value
+  | {warnings; _} ->
+      if not c.conf.quiet then
+        List.iter warnings ~f:(fun w ->
+            let msg = Odoc_model.Error.to_string w in
+            Caml.Format.eprintf
+              "%a:@,Warning: Invalid documentation comment:@,%s\n%!"
```
gpetiot: Maybe this warning should be printed only when `parse-docstrings=true`

Julow:
Thanks for the review. I improved the indentation of code blocks. I keep the behavior of fitting single line blocks.
I cleaned up a little and moved the printing of the warning.

What is not yet implemented is:
- Preserve list syntax
- Preserve whitespaces

I'm not sure if we should and how to preserve whitespaces. This should be done later on its own PR.

gpetiot:
`tools/test_branch.sh odoc_instead_of_octavius 'parse-docstrings=true' 'parse-docstrings=true'` shows the formatting of some docstrings became unstable (probably due to 87375aa)

Julow:
The unstable docstring errors are from the `master` build. They should all be fixed by this PR. This commit is clean.

gpetiot:
> The unstable docstring errors are from the `master` build. They should all be fixed by this PR. This commit is clean.

For example `ocamlformat test/code/infer/infer/src/biabduction/Prop.mli --no-version-check --parse-docstrings` fails with master but also with this PR, so it's not really a regression, but it would be nice if it would be fixed.

But we can fix it in another PR, feel free to merge.

Julow:
I can't reproduce this bug. It fails on master but not on this branch. Do you have an older version of this branch ? (I rebased it 3 days ago)

gpetiot:
> I can't reproduce this bug. It fails on master but not on this branch. Do you have an older version of this branch ? (I rebased it 3 days ago)

No I use your rebased version.

test/passing/doc_comments.mli
hhugo: There are sill new empty lines

test/passing/doc_comments.mli
```diff
@@ -123,9 +140,9 @@ val x : x
     This is preformated code:
 
     {[
-      let verbatim s =
-        s |> String.split_lines |> List.map ~f:String.strip
-        |> fun s -> list s "@," Fmt.str
+let verbatim s =
```
hhugo: Why is still code block indented to the left ?

test/passing/doc_comments.mli
hhugo: ignore me. I didn't see the the reference file.

test/passing/doc_comments.mli
hhugo: ignore me. I didn't see the the reference file.

test/passing/js_source.ml.ref
```diff
@@ -9759,7 +9759,8 @@ if fffffffffffffff aaaaa bb
 then (if b then aaaaaaaaaaaaaaaa ffff)
 else aaaaaaaaaaaa qqqqqqqqqqq
 
-include Base.Fn (** @open *)
+(** @open *)
```
hhugo: Why did the comment move before ?

test/passing/tag_only.ml
```diff
@@ -102,26 +103,33 @@ class b =
 
 [@@@ocamlformat "doc-comments-tag-only=fit"]
 
-open Module  (** @deprecated  *)
+open Module
+(** @deprecated  *)
```
hhugo: Why do we have a line break here ? I though `doc-comments-tag-only=fit` was for keeping theses as one-liner ?

Julow: Must have been lost in a rebase. Fixed.

Julow:
Fixed printing of `@see` tags and tag-only comments and removed unecessary spaces.

Julow:
I tested this with other projects and found some bugs:

- `*` character in the left margin not removed and treated as text

```diff
-val qubesdb: qubesdb typ
-(** For the Qubes target, the Qubes database from which to look up
- *  dynamic runtime configuration information. *)
+val qubesdb : qubesdb typ
+(** For the Qubes target, the Qubes database from which to look up * dynamic
+    runtime configuration information. *)
```

- Balanced brackets do not need to be escaped

```diff
-(** [intersperse xs ~sep] places [sep] between adjacent elements of [xs].  For example,
-    [intersperse [1;2;3] ~sep:0 = [1;0;2;0;3]]. *)
+(** [intersperse xs ~sep] places [sep] between adjacent elements of [xs]. For example,
+    [intersperse \[1;2;3\] ~sep:0 = \[1;0;2;0;3\]]. *)
```

- Breaks in inline code spans are preserved

```diff
-(** [drop l n] returns [l] without the first [n] elements, or the empty list if [n >
-    length l].  [drop l n] is equivalent to [snd (split_n l n)]. *)
+(** [drop l n] returns [l] without the first [n] elements, or the empty list if
+    [n >
+    length l]. [drop l n] is equivalent to [snd (split_n l n)]. *)
```

- Indentation of code blocks is not working when text starts on the same line as the opening

```diff
 (** Abc
-    - {[ a
-         b
+    - {[
+        a
+                b
       ]} *)
```

They may not be bugs or may be fixed later. I'm looking at the last one.

test/passing/doc_comments.mli.ref
```diff
@@ -0,0 +1,331 @@
...
+    ]}
+    {[
+      With empty
+
+      line
+    ]}
+    {[
+      First line
+            on the same line
+            as opening
```
Julow: This requires a fix to Odoc. https://github.com/ocaml/odoc/pull/370

Julow: It is now merged but will be fixed in OCamlformat only after the next release of Odoc.

test/passing/doc_comments.mli.ref
```diff
@@ -330,9 +330,9 @@ end
...
-(** [Multi
-     Line]
+(** [Multi Line]
 
-    [ A lot of    spaces ]
+    [A lot of spaces]
 
-    [Very looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong] *)
+    [Very
+    looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong] *)
```
Julow: Whitespaces are now normalized and allowed to break in code spans.

src/Fmt_odoc.ml
```diff
@@ -145,7 +145,7 @@ let rec fmt_inline_element : inline_element -> Fmt.t = function
       escape $ str_normalized w
   | `Code_span s ->
       let s = escape_brackets s in
-      hovbox 0 (wrap "[" "]" (str s))
+      hovbox 0 (wrap "[" "]" (str_normalized s))
```
gpetiot: The CI is fixed if you use `~escape:false`, is it necessary to escape all these characters?

Julow: It's not necessary to escape all of them, only `[` and `]`. Also `str_normalized` will remove consecutive whitespaces, which may not be what we want.

Julow: I changed to only escape brackets and to keep consecutive whitespaces. The CI is fixed.

gpetiot: I'm not convinced regarding keeping consecutive whitespaces, I think if someone would want to keep the spaces he would use the verbatim blocks instead.

Julow: I reverted that part.

Issue #722, opened 2019-03-27
Uneven spaces inside parentheses


pascutto:
Hi !
Thanks again for your work on ocamlformat !

With default options and master branch, this example code
```
let _ =
  ( module Ephemeron (HHHHHHHHHHHHHHHHHHHHHHHHHH) (HHHHHHHHHHHHHHHHHHHHHHHHHH)
  : Ephemeron.S )
```
gets formatted to
```
let _ =
  (module Ephemeron (HHHHHHHHHHHHHHHHHHHHHHHHHH) (HHHHHHHHHHHHHHHHHHHHHHHHHH)
  : Ephemeron.S )
```

I'm not sure why the space after the open parenthesis was removed while the close one was kept, but that does not seem right to me.

jberdine:
Thanks for reporting, you're right.

PR #724, opened 2019-03-27
Fix uneven spaces around module pack

Julow:
While investigating #722, I accidentally improved formatting of module packs:

```
let v =
  ( module struct
      let x = 3
  end : S )
```

I tried to do this before (#585), the issue have been lost.

This is still a work in progress, there is a few regressions:

- undo #714 for module packs
- alignment of with constraints with `janestreet` profile


gpetiot:
The indentation that you changed is the same that had to be reverted in #714

Julow:
I did not manage to make this PR work but succeed in #735.

PR #729, opened 2019-03-28
Improve: factorize with integrate_module

gpetiot:
Part of the redesign of fmt_module_* functions (more PRs to come).
Maybe the new function `integrate_module` could have a better name?

Diff of test_branch:

```ocaml
diff --git a/testsuite/tests/typing-modules/aliases.ml b/testsuite/tests/typing-modules/aliases.ml
index 8acf72863..7faa80739 100644
--- a/testsuite/tests/typing-modules/aliases.ml
+++ b/testsuite/tests/typing-modules/aliases.ml
@@ -217,7 +217,7 @@ module F (X : sig end) = struct
   module N' = N
 end
 
-module G : functor (X :sig  end) -> sig
+module G : functor (X : sig end) -> sig
   module N' : sig
     val x : int
   end
diff --git a/testsuite/tests/typing-modules/generative.ml b/testsuite/tests/typing-modules/generative.ml
index b38454a1f..1ce7daa5a 100644
--- a/testsuite/tests/typing-modules/generative.ml
+++ b/testsuite/tests/typing-modules/generative.ml
@@ -122,7 +122,7 @@ Error: Signature mismatch:
 
 module F3 () = struct end
 
-module F4 : functor (X :sig  end) -> sig end = F3
+module F4 : functor (X : sig end) -> sig end = F3
 
 (* fail *)
 [%%expect
@@ -145,7 +145,7 @@ module Y (X : sig end) (Y : sig end) (Z : sig end) = struct end
 
 module Z = functor (_ :sig  end) (_ :sig  end) (_ :sig  end) -> struct end
 
-module GZ : functor (X :sig  end) () (Z :sig  end) -> sig end =
+module GZ : functor (X : sig end) () (Z : sig end) -> sig end =
 functor (X :sig  end) () (Z :sig  end) -> struct end
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -61,7 +61,7 @@ let empty =
   ; esp= Fn.const ()
   ; epi= None }
 
-let integrate_module {opn; pro; psp; bdy; cls; esp; epi} ~f =
+let compose_module {opn; pro; psp; bdy; cls; esp; epi} ~f =
   opn $ f (Option.call ~f:pro $ psp $ bdy $ cls $ esp $ Option.call ~f:epi)
 
 (* Debug: catch and report failures at nearest enclosing Ast.t *)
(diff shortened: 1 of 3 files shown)


src/Fmt_ast.ml
```diff
@@ -61,6 +61,9 @@ let empty =
   ; esp= Fn.const ()
   ; epi= None }
 
+let integrate_module {opn; pro; psp; bdy; cls; esp; epi} ~f =
```
jberdine: maybe `compose` instead of `integrate`?

PR #732, opened 2019-03-28
Test: exclude gen_version.ml from test

jberdine:


test/Makefile
```diff
@@ -103,7 +103,7 @@ test_pull:
 	@rm -f $*.pp.mli $*.old.ast $*.new.ast
 	$(REFMT) --print=binary_reason $< | $(OCAMLFORMAT_EXE) $< --reason-intf -o $@
 
-TEST_ML:=$(shell find $(DIRS) -name _build -not -prune -or -path code/ocamlformat/test -not -prune -or -name '*'.ml -and -not -name '*'.pp.ml 2>/dev/null)
+TEST_ML:=$(shell find $(DIRS) -name _build -not -prune -or -name gen_version.ml -not -or -path code/ocamlformat/test -not -prune -or -name '*'.ml -and -not -name '*'.pp.ml 2>/dev/null)
```
Julow: This is not working for me. This works:

```
TEST_ML:=$(shell find $(DIRS) -name '*.ml' ! -name _build ! -name gen_version.ml ! -path 'code/ocamlformat/test/*' ! -name '*.pp.ml')
```

I think it would be simpler to list the source directories explicitly instead of filtering with `find`. What do you think ?

jberdine: Good catch, thanks! I messed up the commit and dropped a `-prune`. In your version, IIUC, find will descend into pruned dirs. I'm not sure about your setups, but I have huge _build dirs under there.

Re listing the dirs explicitly, I don't have a strong view, but it seems harder to keep a whitelist up-to-date since it will fail to pick up files silently.

jberdine:
Sorry, I screwed this up, fixing...

Issue #625, opened 2019-02-12
Bug in functor argument


Julow:
In the `test/passing/js_source.ml.ref`:

```
let magic : 'a 'b. 'a -> 'b =
  fun (type a b) (x : a) ->
   let module M =
     (functor
       (T
         :sig
           
           type 'a t
         end)
       ->
       struct
         let f (Refl : (a T.t, b T.t) eq) = (x :> b)
       end)
       (struct
         type 'a t = unit
       end)
   in
M.f Refl
```

gpetiot:
Fixed by #735

PR #735, opened 2019-03-28
Fix module pack and functor

### LINKED ISSUE #722: Uneven spaces inside parentheses

Julow:
I was struggling with #724 but thanks to #729, there is a simple and clean solution.
Fix #722 and #625.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -62,7 +62,7 @@ let empty =
   ; epi= None }
 
 let compose_module {opn; pro; psp; bdy; cls; esp; epi} ~f =
-  opn $ f (Option.call ~f:pro $ psp $ bdy $ cls $ esp $ Option.call ~f:epi)
+  f (Option.call ~f:pro $ opn $ psp $ bdy $ cls $ esp $ Option.call ~f:epi)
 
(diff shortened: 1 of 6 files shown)


gpetiot:
You forgot to apply `make fmt`

Julow:
> You forgot to apply `make fmt`

Fixed, sorry.

gpetiot:
> But in cases where there are multiple module (struct/sig) args, I'm not sure if we should use the docked style, and instead we should break between the args. What do others think?

I think it would look better if we break between each arg indeed.

PR #740, opened 2019-03-29
Improve: support --name with --inplace

jberdine:
In the case where a temporary file derived from some other file is to
be formatted in-place, it is necessary to specify the original file's
name to look up the config, etc.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1194,8 +1194,7 @@ let name =
     "Name of input file for use in error reporting. Defaults to the input \
      file name. Some options can be specified in configuration files named \
      '.ocamlformat' in the same or a parent directory of $(docv), see \
-     documentation of other options for details. Mutually exclusive with \
-     --inplace."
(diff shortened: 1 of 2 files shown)


src/Conf.ml
```diff
@@ -1489,8 +1489,6 @@ let validate () =
     `Error (false, "Cannot specify stdin together with other inputs")
   else if has_stdin && Option.is_none !name then
     `Error (false, "Must specify name when reading from stdin")
-  else if !inplace && Option.is_some !name then
```
hhugo: Should we fail if `--name` is used with multiple (more than one) inputs instead ?

jberdine: I don't know. In the case of the use in merge-fmt, currently ocamlformat is called three times on one file each time, always using the same name. In this case, a single call to ocamlformat with one name and all three inputs could be made instead. I don't know how over-specific that seems to you, but there is at least one use case.

PR #746, opened 2019-04-01
Implement doc-comments and doc-comments-tag-only for every items

### LINKED ISSUE #723: Doc comment for modules do not follow --doc-comments option
Hi !
Using master, and `--doc-comments=after`,
```
(** This is a comment *)
module Make = Map.Make
```
is left unchanged by ocamlformat, while the comment should be moved after the module.

Julow:
Fix #723

This PR implements `doc-comments=before|after` and `doc-comments-tag-only=default|fit` for every structure, signature and class items:

- Vals and externals
- Type declarations
- Modules
- Includes and opens
- Exceptions
- Classes

Previously, `doc-comments=after` was only implemented for `val` and `external` items
and `doc-comments-tag-only=fit` for `include`, `open` and one-line `module` items.

Also silently implement `doc-comments-padding` for tag-only comments.

Docstring are always placed before in some cases:
- `let` bindings
- Variant declarations
- Module/class declarations that are not "simple"


jberdine:
Note that it is intentional that code items that are often longer than their docstrings, such as type declarations, structures, signatures, etc. are placed after their docstrings while code items that are commonly shorter than their docstrings are placed before their docstrings. This possibility should be preserved. I don't know if only 3 choices (current, all before, all after) are sufficient, or if finer granularity is better.

Julow:
I rebased this PR, after Guillaume's tidying.
I disabled `doc-comments=after` when the module expression or type is not simple (using `Ast.module_*_is_simple`).

src/Ast.ml
```diff
@@ -2215,6 +2213,8 @@ end = struct
     | Ptyp_object _ -> true
     | _ -> false
 end
+(** Operations determining precedence and necessary parenthesization of
+    terms based on their super-terms. *)
```
gpetiot: module_expr_is_simple should not return true for this kind of module, we need to recursively call this function for Pmod_apply and Pmod_constraint (maybe other cases too if it makes sense, I didn't try it)

test/passing/tag_only.mli
```diff
@@ -1,64 +1,80 @@
...
-module A : B (** @deprecated  *)
+module A : B  (** @deprecated  *)
 
 (** @deprecated  *)
 module A : sig
   type t
 end
 
-module type A = B (** @open *)
+module type A = B  (** @open *)
```
gpetiot: If this intended?

test/passing/js_source.ml.ref
```diff
@@ -3629,7 +3629,7 @@ module SSMap = struct
 
   let of_t x = x
   let to_t x = x
-end
+  end
```
gpetiot: We should fix this regression.

src/Ast.ml
Julow: I improved this. I added some arbitrary rules (eg. 2 or more `with`-constraints is not simple, extensions are not simple, etc..). What do you think ?

test/passing/tag_only.mli
Julow: Yes. This is the `doc_comments_padding` option. It was only used on type declarations until now.
@jberdine What do you think of this behavior ?

test/passing/js_source.ml.ref
Julow: Fixed.

src/Translation_unit.mli
```diff
@@ -30,6 +29,7 @@ type 'a t =
       -> Normalize.docstring_error list
   ; normalize: Conf.t -> 'a with_comments -> 'a
   ; printast: Caml.Format.formatter -> 'a -> unit }
+(** Operations on translation units. *)
```
gpetiot: Maybe we need to add a `Ast.type_is_simple` for cases like this?

test/passing/object_type.ml
```diff
@@ -60,8 +59,9 @@ class type a =
     (** about b *)
     method b : int
   end
+(** about a *)
```
gpetiot: Here we need something similar to what is done for modules.

Julow:
I implemented `is_simple` for class expressions and types.
Not for types yet, this would be a regression.

gpetiot:
> I implemented `is_simple` for class expressions and types.
> Not for types yet, this would be a regression.

The output looks better.

We can merge as soon as you took care of:

> TODO:
>
>    Clean up docstring handling, especially for floating documentation

Julow:
I cleaned up the code. This is ready for review.

PR #748, opened 2019-04-01
Tidying fmt ast

gpetiot:
Goals:
- reduce the vertical and horizontal size of Fmt_ast (factorize, externalize nested functions)
- use adequate function if it makes the code more understandable (wrap, box, fmt_if, fmt_or)
- remove unused function parameters
- remove unnecessary variables (record fields)
- merge consecutive formats

More tidying to come in the future but the diff is big enough for now.

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -56,7 +56,7 @@ end = struct
       (List.find_map roots ~f:(fun root ->
            if Itv.contains root elt then
              Hashtbl.find_and_call tbl root
-               ~if_found:( fun children ->
+               ~if_found:(fun children ->
                  parent tbl children ~ancestor:root elt )
                ~if_not_found:Option.some
            else None ))
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -56,7 +56,7 @@ end = struct
       (List.find_map roots ~f:(fun root ->
            if Itv.contains root elt then
              Hashtbl.find_and_call tbl root
-               ~if_found:(fun children ->
+               ~if_found:( fun children ->
                  parent tbl children ~ancestor:root elt )
                ~if_not_found:Option.some
            else None ))
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1696,6 +1696,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
            (list_fl cnd_exps
               (fun ~first ~last (xcnd, xbch, pexp_attributes) ->
                 let parens_bch = parenze_exp xbch in
+                let imd = c.conf.indicate_multiline_delimiters in
                 match c.conf.if_then_else with
                 | `Compact ->
                     hovbox 0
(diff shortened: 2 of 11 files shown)


gpetiot:
@jberdine @hhugo @Julow any opinion about these changes? I think the major point about externalizing inline functions is to make factorization easier in the future, even if it can look like unnecessary noise for now.

gpetiot:

gpetiot:
So far the only changes in the tests are a fix of indicate-multiline-delimiters fot `fun` and `function`.

src/Cmts.ml
```diff
@@ -56,7 +56,7 @@ end = struct
       (List.find_map roots ~f:(fun root ->
            if Itv.contains root elt then
              Hashtbl.find_and_call tbl root
-               ~if_found:(fun children ->
+               ~if_found:( fun children ->
```
jberdine: Note that originally the space for indicate-multiline-delimiters is the one after `(fun`, that is, the whole of `(fun` is taken to be the delimiter since `fun` is not user code and must be followed by a space.

test/passing/js_source.ml.ref
```diff
@@ -3022,7 +3022,9 @@ let rec weaken_alist : type m n. (m, n) alist -> (m succ, n succ) alist = functi
 let rec sub' : type m. m ealist -> m fin -> m term = function
   | EAlist Anil -> var
   | EAlist (Asnoc (s, t, x)) ->
-    comp_subst (sub' (EAlist (weaken_alist s))) (fun t' -> weaken_term (subst_var x t t'))
+    comp_subst
+      (sub' (EAlist (weaken_alist s)))
+      (fun t' -> weaken_term (subst_var x t t'))
```
jberdine: ?

src/Fmt_ast.ml
```diff
@@ -166,44 +156,51 @@ let maybe_disabled_k c (loc : Location.t) l f k =
...
     in
-    k (list l "@\n" (fun s -> str s))
+    k (list l "@\n" str)
 
 let maybe_disabled c loc l f = maybe_disabled_k c loc l f Fn.id
 
 let update_config_maybe_disabled c loc l f =
-  let c = update_config c l in
-  maybe_disabled c loc l f
+  maybe_disabled (update_config c l) loc l f
```
jberdine: It is technically equivalent here, but the previous form is preferable I think since it emphasizes the stateful/monadic nature of update_config, and can be written the same everywhere in the previous form, where in the current one we need to remember to undo this change as soon as another use of `c` is added.

test/passing/js_source.ml.ref
gpetiot: It is exceeding the 90 char margin:
```
$ echo "    comp_subst (sub' (EAlist (weaken_alist s))) (fun t' -> weaken_term (subst_var x t t'))" | wc -c
91
```

src/Cmts.ml
gpetiot: I will revert it.

src/Fmt_ast.ml

src/Params.ml
```diff
@@ -68,9 +68,61 @@ let all ~first ~indent ~parens_here =
...
+type record_type =
+  { docked_before: Fmt.t
+  ; break_before: Fmt.t
+  ; box_record: Fmt.t -> Fmt.t
+  ; sep_before: Fmt.t
+  ; sep_after: Fmt.t
+  ; break_after: Fmt.t
+  ; docked_after: Fmt.t }
+
+let before (c : Conf.t) ~wrap_record =
```
Julow: I think you should move these functions either in a module `Record_type` or, why not, inline them in `get_record_type`.

gpetiot: I inlined the functions, using a module would be contrary to what was decided for `get_cases`.

Julow: Thanks, it's more readable. Could you do the same for `get_cases` ?

gpetiot: Done.

PR #750, opened 2019-04-02
Fix break-fun-decl with janestreet profile

### LINKED ISSUE #728: Missing break after equal with --profile janestreet
```
let f ~first_arg
      ~x (* this is a comment about x *)
      ~last_arg = assert false
```

gpetiot:
Fix #728 
The diff of test_branch for the janestreet profile looks good.

### DIFF
--- a/test/passing/js_source.ml.ref
+++ b/test/passing/js_source.ml.ref
@@ -1458,8 +1458,8 @@ let ty_abc : (([ `A of int | `B of string | `C ] as 'a), 'e) ty =
          ]
 
        method inj
-           : type c.  (int -> string -> noarg -> unit, c) ty_sel * c
-                   -> [ `A of int | `B of string | `C ] =
+             : type c.  (int -> string -> noarg -> unit, c) ty_sel * c
+                     -> [ `A of int | `B of string | `C ] =
          function
          | Thd, v -> `A v
          | Ttl Thd, v -> `B v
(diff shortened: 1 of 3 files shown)


test/passing/js_source.ml.ref
```diff
@@ -1458,8 +1458,8 @@ let ty_abc : (([ `A of int | `B of string | `C ] as 'a), 'e) ty =
          ]
 
        method inj
-           : type c.  (int -> string -> noarg -> unit, c) ty_sel * c
-                   -> [ `A of int | `B of string | `C ] =
+             : type c.  (int -> string -> noarg -> unit, c) ty_sel * c
+                     -> [ `A of int | `B of string | `C ] =
```
gpetiot: I didn't manage to get rid of this regression, it doesn't seem that negative to me in comparison to the other improvements.

Julow: I tried as well but no success. `fmt_kind` deserves a little refactor to return something better than `l, eq, expr`. The `l` makes no sense to me.

gpetiot: I agree, but maybe it can be fixed in another PR (as long as @jberdine @hhugo don't mind about this regression in the meantime)

test/passing/js_source.ml.ref
```diff
@@ -9793,3 +9790,28 @@ let _ =
...
+    ~(* last argument *) last_arg (* last argument *)
+  =
+  assert false
+;;
+
+let f ~first_arg ~x ~(* this is a comment about x *) last_arg (* last argument *) =
+  assert false
+;;
+
+let f ~first_arg ~x ~(* this is a comment about x *) last_arg = assert false
```
Julow: All this format exactly the same on master.

gpetiot: Damn you are right, I'm curious what PR fixed this. I'm closing this one.

PR #752, opened 2019-04-02
Fix grammar: it's -> its

anmonteiro:
Took me longer than I'd like to admit to understand what this option did, until I noticed the mistake.


avsm:
thanks!

PR #755, opened 2019-04-03
Fix inconsistent break before module signature

### LINKED ISSUE #572: Inconsistent break before module signature when there is constraint
When `with type ...` constraints are present, an extra break is added before the `sig`

```
module O :
  sig
	type t
  end
  with type t := t = struct end
```

Without constraints:

```
module O : sig
  type t
end = struct end
```

gpetiot:
Fix #572 
I think the output looks better now than what was proposed in #572, and besides the changes on comments the diff of test_branch looks good as well.


gpetiot:
I added some tests, I would like @jberdine on this PR before merging

PR #758, opened 2019-04-04
Fix: if $HOME/.config is either not set or empty, a default equal…

### LINKED ISSUE #757: XDG_CONFIG_HOME, follow the specification's "should".
The [current code](https://github.com/ocaml-ppx/ocamlformat/blob/5c130c7bc5cb53ef1c4ac2cfc0653fdc4c4855b7/src/Conf.ml#L1739-L1747) does nothing in case `XDG_CONFIG_HOME` is undefined. 

However the [spec](
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) says a default of `$HOME/.config` *should* be used. I'd be glad if that would be the case, on macos the XDG variables are usually undefined and the spec's default are reasonable. 

Also looking at this code it seems it looks for a `.ocamlformat` in there. I don't think there's a need to hide things a second time. It makes things more obscure. E.g. if I simply `ls` `$XDG_CONFIG_HOME/ocamlformat` and see nothing I personally will think no option has been set there.

gpetiot:
… to $HOME/.config should be used, and use `ocamlformat` file instead

Fix #757

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1744,13 +1744,17 @@ let kind_of fname =
 let xdg_config =
   let xdg_config_home =
     match Caml.Sys.getenv_opt "XDG_CONFIG_HOME" with
-    | None | Some "" ->
-        let home = Option.value ~default:"" (Caml.Sys.getenv_opt "HOME") in
-        Fpath.(v home / ".config")
-    | Some xdg_config_home -> Fpath.v xdg_config_home
+    | None | Some "" -> (
+      match Caml.Sys.getenv_opt "HOME" with
+      | None | Some "" -> None
+      | Some home -> Some Fpath.(v home / ".config") )
+    | Some xdg_config_home -> Some (Fpath.v xdg_config_home)
   in
-  let filename = Fpath.(xdg_config_home / "ocamlformat") in
-  if Fpath.exists filename then Some filename else None
+  match xdg_config_home with
+  | Some xdg_config_home ->
+      let filename = Fpath.(xdg_config_home / "ocamlformat") in
+      if Fpath.exists filename then Some filename else None
+  | None -> None
 
 let is_ignored ~quiet ~ignores ~filename =
   let drop_line l = String.is_empty l || String.is_prefix l ~prefix:"#" in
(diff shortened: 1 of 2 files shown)


src/Conf.ml
```diff
@@ -1737,14 +1739,15 @@ let kind_of fname =
...
-        Fpath.(
-          append (v xdg_config_home) (v "ocamlformat" / dot_ocamlformat))
-      in
-      if Fpath.exists filename then Some filename else None
-  | None -> None
+  let xdg_config_home =
+    match Caml.Sys.getenv_opt "XDG_CONFIG_HOME" with
+    | None | Some "" ->
+        let home = Option.value ~default:"" (Caml.Sys.getenv_opt "HOME") in
+        Fpath.(v home / ".config")
```
Julow: If `HOME` is unset, `home` is the empty string and this raises `Invalid_argument`. I think returning `None` is this case is ok.

src/Conf.ml
```diff
@@ -1737,14 +1739,15 @@ let kind_of fname =
...
+    | Some xdg_config_home -> Fpath.v xdg_config_home
+  in
+  let filename = Fpath.(xdg_config_home / "ocamlformat") in
```
Julow: Shouldn't this be `xdg_config_home / "ocamlformat" / dot_ocamlformat` ?

gpetiot: It was suggested in the issue to write the conf in an `ocamlformat` file instead of an hidden `.ocamlformat` file, I agree with that, but if I'm alone on this one we can go back to an `.ocamlformat` file.

jberdine: The rest of the docs refer to `.ocamlformat`, we shouldn't change the name depending on the directory.

gpetiot: The doc has been updated according to this change, but do you want to keep the current architecture: `xdg_config_home / "ocamlformat" / dot_ocamlformat`?

Julow: Sorry I didn't see your suggestion. I agree `xdg_config_home / "ocamlformat"` makes sense.
@jberdine I don't think we should put files starting with a dot in ".config" or sub-directories, I think it is a common pattern to remove the "." inside ".config"

jberdine: I don't agree, but won't make an issue of it.

gpetiot:
> I think this is not working.

What issue do you see exactly? I did some tests and didn't notice anything wrong.

PR #760, opened 2019-04-04
Fix nested parens with no break infix before func

### LINKED ISSUE #754: BUG: ast changed with Pexp_function inside match
The following snippet causes ocamlformat to emit the "BUG: ast changed." error:

```ocaml
match foo with
| 1 ->
  bar >>= (function | _ -> ())
| other -> ()

```

gpetiot:
Fix #754, I don't know if we should bother trying to handle `indicate-multiline-delimiters` since the beginning and the end of the function are not in the same box and we don't want to rely on `Location.is_single_line`?

No diff with test_branch with `break-infix-before-func=false`.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1442,6 +1442,8 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       , [(Nolabel, l); (Nolabel, ({pexp_desc= Pexp_function cs} as r))] )
     when is_infix_id id && not c.conf.break_infix_before_func ->
       Cmts.relocate c.cmts ~src:pexp_loc ~before:loc ~after:loc ;
+      let xpc_rhs = sub_exp ~ctx r in
+      let parens_fct = parenze_exp xpc_rhs in
       wrap_if parens "(" ")"
         (hvbox 0
            ( hvbox 0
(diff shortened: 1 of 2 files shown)


src/Fmt_ast.ml
```diff
@@ -1443,17 +1443,19 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       , [(Nolabel, l); (Nolabel, ({pexp_desc= Pexp_function cs} as r))] )
     when is_infix_id id && not c.conf.break_infix_before_func ->
       Cmts.relocate c.cmts ~src:pexp_loc ~before:loc ~after:loc ;
+      let xpc_rhs = sub_exp ~ctx r in
```
jberdine: Here the `xpc_rhs` copypasta name doesn't really make sense, `xr` is likely better.

src/Fmt_ast.ml
```diff
@@ -1443,17 +1443,19 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+      let parens_fct = parenze_exp xpc_rhs in
```
jberdine: Here it might be better to use `_r` instead of `_fct`. Or to rename `r`.

gpetiot:
> I wonder if it might be better to assume the cases don't fit on a line instead, as I'd guess that more than half of the time that's what happens.

I agree, and the output looks good.

PR #761, opened 2019-04-05
Improve: --disable-outside-detected-project is set by default

### LINKED ISSUE #745: Should `--disable-outside-detected-project` be the default?
It seems that most of the time the safe default is to use `--disable-outside-detected-project`. What do others think?

gpetiot:
Fix #745

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1092,7 +1092,7 @@ let disable_conf_files =
   mk ~default:false
     Arg.(value & flag & info ["disable-conf-files"] ~doc ~docs)
 
-let disable_outside_detected_project =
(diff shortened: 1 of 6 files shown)


src/Conf.ml
```diff
@@ -1549,7 +1547,7 @@ let root =
   Option.map !root ~f:Fpath.(fun x -> v x |> to_absolute |> normalize)
 
 let disable_outside_detected_project =
```
jberdine: It will increase the size of the diff slightly, but I think this should also be renamed.

gpetiot:
I agree, I kept the old option and a warning message is emitted.

PR #762, opened 2019-04-05
Improve: less breaks for break-cases=all but correctly breaks or-patterns

### LINKED ISSUE #597: Replace break-cases=all with break-cases=toplevel ?
(not for 0.9)

The option `break-cases=toplevel` introduced in #536 is a variant of `break-cases=all` where or-patterns are kept on the same line whenever possible.
`break-cases=all` is kept for ocamlformat 0.9 but we could replace it with `toplevel` for the major release.

Any thoughts on this?

gpetiot:
- From https://github.com/janestreet/ocamlformat/commit/a6af924d9bf4184b7c64d697824f4d51d8bd6a29
- also fixes the incorrect OCaml code generated issue of  test/code/ocaml/asmcomp/cmmgen.ml with `break-cases=all`
- Fix #597 

No issue with test_branch with or without `break-cases=all`.
I will change the value of this option in the janestreet profile if the new output is close enought to what @hhugo had in mind.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -866,14 +866,21 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
         (wrap_array c
            (list pats "@;<0 1>; " (sub_pat ~ctx >> fmt_pattern c)))
   | Ppat_or _ ->
+      let has_doc = not (List.is_empty xpat.ast.ppat_attributes) in
       let nested =
         match ctx0 with
-        | Pat {ppat_desc= Ppat_or _} -> true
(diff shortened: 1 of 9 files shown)


src/Fmt_ast.ml
```diff
@@ -950,7 +958,10 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
                   let pro =
                     if first_grp && first then pro0 $ open_hovbox (-2)
                     else if first then proI $ open_hovbox (-2)
-                    else pro2
+                    else
+                      pro2
```
Julow: This is called once, there is no need for the optional argument. And it does not format well.

gpetiot:
There is a regression on js_source if I update the janestreet profile.

PR #763, opened 2019-04-06
Support for OCaml 4.08

### LINKED ISSUE #737: fix for the upcoming ocaml compiler (4.08)
- base : https://github.com/janestreet/base/pull/70
- migrate-parsetree : https://github.com/hhugo/ocaml-migrate-parsetree/tree/408-2
- octavius https://github.com/hhugo/octavius/tree/408

hhugo:
4.08 can be tested with the following pin:
- base : https://github.com/janestreet/base/pull/70
- migrate-parsetree : https://github.com/hhugo/ocaml-migrate-parsetree/tree/408-2
- octavius https://github.com/hhugo/octavius/tree/408

fix #737


test/passing/exceptions.ml
```diff
@@ -1,4 +1,4 @@
-exception EvalError of Error.t [@@deriving sexp]
+exception EvalError of Error.t
```
hhugo: @xclerc, your migrate-parsetree branch might need to be updated.
In 4.07, the following where equivalent:
- `exception EvalError of Error.t [@@deriving sexp]`
- `exception EvalError of Error.t [@deriving sexp]`

It's no longer the case in 4.08, the downgrade 4.08 -> 4.07 is one of the two on the floor.

src/Reason.ml
```diff
@@ -30,12 +30,12 @@ module Binary_reason = struct
      (src/migrate_parsetree_ast_io.ml) *)
   type ast =
     | Impl :
-        (module Migrate_parsetree_versions.OCaml_version
+        (module Migrate_parsetree.Versions.OCaml_version
```
hhugo: We're waiting for a release of migrate-parsetree to push that on master (https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/67)

test/passing/exceptions.ml
hhugo: I've made the change in https://github.com/hhugo/ocaml-migrate-parsetree/commit/941337e273bcb19531be54c3dc22a95cf1a1c1e4

hhugo:
Ready to be merged after we agree on how conditional compilation is implemented.

gpetiot:
I don't know if the lt_408 / ge_408 files are so different that it is justified to have 2 separate files (and each file is rather short), I would have put everything in a single file and used `cppo` instead.

hhugo:
Reasons for the current design are:
- less dependencies
- better integration with tools, in particular, ocamlformat itself.

hhugo:
FYI, conditional compilation mechanism was taken from ocaml-migrate-parsetree.
I do not feel strongly either way. I'll let @jberdine make the call.

src/config/gen.ml
```diff
@@ -0,0 +1,22 @@
+(**********************************************************************
```
Julow: Shouldn't this be in `tools/gen_compat.ml` ? Near `gen_version.ml`.

Makefile
```diff
@@ -55,7 +55,7 @@ cleanbisect:
 	rm -Rf _coverage
 	find ./ -name 'bisect*.out' -delete
 
-SRCS=$(shell \ls {src/{,import/},tools/ocamlformat-diff/}*.ml{,i})
+SRCS=$(shell \ls {src/{,import/,config/,compat/},tools/ocamlformat-diff/}*.ml{,i})
```
Julow: This gives 
```
ls: cannot access 'src/config/*.mli': No such file or directory
ls: cannot access 'src/compat/*.mli': No such file or directory
```
I suggest you change to
```
SRCS=$(shell find src tools/ocamlformat-diff -name '*.ml' -or -name '*.mli')
```

Julow: Also, `tools/gen_version.ml` could be included.

hhugo: We do not include tools/gen_version.ml on purpose (see https://github.com/ocaml-ppx/ocamlformat/pull/732)

src/config/gen.ml
Julow: The other one is in ocamlformat_support, which is kind of an other project. I understand the reason (local) but it feels weird to have it in the `src` directory.

hhugo: You've convinced me.

src/Cmts.ml
```diff
@@ -26,7 +26,7 @@ type t =
     whenever 2 intervals share more than an end-point, then one contains the
     other. *)
 module Non_overlapping_interval_tree (Itv : sig
-  include Hashtbl.Key
+  include Hashtbl.Key[@@ocaml.warning "-3"]
```
jberdine: Tangential, but is it intended that there is no space before the attribute here?

src/compat/ge_408.ml
```diff
@@ -0,0 +1,28 @@
+(**********************************************************************
```
jberdine: I suppose it could be helpful to have a single `compat.mli` to summarize both implementations, right?

src/compat/ge_408.ml
```diff
@@ -0,0 +1,28 @@
...
+ *                            OCamlFormat                             *
+ *                                                                    *
+ *  Copyright (c) 2019-present, Facebook, Inc.  All rights reserved.  *
+ *                                                                    *
+ *  This source code is licensed under the MIT license found in the   *
+ *  LICENSE file in the root directory of this source tree.           *
+ *                                                                    *
+ **********************************************************************)
+
+let _ = Clflags.error_style := Some Misc.Error_style.Short
```
jberdine: Just FMI, `_` instead of `()` for any particular reason?

src/dune
```diff
@@ -15,6 +15,15 @@ Jbuild_plugin.V1.send
...
 
+(rule
+ (copy#
+  compat/%%{read:compat-file}
+  compat.ml))
+
+(rule
+ (targets compat-file)
+ (action  (run %%{ocaml} %%{dep:../tools/gen_compat.ml} %%{ocaml_version})))
+
```
jberdine: I can't help but think that this mechanism looks a lot like dune's virtual libraries, where choosing one based on the compiler version could be a mild extension to the existing [Alternative dependencies](https://dune.readthedocs.io/en/latest/dune-files.html#alternative-deps) mechanism.

src/compat/ge_408.ml
hhugo: none, will fix.

src/Cmts.ml
hhugo: I'll let @Julow or @gpetiot take a look.

Julow: https://github.com/ocaml-ppx/ocamlformat/pull/775

PR #765, opened 2019-04-08
Fix module-item-spacing=sparse

gpetiot:
The selected value for ocamlformat is module-item-spacing=sparse but a lot of spaces were missing, also the implementation should be cleaner now.

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -12,7 +12,6 @@
 (** Abstract syntax tree term *)
 
 open Migrate_ast
-
 open Parsetree
 
 let init, register_reset, leading_nested_match_parens, parens_ite =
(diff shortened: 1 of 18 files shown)


gpetiot:
I would like @jberdine's opinion on this one since it changes the formatting of ocamlformat's codebase.

src/Ast.ml
```diff
@@ -261,26 +266,23 @@ module Structure_item : Module_item with type t = structure_item = struct
...
-        )
-       |Pstr_modtype _, Pstr_modtype _
-       |Pstr_class _, Pstr_class _
-       |Pstr_class_type _, Pstr_class_type _
-       |Pstr_attribute _, Pstr_attribute _
-       |Pstr_extension _, Pstr_extension _ ->
-          true
-      | _ -> false )
-    | _ -> true
+  let allow_adjacent itmI cI itmJ cJ =
```
jberdine: Pairing the item and the config that pertains to it is intended to help clarify the meaning of the arguments. Also, elsewhere the common idiom is to use such pairs. Were they removed here for any particular reason?

jberdine:
But if there is demand, another alternative could be added which has the behavior of sparse in this PR.

gpetiot:
If the produced output is not what is expected (I am not aware of demand from users, I thought I was fixing some bugs) I will juts close this one.

Issue #747, opened 2019-04-01
Mimimize sensitivity to concrete syntax


jberdine:
There are multiple points in the formatting code which are sensitive to concrete syntax that are not about something which necessarily involves the concrete syntax. There are (at least) 2 different uses in the implementation:

1. Features such as preserving string escaping or preserving open lines necessarily need to be sensitive to concrete syntax.

2. Other uses, such as to determine box structure, should not be sensitive to the concrete syntax.

This task is to remove all of the 2nd sort. An initial step will be to remove most uses of `Location.is_single_line`.

PR #767, opened 2019-04-08
Improve: less sensitivity to concrete syntax

gpetiot:
Starts to address #747 (4 calls to `Location.is_single_line` removed, a lot more to go, but the diff is already big enough).

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -197,14 +197,22 @@ let doc_atrs atrs =
   let docs = match docs with [] -> None | l -> Some (List.rev l) in
   (docs, List.rev rev_atrs)
 
-let longident_is_simple (c : Conf.t) x =
+(** [fit_margin c x] returns [true] if and only if [x] does not exceed 1/3
+    of the margin. *)
+let fit_margin (c : Conf.t) x = x * 3 < c.margin
+
+(** [longident_fit_margin c x] returns [true] if and only if [x] does not
+    exceed 2/3 of the margin. *)
+let longident_fit_margin (c : Conf.t) x = x * 3 < c.margin * 2
+
+let longident_is_simple c x =
   let rec length (x : Longident.t) =
     match x with
     | Lident x -> String.length x
     | Ldot (x, y) -> length x + 1 + String.length y
     | Lapply (x, y) -> length x + length y
   in
-  length x * 3 < c.margin * 2
+  longident_fit_margin c (length x)
 
 let module_type_is_simple x =
   match x.pmty_desc with Pmty_signature _ -> false | _ -> true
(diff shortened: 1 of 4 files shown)


test/passing/tag_only.ml
```diff
@@ -126,11 +126,11 @@ module A : sig
...
 (** @deprecated  *)
+open Module.With_veryyyyyy_loooooooooooooooooooooooong_naaaaaaaaaaaaaaaaame
 
-include Module.With_very_loooooooooooooooooooooooong_naaaaaaaaaaaaaaaaame
 (** @deprecated  *)
+include Module.With_very_loooooooooooooooooooooooong_naaaaaaaaaaaaaaaaame
 
-module A = Module.With_very_loooooooooooooooooooooooong_naaaaaaaaaaaaaaaaame
 (** @deprecated  *)
+module A = Module.With_very_loooooooooooooooooooooooong_naaaaaaaaaaaaaaaaame
```
gpetiot: Since the module name is long I think it makes sense to put the docstring before.

src/Ast.ml
```diff
@@ -197,6 +197,28 @@ let doc_atrs atrs =
...
   (docs, List.rev rev_atrs)
 
+let longident_is_simple (c : Conf.t) x =
+  let rec length (x : Longident.t) =
+    match x with
+    | Lident x -> String.length x
+    | Ldot (x, y) -> length x + 1 + String.length y
+    | Lapply (x, y) -> length x + length y
+  in
+  length x * 3 < c.margin
```
Julow: Why `* 3` ?

test/passing/tag_only.ml
Julow: I disagree. It feels arbitrary and may be reported as a bug in the future.

src/Fmt_ast.ml
```diff
@@ -3406,7 +3393,14 @@ and fmt_module c ?epi keyword name xargs xbody colon xmty attributes =
                    | _ -> fmt "" )
                  $ Option.call ~f:epi ) ))
   in
-  (fmt_docstring_around ~loc c doc)
+  let single_line =
+    String.length name.txt * 3 < c.conf.margin
```
Julow: `* 3` ? Maybe module_*_is_simple could also return the length of simple expressions, so that this can be more precise.

src/Ast.ml
jberdine: Note that such a criterion for "simple" is also used elsewhere, and IIRC a third of the margin is used everywhere. We could make the ratio configurable, it is arbitrary to a degree. The original thinking behind it is that it is small enough that only in very extreme cases will the indentation be 2/3 or the margin. Why a third instead of a half or whatever is a good question, and if some empirical evaluation was done that would be good. FTR I tried 1/2 and it seemed to allow some expressions which were long enough that grouping them together in wrapped application arguments started to get potentially hard to read, so 1/3 seemed better.

jberdine: The context here is somewhat different, as we do not expect multiple simple e.g. module aliases on a line, so perhaps something more like 2/3 would be better, to remain basically insensitive to the starting indentation.

gpetiot:
I fixed the regressions and made everything more predictable.

test/passing/tag_only.ml
gpetiot: Fixed.

src/Ast.ml
Julow: I understand. I'm still not convinced that this is a good idea. It seems too arbitrary and someone may report this as a bug later.

For the code, this must be documented. Also, I think it would be more understandable if the maths were done on the rhs (`length x < c.margin / 3`).

gpetiot: I agree it's not ideal but at least it's consistent with the rest of the code. I added new functions and comments for consistency.

src/Ast.ml
```diff
@@ -197,6 +197,29 @@ let doc_atrs atrs =
...
+(** [longident_fit_margin c x] returns [true] if and only if [x] does not
+    exceed 2/3 of the margin. *)
+let longident_fit_margin (c : Conf.t) x = x * 3 < c.margin * 2
+
+let longident_is_simple c x =
+  let rec length (x : Longident.t) =
+    match x with
+    | Lident x -> String.length x
+    | Ldot (x, y) -> length x + 1 + String.length y
+    | Lapply (x, y) -> length x + length y
```
Julow: Missing `+ 3` for the parenthesis and the space.

PR #768, opened 2019-04-10
Improve: new option let-module

### LINKED ISSUE #628: Break before multi-line expression in `let module X =` body
Current formatting:

```
  let module X = Map.Make (struct
	type t = t

	let compare = compare
  end) in
```

Expected output:

```
  let module X =
    Map.Make (struct
      type t = t

      let compare = compare
    end)
  in
```

gpetiot:
Fix #628, the js_source.ml file looks better.

I will wait for feedback before integrating these changes in the unstable branch.


src/Fmt_ast.ml
```diff
@@ -1866,6 +1866,29 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
             $ fmt "@;<1000 0>"
             $ fmt_expression c (sub_exp ~ctx exp) )
         $ fmt_atrs )
+  | Pexp_letmodule (name, pmod, exp) when Poly.(c.conf.let_module = `Sparse)
```
Julow: I don't like the code duplication. I think it would be better to clean up the previous code to allow this feature instead.

src/Fmt_ast.ml
```diff
@@ -1866,6 +1866,29 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+            (parens || not (List.is_empty pexp_attributes))
+            "(" ")"
+            ( hvbox 2
+                ( hovbox 2
+                    ( hovbox 4
+                        ( fmt "let module"
+                        $ fmt_extension_suffix c ext
+                        $ fmt " " $ str name.txt )
+                    $ fmt_or_k c.conf.ocp_indent_compat
+                        (fits_breaks " =" "@;<1000 0>=")
```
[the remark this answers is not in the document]
gpetiot: It comes from the `let` value binding function, but I will add more tests to see if it really makes sense.

src/Conf.ml
```diff
@@ -876,6 +877,24 @@ module Formatting = struct
...
+        , "$(b,compact) does not break a line after the $(i,let module ... \
+           =) and before the $(i,in) if the module declaration does not \
+           fit on a single line." )
+      ; ( "sparse"
+        , `Sparse
+        , "$(b,sparse) breaks a line after $(i,let module ... =) and \
+           before the $(i,in) if the module declaration does not fit on a \
+           single line." ) ]
+    in
+    C.choice ~names:["let-module"] ~all ~doc ~section
```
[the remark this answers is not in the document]
gpetiot: Yes to not have regressions. But we can add `let-module=sparse` to the conventional profile if people ask for it.

src/Fmt_ast.ml
```diff
@@ -1880,14 +1880,29 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+      let let_module =
+        match c.conf.let_module with
+        | `Compact ->
+            fmt_module c keyword name xargs (Some xbody) true xmty []
+              ~epi:(fmt "in")
+        | `Sparse ->
+            hovbox 2
+              ( hovbox 4
+                  ( fmt "let module"
+                  $ fmt_extension_suffix c ext
```
Julow: Should use `keyword` here.

src/Fmt_ast.ml
```diff
@@ -1880,14 +1880,29 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
             (sub_mod ~ctx body_me, Some (sub_mty ~ctx body_mt))
         | _ -> (xbody, None)
       in
+      let let_module =
+        match c.conf.let_module with
+        | `Compact ->
+            fmt_module c keyword name xargs (Some xbody) true xmty []
+              ~epi:(fmt "in")
+        | `Sparse ->
+            hovbox 2
```
Julow: Would it be possible to implement `compact` by sharing the code with `sparse` ? (eg. using the `Params` architecture)
It makes the code a little bit hard to read. Also the `xargs, xbody` code is only used by the `compact` branch.

gpetiot: I tried to do that, but the box layout is very different in the 2 cases, but maybe it can be done later as more of the formatting data is moved to `Params` anyway.

Issue #776, opened 2019-04-16
Possible edge case


chrismamo1:
Trying to use ocamlformat on a large project, I've started getting this error message:

```
ocamlformat: Cannot process "redacted.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: ast changed.
```

I've whittled the problem down to a rather specific-looking case:

```
let%any_extension f = function()
  | _ -> ()
```

Basically, any time a top-level `let`-extension defines a `function`. Am I doing something obviously wrong?

chrismamo1:
I'm doing this for my job, so I'm quite motivated to fix the problem, but I'm totally unfamiliar with this project. If a developer could point me in the right direction then I'll certainly try to fix the issue.

jberdine:
Thanks for your interest!

I suspect the bug is that the extension is formatter [here](https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Fmt_ast.ml#L3807) and then the body is formatted without clearing the extension [here](https://github.com/ocaml-ppx/ocamlformat/blob/master/src/Fmt_ast.ml#L3800). Probably `?ext` in the second one should be removed.

chrismamo1:
I've just encountered a similar issue that doesn't appear to have such a straightforward fix. I get the error `BUG: generating invalid ocaml syntax.` for this file:

```
type%any_extension t =
  < a: 'a >
```

PR #779, opened 2019-04-17
fix minor issue with handling certain edge cases

chrismamo1:
#776


gpetiot:
Thanks! The diff of test_branch.sh does not exhibit any regression.
Can you add a test (like the function of #776) to the tests codebase?
And we can merge as soon as the CLA is signed.

chrismamo1:
Added a test.

PR #780, opened 2019-04-17
Improve: indentation consistency of '<-' and `:=`

### LINKED ISSUE #778: `<-` isn't always indented
see [this line](https://github.com/AestheticIntegration/minisat-ml/blob/06ddb50dd5da52466432c431ddf9d16d5ea34d7c/src/lib/Minisat_solver.ml#L965)

I think

ocaml
```
x <- foo
```

should become, when a break is inserted:

```ocaml
x
  <- foo
``` 

rather than

```
x
<- foo
```

gpetiot:
Fix #778 
The diff of test_branch looks good.
I didn't change the formatting of infix operators as suggested in the issue, it can be done in a separate PR (although I'm not convinced it should be modified at all).


Julow:
What about moving the arrow the end of the previous line ?
To be consistent with `:=`:
```ocaml
let _ =
  entry.logdata.value_end :=
    entry.logdata.value_end - !remove_size + testtesttest
```
and not indented if the rhs is complicated:
```ocaml
let _ =
  (*a*)
  s (*b*).[(*c*)
           (*d*)
           i
           (*e*)]
  := (*f*)
     (*g*)
     x
```

gpetiot:
I think it would be more consistent to put `:=` at the beginning of the next line in the first case (we aim to always have the symbols at the beginning of the line by default)

gpetiot:
I made the same changes for `:=` and I think it looks better (the diff of test_branch looks good too).

jberdine:
I don't have a strong position either way, but note that I was previously convinced to format `:=` unlike other binops and instead more like `=` in `let`. Since it doesn't make sense to have nested assignment expressions (unlike C), the argument for placing `:=` on the left is not so strong as compared to "normal" binops.

I agree that `:=` and `<-` ought to be formatted similarly to each other. Just to note that in the past there was a definite request for the current formatting of `:=`. Is it worth an option?

gpetiot:
The difference is that if there is a `let` at the beginning of the line you immediately see it's a binding, whereas you have to go at the end of the line to look for `:=` in the other case.

```ocaml
let entry.logdata.value_end =
  entry.logdata.value_end - !remove_size + testtesttest

entry.logdata.value_end :=
  entry.logdata.value_end - !remove_size + testtesttest
```

Julow:
> The difference is that if there is a `let` at the beginning of the line you immediately see it's a binding, whereas you have to go at the end of the line to look for `:=` in the other case.

I don't agree. When the next line is indented, I think it's easier to spot the operator on the end of the previous line. Also, I think it is more "conventional".
I'm in favor of keeping the `<-` and `:=` on the end of the line or adding an option for this.

gpetiot:
I added an option and kept the "at the end of line" default behavior.

src/Fmt_ast.ml
```diff
@@ -1081,6 +1081,16 @@ and fmt_body c ?ext ({ast= body} as xbody) =
         $ Cmts.fmt_after c pexp_loc )
   | _ -> (noop, fmt "@ " $ fmt_expression c ~eol:(fmt "@;<1000 0>") xbody)
 
+and fmt_assign_colon c =
```
Julow: Minor comment: These functions do not need to be in the recursive declarations and could be before any functions working with the AST.

test/passing/js_source.ml.ref
```diff
@@ -9675,8 +9675,8 @@ let foo : type a' b' c' t. a' -> b' -> c' -> t = fun a b c -> assert false
 
 let f x =
   x.contents
-  <- (print_string "coucou";
-       x.contents)
+    <- (print_string "coucou";
+         x.contents)
 ;;
```
Julow: Could you add tests for `:=`  ?
Also some tests for the `End_line` case.

PR #781, opened 2019-04-17
Break after anonymous function arrow after infix op

### LINKED ISSUE #769: Option to always break after binding operator RHS function arguments
(Please forgive any terminology faux pas; I'm still new to OCaml.)

As was discussed post-closing in #359, with `--no-break-infix-before-func` I can use bind operators and get closer to the readability of using PPX macros:

```ocaml
let f_test_noppx () =
  somefunction 10000000 20000000 >>= fun a ->
  somefunction 20000000 a >>= fun b -> Ok (b + 42)
;;
```

But as we can see, the line break after `->` isn't systematic.  For example with shorter expressions, there won't be any breaks:

```ocaml
let f_test_noppx () =
  somefunction 1 2 >>= fun a -> somefunction 2 a >>= fun b -> Ok (b + 42)
;;
```

It would improve readability to follow the community's generally adopted standard of always breaking after the arguments of an anonymous function on the right side of a binding operator:

```ocaml
let f_test_noppx () =
  somefunction 1 2 >>= fun a ->
  somefunction 2 a >>= fun b ->
  Ok (b + 42)
;;
```

This would more closely replicate the intent:

```ocaml
let f_test_ppx () =
  let%bind a = somefunction 1 2 in
  let%bind b = somefunction 2 a in
  Ok (b + 42)
;;
```

gpetiot:
Fix #769 
The first commit c4bfb5b should be alright, but I'm not sure about the second one f1532ca maybe I should revert it.

Maybe it should be the default behavior (maybe only if `break-cases` is not equal to `fit` ?).

@samoht for opinion, as you opened #359

### DIFF
--- a/test/passing/infix_bind.ml
+++ b/test/passing/infix_bind.ml
@@ -23,7 +23,11 @@ f x
 ;;
 f x >>= fun y ->
 g y >>= fun () ->
-f x >>= fun y -> g y >>= fun () -> f x >>= fun y -> g y >>= fun () -> y ()
+f x >>= fun y ->
+g y >>= fun () ->
+f x >>= fun y ->
+g y >>= fun () ->
+y ()
 
 ;;
 f x >>= function
@@ -31,19 +35,24 @@ f x >>= function
     g y >>= fun () ->
     f x >>= fun y ->
     g y >>= function
-    | x -> ( f x >>= fun y -> g y >>= function _ -> y () ) )
+    | x -> (
+        f x >>= fun y ->
+        g y >>= function _ -> y () ) )
 
 ;;
-eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x ->
+x
 
 ;;
 eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
-|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
+|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx ->
+x
 
 ;;
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
   eeeeeeeeeeee eeeeeeeeee
-|> fun x -> x
+|> fun x ->
+x
 
 ;;
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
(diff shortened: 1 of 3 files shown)


test/passing/infix_bind.ml
```diff
@@ -23,27 +23,37 @@ f x
...
-    | x -> ( f x >>= fun y -> g y >>= function _ -> y () ) )
+    | x -> (
+        f x >>= fun y ->
+        g y >>= function
+        | _ -> y () ) )
 
 ;;
-eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x ->
+x
```
Julow: Would it make sense to add an exception for these cases ?
It seems a bit too verbose when there is only one bind.

gpetiot: The problem is you can't just put this into an `hvbox` otherwise we have the current behavior where the last infix-bind doesn't break if it fits on 1 line.

Julow: I see. It would be too complicated to detect the "chain" of bind operators ?

gpetiot: It's not impossible to do, but I'm not sure it is worth implementing such a specific case. I will have a look on what it looks like on real code.

gpetiot:
> ```ocaml
> let _ = g x >>= (function a -> b | c -> d) >>= xxx
> 
> let _ =
>   (g x >>= function
>    | a -> b
>    | c -> d)
>   >>= fun () ->
>   xxx
> ```
> 
> Note that the parenthesis are not consistent.

We get:
```ocaml
let _ = g x >>= (function a -> b | c -> d) >>= xxx

let _ = g x >>= (function a -> b | c -> d) >>= fun () -> xxx
```
The parentheses look well placed to me (they are not preserved by only placed where it is necessary to disambiguate), but indeed it is not related to this PR.

gpetiot:
The diff of `tools/test_branch.sh break-after-anonymous-function-arrow-after-infix-op 'break-infix-before-func=false' 'break-infix-before-func=false'` looks good by the way.

gpetiot:
How about keeping only this commit c4bfb5b, as f1532ca can be controlled by `break-cases` anyway?

Julow:
I'm in favor of keeping it to be consistent with the first commit and to avoid surprise when using `function` in the middle of a chain of bind operator.

gpetiot:
@jberdine @samoht @vphantom any opinion on whether or not we should break after the `function` keyword (i.e. whether or not we should keep the 2nd commit) ?

vphantom:
While I understand it may seem inconsistent to break after `expr >>= fun … ->` but not after `expr >>= function`, the two are used differently.  The former is a monadic let binding (without PPX embellishment) while the latter is shorthand syntax for a possibly short pattern match.  To force breaking up that pattern match regardless of `break-cases` and line length would be excessive, in my opinion.

So I believe that `break-cases` is sufficient and the second commit isn't necessary.

Of course, there's also the possibility of adding a configuration option for it, if it is debatable.

gpetiot:
> While I understand it may seem inconsistent to break after `expr >>= fun … ->` but not after `expr >>= function`, the two are used differently. The former is a monadic let binding (without PPX embellishment) while the latter is shorthand syntax for a possibly short pattern match. To force breaking up that pattern match regardless of `break-cases` and line length would be excessive, in my opinion.
> 
> So I believe that `break-cases` is sufficient and the second commit isn't necessary.
> 
> Of course, there's also the possibility of adding a configuration option for it, if it is debatable.

I agree these two fit for different usages, and I'm really not sure we need to break for `function`s.
So for now I'm only merging the first commit, and we may need an option in the future to automatically break for `function`s regardless of `break-cases`, but I don't feel there is the need for that right now.

PR #782, opened 2019-04-17
Fix: type extension

### LINKED ISSUE #776: Possible edge case

gpetiot:
Fix #776


src/Fmt_ast.ml
```diff
@@ -3668,10 +3671,16 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
       let fmt_decl ~first ~last decl =
         let pre =
           if first then
-            if Poly.(rec_flag = Recursive) then "type " else "type nonrec "
+            let pre =
+              if Poly.(rec_flag = Recursive) then "type" else "type nonrec"
+            in
+            if Option.is_some ext then pre else pre ^ " "
```
Julow: Minor comment:
This line is a bit inelegant. The `?pre` argument of `fmt_type_declaration` never needs to be `None` and is always terminated by a space character. (Except for `with` constraints but I get no regression) What do you think of adding the space in `fmt_type_declaration`, in every cases?

gpetiot:
> On signatures I get:
> 
> ```diff
> -type%foo t = < .. >
> +[%%foo: type t = < .. >]
> ```
> 
> There is the exact same code for types in signatures.

I refactored the code and added a test. test_branch looks good.

test/passing/js_source.ml.ref
```diff
@@ -3548,8 +3548,10 @@ end
 module rec Typ : sig
   module type PAIR = sig
     type t
- and t1
- and t2
+
+    and t1
+
+    and t2
```
Julow: This is a fix.

PR #783, opened 2019-04-18
Less sensitivity to concrete syntax to decide whether the comment is on the same line or below

gpetiot:
#747

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -220,6 +220,22 @@ let module_type_is_simple x =
 let module_expr_is_simple x =
   match x.pmod_desc with Pmod_structure l -> List.is_empty l | _ -> true
 
+let signature_item_is_simple x =
+  match x.psig_desc with
+  | Psig_value _ -> false
+  | Psig_type (_, l) -> List.length l <= 1
+  | Psig_typext _ -> true
+  | Psig_exception _ -> true
+  | Psig_module {pmd_type= m} -> module_type_is_simple m
+  | Psig_recmodule _ -> false
+  | Psig_modtype {pmtd_type= m} -> Option.for_all m ~f:module_type_is_simple
+  | Psig_open _ -> true
+  | Psig_include _ -> true
+  | Psig_class _ -> false
+  | Psig_class_type _ -> false
+  | Psig_attribute _ -> true
+  | Psig_extension _ -> false
+
 module type Module_item = sig
   type t
 
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -217,8 +217,12 @@ let longident_is_simple c x =
 let module_type_is_simple x =
   match x.pmty_desc with Pmty_signature l -> List.is_empty l | _ -> true
 
-let module_expr_is_simple x =
-  match x.pmod_desc with Pmod_structure l -> List.is_empty l | _ -> true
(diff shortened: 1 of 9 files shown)


src/Ast.ml
```diff
@@ -217,8 +217,67 @@ let longident_is_simple c x =
...
+  | Psig_extension _ -> false
+
+let value_binding_is_simple c x =
+  match x.pvb_expr.pexp_desc with
+  | Pexp_ident l -> longident_is_simple c l.txt
+  | Pexp_constant _ -> true
+  | Pexp_let _ -> false
+  | Pexp_function [_] -> true
+  | Pexp_function _ -> false
+  | Pexp_fun _ -> false
```
Julow: If you are allowing `function` with one pattern, why not allowing functions with simple body here ? (eg. calling recursively on the body)

src/Ast.ml
```diff
@@ -217,8 +217,67 @@ let longident_is_simple c x =
...
   match x.pmty_desc with Pmty_signature _ -> false | _ -> true
 
-let module_expr_is_simple x =
-  match x.pmod_desc with Pmod_structure _ -> false | _ -> true
+let rec module_expr_is_simple x =
+  match x.pmod_desc with
+  | Pmod_structure _ -> false
+  | Pmod_constraint (e, t) ->
+      module_expr_is_simple e && module_type_is_simple t
+  | _ -> true
```
Julow: You should check `Pmod_functor` and `Pmod_apply`, they can contain sub-expressions (and types).

src/Ast.ml
```diff
@@ -217,8 +217,67 @@ let longident_is_simple c x =
...
+  | Pexp_letexception _ -> false
+  | _ -> true
+
+let structure_item_is_simple c x =
+  match x.pstr_desc with
+  | Pstr_eval _ -> false
+  | Pstr_value (_, [x]) -> value_binding_is_simple c x
+  | Pstr_value (_, _) -> false
+  | Pstr_primitive _ -> true
+  | Pstr_type (_, l) -> List.length l <= 1
```
Julow: Minor style comment: You could pattern match on the list instead, like for `Pstr_value`

test/passing/comment_last.ml
```diff
@@ -1,5 +1,3 @@
 let x = 2
 
-let y = 3
-
-(*comment*)
+let y = 3 (*comment*)
```
Julow: This is a regression

test/passing/js_source.ml.ref
```diff
@@ -9132,7 +9128,9 @@ end
 module OK = struct
   open M1
 
-  let f1 (r : t) = r.x (* ok *)
+  let f1 (r : t) = r.x
+
+  (* ok *)
```
Julow: Regression in the opposite direction

test/passing/comment_last.ml
gpetiot: Only if we want to preserve the empty line (that means relying on the input), otherwise it's a simple structure_item and the comment is attached to it and it fits.

Julow: I think the comment should be treated as a (not simple) structure item otherwise this is changing the meaning of the comment.

gpetiot:
We need to "fix" comments attachment first.

PR #786, opened 2019-04-22
Fix: functor application and break-struct wrap incorrectly

### LINKED ISSUE #487: Functor application and break-struct wrap incorrectly
First of all let me say how excited I am about `ocamlformat`, it's a great piece of work and it has evolved really well since the last time I tried it about 10 months ago. Thanks for putting it together!

I started trying it on our codebase at Cryptosense so I'll probably open a bunch of issues as I found some bugs/quirks and also have a few feature requests.

Anyway, I digress.

Consider the following piece of code:

```ocaml
module A = Make(struct type t = int end)(String)

module B = Make(Some_super_long_module_name_that_will_exceed_margin)(Some_other_shorter_module_name)
```

that gets formatted as:

```ocaml
module A =
  Make (struct
      type t = int
    end)
    (String)

module B =
  Make
    (Some_super_long_module_name_that_will_exceed_margin)
    (Some_other_shorter_module_name)
```

while I would expect the first module expr to be formatted in a similar way as the second one, ie something like:

```ocaml
module A =
  Make
    (struct
      type t = int
    end)
    (String)
```

My guess is that it doesn't cope well with `break-struct=force` (which is the default).

I'm of course happy to help fixing this and contribute to the project!

gpetiot:
Fix #487 
@NathanReb for feedback

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -215,10 +215,10 @@ let longident_is_simple c x =
   longident_fit_margin c (length x)
 
 let module_type_is_simple x =
-  match x.pmty_desc with Pmty_signature _ -> false | _ -> true
+  match x.pmty_desc with Pmty_signature l -> List.is_empty l | _ -> true
 
 let module_expr_is_simple x =
-  match x.pmod_desc with Pmod_structure _ -> false | _ -> true
+  match x.pmod_desc with Pmod_structure l -> List.is_empty l | _ -> true
 
 module type Module_item = sig
   type t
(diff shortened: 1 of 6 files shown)


NathanReb:
Looks great, thanks!

src/Ast.ml
```diff
@@ -215,10 +215,10 @@ let longident_is_simple c x =
   longident_fit_margin c (length x)
 
 let module_type_is_simple x =
-  match x.pmty_desc with Pmty_signature _ -> false | _ -> true
+  match x.pmty_desc with Pmty_signature l -> List.is_empty l | _ -> true
 
 let module_expr_is_simple x =
-  match x.pmod_desc with Pmod_structure _ -> false | _ -> true
+  match x.pmod_desc with Pmod_structure l -> List.is_empty l | _ -> true
```
Julow: Is this required ?
If it is, could you add an example to the test suite?

gpetiot: Yes it is, if we don't have this, we have additional regressions in the testsuite actually.

Julow: I tried again and got 0 regressions. What did I do wrong?
I'm still on `6642e3df`

PR #797, opened 2019-04-25
Improve: break with type constraints

### LINKED ISSUE #790: Alignment of with constraints
With the `janestreet` profile.

Should breaks before the `with` keyword, on module packs in patterns and types:

```ocaml
let x (module M : S with type a = a
                     and type b = b) = ()
```

```ocaml
val x : (module S with type a = a
                   and type b = b) -> unit
```

gpetiot:
Fix #790

The new output is:

```ocaml
let x (module M : S with type a = a and type b = b) = ()

module type X = sig
  val x : (module S with type a = a and type b = b) -> unit
end
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -606,10 +606,10 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
         (list t1N (comma_sep c) (sub_typ ~ctx >> fmt_core_type c))
       $ fmt "@ " $ fmt_longident_loc c lid
   | Ptyp_extension ext -> hvbox 2 (fmt_extension c ctx "%" ext)
-  | Ptyp_package pty ->
+  | Ptyp_package (id, cnstrs) ->
       hvbox 2
-        ( hovbox 0 (fmt "module@ " $ fmt_longident_loc c (fst pty))
-        $ fmt_package_type c ctx pty )
+        ( hovbox 0 (fmt "module@ " $ fmt_longident_loc c id)
+        $ fmt_package_type c ctx cnstrs )
   | Ptyp_poly ([], _) ->
       impossible "produced by the parser, handled elsewhere"
   | Ptyp_poly (a1N, t) ->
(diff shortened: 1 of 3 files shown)


test/passing/source.ml.ref
```diff
@@ -3452,7 +3451,7 @@ module SSMap = struct
 end
 
 let ssmap =
-  ( module SSMap : MapT
+  (module SSMap : MapT
```
gpetiot: This is unavoidable as the opening and closing parentheses are not in the same box (but most of the case the output is consistent with `indicate-multiline-delimiters`).

src/Fmt_ast.ml
```diff
@@ -979,8 +978,10 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
       let ctx = Typ typ in
       hovbox 0
       @@ wrap_if parens "(" ")"
-           ( fmt "module " $ fmt_str_loc c name $ fmt "@ : "
-           $ fmt_package_type c ctx pty ppat_loc )
+           ( hovbox 0
+               ( fmt "module " $ fmt_str_loc c name $ fmt "@ : "
+               $ fmt_longident_loc c (fst pty) )
```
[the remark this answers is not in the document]
gpetiot: > The code looks good but I'm not sure it's a good idea to break arbitrarily at 2 constraints. I fear that we will have to restore the old behavior later.

Fixed (and without using `Location.is_single_line`).

gpetiot:
> The code looks good but I'm not sure it's a good idea to break arbitrarily at 2 constraints. I fear that we will have to restore the old behavior later.

Fixed (and without using `Location.is_single_line`).

src/Fmt_ast.ml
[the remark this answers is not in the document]

test/passing/source.ml.ref
Julow: This is still a bug. There is a space before the closing parenthesis.

gpetiot: I agree, but that's the kind of things that can be improved later I think and it shouldn't delay this PR that improves a lot of things.
We need to rethink the `indicate-multiline-delimiters` because in most of the cases it is not feasible without relying on the concrete input (which is bad).

Julow: I agree.

Issue #792, opened 2019-04-23
Unecessary break before signature in functor types


Julow:
In functor types, with any profile:

```ocaml
module type S' = functor
  (A : A)
  (B :
   sig
     type t
   end)
  -> S with type t = B.t
```

For comparison, the formatting of functor declarations:

```ocaml
module S'
    (A : A) (B : sig
      type t
    end) =
struct end
```

jberdine:
Is the desired output:
```ocaml
module type S' = functor
  (A : A)
  (B : sig
     type t
   end)
  -> S with type t = B.t
```

Julow:
Yes.

gpetiot:
Should
```ocaml
module S'
    (A : A) (B : sig
      type t
    end) =
struct end
```
be modified as well?

gpetiot:
This output looks fine to me, I'm not sure  we should vertically break between each argument (this would break a lot of cases that look fine right now), can we close this issue?

PR #802, opened 2019-04-25
Fix colon break module type functor

gpetiot:
First part of #792

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2939,11 +2939,13 @@ and fmt_extension_constructor c sep ctx ec =
        $ fmt_docstring_padded c doc )
 
 and fmt_functor_arg c (name, mt) =
-  let fmt_mty mt =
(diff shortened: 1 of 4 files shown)


src/Fmt_ast.ml
```diff
@@ -2938,6 +2938,13 @@ and fmt_extension_constructor c sep ctx ec =
        $ fmt_attributes c ~pre:(fmt "@ ") ~key:"@" atrs ~suf
        $ fmt_docstring_padded c doc )
 
+and fmt_functor_arg c (name, mt) =
+  let fmt_mty mt =
+    let mty = fmt_module_type c mt in
+    compose_module mty ~f:(fun k -> fmt " : " $ k)
```
Julow: Maybe `"@ : "` in case of long name ?

gpetiot: Fixed, and no diff with test_branch by the way.

PR #804, opened 2019-04-26
Fix: comment in empty structure

### LINKED ISSUE #798: Comment dropped.
```
$ echo "include A (struct (* a *) end)" | _build/release/src/ocamlformat.exe --name a.ml -
ocamlformat.exe: Cannot process "a.ml".
  BUG: comments dropped.
File "a.ml", line 1, characters 18-25:
Error: Comment (* a *) dropped.
EXIT STATUS 1
```

gpetiot:
Fix #798 
The test_branch diff is empty.

### DIFF
--- a/test/passing/module.ml
+++ b/test/passing/module.ml
@@ -43,3 +43,11 @@ with type t := t
  and type s := s = struct
   let () = ()
 end
+
+include struct
+  (* a *)
+end
+
+include A (struct
+  (* a *)
+end)
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3324,7 +3324,9 @@ and fmt_with_constraint c ctx = function
       $ fmt_longident_loc c m2
 
 and maybe_generative c ~ctx = function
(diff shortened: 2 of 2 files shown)


test/passing/module.ml
```diff
@@ -43,3 +43,11 @@ with type t := t
...
   let () = ()
 end
+
+include struct
+  (* a *)
+end
+
+include A (struct
+  (* a *)
+end)
```
Julow: Shouldn't this be `include A ( (* a *) )` ?

gpetiot: I chose to keep the original behavior that is to have the `struct`/`end` keywords if the structure is non-empty, and the structure is not empty since there is at least one comment.

Julow: I agree. Implementing this is extremely tricky anyway, as the comment inside the parenthesis is placed juste after the lid.

gpetiot: And by the way if we omit the keywords, the comment is placed before `( )` and not inside.

PR #805, opened 2019-04-26
Fix: sugarized extensions

### LINKED ISSUE #796: Invalid syntax generated when trying to always use extension sugar on opens
I get this error:

```
```

when trying to format anything like this:

```ocaml
open%any_extension M
```

with configuration `extension-sugar` set to `always`

gpetiot:
Fix #796 
The modifications on the sequences and let are due to the `extension-sugar=always` mode that crashed for the test `extensions.ml`.


chrismamo1:
Thanks!

Issue #629, opened 2019-02-12
Type annotation on first-class modules


Julow:
This code:

```
let x : (module S) = (module struct end)
```

is rewritten as:
```
let x = (module struct end : S)
```

Extra parenthesis solve the problem:
```
let (x : (module S)) = (module struct end)
```

gpetiot:
Looks related to https://caml.inria.fr/mantis/view.php?id=7758

Julow:
@gpetiot Do you think we should preserve `let x = (module struct end : S)` or is it OK to format it as `let x : (module S) = (module struct end)` every time ?

gpetiot:
I think that just like for #791 this is a case where it's okay to rely on the input to preserve the initial formatting.

gpetiot:
Can be done after #812 to re-use the new functions.

PR #812, opened 2019-04-30
Improve: preserve record type annot

### LINKED ISSUE #791: Coercion in records
In record expressions, type annotations are moved from expressions to field names:

```patch
-let _ = x { a = (a' : string); b = (b' : string) }
+let _ = x { a : string = a'; b : string = b' }
```

In patterns:
```patch
 let x = function
-  | { a = (_ : string); _ } -> ()
+  | { a : string = _; _ } -> ()
```

gpetiot:
Fix #791 
Can replace this patch (https://github.com/janestreet/ocamlformat/commit/9a1a519f4b64018a3b2d6807bbcbc1b9f537541c) in the JS branch.

If everyone agrees on the way it's done we can also apply it for #629 

A big part of the changes are actually due to an improved consistency for all cases (a single function instead of hard-coding each case).

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -854,7 +854,7 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
             $ fmt_if parens ")"
           in
           let fmt_pat ?(parens = false) ~ctx p =
-            fmt "=@ " $ fmt_if parens "("
+            fmt "=@;<1 2>" $ fmt_if parens "("
             $ cbox 0 (fmt_pattern c (sub_pat ~ctx p))
           in
           maybe_cmts field_loc @@ maybe_cmts cnstr_loc
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1928,16 +1928,27 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         let leading_cmt = Cmts.fmt_before c lid1.loc in
         let maybe_cmts = Option.value_map ~default:Fn.id ~f:(Cmts.fmt c) in
         let fmt_field ?field_loc ?cnstr_loc ?ident_loc ?typ ?expr c =
+          let fmt_type ?(parens = false) t =
+            fmt_if Poly.(c.conf.field_space = `Loose) " "
+            $ str ": "
+            $ fmt_core_type c (sub_typ ~ctx t)
+            $ fmt_if parens ")"
+          in
+          let fmt_expr ?(parens = false) e =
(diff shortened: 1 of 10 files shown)


src/Fmt_ast.ml
```diff
@@ -1926,12 +1937,29 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+          @@ cbox 2
+               ( maybe_cmts ident_loc @@ fmt_longident_loc c lid1
+               $
+               match (typ, expr) with
+               | Some t, Some e -> (
+                 match Source.typed_expression t e with
+                 | `Type_first -> fmt_type t $ fmt_expr e
+                 | `Expr_first ->
+                     fmt_expr ~parens:true e $ fmt_type ~parens:true t )
+               | _ -> opt typ fmt_type $ opt expr fmt_expr )
```
Julow: You are not implementing `field-space=loose` in the case there is no type. ([code](https://github.com/ocaml-ppx/ocamlformat/compare/master...Julow:812_loose))

src/Fmt_ast.ml
```diff
@@ -845,41 +845,52 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
...
+               ( maybe_cmts ident_loc @@ fmt_longident_loc c lid1
+               $
+               match (typ, pat) with
+               | Some t, Some p -> (
+                 match Source.typed_pattern t p with
+                 | `Type_first -> fmt_type t $ fmt_pat ~ctx p
+                 | `Pat_first ->
+                     fmt_pat ~ctx ~parens:true p $ fmt_type ~parens:true t )
+               | _ -> opt typ fmt_type $ opt pat (fmt_pat ~ctx) )
+        in
```
Julow: This code is almost identical to the one in `Pexp_record`. I think factorization is required.

gpetiot: I tried to generalize the function but since it operates on different types it is too much trouble, and I'm not sure it would be easier to read.
Or maybe there is a simplification I'm not seeing (feel free to push).

src/Fmt_ast.ml
gpetiot: Done, but we need to apply if for `=` and `:` if I correctly understand the doc.

Julow: I don't think it should be applied to both at the same time, only the first one syntactically. Otherwise there is regressions.

src/Fmt_ast.ml
Julow: I've made a PR to your PR: https://github.com/gpetiot/ocamlformat/pull/1

gpetiot:
I merged your PR, everything looks good (no diff with test_branch). We can merge as soon as it gets approved.

PR #814, opened 2019-05-02
Improve: preserve blank lines in sequences

### LINKED ISSUE #554: Preserve blank lines in function bodies
Given the following code:

```ocaml
let foo x y =
  do_some_setup x ;
  do_some_setup y ;

  (* This is the important bit *)
  important_function x ;
  another_important_function x y ;

  cleanup x y
```

ocamlformat produces:

```ocaml
let foo x y =
  do_some_setup x ;
  do_some_setup y ;
  (* This is the important bit *)
  important_function x ;
  another_important_function x y ;
  cleanup x y
```

Ideally ocamlformat would preserve the blank lines here. Code formatters for other languages collapse consecutive blank lines, but preserve one line ([JS example with prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEqEEAFAJT7AA6U+hx5A3DXS-gEYCGATk217zLNa+fpwBefEWxkj2OSUNkBfEABoQEAA65oAZ2Sge3CAHcACjwQGUnAG4QcAE3Uh23TmADWcGAGUtTxwoAHNkGG50OA0ACxgAWwAbAHUYnHg9QLA4P2t0nDt0gE9kcD0DDWC9OG4Ycw8Q+M5kVE5E6o0AKz0ADwAhD29fP054uAAZYLgWto6Qbp6-YJDEuABFdAh4GfbokEDuau5SrnYixOhXLW5gmGTnGBjkAA4ABg1riGrkjy1S67gRzs0w03DgAEd0DgwfVOI1mkhWrsNNV4jhwpE9nplqsNltpojZnsYJx2PcnI9kAAmDQRTg4RLLADCEHiTVKUGgIJA6GqABVSTYkdVlMogA)).

Possibly related: #521.

gpetiot:
Fix #554 
I changed the `break-sequences` option, but maybe it would be better to have two options instead?


test/passing/sequence_preserve.ml.ref
```diff
@@ -0,0 +1,13 @@
+let foo x y =
+  do_some_setup y ;
+  important_function x
+
+let foo x y =
+  do_some_setup x ;
+  do_some_setup y ;
+
```
Julow: This line contains space characters. Same for the next one.

src/Fmt_ast.ml
```diff
@@ -1978,25 +1978,41 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
       in
       hvbox 0
         (hvbox_if parens 2
            ( wrap_fits_breaks_if ~space:false c.conf parens "(" ")"
                (list grps
                   ( match c.conf.sequence_style with
-                  | `Separator -> " ;@;<1000 0>"
-                  | `Terminator -> ";@;<1000 0>" )
+                  | `Separator -> " ;@;<1000 0>@\n"
+                  | `Terminator -> ";@;<1000 0>@\n" )
```
Julow: This adds trailing space

```diff
-                  | `Separator -> " ;@;<1000 0>@\n"
-                  | `Terminator -> ";@;<1000 0>@\n" )
+                  | `Separator -> " ;\n@;<1000 0>"
+                  | `Terminator -> ";\n@;<1000 0>" )
```

src/Fmt_ast.ml
```diff
@@ -1978,25 +1978,41 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
-          | `Separator -> " ;@ "
-          | `Terminator when c.conf.break_sequences -> ";@;<1000 0>"
-          | `Terminator -> ";@ " )
-          (fmt_expression c)
+        list_pn grp (fun ?prev x ?next:_ ->
+            ( match prev with
+            | Some prev ->
+                let is_simple x = is_simple c.conf width x in
+                let break = not (is_simple prev && is_simple x) in
+                fmt
```
Julow: Minor comment: You could remove an indentation level by calling `fmt` every time.

test/passing/sequence_preserve.ml.ref
gpetiot: fixed

src/Fmt_ast.ml
gpetiot: fixed

src/Fmt_ast.ml
gpetiot: fixed

gpetiot:
cc @Wilfred, @ELLIOTTCABLE for feedback

ELLIOTTCABLE:
I'm sure I'd approve (anything that improves readability and negative space, I'm happy with!); but if you want more thorough commentary, would you mind pasting a few examples (input/before-change/after-change) into the Issue? `[=`

Wilfred:
@gpetiot the logic looks good to me! My sample code from #554:

```
let foo x y =
  do_some_setup x ;
  do_some_setup y ;

  (* This is the important bit *)
  important_function x ;
  another_important_function x y ;

  cleanup x y
```

Is unchanged with `ocamlformat --break-sequences=preserve --enable-outside-detected-project` on this PR, which is exactly what I'd hoped for :). I also like that it collapses blank lines:

Before:

```
let foo x y =
  (* foo *)
  do_some_setup x ;



  cleanup x y
```

After:

```
let foo x y =
  (* foo *)
  do_some_setup x ;

  cleanup x y
```

My only concern is that it doesn't make sense to use `--break-sequences` here. Whether or not the user wants expressions grouped on a line or not (`fit` versus `all`) seems separate to how they want blank lines to be handled.

gpetiot:
> My only concern is that it doesn't make sense to use `--break-sequences` here. Whether or not the user wants expressions grouped on a line or not (`fit` versus `all`) seems separate to how they want blank lines to be handled.

I tend to agree with that, I will change it so it is a separate option.

gpetiot:
I fixed this and added a test.

src/Fmt_ast.ml
```diff
@@ -1970,26 +1970,48 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
-          ( match c.conf.sequence_style with
-          | `Separator when c.conf.break_sequences -> " ;@;<1000 0>"
-          | `Separator -> " ;@ "
-          | `Terminator when c.conf.break_sequences -> ";@;<1000 0>"
-          | `Terminator -> ";@ " )
-          (fmt_expression c)
+        list_pn grp (fun ?prev x ?next:_ ->
+            ( match prev with
+            | Some prev -> (
+              match c.conf.sequence_style with
```
Julow: Minor comment: The matched value only change the leading space so there is code duplication. The break logic should be factorized. Eg.
```ocaml
let break =
  if blank_line prev x then fmt "\n@;<1000 0>"
  else if c.conf.break_sequences then fmt "@;<1000 0>"
  else fmt "@ "
in
match c.conf.sequence_style with
| `Separator -> str " ;" $ break
| `Terminator -> str ";" $ break
```

Also, there is a very similar function a few lines below, could they share the same code ? (eg. with some sort of `force_break_sequence` bool argument)

avsm:
This looks like a good candidate for `conventional`. Collapsing multiple blank lines is good practise in most code.

ELLIOTTCABLE:
@avsm can you elaborate? "ocaml avsm conventional" doesn't yield any interesting search-engine results :P

gpetiot:
This is the name of a preset profile of ocamlformat, you can use it with `--profile=conventional`.

PR #815, opened 2019-05-03
Improve: preserve the position of type annotation in bindings

### LINKED ISSUE #629: Type annotation on first-class modules

gpetiot:
Fix #629 
No regression with test_branch, with or without the janestreet profile.


gpetiot:
> Looks good !
> 
> Should this be implemented for any types ? The issue is only about pack types but the same problem occur with any other types:
> 
> ```diff
>  let a : int = 0
> 
> -let b = (0 : int)
> +let b : int = 0
> ```

Done, this causes a big diff on the existing testsuite but we are now closer to the input.

PR #818, opened 2019-05-06
Improve: remove break before ending paren for anonymous functions

gpetiot:
Follow up on #801, taking care of the remaining cases for `fun` and `function`.
The remaining cases of `indicate-multiline-delimiters` (assert, match cases, constraints, etc) do not seem to be problematic.


test/passing/js_source.ml.ref
```diff
@@ -3037,7 +3037,9 @@ let rec weaken_alist : type m n. (m, n) alist -> (m succ, n succ) alist = functi
 let rec sub' : type m. m ealist -> m fin -> m term = function
   | EAlist Anil -> var
   | EAlist (Asnoc (s, t, x)) ->
-    comp_subst (sub' (EAlist (weaken_alist s))) (fun t' -> weaken_term (subst_var x t t'))
```
gpetiot: Somehow the margin was not respected here.

PR #819, opened 2019-05-06
Improve: space before type constraint in record

gpetiot:


src/Fmt_ast.ml
```diff
@@ -553,9 +553,9 @@ and fmt_record_field c ?typ ?rhs ?(type_first = false) lid1 =
   let fmt_type_rhs =
     match (typ, rhs) with
     | Some t, Some r ->
-        if type_first then field_space $ fmt_type t $ fmt "@ " $ fmt_rhs r
+        if type_first then fmt_type t $ fmt "@ " $ fmt_rhs r
         else field_space $ fmt_rhs ~parens:true r $ fmt_type ~parens:true t
-    | Some t, None -> field_space $ fmt_type t
```
Julow: I think you should keep this, and the one above. Only adding a literal space between the `rhs` and the `type` in the `not type_first` case

gpetiot: fixed

PR #820, opened 2019-05-07
Improve: break-cases=fit-or-vertical

gpetiot:
Request from Jane Street: (replaces the patch https://github.com/janestreet/ocamlformat/commit/9adc31fe7d54c0c0ea4140cd2aa2694d37fc49ff)
- or-patterns fit on the same line if possible, otherwise break all of them (don't wrap)
- if it breaks, don't break after the arrow


gpetiot:
> Would it make sense to change `toplevel` instead ?

I added some tests to show the differences (if the or-patterns do not fit, we vertically break with fit-or-vertical, but we wrap with toplevel and others).

PR #821, opened 2019-05-07
Improve: tidying Fmt_ast.ml

gpetiot:
Mostly factorizations and simplifications.


gpetiot:
More commits to come (each commit can be reviewed independently).

Issue #409, opened 2018-09-09
Performance issue with very large list literal  


hhugo:
see https://github.com/janestreet/core/blob/master/test/bin/ofday_unit_tests_v1.ml

Turning the list into an array improves performances a lot.

gpetiot:
Do you have any idea about where to start?

hhugo:
I would actually lower the priority of that one. I've seen only two problematic files in the entire opam repo

jberdine:
Maybe it's clear, but note that it is reasonable for the perf to change a lot with an array vs a list since internally lists are deeply nested applications rather than a flat aggregate.

As for where to start, I would first see if anything stands out in the data that can be collected by linux `perf`. Even just running `sudo perf top` while processing the input file triggering the slow formatting could be useful.

gpetiot:
In #823 I made the sugaring functions tail-recursive but it doesn't improve the performance.
So the performance issue is actually not in Sugar.list_exp but in Ast.is_sugared_list.
I tried several approaches to optimize the latter:
- using exceptions didn't improve the performance
- assuming a list is always sugared or unsugared breaks the ast

hhugo:
One more optimization in https://github.com/ocaml-ppx/ocamlformat/pull/833

PR #823, opened 2019-05-10
Improve: make Sugar.list_exp and Sugar.list_pat tail-recursive

gpetiot:
The idea was to take care of #409, I made the sugaring functions tail-recursive (and both functions now have the same design), it doesn't improve the performance but it is easier to read I think.
So the performance issue is actually not in `Sugar.list_exp` but in `Ast.is_sugared_list`.
I tried several approaches to optimize the latter:
- using exceptions didn't improve the performance
- assuming a list is always sugared or unsugared breaks the ast

Help regarding improving `Ast.is_sugared_list` is appreciated but it could be done later in another PR anyway.

### DIFF
--- a/src/Sugar.ml
+++ b/src/Sugar.ml
@@ -198,27 +198,23 @@ let rec list_pat cmts pat =
   | _ -> None
 
 let list_exp cmts exp =
-  let rec list_exp_ exp =
+  let rec list_exp_ exp acc =
     let ctx = Exp exp in
     let {pexp_desc; pexp_loc= src} = exp in
     match pexp_desc with
     | Pexp_construct ({txt= Lident "[]"; loc}, None) ->
         Cmts.relocate cmts ~src ~before:loc ~after:loc ;
-        Some ([], loc)
+        Some (List.rev acc, loc)
     | Pexp_construct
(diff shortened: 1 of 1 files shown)


src/Sugar.ml
```diff
@@ -180,45 +180,43 @@ let infix cmts prec xexp =
...
-          Some (([src; loc; pexp_loc], sub_exp ~ctx hd) :: xtl, nil_loc)
-      | _ -> None )
+            { pexp_desc= Pexp_tuple [hd; ({pexp_attributes= []} as tl)]
+            ; pexp_loc
+            ; pexp_attributes= [] } ) ->
+        list_exp_ tl (([src; loc; pexp_loc], sub_exp ~ctx hd) :: acc)
     | _ -> None
   in
-  let r = list_exp_ exp in
-  assert (Bool.equal (Option.is_some r) (is_sugared_list exp)) ;
```
Julow: Why do you remove the assert?

gpetiot: Because if this assert fails, the error would also be caught when checking that the ast did not changed, and this checking would provide a better error message that the assert failure.

Julow: Great.

PR #826, opened 2019-05-12
Fix: performance issue with long list literals

### LINKED ISSUE #409: Performance issue with very large list literal

hhugo:
Fix #409
Building the tree of locations is quadratic when the depth of the ast is very large.

Some relatively simple heuristic reduces the issue a lot.
I'm not 100% confident the heuristic is always correct. I'd like reviewers to think about it.

```
$ wget https://raw.githubusercontent.com/janestreet/core/master/test/bin/ofday_unit_tests_v1.ml
$ time _build/release/src/ocamlformat.exe ofday_unit_tests_v1.ml > /dev/null

real	0m3.470s
user	0m3.434s
sys	0m0.036s
```
Before this change, it was.
```
real	6m37.299s
user	6m36.966s
sys	0m0.120s
```

The heuristic implemented here assumes that two consecutive location (in term of insertion order) will be close to each other in the tree of locations. Instead of looking for where to put the location starting at the roots, one can try to start from ancestors found at the previous iteration.

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -69,8 +69,7 @@ end = struct
 
   let of_list elts =
     let elts_decreasing_width =
-      List.sort ~compare:Itv.compare_width_decreasing elts
-      |> List.remove_consecutive_duplicates ~equal:Poly.equal
+      List.dedup_and_sort ~compare:Itv.compare_width_decreasing elts
     in
     let tree = create () in
     let rec find_in_previous elt = function
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -51,16 +51,16 @@ end = struct
 
   (* Descend tree from roots, find deepest node that contains elt. *)
 
-  let rec parent tbl roots ?ancestor elt =
-    Option.first_some
+  let rec parents tbl roots ~ancestors elt =
+    Option.value ~default:ancestors
       (List.find_map roots ~f:(fun root ->
            if Itv.contains root elt then
              Hashtbl.find_and_call tbl root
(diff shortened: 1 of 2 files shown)


hhugo:
@gpetiot, @jberdine, feel free to take ownership on this PR.

hhugo:
The heuristic implemented here assumes that two consecutive (in term of insertion order) comments will be close to each other in the tree of comments. Instead of looking for the position of a comment starting at the roots, one can try to start from `ancestors` found at the previous iteration.

hhugo:
With the PR formatting js_source.ml goes from 2.7s to 1.5s

hhugo:
The last commit give the following results

`ofday_unit_tests_v1.ml`
```
real	0m2.918s
user	0m2.717s
sys	0m0.085s
```

`ofday_unit_tests_v1.ml` already well formatted
```
real	0m1.533s
user	0m1.310s
sys	0m0.088s
```

hhugo:
`js_source.ml`
```
real	0m0.878s
user	0m0.849s
sys	0m0.029s
```

`js_source.ml` already well formatted

```
real	0m0.296s
user	0m0.279s
sys	0m0.017s
```

hhugo:
Reformatting all sources in ocamlformat
Before:
```
real	0m0.606s
user	0m0.585s
sys	0m0.021s
```

After
```
real	0m0.375s
user	0m0.367s
sys	0m0.008s

PR #827, opened 2019-05-13
Improve: option nested-match=align

### LINKED ISSUE #492: Don't wrap nested match in parenthesis
We have some code like this, using nested `match`:

```ocaml
let () =
  match v with
  | None -> None
  | Some x ->
  match x with
  | None -> None
  | Some x ->
  match x with
  | None -> None
  | Some x -> x
```

When it is reformated, the second and third `match` are parenthesized, which is not very nice and creates some noise. Could it be avoided?

```ocaml
let () =
  match v with
  | None -> None
  | Some x ->
    ( match x with
    | None -> None
    | Some x -> (match x with None -> None | Some x -> x) )
```

ocamlformat config:
```
m=80
doc-comments=before
let-and=sparse
break-cases=fit
break-struct=natural
sequence-style=terminator
cases-exp-indent=2
ocp-indent-compat=true
break-infix=fit-or-vertical
break-separators=after
type-decl=sparse
```

gpetiot:
Fix #492 
No bug with test_branch (the changes are similar to those of the tests).

The change is made for nested `match` and `try` but maybe we should only make for nested `match`.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -45,6 +45,7 @@ type t =
   ; margin: int
   ; max_iters: int
   ; module_item_spacing: [`Compact | `Preserve | `Sparse]
+  ; nested_match: [`Wrap | `Align]
   ; ocp_indent_compat: bool
   ; parens_ite: bool
   ; parens_tuple: [`Always | `Multi_line_only]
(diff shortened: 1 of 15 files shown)


src/Conf.ml
```diff
@@ -990,6 +991,23 @@ module Formatting = struct
...
+    let doc = "Style of let_and." in
+    let names = ["nested-match"] in
+    let all =
+      [ ( "wrap"
+        , `Wrap
+        , "$(b,wrap) classically wraps a match nested in another match \
+           case with parentheses and adds indentation." )
+      ; ( "align"
+        , `Align
+        , "$(b,align) does not indent and wrap a match if it is nested in \
```
Julow: You are using the word `wrap` in the `align` case.
Why is the first case named `wrap` ? Wouldn't it be easier to understand if it was `default` ?

src/Conf.ml
```diff
@@ -990,6 +991,23 @@ module Formatting = struct
       (fun conf x -> {conf with module_item_spacing= x})
       (fun conf -> conf.module_item_spacing)
 
+  let nested_match =
+    let doc = "Style of let_and." in
```
Julow: Copy paste fail ?

src/Conf.ml
gpetiot: The "not" applies to "indent" and to "wrap", but agreed, it can be rephrased.

gpetiot: The "not" applies to "indent" and to "wrap", but agreed, it can be rephrased.

src/Conf.ml
gpetiot: fixed

gpetiot:
@Khady any opinion before we merge? (especially on whether the `try` should be handled like `match` for this feature)

Khady:
We don't really chain `try` blocks. So I don't have a strong opinion about this. The `match` behavior looks like what we need. Thanks!

Issue #630, opened 2019-02-15
Switch to OCaml 4.08 Ast ..


hhugo:
.. once https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/50 is addressed.
 
Here is a non-exhaustive list of interesting changes for ocamlformat: 
1. GPR#1970: fix order of floatting documentation comments in classes  (makes the sorting unnecessary in https://github.com/ocaml-ppx/ocamlformat/blob/85cfc0f9a360e5cfe4268b1aacc364de7d65474a/src/Fmt_ast.ml#L2248 and https://github.com/ocaml-ppx/ocamlformat/blob/85cfc0f9a360e5cfe4268b1aacc364de7d65474a/src/Fmt_ast.ml#L2292). the sort could be implemented in migrate_parsetree when upgrading from 4.07 to 4.08.
2. GPR#1958: allow [module M(_:S) = struct end] syntax (one can lift the restriction on the underscore https://github.com/ocaml-ppx/ocamlformat/blob/85cfc0f9a360e5cfe4268b1aacc364de7d65474a/src/Sugar.ml#L280)
3. GPR#1960: The parser keeps previous location when relocating ast node. (should be able to help with https://github.com/ocaml-ppx/ocamlformat/blob/85cfc0f9a360e5cfe4268b1aacc364de7d65474a/src/Source.ml#L113)
4. GPR#1705: Allow @@attributes on exceptions. (we should be able to address https://github.com/ocaml-ppx/ocamlformat/blob/85cfc0f9a360e5cfe4268b1aacc364de7d65474a/src/Fmt_ast.ml#L3027)
5. GPR#1954: Add locations to toplevel directives
6. GRP#1953: Add locations to attributes in the parsetree.

hhugo:
This makes me think that we might need a flag to specify the targeted version of the OCaml compiler.
For instance, it would be incorrect to use the following syntax in 4.07. 
`module M(_:S) = struct end`

jberdine:
[...]

PR #831, opened 2019-05-14
Improve: switch to ast 4.08

hhugo:
fix https://github.com/ocaml-ppx/ocamlformat/issues/630

- [x] plug the 4.08 ast in the code
- [x] remove hack: sort object fields
- [x] remove hack: dispatch (`[@` vs `[@@`) attributes on exception.    
- [x] process locations to toplevel directives and their arguments (pdir_loc, pdira_loc)
- [x] process locations  in attributes (attr_loc)
- [x] process locations in row_field (prf_loc)
- [x] process locations in object_field (pof_loc)
- [x] process locations in type_extension (ptyext_loc)
- [x] handle letop
- [x] handle extended open 
- [x] handle typesubst
- [x] handle modsubst
- [x] Experiment with `{pexp,ptyp,ppat}_loc_stack` (in another PR #867 )
- [x] Allow `module M(_:S) = struct end` syntax (will break formatting with 4.07) 
- [x] update CI with OCaml 4.08


hhugo:
Some more changes where needed in ocaml-migrate-parsetree. see  
https://github.com/ocaml-ppx/ocaml-migrate-parsetree/pull/71

ocamlformat.opam
```diff
@@ -28,7 +28,7 @@ depends: [
   "cmdliner"
   "dune" {build & >= "1.1.1"}
   "fpath"
-  "ocaml-migrate-parsetree" {>= "1.0.10"}
+  "ocaml-migrate-parsetree" {>= "1.3.0"}
```
hhugo: will probably be 1.3.1 (see https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/73)

hhugo:
@Julow, I've updated the description with a maybe non exhaustive todo list.

Feel free to to push to this branch directly.

Julow:
I updated your branch. I implemented:

- Underscore in sugared functor argument
- Module substitution in signatures
- Type substitution in signatures
- Monadic let operators
- Arbitrary module expression after open in structures and expressions
- Fix some comments dropped
- Exceptions in or-patterns

Some requires a lot more testing (eg. `let open`, comments around new syntaxes).
This is very WIP, the current code must be cleaned before merging.

src/Sugar.ml
```diff
@@ -301,12 +301,15 @@ let rec functor_type cmts ~for_functor_kw ({ast= mty} as xmty) =
 (* The sugar is different when used with the [functor] keyword. The syntax
    M(A : A)(B : B) cannot handle [_] as module name. *)
 let rec functor_ cmts ~for_functor_kw ({ast= me} as xme) =
+  let valid_sugared_name arg =
+    Migrate_parsetree.Versions.OCaml_current.version >= 408
```
hhugo: Formatting a project in a 4.08 switch could make it not build with previous versions..
I don't think we should take `Ocaml_current.version` into account here.

test/passing/open_long.ml.ref
```diff
@@ -0,0 +1,177 @@
...
+end
+
+open (
+  struct
+      type t
+    end :
+    T )
+
+open (
+  struct
```
hhugo: `struct` should have the same indentation as `end`

gpetiot: fixed in f7541bc, but I choose to keep the indentation for the signature constraint after the `:` (similar to what is done after `=` in other cases).

src/Sugar.ml
hhugo: @jberdine, any suggestions on how to handle this case ?

hhugo:
I've implemented some of the missing bullets without much testing.

src/Cmts.ml
```diff
@@ -429,8 +435,8 @@ let init_use_file =
     (List.concat_map ~f:(fun toplevel_phrase ->
          match (toplevel_phrase : toplevel_phrase) with
          | Ptop_def items ->
-             List.map items ~f:(fun {Parsetree.pstr_loc} -> pstr_loc)
-         | Ptop_dir _ -> []))
+             List.map items ~f:(fun {Parsetree.pstr_loc; _} -> pstr_loc)
+         | Ptop_dir {pdir_loc; _} -> [pdir_loc]))
```
hhugo: This will disappear once #861 is merged

src/Migrate_ast.ml
```diff
@@ -78,13 +79,24 @@ module Pprintast = struct
 end
 
 (* Missing from ocaml_migrate_parsetree *)
```
hhugo: We should probably try to upstream that somehow

src/Sugar.mli
```diff
@@ -47,7 +47,7 @@ val fun_ :
      Cmts.t
   -> ?will_keep_first_ast_node:bool
   -> expression Ast.xt
-  -> arg_kind sexp_list * expression Ast.xt
+  -> arg_kind list * expression Ast.xt
```
hhugo: This change is on master already

src/Fmt_ast.ml
```diff
@@ -2166,13 +2187,27 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+            0
+        | _ -> c.conf.indent_after_in
+      in
+      let fmt_expr = fmt_expression c (sub_exp ~ctx body) in
+      let parens = parens || not (List.is_empty pexp_attributes) in
+      fmt_let_op c ctx ~ext ~parens ~fmt_atrs ~fmt_expr (let_ :: ands)
+        ~indent_after_in
 
 and fmt_class_structure c ~ctx ?ext self_ fields =
-  let fields =
```
hhugo: This was moved inside omp

test/passing/break_cases_all.ml.ref
```diff
@@ -162,3 +162,22 @@ let ffffff ~foo =
   | Eeeeeeeeeeeeeeeee ->
       foooooooooooooooooooo
   | Fffffffffffffffff -> fooooooooooooooooo
+
+let _ = function
+  | (exception A)
```
hhugo: Do we need the parens?

gpetiot: It's handled in the same way as `lazy` (which is also parenthesized), it's not a regression of this PR and it can be fixed in another one.

src/Fmt_ast.ml
```diff
@@ -743,7 +743,7 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
                      $ fmt_core_type c (sub_typ ~ctx typ) )
                  $ fmt_docstring_padded c doc
                  $ fmt_attributes c ~pre:(str " ") ~key:"@" atrs )
-        | Oinherit typ -> fmt_core_type c (sub_typ ~ctx typ)
+        | {pof_desc= Oinherit typ; _} -> fmt_core_type c (sub_typ ~ctx typ)
```
hhugo: We probably need to handle `pof_loc` and `pof_attributes` in the `Oinherit` case

src/Fmt_ast.ml
```diff
@@ -770,18 +770,19 @@ and fmt_package_type c ctx cnstrs =
...
-            $ list typs "@ & " (sub_typ ~ctx >> fmt_core_type c) )
+            ( Cmts.fmt c prf_loc
+            @@ ( fmt_str_loc c ~pre:(str "`") name
+               $ fmt_if (not (const && List.is_empty typs)) " of@ "
+               $ fmt_if (const && not (List.is_empty typs)) " & "
+               $ list typs "@ & " (sub_typ ~ctx >> fmt_core_type c) ) )
         $ fmt_attributes c ~key:"@" atrs
         $ fmt_docstring_padded c doc )
-  | Rinherit typ -> fmt_core_type c (sub_typ ~ctx typ)
+  | {prf_desc= Rinherit typ; _} -> fmt_core_type c (sub_typ ~ctx typ)
```
hhugo: We probably need to handle `prf_loc` and `prf_attributes` in the `Rinherit` case

src/Fmt_ast.ml
```diff
@@ -1787,16 +1788,23 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+        ~indent_after_in
   | Pexp_letexception (ext_cstr, exp) ->
       let pre = fmt "let exception@ " in
       hvbox 0
         ( wrap_if
             (parens || not (List.is_empty pexp_attributes))
             "(" ")"
             ( hvbox 0
-                (fmt_exception ~pre c (str ": ") ctx ext_cstr $ fmt "@ in")
+                (* XXX check me *)
```
hhugo: check me

src/Fmt_ast.ml
```diff
@@ -2311,7 +2341,7 @@ and fmt_class_type c ?(box = true) ({ast= typ} as xtyp) =
       in
       hvbox_if box 0 (list xt1N "@;-> " fmt_arg)
   | Pcty_extension ext -> fmt_extension c ctx "%" ext
-  | Pcty_open (flag, lid, cl) ->
+  | Pcty_open ({popen_override= flag; popen_expr= lid; _}, cl) ->
```
hhugo: handle `popen_loc` and `popen_attributes` ..

src/Fmt_ast.ml
```diff
@@ -2383,7 +2414,7 @@ and fmt_class_expr c ?eol ?(box = true) ({ast= exp} as xexp) =
            $ fmt_class_type c (sub_cty ~ctx t) ))
       $ fmt_atrs
   | Pcl_extension ext -> fmt_extension c ctx "%" ext $ fmt_atrs
-  | Pcl_open (flag, lid, cl) ->
+  | Pcl_open ({popen_override= flag; popen_expr= lid; _}, cl) ->
```
hhugo: handle `popen_loc` and `popen_attributes` ..

src/Fmt_ast.ml
```diff
@@ -3273,35 +3308,63 @@ and fmt_module_declaration c ctx ~rec_flag ~first pmd =
...
-    (fmt_module c keyword pmd_name xargs None colon (Some xmty)
+    (fmt_module c keyword pmd_name xargs None ?eqty (Some xmty)
        pmd_attributes)
 
+and fmt_module_substitution c ctx pms =
+  let {pms_name; pms_manifest; pms_attributes; pms_loc} = pms in
+  update_config_maybe_disabled c pms_loc pms_attributes
+  @@ fun c ->
+  let xmty =
+    (* TODO: improve *)
```
hhugo: What needs to be improved here ?

src/Sugar.ml
hhugo: My hunch is that we should have a config for that kind of things where one can mention a minimum version that needs to be supported.

src/Fmt_ast.ml
Julow: Fixed. However I could not find the syntax for this.

src/Fmt_ast.ml
Julow: Fixed. Couldn't find the syntax either but I could fix a dropped comment.

src/Fmt_ast.ml
Julow: Fixed by reusing `fmt_open_description`. Didn't find example for attributes but fix dropped comments.

src/Fmt_ast.ml
Julow: Same as in class types

src/Fmt_ast.ml
Julow: Sorry for the useless comment. The unsatisfying thing is the creation of a `module_type` from a `module_substitution`. The same commit also made changes to `fmt_module` that may deserve to be improved.

src/Sugar.ml
Julow: I changed to preserve the current syntax if an argument is `_` and no longer depend on the version of the parsetree. What do you think?

src/Fmt_ast.ml
```
type t = [ `A | `B ]
type u = [ t [@test] | `C]
```

src/Fmt_ast.ml
```
type t = < a : int >
type u = < (t [@test]); b : int> 
```

Julow: This is working fine on master, because the attribute is attached on the type with ocaml 4.08

src/Fmt_ast.ml
Julow: Same as object types, the attributes is attached to the type `t`, not the `row_field`.

src/Fmt_ast.ml
```
class type a = let open [@aaaaaaaaaaaaa] A in b
```

Julow: Thanks. This was not handled correctly so I fixed it.

src/Migrate_ast.ml
hhugo: Leaving that for later

hhugo:
The new examples in the testsuite that use the 4.08 syntax makes it harder to check regressions with the 4.07 parser. Is this an issue in practice ? Is there anything we can do apart from splitting examples out ?

Julow:
I merged hugo's PR, added some tests and rebased on master.

gpetiot:
I rebased on the last master, it looks good to me and we can merge as soon as the CI is alright.

src/Fmt_ast.ml
```diff
@@ -865,7 +885,7 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
   | Ppat_construct (lid, None) -> fmt_longident_loc c lid
   | Ppat_construct
       ( {txt= Lident "::"; loc}
-      , Some {ppat_desc= Ppat_tuple [x; y]; ppat_attributes= []; ppat_loc}
+      , Some {ppat_desc= Ppat_tuple [x; y]; ppat_attributes= []; ppat_loc; _}
```
hhugo: ppat_loc_stack is ignored here.

src/Fmt_ast.ml
```diff
@@ -1020,10 +1040,11 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
             (if parens then ")" else "")
             (if nested then "" else "@;<1 2>)") )
   | Ppat_constraint
-      ( {ppat_desc= Ppat_unpack name; ppat_attributes= []; ppat_loc}
+      ( {ppat_desc= Ppat_unpack name; ppat_attributes= []; ppat_loc; _}
       , ( { ptyp_desc= Ptyp_package (id, cnstrs)
           ; ptyp_attributes= []
-          ; ptyp_loc= (* TODO: use ptyp_loc *) _ } as typ ) ) ->
+          ; ptyp_loc= (* TODO: use ptyp_loc *) _
+          ; _ } as typ ) ) ->
```
hhugo: loc_stack ignored here.

hhugo:
I've moved the relocation of `*_loc_stack` in cmts.

Julow:
CI fixed (using 4.08.0+rc2).

Julow:
I fixed the bug I added to the CI script. I think this is ready to merge.

gpetiot:
Can the CI failure be due to `uuseg` not being dunified? Looks like `uuseg.string` is not visible from `uuseg`.

Julow:
I think travis tried to merge automatically and got in a weird state. The second build is doing the same commands and is passing.

.travis.yml
```diff
@@ -10,13 +10,16 @@ before_install:
...
+  - opam repo add --dont-select beta https://github.com/ocaml/ocaml-beta-repository.git
+  - opam switch --dry-run ${OCAML_VERSION} 2>/dev/null || opam switch create ${OCAML_VERSION} --repositories=default,beta
+  - opam switch ${OCAML_VERSION}
   - eval `opam config env`
   - export OPAMVERBOSE=1
 install:
-  - opam update --upgrade
-  - opam pin add --no-action ocamlformat .
-  - opam install --deps-only ocamlformat
+  - opam config exec --switch=${OCAML_VERSION} -- opam update --upgrade
```
Julow: I think this is not correct. The `opam update` command will not look at the environment variables and will use the switch selected above.

On my machine, `opam config exec --switch=4.08.0+beta3 -- opam switch show` gives `4.07.1`, which is my current switch.

You should either pass the `--switch` option directly to `opam update` or move the `opam switch ${OCAML_VERSION}` command into the `install` section.

.travis.yml
```diff
@@ -10,13 +10,16 @@ before_install:
   - export PATH=${HOME}:${PATH}
   - export OPAMYES=1
   - export OPAMJOBS=2
-  - opam init --compiler=${OCAML_VERSION} --disable-sandboxing
+  - opam init --bare --disable-sandboxing
+  - opam repo add --dont-select beta https://github.com/ocaml/ocaml-beta-repository.git
+  - opam switch --dry-run ${OCAML_VERSION} 2>/dev/null || opam switch create ${OCAML_VERSION} --repositories=default,beta
+  - opam switch ${OCAML_VERSION}
   - eval `opam config env`
   - export OPAMVERBOSE=1
```
Julow: As Etienne said, these two lines above do nothing and should be removed.

.travis.yml
Julow: You are still using `opam config exec`, I think it make no sense.

hhugo:
Should we wait for ocaml 4.08 to be officially released ? The current PR makes it impossible to run the testsuite with an older version of the compiler.

hhugo:
In term of release, when do you expect the next one ? We probably need a release soon after OCaml 4.08 is release (with support for the 4.08 syntax).

hhugo:
I've tried this branch on js_of_ocaml with both 4.07 and 4.08.
Here is the diff https://github.com/ocsigen/js_of_ocaml/pull/819/files
I found a regression with doc comments on exception  https://github.com/ocsigen/js_of_ocaml/pull/819/files#diff-cbc2fa1c5c44a83f4d10d62fc20cb780R692

hhugo:
4.08 is now officially release. Should we update the CI again :)

PR #832, opened 2019-05-14
Fix exception when calling String.sub

gpetiot:
Fix the following failure:
```
$ echo '(* *)'  | ocamlformat --name f - --debug
([1,0+0]..[1,0+5])   eol


ocamlformat.exe: Cannot process "f".
(Invalid_argument "Negative position: -1")
```

No diff with test_branch.

### DIFF
--- a/src/Source.ml
+++ b/src/Source.ml
@@ -29,7 +29,7 @@ let position_before t (pos : Lexing.position) =
   Option.map pos_cnum_opt ~f:(fun x -> {pos with pos_cnum= x + 1})
 
 let string_between t (l1 : Location.t) (l2 : Location.t) =
-  let pos = l1.loc_end.pos_cnum in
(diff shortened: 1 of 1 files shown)


src/Source.ml
```diff
@@ -29,7 +29,7 @@ let position_before t (pos : Lexing.position) =
   Option.map pos_cnum_opt ~f:(fun x -> {pos with pos_cnum= x + 1})
 
 let string_between t (l1 : Location.t) (l2 : Location.t) =
-  let pos = l1.loc_end.pos_cnum in
+  let pos = max 0 l1.loc_end.pos_cnum in
```
hhugo: Shouldn't we return `None` if pos_cnum is negative ?

gpetiot: you are right, now fixed.

Issue #476, opened 2018-10-31
Mirror ocp-indent options


gpetiot:
Doc of `ocp-indent` options: https://github.com/OCamlPro/ocp-indent/blob/master/.ocp-indent

- [x] `base` (= `let-binding-indent`) #822
- [x] `type` (= `type-decl-indent`) #822
- [x] `in` (= `indent-after-in`) #822
- [x] `with` (= `function-indent`/`match-indent`) #853
- [x] `match_clause` (=  `cases-exp-indent`)
- [x] `ppx_stritem_ext` (= `stritem-extension-indent`) #840 
- [x] `max_indent` (= `max-indent`) #841 
- [x] `strict_with` (= `function-indent-nested`/`match-indent-nested`) #853
- [ ] `strict_else` :
    this option conflicts with `--if-then-else={compact|fit-or-vertical|keyword-first|k-r}`, we have the style of `strict_else=always` in each of the `if-then-else` cases. Would it really make sense to implement `strict_else={auto|never}` for these cases? I have the feeling the current options are more powerful than the ocp-indent option.
- [ ] `strict_comments`
- [ ] `align_ops`
    this option does not make sense for `ocamlformat` because if an expression breaks like this:
    ```ocaml
      let f x = x
        + y
    ```
    the first part of the expression (`x`) won't be on the same line as `let f x =` but will instead be in the same box as `+ y`, so neither `align_ops=true` nor `align_ops=false` are applicable for `ocamlformat`.
- [ ] `align_params`
    same issue as `align_ops`:
    ```ocaml
    match foo with
    | _ -> some_fun
      ^^parameter
    ```

Issue #494, opened 2018-11-08
Don't over indent in lamba


Khady:
When using a pipe and a lamba function, the code of the lambda ends up very indented. I don't understand why exactly. It doesn't happen without the pipe. Would be nice to avoid this extra indentation.

```ocaml
(* original with pipe *)
let () =
  values
  |> List.iter begin fun x ->
    let x = " " ^ x in
    print_endline x
  end

(* ocamlformat with pipe, bad indentation *)
let () =
  values
  |> List.iter (fun x ->
         let x = " " ^ x in
         print_endline x )

(* original no pipe *)
let () =
  List.iter begin fun x ->
    let x = " " ^ x in
    print_endline x
  end values

(* ocamlformat no pipe, good indentation *)
let () =
  List.iter
    (fun x ->
      let x = " " ^ x in
      print_endline x )
    values
```

jberdine:
The indentation of the `fun` body should be indented the same offset, 4, from the beginning of the application (here, the beginning of `List.iter`) irrespective of using `|>`, and irrespective of whether there is a break before `(fun`.

brendanlong:
I'm not sure if this is complicated to implement, but the "before" code above is matching 100% of the existing OCaml code I've read, and 4 or 6 space indentation are both changes that look really inconsistent. I'm trying to switch some code over to ocamlformat, but it's a really big and unwelcome change that suddenly some things are indented by 2 spaces, some indented by 4 spaces, and some are indented by 6 spaces.

Is there any chance of a mode where every time we indent a new line, we just indent by 2 spaces more than the previous indentation level 100% of the time?

jberdine:
Closing this as there isn't anything planned that is similar to a increase indentation ignoring expression structure.

gpetiot:
Fixed with #841

Khady:
Thank you!

PR #841, opened 2019-05-17
Improve: option max-indent

gpetiot:
Alias for option `max_indent` of `ocp-indent`, making progress on #476


gpetiot:
You are right, related to #494 then.

ocamlformat-help.txt
```diff
@@ -243,6 +243,10 @@ OPTIONS (CODE FORMATTING STYLE)
            starts a line. always always apply match-indent. auto applies
            match-indent when seen fit. The default value is never.
 
+       --max-indent=COLS
+           Maximum indentation (COLS columns). The default value is none.
```
Julow: A little ambiguate. Could you explain that it's not the global indentation?

src/Conf.ml
```diff
@@ -377,6 +385,45 @@ end = struct
     store := Pack opt :: !store ;
     opt
 
+  let opt ~type_ ~type_name ~of_string ~to_string ~docv ~names ~doc ~section
```
Julow: Would be great if this could take any `option_decl`: `val opt : 'a option_decl -> 'a option option_decl`.
Doc, parsing function and `Arg.t` should compose, the problem is the way the `term` is wrapped in a function.

gpetiot: I agree it would be nice, the whole Conf file could use a redesign but preferably in another PR.

gpetiot:
For the first example it would maybe look better to break just after the arrow so `method` can be indented relatively to `object` but only in this specific case, we don't want that for any other case so I'm not sure we should change that.

For the second example, the culprit is the indent of 4 columns in `fmt_module` I don't know why this value was chosen since we don't have anything indented of 2 that could confuse the use. We could change that but I'm not sure this PR is the right place to do that. It does not depend on the feature of this PR.

PR #848, opened 2019-05-21
Fix comment dropped in field alias

### LINKED ISSUE #844: Bug: Comment dropped.
```
$ (echo "let x = { aaaaaaaaaa = aaaaaaaaaa "; echo "(* b *); b = b }")  | _build/dev/src/ocamlformat.exe - --name a.ml
ocamlformat.exe: Cannot process "a.ml".
  BUG: comments dropped.
File "a.ml", line 2, characters 0-7:
Error: Comment (* b *) dropped.
EXIT STATUS 1
```

gpetiot:
Fix #844, no regression with test_branch.

### DIFF
--- a/test/passing/record.ml
+++ b/test/passing/record.ml
@@ -106,3 +106,5 @@ let _ = x {a: string = a'; b= (b' : string)}
 let x = function {a= (_ : string); _} -> ()
 
 let x = function {a: string = _; _} -> ()
+
+let x = {aaaaaaaaaa (* b *); b}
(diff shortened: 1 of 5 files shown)


test/passing/record.ml
```diff
@@ -106,3 +106,5 @@ let _ = x {a: string = a'; b= (b' : string)}
 let x = function {a= (_ : string); _} -> ()
 
 let x = function {a: string = _; _} -> ()
+
+let x = {aaaaaaaaaa (* b *); b}
```
hhugo: Can you add an example with a new line like 
```ocaml
let x =
  { aaaaaaaaaa
  (* b *)
  ; b}
```

gpetiot: done

PR #851, opened 2019-05-22
Patch to allow checking of ocamlformat disable directive for disabled files

rudihorn:
We currently have the situation where we would like ocamlformat to be disabled by default, with the option of whitelisting it for certain files for integration with an existing project. Our setup is to create a `.ocamlformat` file with `disable=true` in it, and then individual files can specify `[@@@ocamlformat "enable"]`. 

Currently these files are completely ignored (or processed and then semingly the unchanged version returned. This patch attempts to fix this, with the new behaviour being that if the file is disabled, it will try to run the file and if an error occurs it will return the unchanged version. If it runs successfully, it will respect the value of `disable`. I'm not sure if this is the best way of doing it, but happy to fix the PR if there are better ways.

### DIFF
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -417,14 +417,15 @@ let format xunit (conf : Conf.t) ?output_file ~input_name ~source ~parsed ()
             print_check ~i:(i + 1) ~conf t_new ~source:fmted
   in
   let result =
-    if conf.disable then Ok source
-    else
-      match (parsed : ('a with_comments, exn) Result.t) with
-      | Error exn -> Error (Invalid_source {exn})
-      | Ok t -> (
-        try print_check ~i:1 ~conf t ~source with
-        | Sys_error msg -> Error (User_error msg)
-        | exn -> Error (Ocamlformat_bug {exn}) )
+    match (parsed : ('a with_comments, exn) Result.t) with
+    | Error exn -> 
+        if conf.disable
+        then Ok source 
+        else Error (Invalid_source {exn})
+    | Ok t -> (
+      try print_check ~i:1 ~conf t ~source with
+      | Sys_error msg -> Error (User_error msg)
+      | exn -> Error (Ocamlformat_bug {exn}) )
   in
   ( match result with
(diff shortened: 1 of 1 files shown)


gpetiot:
How many files do you typically ignore/format?
If you don't have too many files to ignore, you can use the `.ocamlformat-ignore` file.

If you have too many files to ignore, and not too many to parse, why not passing these files as parameters when invoking ocamlformat? (and not setting `disable`)
And I don't think it makes sense to use an `.ocamlformat-enable` file when `disabled` is set.

rudihorn:
The issue is two-fold. The first is that the project we are working on wants an explicit "disabled by default, whitelist as required" approach. If it were possible to have a `.ocamlformat-enable` file that would also be fine (but I assume that this is not supported?). The reason is that people involved in the project would like to be able to add files and not require any further intervention to disable ocamlformat. While this approach isn't exactly ideal, the hope is that the project will eventually be switched to an "enabled by default, blacklist as required" setting.

The other issue is that ocamlformat defines a fairly strict hierarchy, where `[@@@ocamlformat ...]` takes precedence over the other files, and the handling of the `disable` flag contradicts this, as it is completely ignored (and therefore has a lower priority value for a disabled value). This behaviour is also undocumented as far as I can tell, and I'm guessing is unintended?

gpetiot:
> If it were possible to have a `.ocamlformat-enable` file that would also be fine (but I assume that this is not supported?).

Yes it is not supported yet, but I guess we can implement it in a similar way to what is done for `.ocamlformat-ignore`.

gpetiot:
I tried another implementation in #854, that is closer to what is already done, and that avoids parsing the attributes at the beginning of the files.

rudihorn:
So I would honestly even be happier with #854 if it could be pulled. I would still find it a little sad if there is a discrepancy in the settings hierarchy, but I would still be fine with this. (Though it would be good to document this in the man page, where I believe it is not mentioned?).

gpetiot:
Closed in favor of #854 (@Julow feel free to reopen if you like this implementation better)

PR #854, opened 2019-05-24
Improve: .ocamlformat-enable files listing files to format when ocamlformat is disabled by default

gpetiot:
This is an alternative to #851

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -488,7 +488,14 @@ let info =
          should ignore. Each line in an $(b,.ocamlformat-ignore) file \
          specifies a filename relative to the directory containing the \
          $(b,.ocamlformat-ignore) file. Lines starting with $(b,#) are \
-         ignored and can be used as comments." ]
+         ignored and can be used as comments."
+    ; `P
+        "An $(b,.ocamlformat-enable) file specifies files that OCamlFormat \
+         should format even when the $(b,disable) option is set. Each line \
+         in an $(b,.ocamlformat-enable) file specifies a filename relative \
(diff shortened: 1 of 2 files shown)


src/Conf.ml
```diff
@@ -488,7 +488,14 @@ let info =
...
          $(b,.ocamlformat-ignore) file. Lines starting with $(b,#) are \
-         ignored and can be used as comments." ]
+         ignored and can be used as comments."
+    ; `P
+        "An $(b,.ocamlformat-enable) file specifies files that OCamlFormat \
+         should format even when the $(b,disable) option is set. Each line \
+         in an $(b,.ocamlformat-enable) file specifies a filename relative \
+         to the directory containing the $(b,.ocamlformat-enable) file. \
+         Lines starting with $(b,#) are ignored and can be used as \
+         comments." ]
```
avsm: Which has precedence if a file is specified in both `.ocamlformat-enable` and `.ocamlformat-ignore`?

gpetiot: With this implementation, `.ocamlformat-enable` is only checked when ocamlformat is disabled (`disable=true`), and `.ocamlformat-ignore` is only checked when ocamlformat is enabled (`disable=false`).

rudihorn:
Thanks for writing this PR! I'm generally happy. Maybe an explicit precedence may be less confusing (i.e. defining an answer to @avsm's question), but I'm happy either way.

src/Conf.ml
avsm: That sounds reasonable -- would be good to define that in the manual pages.

PR #855, opened 2019-05-24
Fix inconsistency of break-separators for wildcards in match cases

### LINKED ISSUE #850: `--break-separators` setting behaves inconsistently for wildcards in match cases
I'd expect code like this:
```ocaml
let _ = match something with
| { very_very_long_field_name_running_out_of_space = 1; another_very_very_long_field_name_running_out_of_space = 2; _ } -> 0
| _ -> 1
```
To be formatted into this with `break-separators` set to `after`:
```ocaml
let _ =
  match something with
  | { very_very_long_field_name_running_out_of_space = 1;
      another_very_very_long_field_name_running_out_of_space = 2;
      _
    } ->
      0
  | _ -> 1
```
However, I am seeing this:
```ocaml
let _ =
  match something with
  | { very_very_long_field_name_running_out_of_space = 1;
      another_very_very_long_field_name_running_out_of_space = 2
    ; _  (* <-- This is not right *)
    } ->
      0
  | _ -> 1
```
The semicolon placed at the beginning of the highlighted line is supposed to be located at the end of the previous line. The same issue exists for `after-and-docked`.

gpetiot:
Fix #850, no diff with test_branch.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -912,7 +912,9 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
         (wrap_if parens "(" ")"
            (wrap_record c.conf
               ( list flds (semic_sep c) fmt_field
-              $ fmt_if Poly.(closed_flag = Open) "@,; _" )))
+              $ fmt_if_k
+                  Poly.(closed_flag = Open)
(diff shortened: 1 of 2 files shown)


gpetiot:
Fixed for lists and arrays, no regression with test_branch.

Issue #835, opened 2019-05-14
Incomplete implementation of break-cases=after-and-docked


grievejia:
I believe currently the first field in the record behaves slightly differently from the rest:
```ocaml
let long_variable_name =
  { xxx = 1;
    yyy = 2;
    zzz = 3 }
in
```
Would it be possible to support a style where all fields get the same indentation? For example,
```ocaml
let long_variable_name =
  {
    xxx = 1;
    yyy = 2;
    zzz = 3
  }
in
```
Or a docked varaint (which I believe is already supported for record type declarations):
```ocaml
let long_variable_name = {
    xxx = 1;
    yyy = 2;
    zzz = 3
  }
in
```

grievejia:
ditto for lists, arrays, etc.

gpetiot:
Would the output of #856 be enough? We cannot put the opening `{` on the previous line right now, due to the current implementation of expressions.

grievejia:
@gpetiot Thank you so much for working on this! Yes as long as the indentations are consistent for all items it would be good enough for me.

PR #856, opened 2019-05-24
Fix inconsistency of break-separators=after-and-docked for record expressions

gpetiot:
Improving #835, only changed the record expressions for now, if the result is alright we can do the same for lists and arrays.
We now get something like:
```ocaml
let long_variable_name =
  {
    xxx= 1;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    yyy= 2;
    zzz= 3
  }
in
foo
```
we cannot do better and put the opening `{` on the previous line due to the current implementation of expressions.

### DIFF
--- a/src/Params.ml
+++ b/src/Params.ml
@@ -96,6 +96,26 @@ let get_record_type (c : Conf.t) ~wrap_record =
       ; break_after= break space (-2)
       ; docked_after= fmt "}" }
 
+type record_expr = {box: Fmt.t -> Fmt.t; sep_before: Fmt.t; sep_after: Fmt.t}
+
+let get_record_expr (c : Conf.t) ~wrap_record =
+  match c.break_separators with
+  | `Before ->
+      { box= (fun k -> hvbox 0 (wrap_record c k))
+      ; sep_before= fmt "@,; "
+      ; sep_after= noop }
+  | `After ->
+      { box= (fun k -> hvbox 2 (wrap_record c k))
+      ; sep_before= noop
+      ; sep_after= fmt ";@ " }
+  | `After_and_docked ->
+      let space = if c.space_around_records then 1 else 0 in
+      { box=
+          (fun k ->
+            hvbox 2 (wrap "{" "}" (break space 0 $ k $ break space (-2))))
+      ; sep_before= noop
+      ; sep_after= fmt ";@ " }
+
 type if_then_else =
   { box_branch: Fmt.t -> Fmt.t
   ; cond: Fmt.t
--- a/src/Params.ml
+++ b/src/Params.ml
@@ -110,10 +110,10 @@ let get_record_expr (c : Conf.t) ~wrap_record =
       ; sep_before= fmt "@,; "
       ; sep_after= noop }
   | `After ->
-      { box= (fun k -> hvbox 2 (wrap_record c k))
-      ; break_after_with= break 1 0
+      { box= (fun k -> hvbox 0 (wrap_record c k))
+      ; break_after_with= break 1 2
       ; sep_before= noop
-      ; sep_after= fmt ";@ " }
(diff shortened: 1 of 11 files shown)


gpetiot:
This PR only takes care of record expressions, I think it's better to take care of array and list expressions in separate PRs (`break-separators=after-and-docked` is not supported yet for array and list types so more work is needed).

test/passing/break_separators_after.ml.ref
```diff
@@ -0,0 +1,249 @@
+type t =
+  { (* fooooooooooooooooooooooooooooooooooooooooooooooooooooooo *)
+    foooooooooooooooooooooooo: foooooooooooooooooooooooooooooooooooooooo;
+    (* foooooooooooooooooooooooooooooooooooooooooooo *)
+    fooooooooooooooooooooooooooooo: fooooooooooooooooooooooooooo }
+
+type x =
+  | B of
+      { (* fooooooooooooooooooooooooooooooooooooooooo *)
+      aaaaaaaaaaaaaaa: aaaaaaaaaaaaaaaa;
```
Julow: Indentation missing (same with after-snd-docked).

src/Params.ml
```diff
@@ -96,6 +96,26 @@ let get_record_type (c : Conf.t) ~wrap_record =
       ; break_after= break space (-2)
       ; docked_after= fmt "}" }
 
+type record_expr = {box: Fmt.t -> Fmt.t; sep_before: Fmt.t; sep_after: Fmt.t}
+
+let get_record_expr (c : Conf.t) ~wrap_record =
+  match c.break_separators with
+  | `Before ->
+      { box= (fun k -> hvbox 0 (wrap_record c k))
```
Julow: I think `wrap_record` should be moved in this module (eg. `type record_pat = { wrap : Fmt.t -> Fmt.t }`)

test/passing/break_separators_after.ml.ref
gpetiot: Fixed.

gpetiot:
No diff for `tools/test_branch.sh fix-inconsistency-break-separators-docked 'break-separators=before' 'break-separators=before'`.

Diff for `tools/test_branch.sh fix-inconsistency-break-separators-docked 'break-separators=after' 'break-separators=after'`:
```
diff --git a/infer/src/IR/Pvar.ml b/infer/src/IR/Pvar.ml
index 9ab16c384..90bb5e56e 100644
--- a/infer/src/IR/Pvar.ml
+++ b/infer/src/IR/Pvar.ml
@@ -26,9 +26,9 @@ type pvar_kind =
       { translation_unit: translation_unit;
         is_constexpr: bool;
         (* is it compile constant? *)
-      is_ice: bool;
+        is_ice: bool;
         (* is it integral constant expression? *)
-      is_pod: bool;
+        is_pod: bool;
         is_static_local: bool;
         is_static_global: bool }  (** global variable *)
```

src/Params.ml
Julow: What I meant is that `wrap_record` is used in `Fmt_ast.ml` for record patterns and inline records. So it could be moved by implementing `get_record_pat` and `get_inline_record` (that would only have the `wrap` function) in `Params`. What do you think?

gpetiot: I moved the `wrap_` functions to Params.

gpetiot:
> The clean up is great !
> 
> `after-and-docker` is still not consistent for record patterns and inline records.

I know, the purpose of this PR was to fix the expressions first, I think it's better to take care of the patterns, and do the same for lists and arrays in other PRs. This PR is already bug enough.

PR #857, opened 2019-05-24
Improve: disambiguate non breaking matching structures

### LINKED ISSUE #423: Ambiguous sequence
```
let () = r := fun () -> f () ; g ()
```
looks ambiguous to me and could mean
```
let () = (r := fun () -> f () ); g ()
```
or
```
let () = r := fun () -> (f () ; g ())
```
or even (but types prevent it in my case)
```
let () = r := ((fun () -> f () ); g() )
```

gpetiot:
Fix #423, if it is indeed more readable (I'm not convinced with the diff of test_branch, no bugs, but it's not clear if it's an improvement) we can add similar options for other cases.

Edit: lots of regression with test_branch, do not merge.


gpetiot:
I'm thinking it would make sense for if-then-else too, is it an overkill to add an option per construct? Or a single option for all is enough?

Julow:
I think a single option for all matching constructs is enough.

The if-then-else is different because it has higher parsing priority than the `;`:

```ocaml
let _ =
  if a then b else c ;
  d

let _ = if a then b else ( c ; d )
```

This should be a different option/PR.

src/Fmt_ast.ml
```diff
@@ -186,6 +186,22 @@ let wrap_exp_if c ~parens ~loc k =
...
+  | `Begin_end -> wrap_fits_breaks_exp_begin_end ~parens k
+
+let wrap_disambiguate_fits_breaks_exp_if ?(space = true) c ~parens ~loc k =
+  match parens_or_begin_end c ~loc with
+  | `Parens ->
+      if c.conf.disambiguate_non_breaking_match then
+        wrap_if_fits_or parens "(" ")" k
+      else wrap_fits_breaks_if ~space c.conf parens "(" ")" k
+  | `Begin_end -> wrap_fits_breaks_exp_begin_end ~parens k
+
```
gpetiot: I am not very happy to add yet another layer to the wrap functions, it's very messy, and I'm not sure it would be better to have a single wrap function parameterized by many booleans (disambiguate, fits/breaks, preserve begin/end, the parens condition, etc.)

src/Fmt_ast.ml
```diff
@@ -176,15 +176,21 @@ let parens_or_begin_end c ~loc =
       let str = String.lstrip (Source.string_at c.source loc) in
       if String.is_prefix ~prefix:"begin" str then `Begin_end else `Parens
 
-let wrap_fits_breaks_exp_if ?(space = true) c ~parens ~loc k =
-  match parens_or_begin_end c ~loc with
-  | `Parens -> wrap_fits_breaks_if ~space c.conf parens "(" ")" k
-  | `Begin_end -> wrap_fits_breaks_exp_begin_end ~parens k
+let wrap_fits_breaks_exp_if ?(space = true) ?(preserve_kw = true)
```
Julow: `preserve_kw` is never used. Can be removed.

src/Fmt_ast.ml
Julow: Indeed looks better.

src/Fmt_ast.ml

gpetiot:
@mbouaziz @trefis any feedback before this PR gets merged?

gpetiot:
I'm reluctant to merge this one, as it would be good to reduce the number of options for the next release, and I'm not sure this one would be widely used.
Regarding the implementation, I don't like making the wrap function more and more complicated, and the disambiguation seems too arbitrary to me.
So I'm not sure what to do with it, @Julow do we close this one? (and what about the parent issue?)

Julow:
Disambiguation is often completely arbitrary. I find this useful and I don't think we should close this only because it is adding a new option.

src/Fmt_ast.ml
```diff
@@ -176,15 +176,20 @@ let parens_or_begin_end c ~loc =
...
 
-let wrap_fits_breaks_exp_if ?(space = true) c ~parens ~loc k =
-  match parens_or_begin_end c ~loc with
-  | `Parens -> wrap_fits_breaks_if ~space c.conf parens "(" ")" k
-  | `Begin_end -> wrap_fits_breaks_exp_begin_end ~parens k
+let wrap_fits_breaks_exp_if ?(space = true) ?(disambiguate = false) c
+    ~parens ~loc k =
+  if Poly.(parens_or_begin_end c ~loc = `Begin_end) then
+    wrap_fits_breaks_exp_begin_end ~parens k
+  else if disambiguate && c.conf.disambiguate_non_breaking_match then
```
Julow: To make this a little more readable you could revert to using a match and write the disambiguate case like this: `` `Parens when disambiguate && c.conf.disambiguate_non_breaking_match``.

PR #863, opened 2019-05-27
fix a bug in test/test.sh

gasche:
The order of printf argument on the changed line was wrong, printing
the following error message with my use-case (addition of
a not-yet-working testfile for an in-progress feature):

```
./test.sh: line 169: printf: multiline_data_delimiters_style: invalid number
[FAILED] [CHANGE: +42] 41
```


hhugo:
duplicate of https://github.com/ocaml-ppx/ocamlformat/pull/858

PR #867, opened 2019-06-02
Ast 408, update with [*_loc_stack]s

hhugo:
extend #831


src/Source.ml
```diff
@@ -112,50 +112,66 @@ let find_after t f (loc : Location.t) =
...
+      let l =
+        find_after t
+          (function
+            | RBRACKET ->
+                if !count = 0 then true else ( Int.decr count ; false )
+            (* It is not clear that an LBRACKET* will ever happen in
+               practice, we're just being defensive here. *)
+            | LBRACKET | LBRACKETBAR | LBRACKETLESS | LBRACKETGREATER
+             |LBRACKETPERCENT | LBRACKETPERCENTPERCENT | LBRACKETAT
+             |LBRACKETATAT | LBRACKETATATAT ->
```
Julow: `LBRACKETBAR` may be closed by `BARRBRACKET` so you should catch it for decrementing.

src/Cmts.ml
```diff
@@ -462,12 +476,16 @@ let relocate t ~src ~before ~after =
     if Conf.debug then
       Format.eprintf "relocate %a to %a and %a@\n%!" Location.fmt src
         Location.fmt before Location.fmt after ;
+    let append_and_sort x acc =
+      List.rev_append x acc
+      |> List.sort ~compare:(Comparable.lift Location.compare_start ~f:snd)
+    in
     update_multi t.cmts_before src before ~f:(fun src_cmts dst_cmts ->
-        List.append src_cmts dst_cmts) ;
+        append_and_sort src_cmts dst_cmts) ;
```
Julow: Arguments flipped ?

test/passing/loc_stack.ml
```diff
@@ -0,0 +1,12 @@
+
```
Julow: This test is not passing with `max-iter=2` (eg. with `make test`)

src/Cmts.ml
```diff
@@ -150,10 +150,22 @@ module Loc_tree = struct
       locs := loc :: !locs ;
       loc
     in
-    map_ast
-      Ast_mapper.{default_mapper with location}
-      (map_ast Ast_mapper.{default_mapper with attribute} ast)
-    |> ignore ;
+    let expr (m : Ast_mapper.mapper) x =
+      let pexp_loc_stack = List.map x.pexp_loc_stack ~f:(m.location m) in
```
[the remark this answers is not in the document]
hhugo: `loc_stack` comes from https://github.com/ocaml/ocaml/pull/1960.
I don't remember why it's not mapped by default. It might be an omission.

src/Cmts.ml
hhugo: I don't follow you.. can you explain ?

src/Source.ml
hhugo: Well this code was not change in that feature (only the indentation changed). Someone should probably open a new PR for this.

src/Cmts.ml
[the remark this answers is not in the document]
Julow: Thanks

src/Cmts.ml
Julow: The 2 next occurrences are `append_and_sort dst_cmts src_cmts`. But this should not be an issue because of the sort.

test/passing/loc_stack.ml
hhugo: I've pushed the `.opts` file

src/Cmts.ml
hhugo: I've cleaned the code a bit.

src/Fmt_ast.ml
```diff
@@ -1915,12 +1931,10 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       in
       match compact with
       | None ->
-          let leading_cmt = Cmts.fmt_before c e0.pexp_loc in
```
hhugo: This will cause regressions on OCaml 4.07 but improve things on 4.08.
Not sure what to do here ..

Julow: I did not find an example. Could you give one?

```
let () = ( (* before *) match (* after *) x with _ -> x)
```

with 4.07
```
let () = match (* before *)
               (* after *) x with _ -> x
```

with 4.08
```
let () = (* before *) match (* after *) x with _ -> x
```

Julow: Thanks. I got surprised by the `dune-workspace` file.

This bug is also present (but comments are placed before) on `master` with 4.07, so at least it's not a regression:

```diff
-let () = ( (* before *) match (* after *) x with _ -> x)
+let () = (* before *)
+         (* after *) match x with _ -> x
```

hhugo: Well, comments were place before on purpose to mitigate the issue. I believe it will be perceived as a regression by people stuck with 4.07.
IIRC, the main issue was

```ocaml
(* long comment
    on mutli
    lines
*)
match x with
| A -> 1
| B -> 0
```

that got reformatted into

```ocaml
match
  (* long comment
    on mutli
    lines
  *)
  x
with
| A -> 1
| B -> 0
```

Julow: Reverting this change doesn't seems to add regressions on 4.08. Maybe we could open an issue to solve this later?

PR #870, opened 2019-06-05
Add an option to control the indentation of nested matches

trefis:
I think the title of the PR (along with the commit messages) says everything there is to say.


trefis:
Actually I wonder whether this new option should actually be merged with `--match-indent-nested` that @gpetiot introduced last week.

:thinking:

PR #874, opened 2019-06-06
Fix: properly interpret indicate-multiline-delimiters for if-then-elses

trefis:
This is a fix for the issue noted in [this comment](https://github.com/ocaml-ppx/ocamlformat/pull/603/files/671324a9869413ae7b5df82d14cd54684f58a96b#pullrequestreview-246550702).

As I said, I think `indicate_multiline_delimiters` is interpreted incorrectly in the context of `if-then-else`s, so I updated the handling of `fit-or-vertical` to something which I believe to be more correct.

There is decent amount of noise in this PR, mostly due to me trying (and succeeding I think!) to avoid trailing whitespaces.


trefis:
(FTR: I just pushed a simpler and more correct implementation)

trefis:
OK, I've now updated all the other cases.

And I added many tests.

PR #876, opened 2019-06-07
Improve: add a new way to indicate multiline delimiters

trefis:
This GPR changes `indicate-multiline-delimiters` from a boolean to a variant with the following cases:
- `space`: equivalent to the previous `true`
- `no`: equivalent to the previous `false`
- `closing-on-separate-line`: forces the closing delimiter to be on a separate line if the expression spans multiple lines.

This isn't as complete as I would like, in particular:
```ocaml
let whatever_labelled a_function_name long_list_one some_other_thing =
  ListLabels.map long_list_one ~f:(fun long_list_one_elt ->
      do_something_with_a_function_and_some_things a_function_name
        long_list_one_elt some_other_thing)
```
is formatted in the same way regardless of this option.
I might look into that in a follow up PR, but this PR already brings several changes so it seems reasonable to stop here.

Also, this is obviously a breaking change. We could mitigate that by changing `space` to `true` and `no` to `false`, that way only the command line flag would break but config files would keep working (indeed if I do that, I can run `make test` without issues in the test directory).
Or we could just add a new option (but I don't think that'd be the right choice).


gpetiot:
Not important but I would change the name of the option to `multiline-delimiters` and use `none` instead of `no`.

gasche:
This is very cool, thanks! I would definitely use the feature.

Is it intended to also apply to data delimiters for lists and arrays and records? (For records, having the `}` on a separate line is already possible.) I tried to implement that feature but I got lost somewhere in the mist of fits_or_breaks combinators and delimiter-related options.

trefis:
> Is it intended to also apply to data delimiters for lists and arrays and records?

It doesn't right now, and as far as I can see doesn't apply to multiline tuples either.
I believe it should be made to, but let's consider this PR a first-step in that direction.

> Not important but I would change the name of the option to multiline-delimiters and use none instead of no.

Meh. For one thing, it's breaking even more than my current proposition (but I don't know what the policy is regarding breaking changes, so it might be OK). But also, I'm not really convinced, for example I would expect `multiline-delimiters=none` to mean that no parentheses will be added around multiline tuples, but that's not what the option is about. That being said, I'm not claiming the name of the option (or of any of the choices for that matter) is good.

trefis:
> I believe it should be made to, but let's consider this PR a first-step in that direction.

That doesn't sound quite right actually.
Currently, we have the following boolean options: `space-around-arrays`, `space-around-lists`, `space-around-records`, `space-around-variants` (whatever that is); surprisingly enough, there is no such option for tuples. We could give these the same treatement as delimiters in this PR.

Again, it seems reasonable to do that in a separate PR.

src/Fmt_ast.ml
```diff
@@ -1674,15 +1678,19 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         $ fmt_atrs )
   | Pexp_assert e0 ->
       let paren_body = parenze_exp (sub_exp ~ctx e0) in
+      let closing_paren =
+        match c.conf.indicate_multiline_delimiters with
+        | `Space -> "@ )"
+        | `No -> "@,)"
+        | `Closing_on_separate_line -> "@;<1000 -2>"
```
gpetiot: The parenthesis is missing.

trefis: Good catch fixed.

trefis:
I will have a look at that `test_branch.sh` test, thanks.

As for the indentation variation: I think the slight inconsistency is due to the (weird IMO) indentation behavior of nested matches.
What I would want is something like this:
```
the_context (
  inner_expr
)
```
That is, an indentation of `-2`.
So if you're in the context of a case, and the inner expr is also a match, you'd get:
```
| some_pat -> (
    match ... with
    | ...
  )
| some_other_pat -> ...
```
Except that for nested matches, by default the indentation (which can be changed since #870) you'd get is:
```
| some_pat -> (
  match ... with
  | ...
)
| some_other_pat ->
```
Which I find less readable.
So in that case I chose to have:
```
| some_pat -> (
  match ... with
  | ...
  )
| some_
```
That is, an indentation of `0` for the closing parenthesis.

trefis:
To give a bit more context: I tried choosing the indentation based on the indent of the enclosing box, which could be non-obvious for matches.
Also, it turns out that I got it wrong on some cases, like in the `Pexp_assert` (I noticed that today while using this branch on some code with assertions). I'll fix that and add a regtest.

src/Conf.ml
```diff
@@ -975,12 +975,21 @@ module Formatting = struct
...
+      "How to indicate that two matching delimiters live on different lines."
     in
     let names = ["indicate-multiline-delimiters"] in
-    let default = true in
-    C.flag ~names ~default ~doc ~section
+    let all =
+      [ ( "space"
+        , `Space
+        , "$(b,space) prints a space inside the delimiter to indicate the \
+           matching one is on a different line" )
```
Julow: Could you add dots at the end of the sentences ? Otherwise the generated doc is hard to read.

src/Fmt_ast.ml
```diff
@@ -1759,11 +1771,16 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                , fmt_after_cmts
                , (fmt_op, [(Nolabel, arg)]) ))) )
   | Pexp_construct (({txt= Lident "::"; loc= _} as lid), Some arg) ->
+      let opn, cls =
+        let ( ! ) = Caml.format_of_string in
```
Julow: Minor: Instead you could partially apply `wrap` to force the type without annotation and introducing a new syntax.

src/Fmt_ast.ml
```diff
@@ -1759,11 +1771,16 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+        match c.conf.indicate_multiline_delimiters with
+        | `No -> (!"(", !")")
+        | `Space -> (!"( ", !" )") (* Why is a space used instead a hint? *)
+        | `Closing_on_separate_line -> (!"(", !"@;<1000 -2>)")
```
Julow: This is around the `( :: )` constructor so this should be like `No`. Assuming this is not correct (but should never happen):
```ocaml
a (::
) b
```

test/passing/ite_kr_closing.ml.ref
```diff
@@ -0,0 +1,182 @@
...
+    this is more
+  ) else
+    e
+
+;;
+f
+  ( if loooooooooooooooooooooooooooooooooooooooooooooooooooooooooong then
+    ()
+  else
+    ()
```
Julow: Should be indented by 2 ?

gpetiot: No actually the indentation is consistent with what is already done, see Josh's answer: https://github.com/ocaml-ppx/ocamlformat/pull/603#discussion_r262851548

Julow: Got it.

src/Fmt.ml
```diff
@@ -142,12 +142,16 @@ let wrap_if_fits_and cnd pre suf k =
   fits_breaks_if cnd pre "" $ k $ fits_breaks_if cnd suf ""
 
 let wrap_fits_breaks_if ?(space = true) c cnd pre suf k =
```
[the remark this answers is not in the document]
gpetiot: Agreed but it should be done in another PR.

src/Fmt_ast.ml
gpetiot: Agreed, fixed.

src/Fmt_ast.ml

src/Conf.ml
gpetiot: Fixed.

trefis:
Thanks @gpetiot!
I'm sorry that I haven't had time to look at this recently.

PR #880, opened 2019-06-07
Toplevel with seperate or behaviour

rudihorn:
This is a patch to add a break-cases option that behaves like toplevel, except when nested _or_ patterns are encountered and then newlines are added between the _or_ patterns.

Example output:
```
let v =
  match x with
  | 0
   |1
   |2 ->
      3
  | 5 -> 6
```

Not sure if this is the best way to do it, open to suggestions. Would also possibly be nice to be able to have the `3` on the same line, but this is also difficult I believe.


gpetiot:
This is the same as `break-cases=all`, are you sure you didn't forget a commit?

rudihorn:
Ah I'm seeing the issue is that this is because the behaviour of `break-cases=all` changed since the latest release. The old `ocamlformat` would format the code as:

```ocaml
let v =
  match x with
  | 0
   |1
   |2 ->
      3
  | 5 ->
      6
```

PR #883, opened 2019-06-09
Improve: align cases horizontally

### LINKED ISSUE #360: Align match cases
Would it be possible for ocamlformat to align match cases?

e.g. having

```ocaml
match f with
| Foo -> toto
| Bar ijx -> bar ijx
```

turned into

```ocaml
match f with
| Foo     -> toto
| Bar ijx -> bar ijx
```

gpetiot:
Fix #360 (also see https://github.com/owlbarn/owl/pull/416#discussion_r291764914)
I try to avoid weird cases when there are comments and it looks good in most cases (see added tests).


src/Fmt_ast.ml
```diff
@@ -2654,6 +2669,45 @@ and fmt_class_type_field c ctx (cf : class_type_field) =
...
+    | Ppat_var {txt= s; _}
+     |Ppat_constant (Pconst_integer (s, _))
+     |Ppat_constant (Pconst_float (s, _))
+     |Ppat_construct ({txt= Lident s; _}, None) ->
+        Some (String.length s)
+    | Ppat_construct ({txt= Lident s; _}, Some arg) -> (
+      match pattern_len ~parens:2 arg with
+      | Some arg -> Some (parens + String.length s + 1 + arg)
+      | None -> None )
+    | Ppat_constant (Pconst_char _) -> Some 4 (* when the char is escaped *)
```
Julow: We can't make approximation:

```ocaml
let _ =
  match f with
  | 'a'    -> 1
  | '\n'    -> 2
  | '\x12'    -> 2
  | pattern -> 3
```

This depends on concrete syntax and `escape-chars` option. I suggest making `fmt_char_escaped` return a string instead and using it to measure what will be printed.

src/Fmt_ast.ml
```diff
@@ -2654,6 +2669,45 @@ and fmt_class_type_field c ctx (cf : class_type_field) =
...
+    | Ppat_variant (s, None) -> Some (String.length s + 1)
+    | Ppat_variant (s, Some arg) -> (
+      match pattern_len ~parens:2 arg with
+      | Some arg -> Some (parens + 1 + String.length s + 1 + arg)
+      | None -> None )
+    | Ppat_constant (Pconst_string (s, _)) -> Some (String.length s + 2)
```
Julow: This will not work if the string contains chars to be escaped or is multi-line.

Julow: Also, the second argument of `Pconst_string` is the delimiter of `{delim| ... |delim}` strings.

src/Fmt_ast.ml
```diff
@@ -2908,7 +2986,8 @@ and fmt_constructor_declaration c ctx ~first ~last:_ cstr_decl =
       ( hovbox 2
           ( hvbox 2
               ( Cmts.fmt c loc
-                  (wrap_if (is_symbol_id txt) "( " " )" (str txt))
+                  (wrap_if (is_symbol_id txt) "( " " )"
+                     (str txt $ fmt_padding))
```
Julow: The padding is before the closing parenthesis of symbol constructors:

```ocaml
type t =
  | ( ::               ) of a * b
  | Long_constructor of loooooooooooooooooooooooong_break
```

src/Fmt_ast.ml
```diff
@@ -2806,10 +2869,13 @@ and fmt_type_declaration c ?ext ?(pre = "") ?(brk = noop) ctx ?fmt_name
           (fmt_manifest ~priv:Public mfst $ str " =" $ fmt_private_flag priv)
         $ fmt "@ |"
     | Ptype_variant ctor_decls ->
+        let max acc d = max acc (String.length d.pcd_name.txt) in
```
Julow: This is not counting the parenthesis of some symbol constructors:

```ocaml
type t =
  | ( :: ) of a * b
  | [] of looooooooooooooooooooooooooooooooooooooong_break
```

But it's maybe not worth the complexity.

src/Fmt_ast.ml

src/Fmt_ast.ml

src/Fmt_ast.ml

src/Fmt_ast.ml
gpetiot: Now disabled instead of having something incorrect.

Julow: Good for me.

src/Fmt_ast.ml
```diff
@@ -2770,6 +2786,31 @@ and fmt_class_type_field c ctx (cf : class_type_field) =
...
+  let case_len case =
+    match case.ppat_desc with
+    | Ppat_any -> Some 1
+    | Ppat_var {txt= s; _}
+     |Ppat_constant (Pconst_integer (s, _))
+     |Ppat_constant (Pconst_float (s, _))
+     |Ppat_construct ({txt= Lident s; _}, None) ->
+        Some (String.length s)
+    | Ppat_constant (Pconst_char chr) ->
+        Some (String.length (char_escaped c ~loc:case.ppat_loc chr))
```
Julow: Missing `+ 2`:
```ocaml
let _ = function
  | 'a'      -> 1
  | abcdef -> 2
```

gpetiot:
> At some point I made a PR to your PR to implement tuples and constructors with arguments. Why did you remove that commit? Do you want me to submit that again, fixing eventual issues ?

Oh sorry it must have get lost when I rebased and forced-pushed, I will apply it again.

src/Fmt_ast.ml

test/passing/align_cases.ml
```diff
@@ -0,0 +1,92 @@
...
+  | '\n'    -> 2
+  | '\t'    -> 2
+  | '\x12'  -> 2
+  | pattern -> 3
+
+type t =
+  | ( :: ) of a * b
+  | []     of looooooooooooooooooooooooooooooooooooooong_break
+
+let _ = match (a, b) with A, B   -> a | AA, BB -> b | p      -> c
```
Julow: Should be disabled when the cases fit.

gpetiot: I found a workaround (for pattern matching at least) since we can't change the box structure.

src/Fmt_ast.ml
```diff
@@ -2804,13 +2860,29 @@ and fmt_cases c ctx cs =
...
       in
       let indent = if align_nested_match then 0 else indent in
+      let fmt_padding =
+        fmt_if_k
+          ( c.conf.align_cases
+          && not (Cmts.has_after c.cmts xlhs.ast.ppat_loc) )
+          ( match (max_len_name, pattern_len xlhs.ast) with
+          | Some max_len, Some len ->
+              let leading_len = String.length "let _ = match _ with" in
+              let case_len = max_len + String.length "|  -> " in
```
gpetiot: I did this (lines 2869-2870) to avoid using magic numbers, the compiler should optimize this anyway.

gpetiot:
@samoht since you opened #360 don't hesitate to give us your feedback about this feature.

gpetiot:
The diff of `tools/test_branch.sh align-cases '' 'align-cases=true,align-constructors-decl=true,align-variants-decl=true'` looks good in most cases (some corner cases will be difficult to get right).

src/Fmt_ast.ml
Julow: I don't think this will be optimized but the code is nice to read, don't be ashamed !

gpetiot:
My last 2 commits fix this issue (but still not perfect because we don't take into account the size of the type after `of`).

Julow:
Nice ! Indeed the break prediction is not perfect

```diff
 let _ = function
-  | 'a'    -> 1
+  | 'a' -> 1
   | abcdef -> looooooooooooooooooooooooooooooooooooooong_break

 type t =
   | ( :: ) of a * b
-  | []     of looooooooooooooooooooooooooooooooooooooong_break
+  | [] of looooooooooooooooooooooooooooooooooooooong_break
```

With the current box structure, what we need is a `fits_breaks` that checks breaks in the parent or grandparent box.

gpetiot:
That's a good idea, I modified `fits_breaks` with my last commit, and it works as expected (could need some reword before merging though)

vendor/ocamlformat_support/format_.mli
```diff
@@ -359,8 +359,8 @@ val pp_print_or_newline : formatter -> int -> int -> string -> string -> unit
   not just been split. Otherwise, print the second string. *)
 
 val pp_print_fits_or_breaks :
-  formatter -> string -> int -> int -> string -> unit
-(** [pp_print_fits_or_breaks fmt fits nspaces offset breaks] prints [fits] if
+  formatter -> int -> string -> int -> int -> string -> unit
```
Julow: Shouldn't it be optional ? For compatibility.

gpetiot: Done, but it could still change before being merged into upstream ocaml anyway.

PR #884, opened 2019-06-09
Fix formatting of arguments when break-fun-decl=fit-or-vertical

### LINKED ISSUE #877: function arguments formatted incorrectly
Input:
```
let fffffffffffffffffffffffffffffffffff x
      yyyyyyyyyyyyyyyyyyyyyyyyyyy =
  ()
```

Output:
```
let fffffffffffffffffffffffffffffffffff x
                                        yyyyyyyyyyyyyyyyyyyyyyyyyyy = ()
```

This function shouldn't need to be line-wrapped at all, it fits on one line.

This also isn't how I expect line-wrapping for function arguments to work when it does happen, but that may be down to whatever bug is causing the wrapping, or due to different configuration I'm used to. First I just want to stop this from wrapping.

Command-line:
`ocamlformat --profile=janestreet --enable-outside-detected-project unformatted.ml > formatted.ml`

I've seen this as early as 722240710c8ac5066862188b27a90a709ebb325c and as recently as 9f0bde1a8138eedaff4d844605d048f712703e90; I haven't checked other commits.

gpetiot:
Fix #877, maybe it could be improved to not have `:` alone on a line, but as a whole the formatting is more consistent.
And not using `Location.is_single_line` is a plus.

@Julow since you introduced this option, feel free to directly commit to this PR.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2444,7 +2444,7 @@ and fmt_class_expr c ?eol ?(box = true) ({ast= exp; _} as xexp) =
         (if Option.is_none eol then 2 else 1)
         (wrap_if parens "(" ")"
            ( hovbox 2
-               ( box_fun_decl_args c 4
+               ( box_fun_decl_args c 0
                    ( str "fun "
                    $ fmt_attributes c ~key:"@" pcl_attributes ~suf:(str " ")
                    $ wrap_fun_decl_args c (fmt_fun_args c xargs)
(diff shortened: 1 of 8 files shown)


src/Fmt_ast.ml
```diff
@@ -140,9 +140,7 @@ let fmt_expressions c width sub_exp exprs fmt fmt_expr =
 let wrap_fun_decl_args ~stmt_loc c k =
```
Julow: `~stmt_loc` is not used.

src/Fmt_ast.ml
```diff
@@ -2439,11 +2437,15 @@ and fmt_class_expr c ?eol ?(box = true) ({ast= exp; _} as xexp) =
   | Pcl_fun _ ->
       let xargs, xbody = Sugar.cl_fun c.cmts xexp in
       let stmt_loc = Sugar.args_location xargs in
+      let args_box =
```
Julow: This is duplicated every time. I suggest returning the `args_box` from `wrap_fun_decl_args`.

src/Fmt_ast.ml

src/Fmt_ast.ml

gpetiot:
The diff has now been reduced on tests, I'm running some tests with test_branch but it looks good so far.

gpetiot:
Yes it will be done in another PR.

PR #885, opened 2019-06-11
Improve: synchronize Format with upstream stdlib

gpetiot:
Updating the patched version of `Format` to make sure the changes proposed in the ocaml PR (https://github.com/ocaml/ocaml/pull/1229) will work for ocamlformat.

Do not merge this, we will eventually remove the patched Format files when https://github.com/ocaml/ocaml/pull/1229 and https://github.com/ocaml/ocaml/pull/8719 are merged.


gpetiot:
@Julow since #1229 looks far from being approved I think we should consider merging this one to rely on the most recent API.

vendor/ocamlformat_support/queue.mli
```diff
@@ -0,0 +1,108 @@
+(**************************************************************************)
```
Julow: Why do you need to vendor `Queue` and `Stack` ?

gpetiot: It was needed when we were relying on 4.07.

gpetiot:
Diff with Format upstream:
```
$ diff vendor/ocamlformat_support/format_.mli ~/ocaml/stdlib/format.mli
978,979d977
< val pp_set_max_newline_offset : formatter -> int -> unit
< 
```
```
$ diff vendor/ocamlformat_support/format_.ml ~/ocaml/stdlib/format.ml
165,167d164
<   (* Maximum offset added to a new line in addition to the offset of
<      the previous line. *)
<   mutable pp_max_newline_offset : int;
290,291d286
<   let real_indent =
<     min real_indent (state.pp_current_indent + state.pp_max_newline_offset) in
879,881d873
< let pp_set_max_newline_offset state n =
<   state.pp_max_newline_offset <- n
< 
1020d1011
<     pp_max_newline_offset = pp_margin - pp_min_space_left;
```

Issue #868, opened 2019-06-04
multi-line Format strings get wrapped


raphael-proust:
We use multi-line format “strings” (of the `format` type) like those:

```
$ cat format_strings.ml
let () =
  if true then (* Shrinking the margin a bit *)
    Format.printf
      "@[<v 2>@{<warning>@{<title>Warning@}@}@,@,\
      \        These are @{<warning>NOT@} the Droids you are looking for!@,\
       @,\
      \        Some more text. Just more letters and words.@,\
      \        All this text is left-aligned because it's part of the UI.@,\
      \        It'll be easier for the user to read this message.@]@\n@."
```
The identation and styling of the `format` string makes it obvious that the text is correctly indented. It mimics the output:

```
$ ocaml format_strings.ml
Warning

          These are NOT the Droids you are looking for!

          Some more text. Just more letters and words.
          All this text is left-aligned because it's part of the UI.
          It'll be easier for the user to read this message.
```

`ocamlformat` melts and wraps these multi-line format strings:

```
$ ocamlformat format_strings.ml
let () =
  if true then
    (* Shrinking the margin a bit *)
    Format.printf
      "@[<v 2>@{<warning>@{<title>Warning@}@}@,@,        These are \
       @{<warning>NOT@} the Droids you are looking for!@,@,        Some more \
       text. Just more letters and words.@,        All this text is \
       left-aligned because it's part of the UI.@,        It'll be easier for \
       the user to read this message.@]@\n\
       @."
```

----------------------------

For info:
```
$ ocamlformat --version
0.9
```

gpetiot:
What value of 'wrap_string_literals' are you using? It looks like you are using 'wrap' (by default), maybe you should try another value.

raphael-proust:
I used the default (no `.ocamlformat`) configuration. I considered the multi-line format wrapping to be an issue because with the same setup, I get much more pleasant results on strings:

```
$ cat multiline_strings.ml
let s =
  "Warning\n\
  \        These are NOT the Droids you are looking for!\n\
  \n\
  \        Some more text. Just more letters and words.\n\
  \        All this text is left-aligned because it's part of the UI.\n\
  \        It'll be easier for the user to read this message.\n"
;;
print_endline s
$ ocaml multiline_strings.ml
Warning
        These are NOT the Droids you are looking for!

        Some more text. Just more letters and words.
        All this text is left-aligned because it's part of the UI.
        It'll be easier for the user to read this message.

[...]

PR #887, opened 2019-06-12
Fix break-string-literals=newlines

gpetiot:
Progress on #868 but there is still a lot of work to do, I'm not satisfied with this implementation, the cases are too specific and it still fails on basic examples like `let pp_sep fmt () = F.fprintf fmt ", @,\n"`.
Feel free to commit if anyone has new ideas.

Edit: this bug is fixed, and more tests added.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -332,7 +332,7 @@ let fmt_constant c ~loc ?epi const =
         | _ -> str (escape_string mode s)
       in
       let fmt_lines ?(break_on_newlines = false) mode lines =
-        let delim = ["@,"; "@;"; "@\\n"] in
+        let delim = ["@,"; "@;"] in
         let fmt_line ?prev:_ curr ?next =
           let fmt_next next =
             let not_suffix suffix = not (String.is_suffix curr ~suffix) in
@@ -367,14 +367,12 @@ let fmt_constant c ~loc ?epi const =
       in
       let contains_pp_commands =
         let is_substring substring = String.is_substring s ~substring in
-        List.exists ["@,"; "@;"; "@\\n"] ~f:is_substring
+        List.exists ["@,"; "@;"] ~f:is_substring
       in
       match c.conf.break_string_literals with
       | `Newlines when contains_pp_commands ->
-          let break_on_pp_commands in_ delim =
-            let pat = String.Search_pattern.create delim in
-            let with_ = delim ^ "\n" in
-            String.Search_pattern.replace_all pat ~in_ ~with_
+          let break_on_pp_commands in_ pattern =
+            String.substr_replace_all in_ ~pattern ~with_:(pattern ^ "\n")
           in
           List.fold_left ["@,"; "@;"; "@\n"] ~init:s ~f:break_on_pp_commands
           |> String.split ~on:'\n'
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -331,14 +331,23 @@ let fmt_constant c ~loc ?epi const =
             hovbox_if (List.length words > 1) 0 (list_pn words fmt_word)
         | _ -> str (escape_string mode s)
       in
-      let fmt_lines mode lines =
-        let fmt_next next =
-          if String.is_empty next then str "\\n"
-          else if Char.equal next.[0] ' ' then
-            str "\\n" $ pre_break 0 "\\" (-1) $ if_newline "\\"
-          else str "\\n" $ pre_break 0 "\\" 0
-        in
+      let fmt_lines ?(break_on_newlines = false) mode lines =
+        let delim = ["@,"; "@;"; "@\\n"] in
         let fmt_line ?prev:_ curr ?next =
+          let fmt_next next =
+            let not_suffix suffix = not (String.is_suffix curr ~suffix) in
+            let print_ln =
+              List.for_all delim ~f:not_suffix || not break_on_newlines
+            in
+            if String.is_empty next then
+              if break_on_newlines then
+                fmt_if_k (String.is_empty curr) (str "\\n")
+              else str "\\n"
+            else if Char.equal next.[0] ' ' then
+              fmt_if_k print_ln (str "\\n")
+              $ pre_break 0 "\\" (-1) $ if_newline "\\"
+            else fmt_if_k print_ln (str "\\n") $ pre_break 0 "\\" 0
+          in
           fmt_line mode curr $ opt next fmt_next
         in
         hvbox 1
@@ -357,7 +366,16 @@ let fmt_constant c ~loc ?epi const =
         | _ -> (s, c.conf.escape_strings)
       in
       match c.conf.break_string_literals with
-      | `Newlines | `Wrap -> fmt_lines mode (String.split ~on:'\n' s)
+      | `Newlines ->
+          let break_on_pp_commands in_ delim =
+            let pat = String.Search_pattern.create delim in
+            let with_ = delim ^ "\n" in
+            String.Search_pattern.replace_all pat ~in_ ~with_
+          in
+          List.fold_left ["@,"; "@;"; "@\n"] ~init:s ~f:break_on_pp_commands
+          |> String.split ~on:'\n'
+          |> fmt_lines mode ~break_on_newlines:true
+      | `Wrap -> fmt_lines mode (String.split ~on:'\n' s)
       | `Never -> wrap "\"" "\"" (fmt_line mode s) )
 
 let fmt_variance = function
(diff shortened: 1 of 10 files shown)


gpetiot:
Diff with `tools/test_branch.sh fix-break-string-literals 'break-string-literals=newlines' 'break-string-literals=newlines'` looks good:
```ocaml
diff --git a/src/Translation_unit.ml b/src/Translation_unit.ml
index b1f27c0..531766d 100644
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -240,30 +240,35 @@ let print_error ?(quiet_unstable = false) ?(quiet_comments = false)
                 | Normalize.Moved (loc_before, loc_after, msg) ->
                     if Location.compare loc_before Location.none = 0 then
                       Format.fprintf fmt
-                        "%!@{<loc>%a@}:@,@{<error>Error@}: Docstring (** %s *) added.\n\
+                        "%!@{<loc>%a@}:@,\
+                         @{<error>Error@}: Docstring (** %s *) added.\n\
                          %!"
                         Location.print_loc loc_after (ellipsis_cmt msg)
                     else if Location.compare loc_after Location.none = 0
                     then
                       Format.fprintf fmt
-                        "%!@{<loc>%a@}:@,@{<error>Error@}: Docstring (** %s *) dropped.\n\
+                        "%!@{<loc>%a@}:@,\
+                         @{<error>Error@}: Docstring (** %s *) dropped.\n\
                          %!"
                         Location.print_loc loc_before (ellipsis_cmt msg)
                     else
                       Format.fprintf fmt
-                        "%!@{<loc>%a@}:@,@{<error>Error@}: Docstring (** %s *) moved to @{<loc>%a@}.\n\
+                        "%!@{<loc>%a@}:@,\
+                         @{<error>Error@}: Docstring (** %s *) moved to @{<loc>%a@}.\n\
                          %!"
                         Location.print_loc loc_before (ellipsis_cmt msg)
                         Location.print_loc loc_after
                 | Normalize.Unstable (loc, s) ->
                     Format.fprintf fmt
-                      "%!@{<loc>%a@}:@,@{<error>Error@}: Formatting of (** %s *) is unstable (e.g. parses as a list or not depending on the margin), please tighten up this comment in the source or disable the formatting using the option --no-parse-docstrings.\n\
+                      "%!@{<loc>%a@}:@,\
+                       @{<error>Error@}: Formatting of (** %s *) is unstable (e.g. parses as a list or not depending on the margin), please tighten up this comment in the source or disable the formatting using the option --no-parse-docstrings.\n\
                        %!"
                       Location.print_loc loc (ellipsis_cmt s))
           | `Comment_dropped l when not conf.Conf.quiet ->
               List.iter l ~f:(fun (loc, msg) ->
                   Format.fprintf fmt
-                    "%!@{<loc>%a@}:@,@{<error>Error@}: Comment (* %s *) dropped.\n\
+                    "%!@{<loc>%a@}:@,\
+                     @{<error>Error@}: Comment (* %s *) dropped.\n\
                      %!"
                     Location.print_loc loc (ellipsis_cmt msg))
           | `Cannot_parse ((Syntaxerr.Error _ | Lexer.Error _) as exn) ->
diff --git a/infer/src/pulse/PulseAbductiveDomain.ml b/infer/src/pulse/PulseAbductiveDomain.ml
index 48a1d4662..9c5876115 100644
--- a/infer/src/pulse/PulseAbductiveDomain.ml
+++ b/infer/src/pulse/PulseAbductiveDomain.ml
@@ -305,7 +305,11 @@ module PrePost = struct
 
   let pp_call_state fmt {astate; subst; rev_subst; visited} =
     F.fprintf fmt
-      "@[<v>{ astate=@[<hv2>%a@];@, subst=@[<hv2>%a@];@, rev_subst=@[<hv2>%a@];@, visited=@[<hv2>%a@]@, }@]"
+      "@[<v>{ astate=@[<hv2>%a@];@,\
+      \ subst=@[<hv2>%a@];@,\
+      \ rev_subst=@[<hv2>%a@];@,\
+      \ visited=@[<hv2>%a@]@,\
+      \ }@]"
       pp astate
       (AddressMap.pp ~pp_value:AbstractAddress.pp)
       subst
diff --git a/sledge/src/report.ml b/sledge/src/report.ml
index 65657d998..31b618861 100644
--- a/sledge/src/report.ml
+++ b/sledge/src/report.ml
@@ -11,7 +11,8 @@ let unknown_call call =
   [%Trace.kprintf
     (fun _ -> assert false)
       "@\n\
-       @[<v 2>%a Called unknown function %a executing instruction@;<1 2>@[%a@]@]@."
+       @[<v 2>%a Called unknown function %a executing instruction@;\
+       <1 2>@[%a@]@]@."
       (fun fs call -> Loc.pp fs (Llair.Term.loc call))
       call
       (fun fs (call : Llair.Term.t) ->
@@ -30,7 +31,9 @@ let invalid_access state pp access loc =
   [%Trace.kprintf
     (fun _ -> assert false)
       "@\n\
-       @[<v 2>%a Invalid memory access executing@;<1 2>@[%a@]@ from symbolic state@;<1 2>@[{ %a@ }@]@]@."
+       @[<v 2>%a Invalid memory access executing@;\
+       <1 2>@[%a@]@ from symbolic state@;\
+       <1 2>@[{ %a@ }@]@]@."
       Loc.pp (loc access) pp access Domain.pp state]
```

gpetiot:
@raphael-proust feel free to test this PR on your codebase, maybe we missed some cases?

raphael-proust:
With the option `break-string-literals=newlines` the diffs on format strings look good on our code base. I'm still experimenting on some of the options to minimise our diff.

src/Fmt_ast.ml
```diff
@@ -348,7 +357,20 @@ let fmt_constant c ~loc ?epi const =
...
+        List.exists ["@,"; "@;"; "@\\n"] ~f:is_substring
+      in
       match c.conf.break_string_literals with
+      | `Newlines when contains_pp_commands ->
+          let break_on_pp_commands in_ delim =
+            let pat = String.Search_pattern.create delim in
+            let with_ = delim ^ "\n" in
+            String.Search_pattern.replace_all pat ~in_ ~with_
+          in
+          List.fold_left ["@,"; "@;"; "@\n"] ~init:s ~f:break_on_pp_commands
```
Julow: The 3rd one is not the same as in `contains_pp_commands`: `"@\n"` vs `"@\\n"`. Is it intentional ?

src/Fmt_ast.ml
```diff
@@ -348,7 +357,20 @@ let fmt_constant c ~loc ?epi const =
...
+      let contains_pp_commands =
+        let is_substring substring = String.is_substring s ~substring in
+        List.exists ["@,"; "@;"; "@\\n"] ~f:is_substring
+      in
       match c.conf.break_string_literals with
+      | `Newlines when contains_pp_commands ->
+          let break_on_pp_commands in_ delim =
+            let pat = String.Search_pattern.create delim in
+            let with_ = delim ^ "\n" in
+            String.Search_pattern.replace_all pat ~in_ ~with_
```
Julow: `String.substr_replace_all in_ ~pattern:delim ~with_` is shorter and equivalent since `pat` is used once and discarded.

src/Fmt_ast.ml
```diff
@@ -348,7 +357,20 @@ let fmt_constant c ~loc ?epi const =
           | Some s -> (s, `Preserve) )
         | _ -> (s, c.conf.escape_strings)
       in
+      let contains_pp_commands =
+        let is_substring substring = String.is_substring s ~substring in
+        List.exists ["@,"; "@;"; "@\\n"] ~f:is_substring
```
Julow: You could avoid calculating `contains_pp_commands` by checking if the `break_on_pp_commands` logic below changed the string.

src/Fmt_ast.ml

src/Fmt_ast.ml
gpetiot: I simplified the lists when it was possible, but didn't find a way to use the same lists.

src/Fmt_ast.ml
gpetiot: Tried to do that, but failed because the lists are not the same in the check and the substitution. If we find a better way to write that I would be happy about it, otherwise I think we can just merge.

raphael-proust:
Looking a bit more at our code base, I found that in the current state, we'd probably use `[@@ocamlformat …]` syntax bits to change the value of the `break-string-literals` option. In some places we have pre-formatted (i.e., that have manual line-breaks) strings, and in other places we have long unformatted strings that are passed to a library that formats them automatically. Typically we use the former for printing warnings and legalese, the latter for, say, adding documentation to Cmdliner terms.

I think an option `break-string-literals=newlines-and-wrap` that breaks either when there's a newline in the string (or format string) or when it reaches the margin would be the most useful for us. That might be more complicated though… And if too complicated, we'll just use the `[@@]` option.

Issue #311, opened 2018-08-11
Generalize Fmt.fits_breaks to allow break specs within breaks string


jberdine:
Support for example:
```
    fits_breaks ".(" " in@ "
```
  instead of requiring
```
    fits_breaks ".(" " in" $ fits_breaks "" "@ "
```

gpetiot:
Parsing of the break removed in #894, and so far the pattern described in this issue is not used in ocamlformat, but feel free to re-open if we need that some day.

PR #892, opened 2019-06-17
Improve: option break-before-in

### LINKED ISSUE #506: Position of 'in' keyword for multiline bindings

gpetiot:
Fix #506 
The diff of test_branch for `break-before-in=auto` looks good, no regression for `break-before-in=always`.


src/Fmt_ast.ml
```diff
@@ -3806,9 +3806,14 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
 
 and fmt_let c ctx ~ext ~rec_flag ~bindings ~parens ~attributes ~fmt_atrs
     ~fmt_expr ~indent_after_in =
+  let fmt_in indent =
+    match c.conf.break_before_in with
+    | `Always -> break 1 (-indent) $ fmt "in"
+    | `Auto -> fits_breaks " in" ("@;<1 " ^ Int.to_string (-indent) ^ ">in")
```
Julow: I don't like this. The first thing `fits_breaks` does is parse its second argument, that is not a format bug looks like one. This is unnecessary code that adds confusion.

I suggest adding `fits_breaks'` function that takes the break values as additional arguments for when they are variable.

ocamlformat-help.txt
```diff
@@ -47,6 +47,12 @@ OPTIONS (CODE FORMATTING STYLE)
            assignment expression does not fit on a single line. The default
            value is end-line.
 
+       --break-before-in={always|auto}
+           Whether the line should break before the in keyword of a let
+           binding. always will always break the line before the in keyword.
+           auto will only break the line if the in keyword does not fit on
+           the previous line. The default value is always.
```
Julow: `always` is only when it cannot fit on a single line otherwise there is no break. What about `fit-or-vertical` instead ?

src/Fmt_ast.ml
gpetiot: I'm in favor of modifying `Fmt.fits_breaks` instead of having 2 functions:
```ocaml
val fits_breaks :
     ?force_fit_if:bool
  -> ?force_break_if:bool
  -> ?nspaces:int
  -> ?offset:int
  -> string
  -> string
  -> t
(** [fits_breaks fits nspaces offset breaks] prints [fits] if the enclosing
    box fits on one line, and otherwise prints [breaks], which is a string
    that optionally follows a break hint equivalent to
    ["@;<nspaces offset>"]. *)
```
it is in conflict with #311  but I'm really not sure the new feature proposed in #311 will really be used in ocamlformat (I did a quick check in Fmt_ast).

ocamlformat-help.txt
gpetiot: Fixed.

src/Fmt_ast.ml
Julow: This must be fixed but could be in an other PR. This modified `fits_breaks` looks good.

PR #894, opened 2019-06-18
Improve: specify break hint in fits_breaks

gpetiot:
Removing the parsing of `fits_breaks` second argument.
Using a pair to always specify both nspaces and offset.

### DIFF
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -103,18 +103,16 @@ let pre_break n s o fs = Format.pp_print_pre_break fs n s o
 (** Conditional on breaking of enclosing box ----------------------------*)
 
 let fits_breaks ?(force_fit_if = false) ?(force_break_if = false)
-    ?(nspaces = 0) ?(offset = Int.min_value) fits breaks fs =
+    ?(hint = (0, Int.to_int Int.min_value)) fits breaks fs =
+  let nspaces, offset = hint in
   if force_fit_if then Format.pp_print_string fs fits
   else if force_break_if then (
     if offset >= 0 then Format.pp_print_break fs nspaces offset ;
     Format.pp_print_string fs breaks )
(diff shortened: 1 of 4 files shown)


src/Fmt.ml
```diff
@@ -102,29 +102,17 @@ let pre_break n s o fs = Format.pp_print_pre_break fs n s o
...
-      | ';' -> (
-        try Scanf.sscanf breaks "@;<%d %d>%s" (fun x y z -> (x, y, z))
-        with Scanf.Scan_failure _ | End_of_file -> (1, 0, b) )
-      | ',' -> (0, 0, b)
-      | ' ' -> (1, 0, b)
-      | _ -> (0, Int.min_value, breaks)
-    else (0, Int.min_value, breaks)
-  in
+let fits_breaks ?(force_fit_if = false) ?(force_break_if = false)
+    ?(hint = (0, Int.to_int Int.min_value)) fits breaks fs =
```
Julow: `Int.to_int` ? This is definitively not needed.

src/Fmt_ast.ml
```diff
@@ -731,9 +731,12 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
...
           else "]" )
-          ( if Poly.(c.conf.type_decl = `Sparse) && space_around then
-            "@;<1000 0>]"
-          else "@ ]" )
+          "]"
+          ~hint:
+            ( ( if Poly.(c.conf.type_decl = `Sparse) && space_around then
+                1000
+              else 1 )
+            , 0 )
```
Julow: Minor style comment: Factorising the `, 0` part make the formating a bit ugly.

src/Fmt.ml
gpetiot: Done.

src/Fmt_ast.ml

PR #899, opened 2019-06-24
Improve: add trailing semicolon inside record when break-separators=after-and-docked

### LINKED ISSUE #895: Allow trailing semicolon inside record?
Trailing semicolons on fields always get removed.  For example,
```
ErrorType of {
  field_one;
  field_two;
}
```
gets turned into
```
ErrorType of {
  field_one;
  field_two
}
```
The issue with such behavior is that it messes with the source-control blame when we want to add/remove fields. Would be great if there's an option to turn it off.

gpetiot:
Fix #895 (I didn't add any option, I think we always want this behavior).

Cases like this:
```ocaml
type x =
   | B of
       { aaaaaaaaaaaaaaa: aaaaaaaaaaaaaaaa;
-        bbbbbbbbbbbbbbbbbbbbbbb: bbbbbbbbbbbbbbb }
+        bbbbbbbbbbbbbbbbbbbbbbb: bbbbbbbbbbbbbbb; }
```
will be fixed in another PR (the brackets should obviously be docked but that is not the purpose of this PR).

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3122,7 +3122,7 @@ and fmt_label_declaration c ctx decl ?(last = false) =
     | `Before -> noop
     | `After -> fmt_if (not last) ";"
     | `After_and_docked ->
-        fmt_or_k last (fits_breaks ~level:7 "" ";") (str ";")
+        fmt_or_k last (fits_breaks ~level:5 "" ";") (str ";")
   in
   hovbox 0
     ( Cmts.fmt_before c pld_loc
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3116,6 +3116,13 @@ and fmt_label_declaration c ctx decl ?(last = false) =
     | `Loose -> true
     | `Tight_decl | `Tight -> false
   in
+  let fmt_semicolon =
+    match c.conf.break_separators with
+    | `Before -> noop
+    | `After -> fmt_if (not last) ";"
+    | `After_and_docked ->
+        fmt_or_k last (fits_breaks ~level:7 "" ";") (str ";")
+  in
   hovbox 0
     ( Cmts.fmt_before c pld_loc
     $ hvbox 4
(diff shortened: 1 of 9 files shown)


src/Fmt_ast.ml
```diff
@@ -3116,6 +3116,13 @@ and fmt_label_declaration c ctx decl ?(last = false) =
     | `Loose -> true
     | `Tight_decl | `Tight -> false
   in
+  let fmt_semicolon =
```
Julow: For readability, I suggest to move the formatting of the before case in the same pattern matching. Currently it is done in 2 different ways for records declarations and inline records.

gpetiot:
> This is great !
> But this changes the behavior of an option that should be used by a lot of projects, it is set in the `conventional` profile. Shouldn't this be a new option?

Yes but I think it's a positive change for this `break-separators=after-and-docked` option, and adding a new option that only does something when the `break-separators` has a specific value sounds like a bad idea.

samoht:
Yes please avoid new options :-) I think that's fine to just modify the current style as this could be considered as a "fix".

src/Fmt_ast.ml
gpetiot: I'm started doing that but it requires an extensive rewriting, if think it should be in another PR.

gpetiot: Having the semicolon in `fmt_label_declaration` makes possible to have the semicolon before the comment (the comment at the end of the line is attached to the type) :
```ocaml
-    margin: int;  (** Format code to fit within [margin] columns. *)
-    max_iters: int;
+    margin: int  (** Format code to fit within [margin] columns. *);
+    max_iters: int
         (** Fail if output of formatting does not stabilize within
-            [max_iters] iterations. *)
+            [max_iters] iterations. *);
     module_item_spacing: [`Compact | `Sparse];
-    ocp_indent_compat: bool;  (** Try to indent like ocp-indent *)
+    ocp_indent_compat: bool  (** Try to indent like ocp-indent *);
```

Julow: I agree. I was suggesting moving the semicolon logic inside `fmt_label_declaration`

test/passing/js_source.ml.ref
```diff
@@ -8076,7 +8076,11 @@ end =
 let _ = test 103 (Lazy.force M3.x) 3
 
 (** Pure type-checking tests: see recmod/*.ml  *)
-type t = A of { x : int; mutable y : int }
+type t =
+  | A of
+      { x : int
+      ; mutable y : int
+      }
```
gpetiot: This is a fix since `type-decl=sparse` in the JS profile.

src/Fmt_ast.ml
gpetiot: Now the whole semicolon logic is in fmt_label_declaration.

gpetiot:
I fixed more cases and also fixed the docked brackets, now Params is used in all cases.

grievejia:
As always, thanks for the effort!

PR #901, opened 2019-06-28
Ability to add the 'unrecognize extension' argument to the emacs hook

Zzull:
Some of our files are using unrecognized extensions and need to have the '--intf'  or '--impl' argument given to ocamlformat to be parsed successfully. It's a problem when using the `before-save-hook` of emacs which doesn't give you the choice to do that. So this is an attempt at addressing this issue.

### DIFF
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -68,6 +68,15 @@ a `before-save-hook'."
           (const :tag "None" nil))
   :group 'ocamlformat)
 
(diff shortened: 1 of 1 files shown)


Zzull:
Oulala what the hell. Can I just say that this code is yours and that I don't want any claim on it? Or can you pull and make it yours?

gpetiot:
Thanks! replaced by #905 to use a branch.

PR #916, opened 2019-07-03
Fix attribute when break-infix-before-func=false

### LINKED ISSUE #915: Bug with break-infix-before-func=false and warning annotation
With both OCamlFormat 0.10 and 5f99201, and the following configuration:

```
break-infix-before-func=false
```

Input file:

```ocaml
let _ =
  foo >>= function [@warning "-4"]
    | A -> false
    | B -> true
```

Output:

```
ocamlformat: Cannot process "bug.ml".
  BUG: ast changed.
```

It works if you remove either the infix operator or the annotation.

gpetiot:
Fix #915 
No diff with test_branch.

### DIFF
--- a/test/passing/infix_bind.ml
+++ b/test/passing/infix_bind.ml
@@ -181,3 +181,19 @@ let foo =
       let batch = {repo; tree= old_tree; origin} in
       let b = Batch batch in
       foo
+
+let _ = foo >>= function[@warning "-4"] A -> false | B -> true
+
+let _ =
+  foo >>= function[@warning "-4"]
+  | Afoooooooooooooooooo fooooooooo -> false
+  | Bfoooooooooooooooooooooo fooooooooo -> true
+
+let _ =
+  foo >>= fun [@warning "-4"] x ->
+  fooooooooooooooooooooooo
+
+let _ =
+  foo >>= fun [@warning "-4"] x y ->
+  fooooooooooooooooooooooo fooooooooooooooooooooooo fooooooooooooooooooooooo
+    fooooooooooooooooooooooo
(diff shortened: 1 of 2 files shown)


test/passing/infix_bind.ml
```diff
@@ -181,3 +181,28 @@ let foo =
...
+  foo >>= fun [@warning "-4"] x ->
+  fooooooooooooooooooooooo
+
+let _ =
+  foo >>= fun [@warning "-4"] x y ->
+  fooooooooooooooooooooooo fooooooooooooooooooooooo fooooooooooooooooooooooo
+    fooooooooooooooooooooooo
+
+let _ =
+  foo >>= function(* foo before *) [@warning "-4"] (* foo after *)
```
Julow: Shouldn't be a space after `function` ?

test/passing/infix_bind.ml
```diff
@@ -181,3 +181,28 @@ let foo =
...
+
+let _ = foo >>= function[@warning "-4"] A -> false | B -> true
+
+let _ =
+  foo >>= function[@warning "-4"]
+  | Afoooooooooooooooooo fooooooooo -> false
+  | Bfoooooooooooooooooooooo fooooooooo -> true
+
+let _ =
+  foo >>= fun [@warning "-4"] x ->
```
Julow: Is there a reason to add a space before the attribute on `fun` but not on `function` ?

test/passing/infix_bind.ml
gpetiot: It wouldn't hurt to have a space after the keyword, but there is also no space when `break-infix-before-func=true` so I was thinking at looking at that in another PR (didn't find a quick way to find that just now).

Issue #917, opened 2019-07-03
comments moving in ways that change their meaning


ceastlund:
I am using ocamlformat built from 2c8d4544d794b6ecd8645b11b3a071fd5f04cf83.

This program ...
```
let _ =
  try_with (fun () ->
    match get () with
    | None -> do_something ()
    | Some _ -> () (* do nothing *))
;;
```
... gets rewritten to this.
```
let _ =
  try_with (fun () ->
    (* do nothing *)
    match get () with
    | None -> do_something ()
    | Some _ -> ())
;;
```

This program ...
```
let _ =
  try_with (fun () ->
    a;
    b (* after a *))
;;
```
... gets rewritten to this.
```
let _ =
  try_with (fun () ->
    (* after a *)
    a;
    b)
;;
```

In both cases the comment becomes nonsensical.

ceastlund:
Another example...
```
let _ =
  match x with
  | Some y ->
    (match y with
     | None -> ()
     | Some z -> incr z (* double some *))
  | None -> ()
;;
```
... gets rewritten to:
```
let _ =
  match x with
  | Some y ->
    (match y with
     | None -> ()
     | Some z -> incr z)
  (* double some *)
  | None -> ()
;;
```

ceastlund:
I used ocamlformat built with ocaml 4.08.0+rc2.

PR #919, opened 2019-07-05
Fix position of comments after anonymous functions

gpetiot:
Improving the situation of #917 not sure we can call this a fix yet, but comment positionning has always been tricky, I have the feeling it will always be this way until comments are eventually added to the AST.

### DIFF
--- a/test/passing/comments.ml
+++ b/test/passing/comments.ml
@@ -179,3 +179,25 @@ inherit oo end
 
 class i = object (* test *)
 inherit oo end
+
+let _ =
+  try_with (fun () -> (* comment before *)
+    match get () with
+    | None -> do_something ()
+    | Some _ -> () (* do nothing *))
+;;
+
+let _ =
+  try_with (fun () -> (* comment before *)
+    a;
+    b (* after a *))
+;;
+
+let _ =
+  match x with
+  | Some y ->
+    (match y with
+     | None -> ()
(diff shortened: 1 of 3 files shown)


test/passing/comments.ml.ref
```diff
@@ -269,3 +269,21 @@ class i =
...
   end
+
+let _ =
+  try_with (fun () ->
+      (* comment before *)
+      match get () with None -> do_something () | Some _ -> ())
+
+(* do nothing *)
+
+let _ = try_with (fun () -> (* comment before *)
```
hhugo: We're probably missing a break here.

test/passing/comments.ml
```diff
@@ -179,3 +179,25 @@ inherit oo end
...
+  try_with (fun () -> (* comment before *)
+    match get () with
+    | None -> do_something ()
+    | Some _ -> () (* do nothing *))
+;;
+
+let _ =
+  try_with (fun () -> (* comment before *)
+    a;
+    b (* after a *))
```
hhugo: "after b" would be more accurate.

test/passing/comments.ml.ref
gpetiot: I'm having a look at this but this may take some time, maybe this PR could be approved first and we could fix this bug in another one?

PR #921, opened 2019-07-09
Fix comments' position

hhugo:
attempt to fix #917

PR #923, opened 2019-07-11
Fix comments' placement (do not look at loc_stack)

### LINKED ISSUE #917: comments moving in ways that change their meaning

gpetiot:
Fix #917, will be reverted with #921 

Diff of test_branch:
```
diff --git a/infer/src/base/ProcessPool.ml b/infer/src/base/ProcessPool.ml
index 988426188..b4b541ec0 100644
--- a/infer/src/base/ProcessPool.ml
+++ b/infer/src/base/ProcessPool.ml
@@ -292,7 +292,8 @@ let rec child_loop ~slot send_to_parent receive_from_parent ~f =
                 false )
               else (
                 (* crash hard, but first let the master know that we have crashed *)
-                send_to_parent (Crash slot) ; true )) ) ;
+                send_to_parent (Crash slot) ;
+                true )) ) ;
       child_loop ~slot send_to_parent receive_from_parent ~f
 
 
diff --git a/infer/src/biabduction/Abs.ml b/infer/src/biabduction/Abs.ml
index 71caec743..3bedc6414 100644
--- a/infer/src/biabduction/Abs.ml
+++ b/infer/src/biabduction/Abs.ml
@@ -481,8 +481,7 @@ let typ_get_recursive_flds tenv typ_exp =
           L.(debug Analysis Quiet)
             "@\ntyp_get_recursive_flds: unexpected %a unknown struct type: %a@." Exp.pp typ_exp
             Typ.Name.pp name ;
-          []
-      (* ToDo: assert false *) )
+          [] (* ToDo: assert false *) )
     | Tint _ | Tvoid | Tfun _ | Tptr _ | Tfloat _ | Tarray _ | TVar _ ->
         [] )
   | Exp.Var _ ->
diff --git a/infer/src/biabduction/Rearrange.ml b/infer/src/biabduction/Rearrange.ml
index aff0994dd..ed9729a30 100644
--- a/infer/src/biabduction/Rearrange.ml
+++ b/infer/src/biabduction/Rearrange.ml
@@ -981,7 +981,8 @@ let add_guarded_by_constraints tenv prop lexp pdesc =
         else if (* we don't know if we have the lock or not. *)
                 should_warn pdesc then (
           (* non-private method; can't ensure that the lock is held. warn. *)
-          warn accessed_fld guarded_by_str ; prop )
+          warn accessed_fld guarded_by_str ;
+          prop )
         else
           (* private method. add locked proof obligation to [pdesc] *)
           Attribute.add tenv ~footprint:true prop Alocked [guarded_by_exp]
diff --git a/infer/src/checkers/functionPointers.ml b/infer/src/checkers/functionPointers.ml
index e6e8e35b2..05ebde1b8 100644
--- a/infer/src/checkers/functionPointers.ml
+++ b/infer/src/checkers/functionPointers.ml
@@ -55,7 +55,7 @@ let find_procname var astate =
         Some procname
     | IContainer.More ->
         Some (ProcnameSet.min_elt procnames)
-    (* TODO: handle multiple procnames, e.g. with non-determinism branching *) )
+        (* TODO: handle multiple procnames, e.g. with non-determinism branching *) )
   | None ->
       None
 
diff --git a/compiler/lib/js_output.ml b/compiler/lib/js_output.ml
index bdb38fc8a..f96034bd3 100644
--- a/compiler/lib/js_output.ml
+++ b/compiler/lib/js_output.ml
@@ -990,7 +990,7 @@ struct
           last_semi ();
           PP.end_group f;
           PP.end_group f
-      (* There MUST be a space between the return and its
+          (* There MUST be a space between the return and its
        argument. A line return will not work *)
       )
```


test/passing/loc_stack.ml.ref
```diff
@@ -1,7 +1,6 @@
-let _ =
-  (* a *)
-  (* b *)
-  2
+let _ = 2 (* a *)
+
+(* b *)
```
Julow: This is wrong. The original code was:

```ocaml
let _ =
  ( (* a *) ( (* b *) 2 ))
```

test/passing/reformat_string.ml.ref
```diff
@@ -2,7 +2,7 @@ let _ = 'a'
 
 let _ = 'a'
 
-let _ = (* test *) "asd"
+let _ = "asd" (* test *)
```
Julow: Should be on the left, was:

```ocaml
let _ = ((* test *)"asd" )
```

test/passing/label_option_default_args.ml.ref
```diff
@@ -36,11 +36,7 @@ let (* 0 *) f
 let f ~x:({f; g} as x) = e
 
 let (* 0 *) f
-    ~x:(* 1 *)
-       (* 2 *)
-    ({f; g} (* 3 *) as (* 4 *) x)
-    (* 5 *)
-    (* 6 *) =
+    ~x:(* 1 *) ((* 2 *) {f; g} (* 3 *) as (* 4 *) x (* 5 *)) (* 6 *) =
```
Julow: `(* 1 *)` is still wrong, it should be before `~x`.

PR #929, opened 2019-07-19
Fix break-separators=after-and-docked for record patterns

### LINKED ISSUE #926: incoherent formatting for records
With ocamlformat 0.10, it seems that the records are not formatted the same way when they are pattern matched on or when they are created. Here are two examples. When pattern matching the first field is on the same line than `{`. When creating a new record, the first field is on a different line.

```ocaml
let f () =
  let { aaaaaaaa;
    bbbbbbbbbb;
      ccccccccc;
        dddddddddd;
        eeeeeeeee;
        ffffffffff;
        gggggggggg;
        hhhhhhhhhh;
        iiiiiiiiii;
        jjjjjjjjjjj
      }
    =
    some_value
  in
  {
    aaaaaaaa;
    bbbbbbbbbb;
    ccccccccc;
    dddddddddd;
    eeeeeeeee;
    ffffffffff;
    gggggggggg;
    hhhhhhhhhh;
    iiiiiiiiii;
    jjjjjjjjjjj;
  }

let g () =
  match some_value with
  | { aaaaaaaa;
    bbbbbbbbbb;
      ccccccccc;
      dddddddddd;
      eeeeeeeee;
      ffffffffff;
      gggggggggg;
      hhhhhhhhhh;
      iiiiiiiiii;
      jjjjjjjjjjj
    } ->
    {
      aaaaaaaa;
      bbbbbbbbbb;
      ccccccccc;
      dddddddddd;
      eeeeeeeee;
      ffffffffff;
      gggggggggg;
      hhhhhhhhhh;
      iiiiiiiiii;
      jjjjjjjjjjj;
    }
```

```
break-cases=toplevel
break-infix-before-func=false
break-infix=fit-or-vertical
break-separators=after-and-docked
break-sequences=true
break-struct=natural
cases-exp-indent=2
doc-comments=before
exp-grouping=preserve
field-space=loose
indicate-multiline-delimiters=closing-on-separate-line
let-and=sparse
m=120
max-indent=2
nested-match=align
ocp-indent-compat=true
parens-ite=true
parens-tuple=multi-line-only
sequence-style=terminator
space-around-arrays=true
space-around-lists=true
space-around-records=true
space-around-variants=true
type-decl=sparse
wrap-fun-args=true
```

gpetiot:
Fix #926, the diff of test_branch looks good.
@Khady feel free to give us feedback.

### DIFF
--- a/src/Params.ml
+++ b/src/Params.ml
@@ -144,6 +144,39 @@ let get_record_expr (c : Conf.t) =
       ; sep_after_non_final= fmt ";@;<1 0>"
       ; sep_after_final= fits_breaks ~level:1 "" ";" }
 
+type record_pat =
+  { box: Fmt.t -> Fmt.t
+  ; sep_before: Fmt.t
+  ; sep_after_non_final: Fmt.t
+  ; sep_after_final: Fmt.t
+  ; wildcard: Fmt.t }
+
+let get_record_pat (c : Conf.t) ~ctx =
+  let r = get_record_expr c in
+  let r =
+    { box= r.box
+    ; sep_before= r.sep_before
+    ; sep_after_non_final= r.sep_after_non_final
+    ; sep_after_final= r.sep_after_final
+    ; wildcard= r.sep_before $ str "_" $ r.sep_after_final }
+  in
+  match c.break_separators with
+  | `Before | `After -> r
+  | `After_and_docked ->
+      let space = if c.space_around_records then 1 else 0 in
+      let indent_opn, indent_cls =
+        match ctx with
+        | Ast.Exp {pexp_desc= Pexp_match _ | Pexp_try _; _} -> (-1, -1)
+        | Ast.Exp {pexp_desc= Pexp_let _; _} -> (-2, -2)
+        | _ -> (2, -2)
+      in
+      { r with
+        box=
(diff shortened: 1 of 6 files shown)


src/Params.ml
```diff
@@ -144,6 +144,39 @@ let get_record_expr (c : Conf.t) =
...
+  in
+  match c.break_separators with
+  | `Before | `After -> r
+  | `After_and_docked ->
+      let space = if c.space_around_records then 1 else 0 in
+      let indent_opn, indent_cls =
+        match ctx with
+        | Ast.Exp {pexp_desc= Pexp_match _ | Pexp_try _; _} -> (-1, -1)
+        | Ast.Exp {pexp_desc= Pexp_let _; _} -> (-2, -2)
+        | _ -> (2, -2)
```
Julow: Couldn't the calling code be changed so it's always `0, -2` and avoid looking at the context ?

Khady:
The diff of the tests also looks good to me. I don't have much time to test the change on our whole codebase, sorry about that.

src/Params.ml
gpetiot: I don't think we could avoid looking at the context since I want the indentation to depend on where the pattern is.

PR #930, opened 2019-07-19
Improve: warning 50 handled like an internal error

gpetiot:
From splitting #921

### DIFF
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -43,7 +43,8 @@ exception
     | `Ast_changed
     | `Doc_comment of Normalize.docstring_error list
     | `Comment
-    | `Comment_dropped of (Location.t * string) list ]
+    | `Comment_dropped of (Location.t * string) list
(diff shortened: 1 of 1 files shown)


src/Translation_unit.ml
```diff
@@ -337,13 +339,15 @@ let format xunit (conf : Conf.t) ?output_file ~input_name ~source ~parsed ()
...
+      | exception exn -> (
           let args =
             [("output file", dump_formatted ~suffix:".invalid-ast" fmted)]
             |> List.filter_map ~f:(fun (s, f_opt) ->
                    Option.map f_opt ~f:(fun f -> (s, String.sexp_of_t f)))
           in
-          internal_error (`Cannot_parse exn) args
+          match exn with
+          | Warning50 l -> internal_error (`Warning50 l) args
+          | exn -> internal_error (`Cannot_parse exn) args )
```
Julow: You can define `args` (as a function or lazy) outside the match and merge the matches.

gpetiot: Done.

PR #934, opened 2019-07-24
Improve: parse code in comments

### LINKED ISSUE #924: Formatting cinaps comments
The [cinaps](https://github.com/janestreet/cinaps) tool recognises special comments with OCaml code inside them. I wanted to check what the ocamlformat team thought of adding knowledge of the cinaps syntax in ocamlformat so that the code inside such comment could be automatically formatted?

More precisely, cinaps recognises the following two syntaxes:

- `(*$ <ocaml-code> *)`
- `(*$ <ocaml-code> $*)`

For instance:

```ocaml
(*$ for i = 1 to 3 do Printf.printf "let x%d = %d\n" i i; done *)
let x1 = 1
let x2 = 2
let x3 = 3
(*$*)
```

To give a bit more context, we use this tool at Jane Street as a principled way of generating and committing parts of ml files. Its use has grown quite a lot since it was introduced a couple of years ago. It's now integrated in Dune via the `cinaps` stanza and used in a couple of open source projects outside of Jane Street, though Idon't know how much cinaps is used in the wild at the moment.

gpetiot:
Fix #924
@diml I assumed every cinaps comment was an ocaml expression but it seems it's not always the case, I didn't see the specification on the github repo, is there more info somewhere?
Should it be allowed in .mli and .mlt files as well?


src/Normalize.ml
```diff
@@ -320,6 +320,8 @@ let mapper_ignore_doc_comment = make_mapper ~ignore_doc_comment:true
 
 let mapper = make_mapper ~ignore_doc_comment:false
 
+let expr c = Mapper.expression (mapper c)
```
gpetiot: Should not be needed afterall.

diml:
Thanks for working on this! cinaps comments are actually structure items, and yh they should be allowed everywhere.

gpetiot:
Turns out it is toplevel phrases instead, this PR should be ready to be reviewed and tested now.

ocamlformat-help.txt
```diff
@@ -331,6 +331,9 @@ OPTIONS (CODE FORMATTING STYLE)
            parentheses around tuples patterns. The default value is
            multi-line-only.
 
+       --parse-code-comments
+           Parse and format code in comments.
+
```
samoht: why adding yet-an-other option for this? Can't just it be the default while parsing comments?

gpetiot: I guess you are right, I will enable it by default.

diml:
Ah yes. Although toplevel phrases were for the old way of using cinaps, i.e. when using it as a toplevel. When used via dune, these have to be structure items. Though I guess parsing them as toplevel phrases should be harmless.

I will test that today.

Just one thought: I grepped the dune-universe for `(*$` and realised that it is used for qtest as well. Is that a concern?

gpetiot:
> Just one thought: I grepped the dune-universe for `(*$` and realised that it is used for qtest as well. Is that a concern?

I think testing whether there is a blank character after `(*$` should be enough.

diml:
I'm testing this, it seems to be working well except that a `;;` is systematically inserted when the code span over multiple lines:

```
(*$
  ;;
  let x = 1 in
  x + 1
*)
```

diml:
BTW, it's fine to restrict the parsing to structure items only. I expect that toplevel phrases support will eventually disappear given that the new method of using cinaps via dune is much more convenient.

gpetiot:
I reverted to using structures instead, so toplevel directives such as `#use "import.cinaps";;` won't be parsed anymore.

diml:
Thanks, I just gave it another try and the comments are being well formatted :)

Just one last issue I found: the trailing `$` in `'$*)` is systematically removed.

src/Cmts.ml
```diff
@@ -550,18 +552,41 @@ let fmt_cmt (conf : Conf.t) cmt =
...
+        let format = t.format in
+        try
+          let parsed =
+            Parse_with_comments.parse Migrate_ast.Parse.implementation conf
+              ~source
+          in
+          let source = Source.create source in
+          let cmts = init_impl ~format source parsed.ast parsed.comments in
+          let formatted = format source cmts conf parsed.ast in
+          hvbox 2 (wrap "(*$" "*)" (fmt "@;" $ formatted $ fmt "@;<1 -2>"))
```
gpetiot: Missing closing `$` here if it was in the source.

gpetiot: done

gpetiot:
> Just one last issue I found: the trailing `$` in `$*)` is systematically removed.

This is fixed now.

diml:
Just tested again, all good👌. Cinaps comment are all nicely formatted :)

gpetiot:
@Julow I guess I can submit another PR where I move `parse`to `Parse_with_comments.ml`, that should make this PR easier to read.

Julow:
> @Julow I guess I can submit another PR where I move `parse`to `Parse_with_comments.ml`, that should make this PR easier to read.

This should not be necessary.

src/Cmts.ml
```diff
@@ -550,18 +552,40 @@ let fmt_cmt (conf : Conf.t) cmt =
...
+      | [""] -> str "(* *)"
+      | [text] -> wrap "(*" "*)" (fill_text text)
+      | [text; ""] -> wrap "(*" " *)" (fill_text text)
+      | asterisk_prefixed_lines ->
+          fmt_asterisk_prefixed_lines asterisk_prefixed_lines
+  in
+  match fst cmt with
+  | "" | "$" -> fmt_non_code cmt
+  | str ->
+      if Char.equal str.[0] '$' then
```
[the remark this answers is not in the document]
gpetiot: Good idea, done.

PR #935, opened 2019-07-25
Improve: force break after an infix op only if followed by another one

gpetiot:
Forcing the break was a bit too aggressive and generates unnecessary diffs with 0.9
cc @samoht

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -1614,7 +1614,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                            4 (fmt_fun_args c xargs)
                        $ fmt "@ ->" ) )
                $ pre_body )
-           $ fmt "@;<1000 0>" $ body ))
+           $ fmt "@ " $ body ))
   | Pexp_apply
       ( ( { pexp_desc= Pexp_ident {txt= Lident id; loc}
           ; pexp_attributes= []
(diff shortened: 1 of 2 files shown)


test/passing/infix_bind.ml
```diff
@@ -23,36 +23,27 @@ f x
 ;;
 f x >>= fun y ->
 g y >>= fun () ->
-f x >>= fun y ->
-g y >>= fun () ->
-f x >>= fun y ->
-g y >>= fun () ->
-y ()
+f x >>= fun y -> g y >>= fun () -> f x >>= fun y -> g y >>= fun () -> y ()
```
Julow: This formatting is weird. They should all break.

gpetiot: I think if the chain is long, there is a good chance one of the bodies will be long enough to break the line.
So the cases we want to improve are indeed the smaller ones.

gpetiot: So do you think we should not merge this PR?

Julow: I find this case a bit problematic. What make it hard to fix ?

gpetiot: It should be better now.

src/Fmt_ast.ml
```diff
@@ -1598,6 +1598,17 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       let xargs, xbody = Sugar.fun_ c.cmts (sub_exp ~ctx r) in
       let indent_wrap = if parens then -2 else 0 in
       let pre_body, body = fmt_body c ?ext xbody in
+      let followed_by_infix_op =
+        match xbody.ast.pexp_desc with
+        | Pexp_apply
+            ( { pexp_desc= Pexp_ident {txt= Lident id; loc= _}
+              ; pexp_attributes= []
+              ; _ }
+            , [(Nolabel, _); (Nolabel, {pexp_desc= Pexp_fun _; _})] )
```
Julow: Missing `Pexp_function`.

gpetiot: Done

PR #937, opened 2019-07-26
Improve: support Shell-style regular expressions in .ocamlformat-ignore and .ocamlformat-enable files

gpetiot:


Julow:
This is just handling the literal `*` and not globbing (eg. `dir/*`). Can't the same effect be done with `disable=true` and `disable=false` ?

gpetiot:
It can but I think it may be less intuitive for some users (and the additional code is negligible)

Julow:
Users will expect full globbing support if they see this feature, IMO.

src/Conf.ml
```diff
@@ -1875,18 +1875,30 @@ let is_in_listing_file ~quiet ~listings ~filename =
...
-                        Format.eprintf "File %a, line %d:\nWarning: %s\n"
-                          Fpath.pp listing_file lno msg ;
-                      None))
+                match Fpath.of_string line with
+                | Ok file_on_current_line -> (
+                    let f = Fpath.(dir // file_on_current_line) in
+                    if Fpath.equal filename f then Some (listing_file, lno)
+                    else
+                      try
+                        let re = Re.Glob.glob ~pathname:true line in
```
Julow: You should set `~⁠anchored:true` otherwise `sr` will match `src/Abc.ml`.

src/Conf.ml
```diff
@@ -1875,18 +1875,30 @@ let is_in_listing_file ~quiet ~listings ~filename =
...
+                        let re = Re.compile re in
+                        let s = Fpath.to_string f in
+                        let (_ : Re.Group.t) = Re.exec re s in
```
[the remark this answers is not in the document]
gpetiot: Done

src/Conf.ml
gpetiot: Done, also I used the wrong variable for the filename (fixed as well)

PR #941, opened 2019-07-29
Improve: parse code in docstrings

### LINKED ISSUE #478: Formatting of code blocks inside documentation comments
It would be nice if code blocks inside documentation comments where formatted just like normal code so that they don't have to be formatted manually.

gpetiot:
Fix #478 
Made on top of #934 (hence the diff), re-using some bits of #721 (`fmt_code_block`)


gpetiot:
Rebased, now that we merged #934

src/Fmt_odoc.mli
```diff
@@ -9,7 +9,11 @@
  *                                                                    *
  **********************************************************************)
 
-val fmt : Octavius.Types.t -> Fmt.t
+val init :
+  (Source.t -> Cmts.t -> Conf.t -> Parsetree.structure -> Fmt.t) -> unit
+(** Initialize internal state *)
+
+val fmt : Conf.t -> Octavius.Types.t -> Fmt.t
```
Julow: Instead of passing `Conf.t` as argument and `format` as global state, I think it would be better to create a new type `type conf = { fmt_code : string -> Fmt.t }` that can be extended in the future and abstract a lot of things from this module (`Source`, `Cmts`. `Parsetree`, `Conf`).

gpetiot: Done in 3fe083f are you really sure it's better? if yes I will do the same for the ref in Normalize.ml

Julow: Yes it's definitely better. I had in mind a simpler `string -> Fmt.t` that abstract modules that `Fmt_odoc` don't need (`Source`, `Cmts`, `Conf`, the parsetree and the parsing code)

gpetiot: done

hhugo:
What's the status of the PR ?

gpetiot:
It is waiting for approval.

src/Sugar.mli
```diff
@@ -144,10 +140,10 @@ val polynewtype :
 (** [polynewtype cmts pat exp] returns expression of a type-constrained
     pattern [pat] with body [exp]. e.g.:
 
-    {[
-      let f: 'r 's. 'r 's t = fun (type r) -> fun (type s) -> (e : r s t)
-    ]}
+    {v
+let f: 'r 's. 'r 's t = fun (type r) -> fun (type s) -> (e : r s t)
```
hhugo: Why do we have this diff ?

hhugo:
Should we have more tests for this feature ?

src/Sugar.mli
gpetiot: Because this would be formatted as `let f : type r s. r s t = e`

gpetiot:
I added more tests (parsing, not parsing, handling ocamlformat attributes)

hhugo:
Is it expected to only format code in docstrings and not in other comments ?

src/Normalize.ml
```diff
@@ -65,67 +70,77 @@ let rec odoc_inline_element fmt = function
...
 
-let rec odoc_nestable_block_element fmt : nestable_block_element -> unit =
-  function
+let rec odoc_nestable_block_element c fmt = function
   | `Paragraph elms -> fpf fmt "Paragraph,%a" odoc_inline_elements elms
-  | `Code_block txt -> fpf fmt "Code_block,%a" str txt
+  | `Code_block txt ->
+      let txt =
+        try
+          Migrate_ast.Parse.implementation (Lexing.from_string txt)
```
hhugo: Should we look at comments here ?

src/Sugar.mli
hhugo: ok, Should we fix the indentation then ?

hhugo:
@gpetiot, FYI, I've pushed some refactoring to your branch directly

hhugo:
> Is it expected to only format code in docstrings and not in other comments ?

Replying to myself.
Formatting code in non-doc comments, if we want it, should be done as an other PR.

test/passing/doc_comments.mli
```diff
@@ -351,3 +351,82 @@ end
...
+
+]} *)
+
+
+
+(**
+  This is a comment with code inside
+  {[
+    (** This is a comment with code inside
+        {[ let code inside = f inside ]}
```
hhugo: I've added this example of nested comments. It seems that odoc is not happy about it.

hhugo: https://github.com/ocaml/odoc/issues/375

hhugo:
Formatting the following fails: 
```ocaml
(** This is a comment with code inside
    {[
    let code inside (* comment   *) = { a = [ ] }
    ]} *)
```

src/Sugar.mli
gpetiot: done

src/Fmt_odoc.mli
```diff
@@ -9,13 +9,14 @@
  *                                                                    *
  **********************************************************************)
 
-val fmt : Odoc_parser.Ast.docs -> Fmt.t
+val fmt :
+     Conf.t
```
Julow: This argument is only used to call `fmt_code` so it should be removed and passed to `fmt_code` by calling code.

src/Fmt_odoc.ml
```diff
@@ -63,23 +65,22 @@ let fmt_verbatim_block s =
...
-    let l = String.rstrip l in
-    if first then str l
-    else if String.length l = 0 then str "\n"
-    else fmt "@," $ str l
-  in
-  let lines = String.split_lines s in
-  let box = match lines with _ :: _ :: _ -> vbox 0 | _ -> hvbox 0 in
-  box (wrap "{[@;<1 2>" "@ ]}" (vbox 0 (list_fl lines fmt_line)))
+let fmt_code_block conf s =
+  try hvbox 0 (wrap "{[@;<1 2>" "@ ]}" (conf.fmt_code conf.conf s))
```
Julow: That `fmt_code` is expected to raise should be documented in the `mli`. Or a more explicit type should be used, eg. `(Fmt.t, unit) result`.

PR #943, opened 2019-07-29
Add support for passing arguments to ocamlformat in ocamlformat.el

sonologico:
This is done using the customizable variable ocamlformat-args, which accepts a list of strings to be passed as arguments to the command.

### DIFF
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -39,6 +39,11 @@
   :type 'string
   :group 'ocamlformat)
 
+(defcustom ocamlformat-args nil
+  "List of command line arguments to pass to ocamlformat."
+  :type '(repeat string)
+  :group 'ocamlformat)
+
 (defcustom ocamlformat-enable 'enable
   "Enable or disable ocamlformat."
   :type '(choice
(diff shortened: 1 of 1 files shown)


gpetiot:
I'm not sure it's a good idea to allow every argument to be passed to ocamlformat this way, any specific option you are thinking about?
I'm not sure I understand why you would like to parameterize ocamlformat using emacs instead of using .ocamlformat files.

sonologico:
I was using it to set a profile as I have been on working on ocaml files that do not belong to specific projects quite often. They had no .ocamlformat files around and I didn't want to have to create them.

Having an .ocamlformat file in my home directory wasn't working for me, but your comment prompted me to check the manual again. Adding --enable-outside-detected-project made it work and this is already available in ocamlformat.el, so I'm good without this pull request.

sonologico:
I would suggest adding to the manual that --enable-outside-detected-project is needed to make it work.
The following section is not really clear on this.

    (*) .ocamlformat files in current and all ancestor directories for each
    input file are used, as well as the global ocamlformat file defined in
    $XDG_CONFIG_HOME or in $HOME/.config if $XDG_CONFIG_HOME is undefined.
    The global ocamlformat file has the lowest priority, then the closer
    the directory is to the processed file, the higher the priority.

gpetiot:
You are right, I opened #948

PR #945, opened 2019-07-30
Improve: indent attributes more

### LINKED ISSUE #940: Not so nice @@deriving indentation
Actual:
```ocaml
type t =
  | A of a
  | B of b
[@@deriving sexp]
```
Expected:
```ocaml
type t =
  | A of a
  | B of b
  [@@deriving sexp]
```

gpetiot:
Fix #940 
I agree it looks more readable this way.


gpetiot:
Do you think it is worth adding an option? (I'm not really sure given out new policy of NOT adding new options, and reducing the number of options instead :D)

PR #947, opened 2019-07-30
Improve: new 'before' value for option 'sequence-style'

### LINKED ISSUE #938: break before sequences
Is it possible to format
```ocaml
let _ =
  do_ ;
  job_1 ;
  job_2 ;
  return ()
```
as
```ocaml
let _ = do_
  ; job_1
  ; job_2
  ; return ()
```
Can `ocamlformat` have such option like `sequences-style=[before|after]` or even `infix-style=[before|after]`?

gpetiot:
Fix #938 
The diff of test_branch when setting the new value looks good.

No diff with test_branch with default values by the way.


src/Fmt_ast.ml
```diff
@@ -2261,56 +2314,20 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+          [ { pstr_desc=
+                Pstr_eval
+                  ( ( {pexp_desc= Pexp_sequence _; pexp_attributes= []; _}
+                    as e1 )
+                  , _ )
+            ; pstr_loc= _ } ] )
+    when List.is_empty pexp_attributes
+         && ( Poly.(c.conf.extension_sugar = `Always)
+            || Source.extension_using_sugar ~name:ext ~payload:e1 ) ->
+      fmt_sequence c parens width xexp pexp_loc fmt_atrs
```
Julow: Can this function still be called on sequences with `ext` set ? I could not find an example but this could cause a bug in the future. Getting rid of the `ext` mechanism seems hard so is it possible to continue to use `ext` here instead ?

src/Conf.ml
```diff
@@ -24,6 +24,7 @@ type t =
   ; break_fun_decl: [`Wrap | `Fit_or_vertical | `Smart]
   ; break_fun_sig: [`Wrap | `Fit_or_vertical | `Smart]
   ; break_separators: [`Before | `After | `After_and_docked]
+  ; break_sequence_separators: [`Before | `After]
```
Julow: Instead of adding an option, what do you think of merging this with `sequence_style` ?
It is not used in the `before` case and is used only here.

gpetiot: Good idea, done.

src/Fmt_ast.ml
gpetiot: It could not, but I modified the code to pass this argument and added an assertion checking that it is equal to the one extracted from Sugar.

src/Fmt_ast.ml
```diff
@@ -1449,43 +1506,6 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
-               (wrap "[" "]"
-                  ( str "%"
-                  $ hovbox 2
-                      ( fmt_str_loc c name $ str " fun "
-                      $ fmt_attributes c ~suf:(str " ") call.pexp_attributes
-                          ~key:"@"
-                      $ fmt_fun_args c xargs $ fmt "@ ->" )
-                  $ fmt "@ " $ fmt_expression c xbody ))
-           $ fmt "@ ;@ "
-           $ list grps " ;@;<1000 0>" fmt_grp ))
```
gpetiot: I removed this special case, I don't see what made it necessary, before we had:
```ocaml
let _ =
  [%foooo fun foooooooooooooooooooo fooooooooooooooo fooo -> fooooooooo]
  ;
  foooooooooooooooooooooooo
```
now we have:
```ocaml
let _ =                
  [%foooo fun foooooooooooooooooooo fooooooooooooooo fooo -> fooooooooo] ;
  foooooooooooooooooooooooo
```
and it benefits from the changes of this PR.

Julow: Looks good.

src/Fmt_ast.ml
```diff
@@ -1305,6 +1305,63 @@ and fmt_args ~first:first_grp ~last:last_grp c ctx args =
...
+            (str " ")
+    | `Separator -> str " ;" $ fmt_extension_suffix c ext $ break
+    | `Terminator -> str ";" $ fmt_extension_suffix c ext $ break
+  in
+  let is_simple x = is_simple c.conf width x in
+  let break (_, xexp1) (_, xexp2) =
+    not (is_simple xexp1 && is_simple xexp2)
+  in
+  let elts = Sugar.sequence c.conf c.cmts xexp in
+  let first_ext =
```
Julow: That variable is not needed after the assert so you could move the assert inside the match and transform that into an `()`.

PR #948, opened 2019-07-30
Doc: enable-outside-detected-project necessary for global conf file

gpetiot:
As suggested in #943


src/Conf.ml
```diff
@@ -118,7 +118,8 @@ let info =
          $(b,\\$HOME/.config) if $(b,\\$XDG_CONFIG_HOME) is undefined. The \
          global $(b,ocamlformat) file has the lowest priority, then the \
          closer the directory is to the processed file, the higher the \
-         priority."
+         priority. The global $(b,ocamlformat) file is only used when the \
```
Julow: You could just say "It".

gpetiot: I think it is clearer this way, sinceit is a long sentence and the previous "global ocamformat file" is really early in the sentence.

PR #950, opened 2019-07-31
Project: create issue templates

gpetiot:
(created with the github web interface)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -10,11 +10,10 @@ assignees: ''
 **Is your feature request related to a problem? Please describe.**
 A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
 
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
```
Julow: I think users will not necessarily keep track of what they tried or may not want to spend the time writing it down.
We could ask for that in the first point ? so users will not feel forced to write that.

gpetiot: done

PR #951, opened 2019-08-01
Improve: tidying up Conf.ml

gpetiot:
Just some cleaning and an optimization

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1746,8 +1746,9 @@ let dot_ocamlformat_enable = ".ocamlformat-enable"
 let rec collect_files ~segs ~ignores ~enables ~files =
   match segs with
   | [] | [""] -> (ignores, enables, files, None)
-  | "" :: segs -> collect_files ~segs ~ignores ~enables ~files
-  | _ :: segs ->
+  | "" :: upper_segs ->
+      collect_files ~segs:upper_segs ~ignores ~enables ~files
+  | _ :: upper_segs ->
       let sep = Fpath.dir_sep in
       let dir = String.concat ~sep (List.rev segs) |> Fpath.v in
       let ignores =
@@ -1767,7 +1768,7 @@ let rec collect_files ~segs ~ignores ~enables ~files =
       in
       if is_project_root dir && not enable_outside_detected_project then
         (ignores, enables, files, Some dir)
-      else collect_files ~segs ~ignores ~enables ~files
+      else collect_files ~segs:upper_segs ~ignores ~enables ~files
 
 let read_config_file conf filename_kind =
   match filename_kind with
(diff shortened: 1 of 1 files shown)


src/Conf.ml
```diff
@@ -1766,9 +1760,11 @@ let rec collect_files ~segs ~ignores ~enables ~files =
         if Fpath.exists filename then filename :: enables else enables
       in
       let files =
-        let filename = Fpath.(dir / dot_ocp_indent) in
-        if Fpath.exists filename then `Ocp_indent filename :: files
-        else files
+        let f_1 = Fpath.(dir / dot_ocamlformat) in
+        let f_2 = Fpath.(dir / dot_ocp_indent) in
+        files
+        |> (fun k -> if Fpath.exists f_1 then `Ocamlformat f_1 :: k else k)
```
Julow: I think ``let files = if Fpath.exists f_1 then `Ocamlformat f_1 :: files else files in`` would be a little more readable. That way `f_1` and `f_2` could be defined just before they are used.

src/Conf.ml
```diff
@@ -1839,14 +1834,13 @@ type action =
...
-  | Some kind -> kind
-  | None -> (
-    match Filename.extension fname with
-    | ".ml" -> `Impl
-    | ".mli" -> `Intf
-    | ".mlt" -> `Use_file
-    | _ -> `Impl )
+  Option.value !kind
+    ~default:
+      ( match Filename.extension fname with
```
Julow: This was computed only when needed so this is slightly less good.

src/Conf.ml
```diff
@@ -1872,9 +1866,9 @@ let is_in_listing_file ~quiet ~listings ~filename =
       try
         In_channel.with_file (Fpath.to_string listing_file) ~f:(fun ch ->
             let lines =
-              In_channel.input_lines ch
-              |> List.mapi ~f:(fun i s -> (i + 1, String.strip s))
-              |> List.filter ~f:(fun (_, l) -> not (drop_line l))
+              List.filter_mapi (In_channel.input_lines ch) ~f:(fun i s ->
+                  let l = String.strip s in
+                  Option.some_if (not (drop_line l)) (i + 1, l))
```
Julow: `Option.some_if` is only 6 characters shorter than an `if then Some ... else None` so I'm not sure it's worth the complexity. We are using it a lot in OCamlformat so it's OK.

src/Conf.ml
```diff
@@ -1933,39 +1923,33 @@ let build_config ~file =
...
     List.for_all files ~f
   in
+  let quiet = conf.quiet in
   if no_ocamlformat_files && not enable_outside_detected_project then (
     ( if not conf.quiet then
-      let reason =
-        match project_root with
-        | Some root ->
+      let why =
+        Option.value_map project_root ~default:"no project root was found"
```
Julow: I think this is adding a lot of complexity and is only a single line shorter

src/Conf.ml

src/Conf.ml

src/Conf.ml

src/Conf.ml

PR #952, opened 2019-08-01
Improve: generalize API of Config_option

gpetiot:
Improving Config_option API.


gpetiot:
> Other decls are still doing the parsing themselves (`int`, `choice`, `flag`). Would it make sense to implement a `required : 'a Arg.conv -> ... -> 'a option_decl` that is implemented the same as `opt` ?

`any` and `choice` are now factorized (and `opt` is removed), but it's more complicated for `flag`, I think it's better to keep it like that for now.

Issue #483, opened 2018-11-05
Closing `*)` of comments does not respect margin


jberdine:
With the margin set to 20, we see the following output:
```
(* 00000011111111112
 * 45678901234567890
 *)

(* 1. xxxx x xxxx *)
(* 2. xxxx xx xxxx *)
(* 3. xxxx xxx xxxx *)
(* 4. xxxx xxxx
   xxxx *)
```
Note that cases 1. and 4. are handled correctly, but cases 2. and 3. do not respect the margin and should break.

gpetiot:
I have no idea what's wrong here, the margin is correctly set, and each word of the comment is correctly separated by `@ `, so the line should be split before the last `xxxx` word.
Setting a different margin for the formatter on-the-fly messes up the whole output (and flushing or not does not help).

jberdine:
Yes, changing the margin on-the-fly is a no-go, `Format` has way too much internal state to make that close to feasible.

I suspect that this issue is due to how the comments are wrapped and formatted, and then in `Cmts.fmt_cmts` we add the ending `*)` without a break and more importantly without having earlier taken its length into account.

gpetiot:
Related problem: let's say the margin is limited to 80 characters, it currently only allows to print 79 characters (the 80th being `\n`), is this intended or do we want to write `set_margin (n + 1)` to really be able to write 80 characters and then the `\n` ?

jberdine:
Do you mean that e.g.:
```
(* 5. xxxx xxxx xxxx
   xxxx xxxx *)
```
should be preserved, while currently it formats to:
```
(* 5. xxxx xxxx
   xxxx xxxx xxxx *)
```
That does look like a related issue.

This is limited to comments as far as I'm aware, or have you seen other cases?

jberdine:
I see, thanks. Some lines do reach the margin, but perhaps that is due to other bugs rather than the newline not being included in the line length, I'm not sure. There is another possible explanation, which involves the Format module's max_indent value. AFAIU, if a box is opened after this indentation, the line is broken, even if the opened box ends up being empty. It does not seem possible to set max_indent to the margin, so ocamlformat sets it to one less. It could be that we bump into that.

gpetiot:
I know it's not high priority but I'm looking into this issue from time to time. Do you think it could be an issue from the vendored Format we are using? It could be a bug introduced with the fork or a bug in the original code but I didn't find any related bug in the ocaml mantis and I think someone would have noticed.

[...]

PR #957, opened 2019-08-07
Fix the margin, and correctly breaks comments

gpetiot:
Fix the first part of #483 

I checked the length for the files in `src/*`, but not yet in `test/passing/*`


src/Fmt.ml
```diff
@@ -238,14 +235,14 @@ let fill_text text =
...
+  $ hovbox 0
+      ( list_pn lines (fun ?prev:_ curr ?next ->
+            fmt_line curr
+            $
+            match next with
+            | Some str when String.for_all str ~f:Char.is_whitespace ->
+                close_box $ fmt "\n@," $ open_hovbox 0
+            | Some _ when not (String.is_empty curr) -> fmt "@ "
+            | _ -> noop)
+      $ fmt_if (Char.is_whitespace text.[String.length text - 1]) " "
```
Julow: Shouldn't this be `@ ` ?

test/passing/compact-lists-arrays.ml
```diff
@@ -18,21 +18,20 @@ let length =
...
 let length =
-  [ 0; 1; 2; 3; 4; 5; 6; 7; 8; 8; 9; 9; 10; 10; 11; 11; 12; 12; 12; 12; 13
-  ; 13; 13; 13; 14; 14; 14; 14; 15; 15; 15; 15; 16; 16; 16; 16; 16; 16; 16
-  ; 16; 17; 17; 17; 17; 17; 17; 17; 17; 18; 18; 18; 18; 18; 18; 18; 18; 19
-  ; 19; 19; 19; 19; 19; 19; 19; 20; 20; 20; 20; 20; 20; 20; 20; 20; 20; 20
-  ; 20; 20; 20; 20; 20; 21; 21; 21; 21; 21; 21; 21; 21; 21; 21; 21; 21; 21
-  ; 21; 21; 21; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22; 22
-  ; 22; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 23; 24
+  [ 0; 1; 2; 3; 4; 5; 6; 7; 8; 8; 9; 9; 10; 10; 11; 11; 12; 12; 12; 12; 13; 13
+  ; 13; 13; 14; 14; 14; 14; 15; 15; 15; 15; 16; 16; 16; 16; 16; 16; 16; 16
```
Julow: `; 17` could fit at the end of this line.

test/passing/source.ml.ref
```diff
@@ -1499,12 +1493,12 @@ let ty_list : type a e. (a, e) ty -> (a vlist, e) ty =
 (* type (_,_) ty_assoc = | Anil : (unit,'e) ty_assoc | Acons : string *
    ('a,'e) ty * ('b,'e) ty_assoc -> ('a -> 'b, 'e) ty_assoc
 
-   and (_,_) ty_pvar = | Pnil : ('a,'e) ty_pvar | Pconst : 't * ('b,'e)
-   ty_pvar -> ('t -> 'b, 'e) ty_pvar | Parg : 't * ('a,'e) ty * ('b,'e)
-   ty_pvar -> ('t * 'a -> 'b, 'e) ty_pvar *)
+and (_,_) ty_pvar = | Pnil : ('a,'e) ty_pvar | Pconst : 't * ('b,'e) ty_pvar
+-> ('t -> 'b, 'e) ty_pvar | Parg : 't * ('a,'e) ty * ('b,'e) ty_pvar -> ('t *
+'a -> 'b, 'e) ty_pvar *)
```
Julow: Leading spaces are lost after an empty line.

src/Fmt.ml
gpetiot: No because we don't want to break just before `*)`

test/passing/source.ml.ref
gpetiot: Fixed.

src/Fmt.ml
```diff
@@ -238,14 +235,16 @@ let fill_text text =
...
-              $
-              match next with
-              | Some str when String.for_all str ~f:Char.is_whitespace ->
-                  close_box $ fmt "\n@," $ open_hovbox 0
-              | Some _ when not (String.is_empty curr) -> fmt "@ "
-              | _ -> noop)))
-  $ fmt_if (Char.is_whitespace text.[String.length text - 1]) " "
+  $ hovbox 0
+      ( hvbox 0
+          (hovbox 0
```
Julow: Is this box necessary ?

test/passing/source.ml.ref
```diff
@@ -4814,9 +4800,7 @@ module A_alias_expanded = struct
   include A_alias
 end
 
-let _ = f (module A_alias_expanded : S with type t = (module A.A_S))
-
-(* ok *)
+let _ = f (module A_alias_expanded : S with type t = (module A.A_S)) (* ok *)
```
Julow: Is this intended ? (Comment was on its line, separated by an empty line)

gpetiot: In source.ml we have `let _ = f (module A_alias_expanded : S with type t = (module A.A_S)) (* ok *)` is also on the same line so this is a fix.

src/Fmt.ml
gpetiot: Yes these boxes are necessary, the output is slightly worse if we remove the hvbox (since the inner hovbox is sometimes closed and re-opened, it is not a single hovbox)

test/passing/compact-lists-arrays.ml
gpetiot: So far I have no idea what is causing this, it looks like a Format issue.

gpetiot:
It was supposed to fix the first part of #483 (the comments) but it seems it doesn't.

Issue #961, opened 2019-08-08
Odd behaviour when mixing infix operators with comments


CraigFe:
OCamlformat currently converts the following input
```ocaml
let () =
  (* Open the repo *)
  initialise >>=
  (* Perform a subsequent action *)
  subsequent_action >|=
  (* Keep going... *)
  another_action |> fun t ->
  (* And finally do this *)
  final_action t
```
to this output:
```ocaml
let () =
  (* Open the repo *)
  initialise >>= (* Perform a subsequent action *)
                 subsequent_action >|= (* Keep going... *)
                                       another_action
  |> fun t ->
  (* And finally do this *)
  final_action t
```
which looks odd to me.

gpetiot:
Do you have some feedback on how #986 looks on your code?

CraigFe:
The behaviour on the code snippet above is now:
```ocaml
let () =
  (* Open the repo *)
  initialise
  >>= (* Perform a subsequent action *)
  subsequent_action
  >|= (* Keep going... *)
  another_action
  |> fun t ->
  (* And finally do this *)
  final_action t
```

As a user, I see this as OCamlformat being opinionated about where I should put my binds (which I don't particularly mind, save for some concerns about compatibility with `let+` syntax). If I listen to OCamlformat and move my comments to be before the infix operators, I would give this as input:
```ocaml
let () =
  (* Open the repo *)
  initialise
  (* Perform a subsequent action *)
  >>= subsequent_action
  (* Keep going... *)
  >|= another_action
  (* And finally do this *)
  |> fun t -> final_action t
```
Which unfortunately becomes this:
```ocaml
let () =
  (* Open the repo *)
  initialise (* Perform a subsequent action *) >>= subsequent_action
  (* Keep going... *)
  >|= another_action
  (* And finally do this *)
  |> fun t -> final_action t
```
If OCamlformat is to force bind operators to the start of lines, this seems like a very unpleasant interaction with comments.

Julow:
The incorrect formatting is fixed in https://github.com/ocaml-ppx/ocamlformat/pull/986. Should we close this (by merging) and open new issues for the formatting and for the placement of operators ?

gpetiot:
I improved the case when the comment is before the operator.

CraigFe:
The new behaviour seems good to me; thanks :slightly_smiling_face:

Julow:
[...]

PR #970, opened 2019-08-15
Fix position of expressions regarding of comments in infix-op expressions

gpetiot:
Improve this part of the code before working on #961

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -50,14 +50,6 @@ let is_kwdopchar = function
       true
   | _ -> false
 
-let is_monadic_last_char = function '=' | '|' | '>' -> true | _ -> false
-
-let is_monadic_id s =
-  String.length s >= 2
-  && String.for_all s ~f:is_kwdopchar
-  && is_monadic_last_char s.[String.length s - 1]
-  && not (List.mem ~equal:String.equal [":="; "||"] s)
-
 let is_monadic_binding_id s =
   String.length s > 3
   && (String.is_prefix s ~prefix:"let" || String.is_prefix s ~prefix:"and")
@@ -2437,3 +2429,6 @@ end
 
 include In_ctx
 include Requires_sub_terms
+
+let is_monadic_expr e =
+  match prec_ast (Exp e) with Some InfixOp0 -> true | _ -> false
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -50,6 +50,13 @@ let is_kwdopchar = function
       true
   | _ -> false
 
+let is_monadic_last_char = function '=' | '|' -> true | _ -> false
+
+let is_monadic_id s =
+  String.length s > 2
+  && String.for_all s ~f:is_kwdopchar
+  && is_monadic_last_char s.[String.length s - 1]
+
 let is_monadic_binding_id s =
   String.length s > 3
   && (String.is_prefix s ~prefix:"let" || String.is_prefix s ~prefix:"and")
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -50,12 +50,13 @@ let is_kwdopchar = function
       true
   | _ -> false
 
-let is_monadic_last_char = function '=' | '|' -> true | _ -> false
+let is_monadic_last_char = function '=' | '|' | '>' -> true | _ -> false
 
 let is_monadic_id s =
-  String.length s > 2
+  String.length s >= 2
   && String.for_all s ~f:is_kwdopchar
   && is_monadic_last_char s.[String.length s - 1]
+  && not (List.mem ~equal:String.equal [":="; "||"] s)
 
 let is_monadic_binding_id s =
   String.length s > 3
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -861,6 +861,8 @@ and Requires_sub_terms : sig
   val parenze_pat : pattern In_ctx.xt -> bool
 
   val parenze_exp : expression In_ctx.xt -> bool
+
+  val parenze_nested_exp : expression In_ctx.xt -> bool
 end = struct
   open In_ctx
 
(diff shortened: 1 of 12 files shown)


test/passing/apply.ml
```diff
@@ -51,7 +51,7 @@ let whatever_labelled a_function_name long_list_one some_other_thing =
 
 let cartesian_product' long_list_one long_list_two =
   List.concat
-    (long_list_one
+    ( long_list_one
```
gpetiot: The space should have been there since `indicate_multiline_delimiters = closing-on-separate-line`.

gpetiot:
55f3f32 should fix #961

CraigFe:
Hmm... I'm getting this behaviour now:
```
let () =
  (* Open the repo *)
  initialise
  >>= (* Perform a subsequent action *)
      subsequent_action
  >|= (* Keep going... *)
      another_action
  |> fun t ->
  (* And finally do this *)
  final_action t
```

Which is perhaps not ideal either? The indentation on the line after the infix operators looks off to me.

gpetiot:
I think it is fixed and ready to be reviewed now, here is the diff with test_branch:
```ocaml
diff --git a/infer/src/biabduction/interproc.ml b/infer/src/biabduction/interproc.ml
index 0e2dff723..2d6baa9db 100644
--- a/infer/src/biabduction/interproc.ml
+++ b/infer/src/biabduction/interproc.ml
@@ -707,8 +707,9 @@ let prop_init_formals_seed tenv new_formals (prop : 'a Prop.t) : Prop.exposed Pr
     List.map ~f:do_formal new_formals
   in
   let sigma_seed =
-    create_seed_vars ((* formals already there plus new ones *)
-                      prop.Prop.sigma @ sigma_new_formals)
+    create_seed_vars
+      ( (* formals already there plus new ones *)
+        prop.Prop.sigma @ sigma_new_formals )
   in
   let sigma = sigma_seed @ sigma_new_formals in
   let new_pi = prop.Prop.pi in
diff --git a/infer/src/clang/ClangWrapper.ml b/infer/src/clang/ClangWrapper.ml
index 450b693b3..234ccdcbd 100644
--- a/infer/src/clang/ClangWrapper.ml
+++ b/infer/src/clang/ClangWrapper.ml
@@ -89,8 +89,9 @@ let clang_driver_action_items : ClangCommand.t -> action_item list =
       CanonicalCommand
         ( (* massage line to remove edge-cases for splitting *)
         match
-          "\"" ^ line ^ " \"" |> (* split by whitespace *)
-                                 Str.split (Str.regexp_string "\" \"")
+          "\"" ^ line ^ " \""
+          |> (* split by whitespace *)
+             Str.split (Str.regexp_string "\" \"")
         with
         | prog :: args ->
             ClangCommand.mk ~is_driver:false ClangQuotes.EscapedDoubleQuotes ~prog ~args
diff --git a/compiler/lib/generate.ml b/compiler/lib/generate.ml
index 728cf7e7b..22c886acf 100644
--- a/compiler/lib/generate.ml
+++ b/compiler/lib/generate.ml
@@ -1835,9 +1835,9 @@ let generate_shared_value ctx =
             | None -> []
             | Some v ->
                 [J.V v, Some (J.EDot (s_var Constant.global_object, "jsoo_runtime"), J.N)])
-          @ List.map
-              (StringMap.bindings ctx.Ctx.share.Share.vars.Share.strings)
-              ~f:(fun (s, v) -> v, Some (str_js s, J.N))
+           @ List.map
+               (StringMap.bindings ctx.Ctx.share.Share.vars.Share.strings)
+               ~f:(fun (s, v) -> v, Some (str_js s, J.N))
            @ List.map
                (StringMap.bindings ctx.Ctx.share.Share.vars.Share.prims)
                ~f:(fun (s, v) -> v, Some (runtime_fun ctx s, J.N))))
```

src/Ast.ml
```diff
@@ -50,6 +50,13 @@ let is_kwdopchar = function
       true
   | _ -> false
 
+let is_monadic_last_char = function '=' | '|' -> true | _ -> false
```
CraigFe: I'm not familiar with the existing code but this makes me uneasy; what about `|>`?

gpetiot: at first I wanted to avoid doing it for `|>` because it makes a long diff but it is more consistent indeed, fixed with my last commit.

src/Ast.ml
```diff
@@ -50,6 +50,14 @@ let is_kwdopchar = function
       true
   | _ -> false
 
+let is_monadic_last_char = function '=' | '|' | '>' -> true | _ -> false
+
+let is_monadic_id s =
```
Julow: I don't think we should use the operators characters to guess that. This is accepting operators like `||` and `:=` and it's rejecting a lot of potential monadic operators.

src/Ast.ml
```diff
@@ -2393,7 +2395,40 @@ end = struct
...
+        false
+    | None, _ | _, None -> false
+    | Some p1, Some p2 -> Poly.(p1 <> p2)
 end
 
 include In_ctx
 include Requires_sub_terms
+
+let is_monadic_expr e =
+  match prec_ast (Exp e) with Some InfixOp0 -> true | _ -> false
```
Julow: This causes a break just after `|>` operators 
```diff
-      |> List.sort
-           ~compare:(Comparable.lift Location.compare_start ~f:Cmt.loc)
+      |>
+      List.sort ~compare:(Comparable.lift Location.compare_start ~f:Cmt.loc)
 ```

Comment (gpetiot) [reply src/Ast.ml]: Yes, that is the new behavior for all these operators, that's why I was filtering a bit (size > 2, checking the last character), so should `|>` be treated like the rest or not?

Comment (CraigFe) [reply src/Ast.ml]: FWIW, I also think this break looks odd. My toy example was using infix operators at the end of each line, a la:
```ocaml
(* Open the repo *)
initialise >>=
(* Perform a subsequent action *)
subsequent_action >|=
```
but when I'm writing this sort of code personally, I keep the infix operators at the start:
```ocaml
x
>>= foo
>>= bar
```
(or the same with `|>`). I don't really mind which of the two OCamlformat chooses to pick (this is probably a reasonably important decision), but I do care about its interaction with comments. i.e., this looks wrong to me:
```ocaml
initialise
>>= (* Perform a subsequent action *)
    subsequent_action 
>|= (* Keep going... *)
    another_action
```
And if OCamlformat chooses to enforce one or the other, there is the question of what to do with input flies that are doing it the other way (comments before/after the infix operator).

CraigFe: (this may be an instance where it's worth having an option to select between prepending or appending infix operators)

gpetiot: Yes that would be good to have that but it requires a bigger rewrite of the code, the sequences of infix operations are currently sugared into a weird structure that mix different operators (as long as they have the same precedence, e.g. `>>=` and `>|=`), so it's not as simple as it was for sequences of items separated by `;` only.

Julow:
I don't mind the indentation in
```ocaml
initialise
>>= (* Perform a subsequent action *)
    subsequent_action 
>|= (* Keep going... *)
    another_action
```

Multiline expressions will render the same and here the first comment is part of the `subsequent_action` expression.

I don't think it's worth adding a special case that move the operator at the end of the line. It will probably be considered as a bug by users. Comments could be moved before instead:

```ocaml
initialise
(* Perform a subsequent action *)
>>= subsequent_action 
(* Keep going... *)
>|= another_action
```

(which does not work currently: if the whole expression fit, the comment will not be on its line)

gpetiot:
I think I will merge after https://github.com/ocaml-ppx/ocamlformat/pull/970/commits/bf9b0465b20f5b738446dc15f73fa2b69e09cc34 because this is clearly an improvement, but keep the issue open, so we can maybe improve it again later.

gpetiot:
Replaced by #986

PR #974, opened 2019-08-20
API: change the type of fmt-code

gpetiot:
Follow-up on #941 
Change the signature of the `fmt_code` function

from `Conf.t -> string -> Fmt.t`

to `string -> (Fmt.t, unit) Result.t`.

### DIFF
--- a/src/Cmts.mli
+++ b/src/Cmts.mli
@@ -56,7 +56,7 @@ val relocate :
 val fmt_before :
      t
   -> Conf.t
-  -> fmt_code:(string -> Fmt.t)
+  -> fmt_code:(string -> (Fmt.t, unit) Result.t)
   -> ?pro:Fmt.t
   -> ?epi:Fmt.t
   -> ?eol:Fmt.t
@@ -69,7 +69,7 @@ val fmt_before :
 val fmt_after :
      t
   -> Conf.t
-  -> fmt_code:(string -> Fmt.t)
+  -> fmt_code:(string -> (Fmt.t, unit) Result.t)
   -> ?pro:Fmt.t
   -> ?epi:Fmt.t
   -> Location.t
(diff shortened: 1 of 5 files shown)


src/Cmts.mli
```diff
@@ -56,7 +56,7 @@ val relocate :
 val fmt_before :
      t
   -> Conf.t
-  -> fmt_code:(Conf.t -> string -> Fmt.t)
+  -> fmt_code:(string -> (Fmt.t, unit) Result.t)
```
Julow: `Cmts.ml` have the `Conf.t` already so it makes sense to pass it to `fmt_code` here, to be sure the conf is uptodate and avoid having 2 ways to pass the same thing.

gpetiot: Yes but I find this more consistent with Fmt_odoc. And it also avoids the conf to be modified somewhere in Cmts. (currently it is just passed around each function, but better be safe, it shouldn't be modified since the call to `Cmts.fmt_*` from Fmt_ast).

src/Fmt_ast.ml
```diff
@@ -4417,14 +4417,16 @@ let fmt_file ~ctx ~f ~fmt_code source cmts conf itms =
...
+  try
+    let ({ast; comments; _} : _ Parse_with_comments.with_comments) =
+      Parse_with_comments.parse Migrate_ast.Parse.implementation conf
+        ~source:s
+    in
+    let source = Source.create s in
+    let cmts = Cmts.init_impl source ast comments in
+    let ctx = Pld (PStr ast) in
+    Ok (fmt_file ~f:fmt_structure ~ctx source cmts conf ast ~fmt_code)
+  with _ -> Error ()
```
hhugo: Can we catch exceptions around `Parse_with_comments.parse` only.

src/Cmts.mli
Julow: Nothing prevent the config from changing inside `Cmts.fmt_*` and it's usually changing when used in `Fmt_ast`, we should not add implicit assumption.
I think we should pass it to `fmt_code`, for consistency with how `Cont.t` is used elsewhere.

src/Fmt_ast.ml

src/Cmts.mli
gpetiot: I reverted this commit

PR #975, opened 2019-08-20
Fix formatting of custom indexing operators

### LINKED ISSUE #972: Bug: Custom Indexing Operators are sometimes re-ordered
**Describe the bug**
`ocamlformat` moves around the extended indexing operators (https://caml.inria.fr/pub/docs/manual-ocaml/manual042.html) if the module containing them is not open in a bigger scope. I attach here the diff from a recent `owl` example, and below a small repro:

```
-        sol_planets.Mat.${[ []; [ 3*i ] ]}
-        sol_planets.Mat.${[ []; [ 3*i + 1 ] ]};
-
+        (Mat.( .${} ) sol_planets [ []; [ 3 * i ] ])
+        (Mat.( .${} ) sol_planets [ []; [ (3 * i) + 1 ] ]);
```

**How to Reproduce**

- What is the formatted file? Please try to minimize the code snippet.

The following code
```ocaml
open Owl

let () =
  let m = Mat.zeros 5 5 in
  m.Mat.${[[2];[5]]} |> ignore;
  let open Mat in
  m.${[[2];[5]]} |> ignore;
```
produces the following diff:
```
@@ -2,6 +2,6 @@ open Owl
 
 let () =
   let m = Mat.zeros 5 5 in
-  m.Mat.${[[2];[5]]} |> ignore;
+  Mat.( .${} ) m [ [ 2 ]; [ 5 ] ] |> ignore;
   let open Mat in
-  m.${[[2];[5]]} |> ignore;
+  m.${[ [ 2 ]; [ 5 ] ]} |> ignore
```

- What are the selected options?

My `.ocamlformat` file is:
```
profile=janestreet
wrap-comments=false
let-binding-spacing=sparse
```

gpetiot:
Fix #972, no regression with test_branch.

I created new functions to minimize the diff in Ast.ml (because the old ones are heavily used in Ast.ml and the types are different).

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -88,28 +88,26 @@ let parens_kind i =
         Some (txt, opn, cls)
     | _ -> None
 
-let split_index_op s =
-  match String.rindex s '.' with
-  | Some i when String.length s > i + 2 ->
-      let before = String.sub s ~pos:0 ~len:i in
-      let after = String.sub s ~pos:i ~len:(String.length s - i) in
-      Some ((if String.is_empty before then "" else "." ^ before), after)
-  | _ -> None
-
 let index_op_get i =
-  match split_index_op i with
-  | None -> None
-  | Some (before, after) -> (
-    match parens_kind after with
-    | Some (s, o, c) -> Some (before ^ s, o, c)
-    | None -> None )
+  match (i.[0], parens_kind i) with
+  | '.', Some (s, o, c) -> Some (s, o, c)
+  | _ -> None
 
 let index_op_set i =
   match String.chop_suffix i ~suffix:"<-" with
   | None -> None
   | Some i -> index_op_get i
 
-let index_op_lid i ~f = f (String.concat ~sep:"" (Longident.flatten i))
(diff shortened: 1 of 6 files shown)


src/Ast.ml
```diff
@@ -88,18 +88,32 @@ let parens_kind i =
...
 let index_op_set i =
   match String.chop_suffix i ~suffix:"<-" with
   | None -> None
-  | Some i -> (
-    match (i.[0], parens_kind i) with
-    | '.', Some (s, o, c) -> Some (s, o, c)
-    | _ -> None )
+  | Some i -> index_op_get i
+
+let index_op_lid i ~f = f (String.concat ~sep:"" (Longident.flatten i))
```
Julow: Why convert it to a string and then parse it in `split_index_op` ?
Why not inspect the long indent value (or the result of `Longident.flatten`) ?

gpetiot: I reduced the diff, but it is indeed easier to convert the identifiers to strings and re-use the existing `index_op_get` function.

PR #977, opened 2019-08-20
Change from symbolic links for Windows compat

ulrikstrid:
When installing ocamlformat on Windows with `esy` we just get errors because of the symlinks. This replaces all the symlinks with the real files instead. If there is another workaround please let me know.


gpetiot:
What kind of error do you observe, is it when building or when running the tests?

ulrikstrid:
I get this during `esy install` on Windows, I think all the files I touched in this PR is in the error but they are all the same:
```
tar: ocamlformat-0.11.0/test/passing/break_before_in_auto.ml: Cannot create symlink to 'break_before_in_fit_or_vertical.ml': No such file or directory
```

gpetiot:
From the issue you created on esy (https://github.com/esy/esy/issues/389) it sounds like it should work if you have administrator privileges when executing the commands, does it?

To remove the symbolic links in ocamlformat, we would rather change the code of `test/passing/dune` to look for the most probable .ml file associated for a .ml.ref, for example `break_before_in_auto.ml.ref` would be associated to `break_before_in_fit_or_vertical.ml` instead of `break_before_in_auto.ml`.

ulrikstrid:
That is not the same issue sadly. This issue is that Windows can't handle untaring when there are symlinks in the package, this seems to be true both for cygwin and WSL.

Is this something that is hard to do or can I get some pointers on how to do it?
It's currently blocking to use this with `esy` on Windows, not sure about `opam`.

gpetiot:
There is not a lot of work, I started working on this, but renaming a lot of test files to comply with a new naming convention is a bit tedious.

gpetiot:
Replaced by #980

Issue #965, opened 2019-08-12
Bug: exp-grouping=preserve doesn't affect else blocks


Wilfred:
```
let add_watch env path =
  if SSet.mem path env.wpaths
  then None
  else begin
    env.wpaths <- SSet.add path env.wpaths;
    Some path
  end
```

Formatting produces:

```
$ ocamlformat --enable-outside-detected-project --exp-grouping=preserve /tmp/foo.ml
let add_watch env path =
  if SSet.mem path env.wpaths then None
  else (
    env.wpaths <- SSet.add path env.wpaths ;
    Some path )
```

I expected the `begin` and `end` to be preserved here.

PR #978, opened 2019-08-20
Improve: preserve begin-end keywords in if-then-else

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/965

### DIFF
--- a/src/Params.ml
+++ b/src/Params.ml
@@ -255,7 +255,8 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
     | `Parens -> wrap "(" ")" (wrap_breaks k)
     | `Begin_end -> wrap "begin" "end" (wrap_breaks k)
   in
-  let get_parens_breaks ~opn_hint:(oh_space, oh_other) ~cls_hint:(ch_sp, ch_sl) =
+  let get_parens_breaks ~opn_hint:(oh_space, oh_other)
+      ~cls_hint:(ch_sp, ch_sl) =
     let brk hint = fits_breaks "" ~hint "" in
     match (exp_grouping_bch, imd) with
     | `Parens, `Space -> wrap_k (brk oh_space) (brk ch_sp)
@@ -289,7 +290,11 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
       ; box_keyword_and_expr= Fn.id
       ; branch_pro= fmt_or parens_bch " " "@ "
       ; wrap_parens=
-          wrap_parens ~wrap_breaks:(get_parens_breaks ~opn_hint:((1, 0), (0, 0)) ~cls_hint:((1, 0), (1000, -2)))
+          wrap_parens
+            ~wrap_breaks:
+              (get_parens_breaks
+                 ~opn_hint:((1, 0), (0, 0))
+                 ~cls_hint:((1, 0), (1000, -2)))
       ; expr_pro= None
       ; expr_eol= None
       ; break_end_branch= noop
@@ -299,8 +304,7 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
       ; cond= cond ()
       ; box_keyword_and_expr= Fn.id
       ; branch_pro
-      ; wrap_parens=
-          wrap_parens ~wrap_breaks:(wrap_k (break 1000 2) noop)
+      ; wrap_parens= wrap_parens ~wrap_breaks:(wrap_k (break 1000 2) noop)
       ; expr_pro= None
       ; expr_eol= Some (fmt "@;<1 2>")
       ; break_end_branch= fmt_if_k (parens_bch || not last) (break 1000 0)
@@ -315,9 +319,11 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
       ; box_keyword_and_expr= Fn.id
       ; branch_pro
       ; wrap_parens=
-          wrap_parens ~wrap_breaks:(get_parens_breaks
-            ~opn_hint:((1, 2), (0, 2))
-            ~cls_hint:((1, 0), (1000, 0)))
+          wrap_parens
+            ~wrap_breaks:
+              (get_parens_breaks
+                 ~opn_hint:((1, 2), (0, 2))
+                 ~cls_hint:((1, 0), (1000, 0)))
       ; expr_pro=
           Some
             (fmt_if_k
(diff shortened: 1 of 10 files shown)


src/Params.ml
```diff
@@ -13,6 +13,32 @@ module Format = Format_
...
+  str "(" $ break_before $ k $ break_after $ str ")"
+
+let wrap_beginend ~opn_hint:(_, oh) ~cls_hint:(_, ch) k =
+  str "begin"
+  $ fits_breaks "" ~hint:oh ""
+  $ k
+  $ fits_breaks "" ~hint:ch ""
+  $ str "end"
+
+let wrap_parens_or_beginend_if c ~parens_bch ~exp_grouping_bch ~opn_hint
```
Julow: Similar logic is already implemented in `Fmt_ast` but I did not find it worth it to factorize it. What do you think?

src/Params.ml
```diff
@@ -235,7 +261,7 @@ type if_then_else =
 
 let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
     ~parens_prev_bch ~xcond ~expr_loc ~fmt_extension_suffix ~fmt_attributes
-    ~fmt_cond ~exp_grouping =
+    ~fmt_cond ~exp_grouping ~exp_grouping_bch =
```
Julow: Adding again a new argument to this monster function. Formatting of if-then-else is too complex and not defined.

src/Params.ml
gpetiot: I think it would be nice to factorize this with `wrap_fits_breaks_exp_begin_end`, `parens_or_begin_end`, ` wrap_fits_breaks_exp_if` and `wrap_exp_if` if possible for sure, it makes sense to me to define these functions in Params since they depend on a lot of options.

src/Params.ml
```diff
@@ -291,7 +301,10 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
       ; cond= cond ()
       ; box_keyword_and_expr= Fn.id
       ; branch_pro
-      ; wrap_parens= wrap_if parens_bch "(@;<0 2>" ")"
+      ; wrap_parens=
+          wrap_parens_or_beginend_if c ~parens_bch ~exp_grouping_bch
+            ~opn_hint:((0, 2), (0, 2))
+            ~cls_hint:((0, 0), (0, 0))
```
gpetiot: Maybe it makes sense to have opn_hint and cls_hint (the names could be improved) in the record instead of wrap_parens?

Julow: They are used differently for parens and for begin-end. Also, there is the `indicate_multiline_delimiters` option that should stay in this module.

src/Params.ml
Julow: Moved parens wrapping to `Params`. Did not manage to share code with `if-then-else`.

gpetiot:
Regressions with `test_branch.sh`:

for `if-then-else=k-r` (and both values of `exp-grouping`):
```ocaml
diff --git a/infer/src/clang/Capture.ml b/infer/src/clang/Capture.ml
index 15f0c1bf3..4b184e114 100644
--- a/infer/src/clang/Capture.ml
+++ b/infer/src/clang/Capture.ml
@@ -190,6 +190,7 @@ let cc1_capture clang_cmd =
     L.(debug Capture Quiet) "@\n Skip the analysis of source file %s@\n@\n" source_path ;
     (* We still need to run clang, but we don't have to attach the plugin. *)
     run_clang clang_cmd Utils.consume_in
+  
   ) else if
       Config.skip_analysis_in_path_skips_compilation && CLocation.is_file_blacklisted source_path
     then (
diff --git a/compiler/lib/base64.ml b/compiler/lib/base64.ml
index 7d4999ed1..345b335dc 100644
--- a/compiler/lib/base64.ml
+++ b/compiler/lib/base64.ml
@@ -184,8 +184,7 @@ let decode_sub ?(pad = true) {dmap; _} ?(off = 0) ?len input =
     let n' = n // 4 * 3 in
     let res = Bytes.create n' in
     let get_uint8_or_padding =
-      if pad then (
-        fun t i ->
+      if pad then (fun t i ->
       if i >= len then raise Out_of_bounds;
       get_uint8 t (off + i)
       ) else
```

with `if-then-else=compact` (and both values of `exp-grouping`):
```ocaml
diff --git a/lambda/matching.ml b/lambda/matching.ml
index a9c4929fa..afba50ba8 100644
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -1210,7 +1210,7 @@ and split_no_or cls args def k =
           assert (rev_no = []);
           let yes = List.rev rev_yes in
           insert_split group_discr yes (cl :: rem) def k
-          )
+        )
         else collect group_discr rev_yes (cl :: rev_no) rem
     | [] ->
         let yes = List.rev rev_yes and no = List.rev rev_no in
@@ -1388,7 +1388,7 @@ let split_and_precompile argo pm =
     Format.eprintf "** SPLIT **\n";
     pretty_pm pm;
     pretty_precompiled_res next nexts
-    );
+  );
   (next, nexts)
 
 (* General divide functions *)
@@ -2318,7 +2318,7 @@ let reintroduce_fail sw =
           if c > !max then (
             i_max := i;
             max := c
-            ))
+          ))
         t;
       if !max >= 3 then
         let default = !i_max in
@@ -2482,7 +2482,7 @@ let mk_failaction_pos partial seen ctx defs =
     Format.eprintf "**POS**\n";
     Default_environment.pp defs;
     ()
-    );
+  );
   let rec scan_def env to_test defs =
     match (to_test, Default_environment.pop defs) with
     | [], _
@@ -2518,9 +2518,9 @@ let mk_failaction_pos partial seen ctx defs =
     if dbg then (
       eprintf "POSITIVE JUMPS [%i]:\n" (List.length fail_pats);
       Jumps.eprintf jmps
-      );
+    );
     (None, fail, jmps)
-    )
+  )
   else (
     (* Too many non-matched constructors -> reduced information *)
     if dbg then eprintf "POS->NEG!!!\n%!";
@@ -2532,7 +2532,7 @@ let mk_failaction_pos partial seen ctx defs =
         | Some lam -> string_of_lam lam
         );
     (fail, [], jumps)
-    )
+  )
```
it looks better but shouldn't the closing parens be at the end of the previous line when `if-then-else=compact`?

Other combinations of `exp-grouping` and `if-then-else` are fine.

Julow:
Regressions fixed.

gpetiot:
> Regressions fixed.

They are fixed for `if-then-else=k-r` but I still get them for `if-then-else=compact`.

Julow:
Looks like this behavior is intended and this PR fixes it. I added a test for it: test/passing/ite-compact_closing.ml.ref

PR #980, opened 2019-08-21
Tests: remove symbolic links and change naming convention of tests

gpetiot:
Replace #977 

New naming convention: `-` is now used to separate the original file name (without extension) and the suffix used to differentiate the variants:  `original_file_prefix-variant-suffix.extension`, for example:
- `foo.ml` : base file
- `foo.ml.ref` : output file
- `foo.ml.opts` : options file
- NO `foo-bar.ml` anymore (was a symbolic link to foo.ml)
- `foo-bar.ml.ref` : output for variant based on foo.ml
- `foo-bar.ml.opts` : options for variant based on foo.ml


ulrikstrid:
Thank you for doing this

test/passing/dune
```diff
@@ -16,19 +15,54 @@ let read_lines fn =
...
+          let ref_exists {ref; _} =
+            Option.equal String.equal ref (Some ref_name)
+          in
+          let outputs =
+            if List.exists ref_exists setup.outputs then
+              let opts = Some fname in
+              let update o = if ref_exists o then {o with opts} else o in
+              List.map update setup.outputs
+            else {ref= None; opts= Some fname} :: setup.outputs
+          in
```
Julow: This code looks too complicated. Is it necessary to pair `ref` and `opts`, even when one can be `None` ? Can't each variations of a tests be in the `tests` table and having a single `out` ?

gpetiot: I reduced the diff.

Julow: I now have a lot of errors like
`Error: No rule found for test/passing/break_before_in.ml.ref (context
coverage)`
A lot of them is because the rule is using `.ref` or `.opts` files that don't exists

gpetiot: It should be fixed now, I forgot to clean the env before testing this locally.

test/passing/dune
```diff
@@ -16,14 +17,34 @@ let read_lines fn =
...
+          match String.index_opt test_name '-' with
+          | None ->
+              let s =
+                { has_ref= false
+                ; has_opts= false
+                ; base_file= None
+                ; extra_deps= [] }
+              in
+              Hashtbl.add tests src_test_name s ;
+              s
```
Julow: You could move that to a `add_test ?base_file src_test_name` function to avoid duplication.

gpetiot: Done, I also re-enabled the `.mlt` tests (don't know since when they are disabled)

andschwa:
Thank you so much @ulrikstrid and @gpetiot ❤️

PR #984, opened 2019-08-26
Improve: allow to ignore invalid options

hhugo:
- [x] allow to ignore invalid options
- [x] disable warning 3 in the lexer in quiet mode
- [x] don't warn if quiet

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -84,6 +84,27 @@ let profile_option_names = ["p"; "profile"]
 
 open Cmdliner
 
+let warning, collect_warnings =
+  let delay_warning = ref false in
+  let delayed_warning_list = ref [] in
+  let warning s =
+    if !delay_warning then delayed_warning_list := s :: !delayed_warning_list
+    else Format.eprintf "%s" s
+  in
+  let collect_warnings f =
+    let old_flag, old_list = (!delay_warning, !delayed_warning_list) in
+    delay_warning := true ;
+    delayed_warning_list := [] ;
+    let res = f () in
+    let collected = List.rev !delayed_warning_list in
+    delay_warning := old_flag ;
+    delayed_warning_list := old_list ;
+    (res, collected)
+  in
+  (warning, collect_warnings)
+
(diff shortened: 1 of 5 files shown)


src/Conf.ml
```diff
@@ -1777,6 +1784,8 @@ let read_config_file conf filename_kind =
                 let from = `File (filename, num) in
                 match parse_line conf ~from line with
                 | Ok conf -> (conf, errors, Int.succ num)
+                | Error _ when !ignore_invalid_options ->
+                    (conf, errors, Int.succ num)
```
Julow: Should we warn in this case ? Some users (eg. editor plugins) may want to always enable this but still show warnings to the user.

hhugo: Good point.

The `don't warn if quiet` bit doesn't work well as the command line (containing `--quiet`) is parsed after the files. Any suggestion on how to fix this ?

Julow: Command line arguments that are not formatting options (eg. `inplace`, `debug`) are parsed before but `quiet` is an exception because it can be set in `.ocamlformat` files.
Maybe we could set `conf.quiet` both before and after parsing files and env (`quiet` can be set in env too) ? Otherwise, we have to accumulate warnings and print them after.

src/Conf.ml
```diff
@@ -1921,7 +1928,12 @@ let build_config ~file =
   in
   let files = if !disable_conf_files then [] else files in
   let conf =
-    List.fold files ~init:ocamlformat_profile ~f:read_config_file
+    let init =
+      match C.get_from_cmdline quiet with
+      | None | Some false -> ocamlformat_profile
+      | Some true -> {ocamlformat_profile with quiet= true}
+    in
```
Julow: Should call `update_using_env` here (env variable and `--config`)

src/Config_option.mli
```diff
@@ -45,6 +47,8 @@ module Make (C : CONFIG) : sig
 
   val default : 'a t -> 'a
 
+  val get_from_cmdline : 'a t -> 'a option
```
gpetiot: Not used anymore.

Issue #985, opened 2019-08-26
Bugs in opam repository


hhugo:
Note that this experiment is based on https://github.com/ocaml-ppx/ocamlformat/pull/984

**How to reproduce**
```bash
mkdir all-sources
cd all-sources
opam list -a --columns name | grep -v "^#" | xargs -n 1 -I {} opam source {}
find ./ -type f -name "*.ml" -o -name "*.mli" | parallel -- ../ocamlformat/_build/release/src/ocamlformat.exe --enable-outside-detected-project --no-version-check --quiet --no-comment-check --ignore-invalid-option > /dev/null
```
gives, with some manual cleanup, the following errors:
- [x] `./bap-objdump.1.6.0/lib/x86_cpu/x86_types.ml: formatting did not stabilize after 10 iterations.` (fixed by https://github.com/ocaml-ppx/ocamlformat/pull/998)
- [x] `./boomerang.1.1.0/lib/bakery.ml: ast changed.` (issue https://github.com/ocaml-ppx/ocamlformat/issues/994, fixed by https://github.com/ocaml-ppx/ocamlformat/pull/995)
- [x] `./frenetic.5.0.3/test/Test_Frenetic_NetKAT_Compiler.ml: ast changed.` (fixed by https://github.com/ocaml-ppx/ocamlformat/pull/995)
- [x] `./mldonkey.3.1.6/src/daemon/common/commonMessages.ml: ast changed.` (fixed by https://github.com/ocaml-ppx/ocamlformat/pull/995)
- [x] `./mldonkey.3.1.6/src/daemon/common/commonPictures.ml: ast changed.` (fixed by https://github.com/ocaml-ppx/ocamlformat/pull/995)
- [ ] `./ocaml-variants.4.10.0+trunk+flambda/testsuite/tests/parsetree/source.ml: ~~generating invalid ocaml syntax~~ ast changed.` (~~fixed by https://github.com/ocaml-ppx/ocamlformat/pull/996~~)
- [x] `./ocaml_at_p.1.1.0/test/test.ml: ast changed.` (#987)
- [x] `./ppx_cstubs.0.2.0/examples/inline.c.ml: generating invalid ocaml syntax.` (#988)
- [x] `./ppx_monadic.2.3.0/tests/test_beginm.ml: unhandled exception. Use [--debug] for details.`
[...]

PR #988, opened 2019-08-27
Fix: escape special characters in external declaration

Julow:
Escape special characters in the primitive name. This code appear in `ppx_cstubs`:

```ocaml
external%c print: str:string -> d:int -> void = {|
  printf("%s (%d)\n",$str,$d);
  fflush(stdout);
|}
```

The formatting is ugly, should we improve this ?

Found by https://github.com/ocaml-ppx/ocamlformat/issues/985

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2992,6 +2992,11 @@ and fmt_value_description c ctx vd =
   let doc_before, doc_after, atrs =
     fmt_docstring_around_item c pval_attributes
   in
+  let fmt_val_prim s =
+    if List.length (String.split_on_chars s ~on:[' '; '\n']) > 1 then
+      wrap "{|" "|}" (str s)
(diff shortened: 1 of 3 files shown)


gpetiot:
I pushed a commit to use `{| |}` when the string is long (could be more precise but looks good in practice).

src/Fmt_ast.ml
```diff
@@ -2992,6 +2992,11 @@ and fmt_value_description c ctx vd =
   let doc_before, doc_after, atrs =
     fmt_docstring_around_item c pval_attributes
   in
+  let fmt_val_prim s =
+    if List.length (String.split_on_chars s ~on:[' '; '\n']) > 1 then
```
hhugo: `String.exists` ?

gpetiot: Done

Julow:
Your improvement looks good. We can merge.

PR #1003, opened 2019-09-02
Fix infix operators in pattern

Julow:
Extracted from https://github.com/ocaml-ppx/ocamlformat/pull/921, original commit by @gpetiot

I tried to use `fmt_infix_op_args` for both expressions and patterns.

Still WIP, there is regressions.


Julow:
This work is probably obsolete, we should start again from https://github.com/ocaml-ppx/ocamlformat/pull/921, depending on what still makes sense now

PR #1005, opened 2019-09-03
Improve: remove the bisect_ppx dependency and clean Makefile

Julow:
- Avoid building and running tests twice
- Avoid using paths to `_build`
- Remove `dune-workspace` that was making `dune` hard to use on the command line
- Remove plugin code from `src/dune`
- Remove bisect support
- Improve `test_branch` to use a separate tree, so the script does not change while executing (UB) and does not change user's worktree

Everyone that built OCamlformat before must remove its `dune-workspace` file.
That file must be removed everytime we switch from before to after this PR is eventually merged (eg. reviewing PRs, git bisect)

With this, `dune build` and `dune exec -- ocamlformat` will work correctly.


hhugo:
- I believe this will make ocamlformat slower. The "release" profile currently removes asserts by passing the `-noassert` flag. There are pretty expensive checks in Ast.ml. @jberdine, do you have any though on this ?
- I'm a bit sad to see bisect support disappear. I was hoping that we build something with it. Like making sure that new feature come with good tests (covering the new code paths)
- can you explain why `dune-workspace` is make using dune hard ?

Julow:
I readded the `-noassert` flag in release mode.

The `dune-workspace` was problematic because:
- It is generated by the Makefile
- It references the current opam switch but is not regenerated if it changes
- If it is not present, the contexes it defines will not be built. So `test-extra/Makefile` won't find the binary
- It removes the `default` context. `dune exec` must be called with `--context=dev`

The last 2 could be fixed by renaming the `dev` context to `default`.

Removing bisect is not great, I agree. I'll try to add it back.

Julow:
Bisect is quite easy to use manually, I'm not sure it's worth the trouble adding it to our build system:

- Add `(preprocess (pps bisect_ppx))` to `src/dune`
- `dune runtest`
- `bisect-ppx-report -ignore-missing-files -html coverage _build/default/test/*/bisect*.out`

The first step could be automated using either a diff patch or a template comment.

Julow:
I added `tools/bisect.sh` that builds and runs ocamlformat with ppx_bisect in a temporary worktree.

gpetiot:
It looks good to me, if it suits Hugo as well you can merge.

Julow:
I added a commit to remote the temporary worktrees in `test_branch.sh` to avoid filling `/tmp`.

gpetiot:
Can be rebased now that the odoc dependency has been updated, and merged.

Julow:
Rebased.

jberdine:
Isn't `dune exec` not working with any context a dune bug?

After this change, what is the intended workflow for testing multiple compiler versions, etc. With no workspace, it seems that the only reasonable workable thing to do is to use a local opam switch so that it is automatically selected. But then there can be only one. Which makes the process of e.g. changing the compiler version very very slow.

I don't understand how this change isn't seen as a huge developer experience regression.

Julow:
I think the usual workflow for that is simply changing the current switch (`opam switch <switch>`).
If you must not change your current global switch, you can use `opam switch link <switch>`. It symlinks a global switch as a local switch in your current directory.

emillon:
@jberdine there's also another way to test on multiple compiler versions, which is having a `dune-workspace.dev` file. That way, it's not picked by default, but it's still possible to test all versions at once. There's [a setup like this in dune](https://github.com/ocaml/dune/blob/de06f8fb1fcdd2b11496ea67db3c5f0d19a56597/Makefile#L49-L50), though in practice we often let CI deal with these compatibility issues.

jberdine:
@Julow thanks, I didn't know about `opam switch link`. But especially for testing, when the config is not checked in, it is hard to make it reproducible across developers and across time, e.g. while bisecting.

@emillon, with `dune-workspace.dev`, do you mean basically to add back the `dune-workspace` file removed by this PR, but to add `.dev`?

emillon:
@jberdine yes, something like that (I'm not familiar with the context of this PR or how the file was generated)

Issue #1010, opened 2019-09-10
Style suggestion: consistent placing of structure item doc comments


CraigFe:
**What is the reason for this new style?**
Inconsistency surrounding doc comments and signatures. Consider the following example:

```ocaml
(** This comment goes before *)
module S_ext : sig
  type t
end

module Index : Index.S
(** This one goes after *)

module Index2
    (Paramater_module : BAR_LONG_MODULE_TYPE_NAME)
    (Foo : BAR)
    (Foo : BAR)
    (Foo : BAR)
    (Foo : BAR)
    (Foo : BAR) : sig end
(** This one _still_ goes after *)

module Make (Config : sig
  val blah : string

  (* this could be a really long signature *)
end) : S
(** Doc comment still goes after *)
```
With `profile=conventional`, Ocamlformat will try to put structure doc comments _after_ the struct by default (presumably to keep symmetry with the value signatures). When the structure is too large, it will instead put the comment at the start. I like this behaviour; however, it isn't consistent about this w.r.t. functors whose parameters span multiple lines.

**Describe the solution you'd like**
I see two solutions:
 1. always put doc comments on structures before, regardless of the size of the structure.
 1. refine the size heuristic to account for potential multi-line functor parameters.

CC. @Julow

Julow:
Your 2nd suggestion is easy to implement but the inconsistency still remains.

We could get rid of this rule and:
- Always put doc comments before modules, as you suggested
  This would be inconsistent with `val`s, but at least every forms of `module` would be consistent
- Follow the `doc-comments=before|after` option strictly
  I think the argument against that is the comment would be too far down to be noticed.
  My opinion is that this comment is about how this module fits with the rest of the parent module and does not matter when reading the module implementation.
  ```ocaml
  module Foo = struct
    (** Implement that algorithm
        ... long comment *)

    type t = ...
  end
  (** This module is used represent x *)

  type t = { x : Foo.t }
  (** t *)
  ```

Also, there is an other special case, doc comments cannot be placed after a variant declaration:

```ocaml
type t = int
(** t *)

(** t' *)
type t' = Foo
```

Otherwise, the comment would be attached to the last constructor (by OCaml's parser) and the AST would change:

```diff
-type t' = Foo
-(** t' *)
+type t' = Foo  (** t' *)
```

CraigFe:
[...]

PR #1012, opened 2019-09-10
Add the doc-comments-val option

Julow:
Partial fix to https://github.com/ocaml-ppx/ocamlformat/issues/1010

~~Add `doc-comments=after-val` that is like `after` for `val`s and `external`s and like `before` for other items.~~
Edit: This PR adds `doc-comments-val=before|after` to control placement of doc comments on `val` and `external` only.

This removes the inconsistency caused by the `is_simple` heuristic on modules and classes or by the attachments of doc comments on constructor by OCaml's parser:

```ocaml
(** This comment goes before *)
module S_ext : sig
  type t
end

module Index : Index.S
(** This one goes after *)
```

```ocaml
type t = int
(** t *)

(** t' *)
type t' = Foo
```

This is added to the `conventional` profile.


CraigFe:
Adding my (underqualified) opinions given offline:
 - I think this is a good short-term solution for backwards compatibility reasons, but that `after` should be deprecated as a confusing option.
 - I still think the most intuitive interface would be something like 
   - `doc-comments-val={before,_after_}`
   - `doc-comments-struct={_before_,after,auto}`
  potentially with the 'auto' option being deprecated depending on how much people care about this heuristic. If the heuristic is to be kept, it could benefit from improvements re. functors as suggested in #1010.

Also worth noting that this is does nothing to resolve the variants issue highlighted by @Julow [here](https://github.com/ocaml-ppx/ocamlformat/issues/1010#issuecomment-529948458), since that's an OCaml parser problem.

test/passing/doc_comments-after_val.ml.ref
```diff
@@ -0,0 +1,264 @@
...
+  class b =
+    object
+      (** Method *)
+      method f = 0
+
+      (** Inherit *)
+      inherit a
+
+      (** Val *)
+      val x = 1
```
gpetiot: should be after?

gpetiot:
Having 2 options makes sense to me, to avoid having weird option values like `after-val` that would be confusing to the users.

I have a question though, why is this new behavior applied to `val`/`external` in signatures but not to the `type` definitions or other signature elements? What makes the `val` so specific?

It seems like it is only applied for toplevel/module values, and not for class values, is this intended?

gpetiot:
I'm just not convinced with the name, maybe `auto` would be better, since we basically decide what's better instead of the user, depending of the size and kind of the item.

Julow:
I don't like `auto` because the rules are simple and the user will know exactly how it will format.
Do you mean renaming `after` to `auto` ? This would make sense, yes.

Julow:
Just rebased.

gpetiot:
If replacing `after` by `auto` makes sense, why not.
But to me `after-val` sounds way too close to `after` and yet very specific, do `after` and `after-val` really need to co-exist? Would it be feasible to keep only one of them?
I'm more reluctant to add more options/values now that we experienced how difficult it is to remove/modify it when it's released out there.

Julow:
The reason for this new value is because `after` has too arbitrary and hard to understand. Users can't predict where the docstring will be placed and just assume this is a bug.
`after-val` has a much simpler rule, it places docstrings after `val`/`external` declarations and before everything else.
I agree the name is not nice and looks like a special case.

An alternative would be to deprecate `doc-comments`, making it default to `before` and adding a new `doc-comments-val=before|after` to control just that.

gpetiot:
> An alternative would be to deprecate `doc-comments`, making it default to `before` and adding a new `doc-comments-val=before|after` to control just that.

This sounds better indeed, as long as there is no regression for the `janestreet` profile, and only improvements for the `conventional` and `ocamlformat` profiles I'm all for it.

Julow:
I changed to keep `doc-comments` as-is and added `doc-comments-val=before|after`.
In conventional it's set as `doc-comments=before` and `doc-comments-val=after`.

Only the `ocamlformat` profile uses `doc-comments=after`.
Should we deprecate `doc-comments` ?
If yes, should we set change the `ocamlformat` profile to allow to remove that option ?

gpetiot:
> Only the `ocamlformat` profile uses `doc-comments=after`.
> Should we deprecate `doc-comments` ?
> If yes, should we set change the `ocamlformat` profile to allow to remove that option ?

How does the diff look like? Docstrings for big elements should already be before.

Julow:
The diff is huge on infer, mainly because of doc comments on types.
A more reasonable plan would be to deprecate `after` to later rename it to something like `auto`. What do you think?

gpetiot:
We can avoid going through a deprecation step if `doc-comments` is in the way of having a consistent `doc-comments-val`, so to summarize we will have:
- `doc-comments-val={before,after}`
- `doc-comments` is removed and its default will be the current `before`

but printing a custom message when this option is parsed would be useful for the user

Julow:
I rebased and cleaned the history a bit. There quite a bit of changes with test_branch but they all are expected.
I'm Ok with removing `doc-comments` without warning. We should do it in an other PR, as we may want to log that properly.

lib/Fmt_ast.ml
```diff
@@ -2907,7 +2918,7 @@ and fmt_value_description c ctx vd =
   @@ fun c ->
   let pre = if List.is_empty pval_prim then "val" else "external" in
   let doc_before, doc_after, atrs =
-    fmt_docstring_around_item c pval_attributes
+    fmt_docstring_around_item ~is_val:true c pval_attributes
```
Julow: This applies `doc-comments-val` on `external`s in structures. Is this a problem ?

gpetiot: It is what I would expect, since they are similar to vals, what do you think @craigfe ?

CraigFe: It's also what I would expect, and I think the documentation makes this sufficiently clear.

PR #1013, opened 2019-09-11
Check functor arguments when computing placement of doc comments

Julow:
Partial fix to https://github.com/ocaml-ppx/ocamlformat/issues/1010

Functor arguments are not checked to compute the `force_before` value.

In this example, the comment should be before:

```ocaml
module Make (Config : sig
  val blah : string

  (* this could be a really long signature *)
end) : S
(** Doc comment still goes after *)
```

### DIFF
--- a/test/passing/doc_comments-after.ml.ref
+++ b/test/passing/doc_comments-after.ml.ref
@@ -107,6 +107,30 @@ module Comment_placement : sig
   (** A *)
   external a : b = "double_comment"
   (** B *)
+
+  (** This comment goes before *)
+  module S_ext : sig
+    type t
+  end
+
+  module Index : Index.S
+  (** This one goes after *)
+
+  module Index2
+      (Paramater_module : BAR_LONG_MODULE_TYPE_NAME)
+      (Foo : BAR)
+      (Foo : BAR)
+      (Foo : BAR)
+      (Foo : BAR)
+      (Foo : BAR) : sig end
+  (** This one _still_ goes after *)
+
+  module Make (Config : sig
+    val blah : string
+
+    (* this could be a really long signature *)
+  end) : S
+  (** Doc comment still goes after *)
 end = struct
   type t = {a: int}
   (** Type *)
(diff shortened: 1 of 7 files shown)


test/passing/doc_comments-after.ml.ref
```diff
@@ -107,6 +107,30 @@ module Comment_placement : sig
...
+  (** This one goes after *)
+
+  module Index2
+      (Paramater_module : BAR_LONG_MODULE_TYPE_NAME)
+      (Foo : BAR)
+      (Foo : BAR)
+      (Foo : BAR)
+      (Foo : BAR)
+      (Foo : BAR) : sig end
+  (** This one _still_ goes after *)
```
Julow: Should we do something to this ?
The `*_is_simple` functions could easily be modified to make an estimation of the size of the line, as any corner cases will be not "simple".

gpetiot: It would be nice to do something, maybe just considering that `Pmty_functor (_, Some t1, t2)` is not simple, even if `t1` is simple.

Julow: What about `Pmty_functor (_, None, t2)` ? I changed so any functor is not simple. I have no diffs with `test_branch`, even on [Irmin](https://github.com/mirage/irmin/).

gpetiot: This case is something like `module Foooo () : t2` so we can keep checking if t2 is simple like it's currently done (and add some tests).

Julow: Modifying the _is_simple function won't work because it's not called on the whole module expr/type. But this case can be handled easily in `fmt_module`. I wonder if it's worth adding a "ctx" argument to it to be able to implement better the `single_line` logic ?

I made generative functors simple in `fmt_module`.

gpetiot: > I wonder if it's worth adding a "ctx" argument to it to be able to implement better the `single_line` logic ?

I'm not sure we should over-engineer this function, the output looks good as it is now.

PR #1014, opened 2019-09-11
Improve: dock collection brackets

gpetiot:
- change `break-separators` to `={before|after}` only (remove `after-and-docked`)
- add the `dock-collection-brackets` option
- the formatting of some cases of "before and docked" can be discussed, especially in the type declarations (I think the `;` should be aligned with the closing bracket but it looks weird when it results in a 0 column indentation)

benefits:
- we can have a "before and docked" configuration, I remember someone has a need for that (can't put my finger on it, it was either on the slack or on the discuss)
- code easier to maintain

The formatting of docked collections in let-bindings and in pattern-matching is not yet fixed in this PR.


Wilfred:
Agreed that the split makes sense, but it's worth mentioning this is a breaking change. Existing configurations using `after-and-docked` now give me:

```
$ ocamlformat foo.ml
  ("bad value for"
    (break-separators
      "invalid value `after-and-docked', expected either `after' or `before'")))
Raised at file "src/error.ml" (inlined), line 9, characters 14-30
Called from file "src/error.ml", line 11, characters 19-40
Called from file "src/exn.ml", line 66, characters 8-11
Re-raised at file "src/exn.ml", line 69, characters 6-15
Called from file "src/Conf.ml", line 1890, characters 6-1023
Called from file "list.ml", line 121, characters 24-34
Called from file "src/list0.ml" (inlined), line 22, characters 40-81
Called from file "src/Conf.ml", line 2040, characters 4-45
Called from file "src/Conf.ml", line 99, characters 14-18
Called from file "src/Conf.ml", line 2077, characters 4-61
Called from file "src/Conf.ml", line 2120, characters 20-69
```

gpetiot:
Yes sorry about that, we may take the liberty of sometimes renaming some options/values before the 1.0 release (we will be more stable after). We will document every breaking change in the next release patchnote.

emillon:
@Wilfred I improved the error message for this case in #1089!

PR #1018, opened 2019-09-12
Improve: don't require --name, require kind, forbid --inplace, allow --check, make --enable-outside-detected-project implicit when reading from stdin

### LINKED ISSUE #1015: Feature request: make --enable-outside-detected-project implicit when inputting via std.in
When running OCamlformat with standard input, it is required to pass `--enable-outside-detected-project`.

```sh
ᐅ echo "let x = ()" | ocamlformat --profile=conventional --name foo.ml -
File "foo.ml":
Warning: Ocamlformat disabled because [--enable-outside-detected-project] is not set and no project root was found
let x = ()
```

Perhaps `--enable-outside-detected-project` should be implicit when running OCamlformat this way? I think if the user is providing input from `std.in`, it's likely they're not getting this code from an enclosing OCamlformatted project.

CC. @Julow

### LINKED ISSUE #1016: Feature request: allow omitting --name flag when reading from std.in
Related to https://github.com/ocaml-ppx/ocamlformat/issues/1015, perhaps the {`--impl`,`--intf`,`--use-file`} flags should avoid the need to pass a filename when reading from standard input:
```sh
ᐅ echo "let x = ()" | ocamlformat --profile=conventional --impl -

ocamlformat: Must specify name when reading from stdin

```

Make `--impl`, `--intf`, `--use-file` override the need to pass a `--name` option. Relatedly, I think the error message when no flags are passed should suggest using one of the former rather than suggesting passing a name e.g.:

```sh
ᐅ echo "let x = ()" | ocamlformat --profile=conventional -

ocamlformat: Must use one of {--impl, --intf, --use-file} to specify the file-type 
             when reading from stdin.
```

(as a user, I was initially confused as to why OCamlformat needed me to pass a `--name` in this case)

CC. @Julow

gpetiot:
Fix #1016 and fix #1015 

- [x] do not force the user to provide `--name` when the input is read from stdin, force the use of `--impl` or `--intf` or `--use-file` instead
- [x] check that we do not use `--inplace` when using stdin
- [x] allow `--check` when using stdin
- [x] better error messages when the input is read from stdin (use `<standard input>` instead of `-` in messages, can still be overriden by `--name`)
- [x] make `--enable-outside-detected-project` implicit when reading from stdin

each commit can be reviewed separately

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1629,8 +1629,8 @@ let validate () =
     `Error (false, "Must specify at least one input file, or `-` for stdin")
   else if has_stdin && inputs_len > 1 then
     `Error (false, "Cannot specify stdin together with other inputs")
-  else if has_stdin && Option.is_none !name then
-    `Error (false, "Must specify name when reading from stdin")
+  else if has_stdin && !inplace then
+    `Error (false, "Cannot specific stdin together with --inplace")
   else if !inplace && Option.is_some !output then
     `Error (false, "Cannot specify --output with --inplace")
   else if !check && !inplace then
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -1641,6 +1641,11 @@ let validate () =
     `Error (false, "Must specify at least one input file, or `-` for stdin")
   else if has_stdin && inputs_len > 1 then
     `Error (false, "Cannot specify stdin together with other inputs")
+  else if has_stdin && Option.is_none !kind then
+    `Error
+      ( false
+      , "Must specify at least one of --impl, --intf or --use-file when \
+         reading from stdin" )
   else if has_stdin && !inplace then
     `Error (false, "Cannot specify stdin together with --inplace")
   else if !inplace && Option.is_some !output then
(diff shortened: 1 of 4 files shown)


src/Conf.ml
```diff
@@ -1865,22 +1882,31 @@ let update_using_env conf =
...
   match !kind with
   | Some kind -> kind
   | None -> (
-    match Filename.extension fname with
-    | ".ml" -> `Impl
-    | ".mli" -> `Intf
-    | ".mlt" -> `Use_file
-    | _ -> `Impl )
+    match file with
+    | Stdin -> impossible "checked by validate"
```
Julow: I don't like having this unchecked. We could have the kind and display name of a file computed before validate and passed safely (eg. `Stdin of kind | File of name * kind`). This would be a big change on how all this code works, so we should do it in an other PR.

gpetiot: I agree but this assumes the kind of file arg is passed before the input files, otherwise it needs to be done afterwards and it's not much cleaner than what is currently done.

Issue #1011, opened 2019-09-10
Feature request: Support #use toplevel directives in .ml files


Wilfred:
**Is your feature request related to a problem? Please describe.**
It would be really useful if `ocamlformat` could handle files with `#use "foo.ml"` in them.

#523 mentions that shebangs `#!/usr/bin/env ocaml` and `#require "Foo"` are supported, and #731 mentions some others, but `#use` currently isn't supported.

```ocaml
#use "bar.ml"
```

```
$ ocamlformat --enable-outside-detected-project ~/tmp/foo.ml
ocamlformat: ignoring "/Users/wilfred/tmp/foo.ml" (syntax error)
File "/Users/wilfred/tmp/foo.ml", line 1, characters 0-1:
Error: Syntax error
```

**Describe the solution you'd like**
Format files with `#use "foo.ml" in them. I don't think ocamlformat needs to anything with the directive, but it would be nice if it processed the rest of the file.

Julow:
OCamlformat parses directives only in `use-file` mode. You can either rename your file to `foo.mlt`, use `--use-file` or `--name foo.mlt` to tell OCamlformat how to parse this file.

Wilfred:
Is this a common convention for putting `#use` in `.mlt` files? I couldn't find anything when searching for "ocaml mlt" except https://github.com/janestreet/mlt_parser, which doesn't seem to be the same syntax.

PR #1020, opened 2019-09-13
Parse toplevel directives

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1011

This PR fixes the formatting of use-files and parse/format implementations as use-files.

This is WIP

### DIFF
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -696,6 +696,7 @@ let break_between s ~cmts ~has_cmts_before ~has_cmts_after (i1, c1) (i2, c2)
   | Mod _, Mod _ ->
       break_between_modules ~cmts ~has_cmts_before ~has_cmts_after (i1, c1)
         (i2, c2)
+  | Top, _ | _, Top -> true
   | _ -> assert false
 
 (** Precedence levels of Ast terms. *)
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -563,6 +563,9 @@ let has_trailing_attributes_mty {pmty_attributes; _} =
 let has_trailing_attributes_mod {pmod_attributes; _} =
   List.exists pmod_attributes ~f:is_doc
 
+type toplevel_item =
+  [`Item of structure_item | `Directive of toplevel_directive]
+
 (** Ast terms of various forms. *)
 module T = struct
   type t =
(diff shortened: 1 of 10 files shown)


src/ocamlformat.ml
```diff
@@ -22,13 +22,13 @@ let moved_docstrings f c a b =
...
-  ; moved_docstrings= moved_docstrings Normalize.moved_docstrings_impl
-  ; normalize= normalize Normalize.impl
-  ; printast= Migrate_ast.Printast.implementation }
+  { parse= Migrate_ast.Parse.use_file
+  ; init_cmts= Cmts.init_use_file
+  ; fmt= Fmt_ast.fmt_use_file
+  ; equal= equal Normalize.equal_use_file
+  ; moved_docstrings= moved_docstrings Normalize.moved_docstrings_use_file
+  ; normalize= normalize Normalize.use_file
+  ; printast= Migrate_ast.Printast.use_file }
```
Julow: I'm tempted to change "use_file" to "toplevel_structure" or "impl" in the name of all this functions. But they would no longer match the name of the functions from OCaml's parser.

gpetiot: My opinion would be that being consistent in ocamlformat itself is more important than being consistent with the parser's API, so we could have `impl` everywhere (but I don't mind `use-file` instead, as long as we are consistent)

Julow: I renamed "use_file" to "toplevel" except in `Migrate_ast`.

src/Ast.ml
```diff
@@ -664,7 +671,7 @@ let location = function
   | Mod x -> x.pmod_loc
   | Sig x -> x.psig_loc
   | Str x -> x.pstr_loc
-  | Top -> Location.none
+  | Top | Tli _ -> Location.none
```
gpetiot: There is a loc for `Item` and for `Directive`

Julow: Done

Julow:
Fixed ocamlformat-help.txt. I'm not sure it's worth adding tests, there is a test already for directives. The real test here is that it works as well as before.

gpetiot:
> Fixed ocamlformat-help.txt. I'm not sure it's worth adding tests, there is a test already for directives. The real test here is that it works as well as before.

This test ia a .mlt file, we can just rename it to a .ml file.

Issue #1019, opened 2019-09-13
Bug: Extra empty line in sequences containing comments 


pascutto:
Hi !
When writing a sequence of instructions with interleaved comments, `ocamlformat` with profile `conventional` adds a extra line before the comment:
```ocaml
let () =
  print_string "Hello, world!";
  (* This is a comment *)
  print_newline ()
```
is formatted to
```ocaml
let () =
  print_string "Hello, world!";

  (* This is a comment *)
  print_newline ()
```

PR #1021, opened 2019-09-13
Fix extra blank line in sequence

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1019

### DIFF
--- a/src/Cmts.mli
+++ b/src/Cmts.mli
@@ -131,7 +131,7 @@ val remaining_comments : t -> (Cmt.t * string * Sexp.t) list
 val remaining_locs : t -> Location.t list
 
 val remaining_before : t -> Location.t -> Cmt.t list
-(** Returns comments before [loc] *)
+(** [remaining_before c loc] returns the comments before [loc] *)
 
 val diff :
   Conf.t -> Cmt.t list -> Cmt.t list -> (string, string) Either.t Sequence.t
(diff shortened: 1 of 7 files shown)


src/Fmt_ast.ml
```diff
@@ -541,6 +541,23 @@ let fmt_assign_arrow c =
...
+  | `Preserve_one ->
+      let l1 = xe1.ast.pexp_loc.loc_end.pos_lnum in
+      let rec blank_lines l2 =
+        let open Location in
+        let n = l2.loc_start.pos_lnum - l1 in
+        if n <= 0 then 0
+        else if Cmts.has_before c.cmts l2 then
+          let s = l2.loc_start in
+          let loc_start = {s with pos_lnum= s.pos_lnum - 1} in
+          blank_lines {l2 with loc_start}
```
Julow: I'm not a fan of constructing a record from OCaml's parser but it's the only way to talk to `Cmts`.

gpetiot: Could we instead use the location `Cmt.loc`? (not sure iterating on the comments from one location is much cleaner though)

Julow: I completely overlooked that. I have a much cleaner implementation now that only compare line numbers and is more readable.

src/Cmts.mli
```diff
@@ -130,6 +130,9 @@ val remaining_comments : t -> (Cmt.t * string * Sexp.t) list
 
 val remaining_locs : t -> Location.t list
 
+val remaining_before : t -> Location.t -> Cmt.t list
+(** Returns comments before [loc] *)
```
gpetiot: I would maybe change to something like: `[remaining_before c loc] returns comments before [loc].`

Julow: Yes that's better.

PR #1023, opened 2019-09-19
Project: use a CHANGES.md log file again

gpetiot:


avsm:
Thanks, this is very useful to have for other tools like dune-release, as well as for grepping for changes quickly.

HACKING.org
```diff
@@ -4,6 +4,8 @@
 - dune subst
 - ocaml tools/gen_version.mlt src/Version.ml M.N.O
 - git add src/Version.ml
+- update CHANGES.md by editing output of tools/gen_changes.sh vCURR vM.N.O
```
Julow: I don't think we should generate this. These are the PR names, which are often badly written.
We may want to add one more line to explain the feature or add more than 1 contributor (for example, reviewers, reporter, more than one contributor).
Also, we may want to sort entries, for example, bugs fixed, breaking changes, new options, etc...

What do you think ?

gpetiot: Yes but automatically generating the list is convenient (it's up to us to have good names for the PRs), and of course we can manually add more details and categorize the items afterwards.

Julow: Of course we should generate the list now.
But I don't think we should update it at every release but instead in every PRs, by hand.

gpetiot: I agree it would be better to do it for each PR (I don't remember how it was done by Josh before the changelog file was removed in the first time), the line in HACKING.org is merely a reminder to keep the file up-to-date before a release.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: c
 env:
-  - OCAML_VERSION=4.08.0
+  - CI_KIND=build OCAML_VERSION=4.08.0
+  - CI_KIND=changes OCAML_VERSION=4.08.0
```
gpetiot: it looks like it is running twice the same build and not checking the changelog at all

Julow: Why would `CI_KIND` have any effect ? That's just passed to our scripts, we have to interpret it.

Julow:
I added a command to check that CHANGES.md is updated.

The CI fail at `opam upgrade` with:
```
[WARNING] Upgrade is not possible because of conflicts or packages that are no
          longer available:
  - No available version of odoc satisfies the constraints
The following dependencies are the cause:
  - ocamlformat → odoc = 1.4.1
```

gpetiot:
Should we just give up 4.08, upgrade to 4.09, and update the tests in this PR to fix the CI?

HACKING.org
gpetiot: I reverted these changes.

PR #1026, opened 2019-09-24
Improve: format invalid files

### LINKED ISSUE #906: Do not fail on incorrect files
Files that cannot be fully parsed like:
```ocaml
let fooooooooooo =
  let a = 1 in
  let fooooooooooooooo =
    let b = 2 in
    let foooooooooooooo =
```
should be processed by `ocamlformat` (like it is the case with `ocp-indent`).

No pressure yet, but it would be a killer feature for 1.0

cc @let-def, any new insights since the last time we discussed about it is welcomed.

gpetiot:
Fix #906


gpetiot:
@let-def can you confirm it's the wrong approach and we should close this one?

gpetiot:
I managed to make ocamlformat compile, no regression for non-failing files.
But for an invalid file like this:
```ocaml
let f =
  let foooooooooo =
    let fooooooooooooo = foooooooooooooo in
    fooooooooooooo
  in
```
an extension node is added like this:
```ocaml
let f =
  let foooooooooo =
    let fooooooooooooo = foooooooooooooo in
    fooooooooooooo
  in
  [%merlin.hole]
```

there is still some work to modify this into an extension node containing the verbatim of the input item.

gpetiot:
@emillon in f6f9515 I had to remove the .mli file (that looked correct) to fix the following error (first time seeing this one):
```
Error: Module Raw_parser in directory _build/default/lib depends on
Ocamlformat_lib.
This doesn't make sense to me.

Ocamlformat_lib is the main module of the library and is the only module
exposed outside of the library. Consequently, it should be the one depending
on all the other modules in the library.
```
do you have an idea about what is the most likely cause?

emillon:
seems weird, I commented on the related issue.

gpetiot:
This branch will have to be modified based on what is done in https://github.com/gpetiot/parse-wyc: we have to modify (wrap in verbatim) the original source based on the locations of invalid chunks provided by the parser.

gpetiot:
@emillon @Julow I reverted previous changes and then rebased on top of #1188 (I thought it would be necessary, but turns out it's not, I can rebase on master instead). We have now a prototype that formats invalid files (would be improved in https://github.com/gpetiot/parse-wyc).

Should we add an option to keep the former behavior (`--safe/--unsafe`, `--valid-only`) ? I don't know what to do of the tests that are expected to fail: `test/cli/fmterr_stdin_and_name.output`, `test/cli/fmterr_file_and_name.output` and `test/passing/error1.ml.output`.

emillon:
Great work!

Yes, I think that the default should be to only accept valid ASTs and add a `--partial` flag to opt into this new (beta) behavior.

gpetiot:
I made the integration of the parser in `Translation_unit.ml` simpler and easier to read (I don't think it could be shorter), the only missing thing is to make the parser lib available, either on opam or vendored in ocamlformat.

gpetiot:
I vendored `parse-wyc`, but its failing tests cause the whole CI checks to fail, if it's possible to mute them (and fix them in later PRs) it's good to be reviewed.

gpetiot:
The ocaml-ci target `lint` weirdly fails, does anyone has the same failure locally?

lib/Conf.ml
```diff
@@ -1371,6 +1371,10 @@ let output =
       & opt (some string) default
       & info ["o"; "output"] ~doc ~docs ~docv)
 
+let partial =
+  let doc = "Ignore invalid (unparsable) parts of the input." in
+  mk ~default:false Arg.(value & flag & info ["partial"] ~doc ~docs)
```
emillon: I'm not too sure about the name. `--partial` can convey the idea that only part of the file will be kept, or that formatting will be done only on one part of the file. I think that the idea of the configuration key is to enable this particular feature, so what do you think about (combinations of) the following options:
- `--support-invalid-files`
- `--recover-parse-errors`
- `--format-invalid-code`

By the way, it seems that historically all flags (boolean options) tend to evolve into something non-boolean. For this particular feature, I could imagine that we expose different granularities of recovery when we try to introduce recovery at the expression level later... but not sure.

Should we account for this and start with a `<never> | <auto>` enum?

lib/Translation_unit.ml
```diff
@@ -297,7 +298,10 @@ let format xunit ?output_file ~input_name ~source ~parsed conf opts =
         |> List.filter_map ~f:(fun (s, f_opt) ->
                Option.map f_opt ~f:(fun f -> (s, String.sexp_of_t f)))
       in
-      match parse xunit.parse conf ~source:fmted with
+      let fmted_parsable =
+        if opts.Conf.partial then xunit.make_parsable fmted else fmted
```
Julow: Shouldn't we do that only if the parse failed ?

ocamlformat-help.txt
```diff
@@ -579,6 +579,9 @@ OPTIONS
            making some syntactic gotchas visually obvious. The janestreet
            profile is used at Jane Street. The default value is default.
 
+       --partial
+           Ignore invalid (unparsable) parts of the input.
```
emillon: I think it'll be worth expanding a bit, explaining that formattable bits get formatted, and parts with errors are passed verbatim. Also this can mentioned that it is a pretty preliminary implementation.

lib/Translation_unit.mli
```diff
@@ -15,6 +15,7 @@ type 'a t =
   { init_cmts: debug:bool -> Source.t -> 'a -> Cmt.t list -> Cmts.t
   ; fmt: debug:bool -> Source.t -> Cmts.t -> Conf.t -> 'a -> Fmt.t
   ; parse: Lexing.lexbuf -> 'a
+  ; make_parsable: string -> string
```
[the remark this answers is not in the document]
gpetiot: I agree on renaming the function, but I don't see what having an option type here will achieve, it will force us to test both option values even though we know the appropriate recover function is always provided

lib/Translation_unit.ml
Julow: Also, this seems to ignore parsing errors. Should we show them ? (eg. as warnings)

lib/Conf.ml
gpetiot: Agreed, I renamed to `--format-invalid-files={never,auto}`

ocamlformat-help.txt
gpetiot: I improved the doc a bit

lib/Translation_unit.ml
gpetiot: If this option is set, I think the parsing errors should be ignored.
The code is written this way to avoid having to duplicate too much code and have a minimal impact on the current implementation. (as this PR is already hard enough to review)

Julow: Why would this make duplicated code or a big impact ?

Showing all the parsing errors can be unhelpful indeed. What about showing a single warning in this case ? (eg. "This file contains parsing errors and is not fully formatted")

If you are concerned about the emacs plugin opening buffers, note that we can modify it.

emillon: I think we can assume that errors are already reported by the editor.

vendor/parse-wyc/lib/parse_wyc.ml
```diff
@@ -0,0 +1,238 @@
...
+      attribute;
+      expr;
+      class_expr;
+      structure_item;
+      signature_item;
+    }
+  in
+  let mapper = make_mapper () in
+  ignore ((m mapper) ast);
+  normalize_locs !loc_list
```
Julow: This function is collecting the locs of invalid parts of the code.
But we have an ast here, why not use it instead of going back to a string in the next function ?
The `Ast_mapper` and `Stack` tricks used here can be used to map "structure item that contains an annot, deeply" to "invalid ast ext node" directly.

lib/Translation_unit.ml
Julow: Not necessarily, OCamlformat must parse the entire file so it's sensitive to all parsing errors that the editor won't see. OCamlformat may be called more often than the compiler.

Also, the editor will most probably always pass this option. The user may be confused by the "Applied OCamlformat" message but nothing changed.

vendor/parse-wyc/lib/parse_wyc.ml
gpetiot: I thought to have a string output was a simpler interface for ocamlformat, as it is what is processed in Translation_unit, and returning the ast types is tricky as we have an OMP layer which is not the same as the one of ocamlformat (Migrate_ast is duplicated)

Julow: Having two different AST/parser seems like something that could cause bugs. How hard to implement and what are the drawback to use a single implementation ?

Having control on the parser's code may solve a few issues too (eg. to modify it or to avoid changes in behavior between OMP's versions, which could cause formatting changes between the same version of OCamlformat)

gpetiot: It's necessary if we want this parser (pwyc) to be independent from ocamlformat, which is the what we agreed upon. Maybe the Migrate_ast could be factorized in the future, but I don't think it should be done in this PR.

lib/Translation_unit.ml
gpetiot: I applied the recovery only when the parsing failed and a warning is emitted (maybe more than once, should be fixed in another PR)

vendor/parse-wyc/lib/parse_wyc.ml
Julow: My question was about using the AST and parser from pwyc in OCamlformat, dropping OMP. The dependency stays the same. 
My question is, what are the drawbacks of doing that ?

lib/Translation_unit.ml
```diff
@@ -297,86 +299,108 @@ let format xunit ?output_file ~input_name ~source ~parsed conf opts =
...
+      | exception exn ->
+          if opts.Conf.format_invalid_files then (
+            match parse xunit.parse conf ~source:(xunit.recover fmted) with
+            | exception exn ->
+                internal_error (`Cannot_parse exn) (exn_args ())
+            | t_new ->
+                Format.fprintf Format.err_formatter
+                  "Warning: %s is invalid, recovering.\n%!" input_name ;
+                Ok t_new )
+          else internal_error (`Cannot_parse exn) (exn_args ())
```
Julow: This code is also in `parse_result` below. Let's share it.

lib/Translation_unit.ml
```diff
@@ -297,86 +299,108 @@ let format xunit ?output_file ~input_name ~source ~parsed conf opts =
...
+        (* All good, continue *)
+        print_check ~i:(i + 1) ~conf t_new ~source:fmted
   in
   try print_check ~i:1 ~conf parsed ~source with
   | Sys_error msg -> Error (User_error msg)
   | exn -> Error (Ocamlformat_bug {exn})
 
+let parse_result xunit conf (opts : Conf.opts) ~source ~input_name =
+  match parse xunit.parse conf ~source with
+  | exception exn ->
```
Julow: Looking at `format` above, `parse` could raise several kinds of exceptions:
```
      | exception Sys_error msg -> Error (User_error msg)
      | exception Warning50 l -> internal_error (`Warning50 l) (exn_args ())
      | exception exn -> internal_error (`Cannot_parse exn) (exn_args ())
```
The corresponding errors are no longer reported correctly. I'm not sure it makes sense to apply recovering when the first two (`Sys_error`, `Warning50`) happen.

emillon:
I'd like to explain the dual parser design a bit. The idea is generally what's described in #906. One important aspect is that it leave the secondary parser out of the happy path. This ensures that:
- we keep using the system parser and OMP for all valid code.
- we do not have to maintain a fork of the parser. When a new version comes out, updating the recovery parser is not on the critical path.
- this feature is easy to revert.
- the `string -> string` API allows a clear separation between the ocaml parser and the rest of ocamlformat. This is important for both technical and licensing reasons.

emillon:
I'd like to explain the dual parser design a bit. The idea is generally what's described in #906. One important aspect is that it leave the secondary parser out of the happy path. This ensures that:
- we keep using the system parser and OMP for all valid code.
- we do not have to maintain a fork of the parser. When a new version comes out, updating the recovery parser is not on the critical path.
- this feature is easy to revert.
- the `string -> string` API allows a clear separation between the ocaml parser and the rest of ocamlformat. This is important for both technical and licensing reasons.

lib/Translation_unit.mli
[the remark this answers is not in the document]
emillon: That was an API design comment. `string -> string` is great and it stay that way. Let me explain what I meant:

Suppose you're a user of that API and you want to create a value of type `'a t`. You have to provide a `make_parsable` function of type `string -> string` for your language right? Suppose you do not care about that and want to not use that feature. What should you pass? `fun x -> x`? `fun _ -> ""`? `fun _ -> assert false`? Difficult to tell without looking at the other uses! Asking for a `(string -> string) option` makes that a bit more obvious - you can pass `None`. It clarifies the fact that there's a natural default value. Of course, you need to add a `Option.value ~default:Fn.id` where you use that function. (it's also possible to keep storing `string -> string` but do this default value thing in a constructor)

emillon:
HUGE thanks to @gpetiot who's been tirelessly working on this feature for several months! :tada:

gpetiot:
> I think this is good to go. There are probably tons of ways we can improve this but as long as this works in is properly isolated this can happen on master.

Thanks! I will merge it now, but everyone can of course improve it in subsequent PRs

Julow:
This is awesome ! Congrats 

I feel like this has been merged too fast, however. I think the error handling may be broken and the code in `Translation_unit` needs improvement.

dune-project
```diff
@@ -13,3 +13,4 @@
 
 (name ocamlformat)
 (using fmt 1.2)
+(using menhir 2.1)
```
jberdine, line 16: @gpetiot  Is it only me for whom this causes dune to throw a warning:
```
File "dune-project", line 16, characters 14-17:
16 | (using menhir 2.1)
                   ^^^
Warning: Version 2.1 of the menhir extension is not supported until version
2.2 of the dune language.
Supported versions of this extension in version 1.11 of the dune language:
- 1.0 to 1.1
- 2.0
```

PR #1029, opened 2019-09-25
Fix unstable comments around attributes

### LINKED ISSUE #1025: Bug: unstable comment between attributes
**Describe the bug**
If I run ocamlformat on the following file...
```
module M = struct
  type t
  [@@immediate]
  (* ______________________________________ *)
  [@@deriving variants, sexp_of]
end
```

...I get the following error:
```
ocamlformat: Cannot process "app/apply-style/feedback/comment_between_attributes.ml".
```

**How to Reproduce**
This is using ocamlformat built from 63e1d3ab131ca8b5f2809c0d6fdfafdf749fe1b9, using the flags `--profile=janestreet --enable-outside-detected-project --disable-conf-files`.

gpetiot:
Fix #1025
also contains some simplfications to `fmt_type_declaration` that are small enough to be squeezed in this PR (I thought the issue was specific to type declarations at first so I started cleaning a bit).

No diff with test_branch with and without using the janestreet profile (the issue is due to the `margin=90` actually).

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3001,7 +3001,7 @@ and fmt_type_declaration c ?ext ?(pre = "") ctx ?fmt_name ?(eq = "=") decl =
   in
   update_config_maybe_disabled c ptype_loc ptype_attributes
   @@ fun c ->
-  let fmt_manifest ~priv manifest decl =
+  let fmt_manifest ~priv decl =
     let break_before_manifest_kind =
       match ptype_kind with
       | Ptype_abstract -> fmt "@ "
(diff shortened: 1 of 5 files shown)


src/Fmt_ast.ml
```diff
@@ -616,7 +616,8 @@ and fmt_attributes c ?(pre = noop) ?(suf = noop) ~key attrs =
   let num = List.length attrs in
   let fmt_attr ~first ~last {attr_name; attr_payload; attr_loc} =
     fmt_or_k first (open_hvbox 0) (fmt "@ ")
-    $ Cmts.fmt c attr_loc (fmt_attribute c key (attr_name, attr_payload))
+    $ hvbox 0
+        (Cmts.fmt c attr_loc (fmt_attribute c key (attr_name, attr_payload)))
```
gpetiot: The fix is here, other changes are simplifications.

gpetiot:

PR #1030, opened 2019-09-25
Fix dropped comment after infix op

### LINKED ISSUE #1028: Bug: Failure to format a file
**Describe the bug**
I accidentally stumbled upon this failure to format a file.

**How to Reproduce**
- What is the formatted file? Please try to minimize the code snippet.

```
let (>>=) = Caml.Result.bind

let f = (Ok ()) >>= (* *)
  (fun _ -> Ok ())
```

The problem seems to be that the comment gets dropped.

- What are the selected options?

```
(...)
break-infix-before-func=false (file .ocamlformat:21)
(...)
```

gpetiot:
Fix #1028. No diff with test_branch with `break-infix-before-func=false`


gpetiot:

Julow:
This does not stabilize:
```ocaml
[@@@ocamlformat "break-infix-before-func=false"]
let f =
  Ok () >>= 
  (* fooooooooooooooo fooooooooooooooo fooooooooooooooo foooooooooooooooo *)
  function
  | Foo -> Ok foooooooooooooooooooooooooooooooooooooooooooooooooo
```

This seems to be because of the `~adj:noop`

gpetiot:
> This seems to be because of the `~adj:noop`

Thanks, this is fixed now.

PR #1031, opened 2019-09-25
Fix: no newline if the input is empty

### LINKED ISSUE #1004: Bug: empty files get an empty line appended
**Describe the bug**

When formatting an empty file with ocamlformat, it append a trailing line.

**How to Reproduce**

```
$ touch ~/tmp/empty.ml
$ ocamlformat --enable-outside-detected-project -i ~/tmp/empty.ml
```

`empty.ml` ends up with an extra line in it. `wc -l empty.ml` shows different counts before and after formatting.

Tested against ocamlformat at e94fa814ae935f9084ef4e4e0a3b2b6127db230f.

gpetiot:
Fix #1004 (no regression with test_branch).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ### (master)
 
+  + Fix: no newline if the input is empty (#1031) (Guillaume Petiot)
   + Fix unstable comments around attributes (#1029) (Guillaume Petiot)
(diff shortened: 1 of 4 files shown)


gpetiot:

gpetiot:
> Looks good.
> 
> Does not work if the file only contains characters that would be removed (eg. spaces) but that's ok.

Fixed.

PR #1032, opened 2019-09-25
Fix test_branch.sh and CI checking of CHANGES.md

Julow:
`test_branch.sh` never report any diffs.
The path to ocamlformat binary was not passed correctly and it was using the same binary twice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - opam pin add --no-action ocamlformat .
   - opam install --deps-only ocamlformat
 script:
-  - ! git diff --exit-code "$TRAVIS_MERGE_BASE..$TRAVIS_PR_HEAD" -- CHANGES.md >/dev/null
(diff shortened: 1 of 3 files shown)


Julow:
I also fix the checking of CHANGES.md in CI.
Should be ready for review.

.travis.yml
```diff
@@ -18,7 +18,7 @@ install:
   - opam pin add --no-action ocamlformat .
   - opam install --deps-only ocamlformat
 script:
-  - ! git diff --exit-code "$TRAVIS_MERGE_BASE..$TRAVIS_PR_HEAD" -- CHANGES.md >/dev/null
+  - if [ $TRAVIS_BRANCH = master ]; then ! git diff --exit-code $TRAVIS_BRANCH...$TRAVIS_COMMIT -- CHANGES.md >/dev/null; fi
```
gpetiot: The build running after a PR is merged is failing, maybe we should also test `TRAVIS_PULL_REQUEST != false` to only check the changelog for PRs ?

Julow: Yes we should do that. I'll make a PR

PR #1033, opened 2019-09-26
Improve: use the same sets of options for both branches by default in test_branch.sh

gpetiot:
The goal is to avoid writing the same list of options twice (error-prone)

Tested on the docked-bindings branch:
- `test_branch.sh docked-branch` no options for both branches
- `test_branch.sh docked-branch '' ''` no options for both branches
- `test_branch.sh docked-branch ''` no options for both branches
- `test_branch.sh docked-branch 'some options' ''` some options only for the base branch
- `test_branch.sh docked-branch 'some options'` some options for both branches
- `test_branch.sh docked-branch 'some options' 'different options'` some options for the base branch, different options for the tested branch

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ### (master)
 
-  + Improve: use the same sets of options for both branches by default (#1033) (Guillaume Petiot)
+  + Improve: use the same sets of options for both branches by default in test_branch.sh (#1033) (Guillaume Petiot)
   + Fix test_branch.sh and CI checking of CHANGES.md (#1032, #1034) (Jules Aguillon)
   + Fix flag of git-worktree in test_branch.sh and bisect.sh (#1027) (Guillaume Petiot)
   + Improve: remove the bisect_ppx dependency and clean Makefile (#1005) (Jules Aguillon)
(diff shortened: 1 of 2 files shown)


tools/test_branch.sh
```diff
@@ -29,6 +31,14 @@ else
     branch=$(git rev-parse HEAD)
 fi
 
+opts_base="$2"
+
+if [[ $# -lt 3 ]]; then
+    opts_branch="$opts_base"
+else
+    opts_branch="$3"
+fi
```
Julow: This can be written as `opts_branch=${3-$opts_base}`

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ### (master)
 
+  + Improve: use the same sets of options for both branches by default (#1033) (Guillaume Petiot)
```
Julow: You should say it's for `test_branch.sh`

tools/test_branch.sh
gpetiot: I didn't know this syntax ([here is the doc](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02)), but I think the conditional is easier to read.

Issue #1035, opened 2019-09-28
Bug: ocamlformat_reason ~dev doesn't compile (opam 2.0.5 / ocaml 4.07.1)


RogerTarani:
Any idea for installing ocamlformat_reason ~dev?

Not ok with `ocamlformat_reason ~dev`:
```
$ opam pin add ocamlformat_reason https://github.com/ocaml-ppx/ocamlformat.git
Package ocamlformat_reason does not exist, create as a NEW package? [Y/n] y
[ocamlformat_reason.~dev] synchronised from git+https://github.com/ocaml-ppx/ocamlformat.git
ocamlformat_reason is now pinned to git+https://github.com/ocaml-ppx/ocamlformat.git (version ~dev)

The following actions will be performed:
  ∗ install ocamlformat_reason ~dev*
Do you want to continue? [Y/n] y
[ocamlformat_reason.~dev] synchronised from git+https://github.com/ocaml-ppx/ocamlformat.git

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of ocamlformat_reason failed at "/home/rt/.opam/opam-init/hooks/sandbox.sh build
        dune build -p ocamlformat_reason -j 1".

#=== ERROR while compiling ocamlformat_reason.~dev ============================#
# context     2.0.5 | linux/x86_64 | ocaml-base-compiler.4.07.1 | pinned(git+https://github.com/ocaml-ppx/ocamlformat.git#ce9e5393)
# path        ~/.opam/4.07.1/.opam-switch/build/ocamlformat_reason.~dev
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ocamlformat_reason -j 1
# exit-code   1
# env-file    ~/.opam/log/ocamlformat_reason-10335-d6d332.env
# output-file ~/.opam/log/ocamlformat_reason-10335-d6d332.out
### output ###
#        ocaml (internal) (exit 2)
# (cd test/passing && /home/rt/.opam/4.07.1/bin/ocaml -I +compiler-libs /home/rt/.opam/4.07.1/.opam-switch/build/ocamlformat_reason.~dev/_build/.dune/default/test/passing/dune.ml)
# File "test/passing/dune", line 54, characters 23-35:
# Error: Unbound module Option


<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
┌─ The following actions failed
[...]

PR #1036, opened 2019-09-30
CI: check the build with OCaml 4.07.1 and 4.08.0

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1035

Small tweak to build with 4.07.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: c
 env:
-  - OCAML_VERSION=4.09.0
+  - TO_TEST=install OCAML_VERSION=4.07.1
+  - TO_TEST=install OCAML_VERSION=4.08.0
+  - TO_TEST=tests OCAML_VERSION=4.09.0
 cache:
   directories:
     - ${HOME}/.opam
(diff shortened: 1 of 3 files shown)


gpetiot:
It looks good to me, if you approve my travis changes you can merge.

.travis.yml
```diff
@@ -1,27 +1,50 @@
...
-  - opam switch ${OCAML_VERSION}
-  - opam update --upgrade
-  - opam pin add --no-action ocamlformat .
-  - opam install --deps-only ocamlformat
+  - |
+    case "$TO_TEST" in
+      install|tests)
+        opam switch ${OCAML_VERSION}
+        opam update --upgrade
+        opam pin add --no-action ocamlformat .
```
Julow: We should do that below, in `script`. Otherwise it may be cached.

.travis.yml
```diff
@@ -1,27 +1,50 @@
...
+      install|tests)
+        wget -O ${HOME}/opam https://github.com/ocaml/opam/releases/download/2.0.2/opam-2.0.2-x86_64-linux
+        chmod +x ${HOME}/opam
+        export PATH=${HOME}:${PATH}
+        export OPAMYES=1
+        export OPAMJOBS=2
+        opam init --bare --disable-sandboxing
+        opam switch --dry-run ${OCAML_VERSION} 2>/dev/null || opam switch create ${OCAML_VERSION}
+        ;;
+    esac
```
Julow: You should add a case for the empty string, same in `install` and `script`.
If `TO_TEST` is not set correctly, nothing will happen, that should be an error.

gpetiot: done

.travis.yml
gpetiot: done

Issue #1037, opened 2019-09-30
Bug: "ast changed" error with certain infix operators


ceastlund:
**Describe the bug**
```
ocamlformat: Cannot process "ast_changed.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: ast changed.
```

**How to Reproduce**
`ocamlformat --profile=janestreet --enable-outside-detected-project --disable-conf-files ast_changed.ml`
produces:
```
(x.@(y) <- z) @ []
```

PR #1039, opened 2019-10-02
Fix indexing operators precedence

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1037

The precedence of indexing operators was not right (was Comma). Also, custom index operators are not represented the same as for array or string by the OCaml parser and were not handled correctly.


Julow:
The diff is from js_of_ocaml which uses `parens-tuple=multi-line-only` and `infix-precedence=parens`, so this is a fix.

PR #1046, opened 2019-10-03
Fix alignment of cases

### LINKED ISSUE #1041: Bug: Wrong alignment of patterns
On `master` with `--align-cases`:

```ocaml
let _ =
  match (foooooooooooooo, foooooooooooooo) with
  | Some a, Some b -> a + b
  | None, _            -> 1
  | Some _, None     -> 2
```

Arrows are not aligned.

An other example:

```ocaml
let _ =
  match (foooooooooooooo, foooooooooooooo) with
  | [x]           -> x
  | [_; x] -> x
  | _                    -> 0
```

gpetiot:
Fix #1041, the diff of `tools/test_branch.sh fix-align-cases 'align-cases=true'` is too big to check everything but a lot of cases are fixed.

Using the direct output of the formatting instead of duplicating the logic of what does this option do in this case, etc. or approximating.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -2812,7 +2812,9 @@ and fmt_cases c ctx cs =
   let pattern_len pat =
     let xpat = sub_pat ~ctx pat in
     let fmted = Cmts.preserve (fmt_pattern c) xpat in
-    if String.contains fmted '\n' then None else Some (String.length fmted)
+    let len = String.length fmted in
+    if len * 3 >= c.conf.margin || String.contains fmted '\n' then None
+    else Some len
   in
   let fold_pattern_len ~f ps =
     List.fold_until ~init:0 ps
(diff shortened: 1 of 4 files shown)


src/Fmt_ast.ml
```diff
@@ -2810,47 +2810,25 @@ and fmt_class_type_field c ctx (cf : class_type_field) =
...
-        None
-  and fold_pattern_len ?parens ?(init = 0) ~f ps =
-    List.fold_until ~init ps
-      ~f:(fun acc pat ->
-        match pattern_len ?parens pat with
+  let pattern_len {pc_lhs; pc_guard; _} =
+    if Option.is_some pc_guard then None
+    else
+      let xpat = sub_pat ~ctx pc_lhs in
+      let fmted = Cmts.preserve (fmt_pattern c) xpat in
```
Julow: We wanted to avoid doing that but I can't think of any valid reason. At least now it's more robust and handle every patterns.

PR #1050, opened 2019-10-07
Improve: inline extensions that do not break

### LINKED ISSUE #1048: Bug: top-level extension points are misindented based on what precedes them
**Describe the bug**
I get this formatting:
```
let this_function_has_a_long_name plus very many arguments = "and a kind of long body"

[%%expect
  {||}]
```
The `[%%expect {||}]` extension should be all on one line, not split onto two lines.

**How to Reproduce**
I got this using ocamlformat built from ce9e53933cf80fe164bc12dac31913d33b0f6e3d, using `--profile=janestreet`.

gpetiot:
Fix #1048 
We put a box around non-breaking payload of the extensions (the breaking ones must not be boxed)

I think in the future we should replace the calls to `Location.is_single_line` and their friends with `fits` whenever possible, so we rely on the output instead of the input.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -562,6 +562,14 @@ let sequence_blank_line c xe1 xe2 =
       b - a - commented_lines > 1
   | `Compact -> false
 
+let fits c x ~f =
+  let fmted = Cmts.preserve f x in
+  3 * String.length fmted < c.conf.margin
+  &&
+  match String.rindex fmted '\n' with
+  | Some i -> i = 0 (* if the first character is '\n' we ignore it *)
+  | None -> true
+
 let rec fmt_extension c ctx key (ext, pld) =
   match (pld, ctx) with
   | ( PStr [({pstr_desc= Pstr_value _ | Pstr_type _; _} as si)]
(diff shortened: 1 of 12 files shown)


src/Fmt_ast.ml
```diff
@@ -562,14 +562,24 @@ let sequence_blank_line c xe1 xe2 =
...
 let rec fmt_extension c ctx key (ext, pld) =
   match (pld, ctx) with
   | ( PStr [({pstr_desc= Pstr_value _ | Pstr_type _; _} as si)]
     , (Pld _ | Str _ | Top) ) ->
       fmt_structure_item c ~last:true ~ext (sub_str ~ctx si)
   | PSig [({psig_desc= Psig_type _; _} as si)], (Pld _ | Sig _ | Top) ->
       fmt_signature_item c ~ext (sub_sig ~ctx si)
-  | _ -> fmt_attribute_or_extension c key Fn.id (ext, pld)
+  | _ ->
+      let fits = fits c pld ~f:(fmt_payload c (Pld pld)) in
```
Julow: Why is this necessary ? Is this not possible to implement with `fits_breaks` ?

test/passing/doc_comments-after.ml.ref
```diff
@@ -232,8 +232,7 @@ end = struct
   (* [@@@some attribute] *)
   (* (** Attribute *) *)
 
-  (** Extension *)[%%some
-  extension]
+  (** Extension *)[%%some extension]
```
Julow: Not related but there should be a break after the doc comment.

src/Fmt_ast.ml
gpetiot: No because we would need to open the box before formatting (before using fits_breaks)

test/passing/doc_comments-after.ml.ref
gpetiot: It can indeed be fixed in another PR.

PR #1051, opened 2019-10-07
Fix: add missing cut before attributes in type declarations

### LINKED ISSUE #1049: Bug: "long but not long enough" doc comments mess with type attributes
## Describe the bug
"Not long enough" doc-comments make the type attribute's closing bracket wrap to the next line, along with the wrapping parenthesis, but indented really close to the margin

## How to Reproduce
(in the following snippets, the margin is indicated with a commented pipe below the declaration, those comments are irrelevant to the issue)

This snippet causes the issue:
```ocaml
type t =
  ([ `foo
   | `bar  (** 58 chars.................................................. *) ][@js.enum
                                                                       ])
                                                                             (* | *)

```

This one doesn't (included for comparison with expected output)
```ocaml
type t =
  ([ `foo
   | `bar  (** 59 chars................................................... *)
   ][@js.enum])
                                                                             (* | *)

```

### Config
Ocamlformat config file:
```
break-cases = fit
break-collection-expressions = wrap
break-fun-decl = wrap
break-infix = wrap
break-infix-before-func = false
break-separators = after-and-docked
break-string-literals = never
break-struct = force
doc-comments = after
extension-sugar = preserve
indicate-multiline-delimiters = no
infix-precedence = indent
let-and = compact
let-binding-spacing = compact
let-open = auto
module-item-spacing = compact
parens-tuple = multi-line-only
parens-tuple-patterns = multi-line-only
sequence-style = terminator
single-case = compact
type-decl = compact
if-then-else = keyword-first
field-space = loose
ocp-indent-compat = true
wrap-comments = true
parse-docstrings = true
align-variants-decl = true
```

Margin is at 80

gpetiot:
Fix #1049, no regression with test_branch.
The `@,` could be replaced by a `@ `, I don't have a preference between `foo[@ext]` and `foo [@ext]`, but nobody complained about the first style so far, so I'm keeping it like that.


Julow:
You need to add a CHANGES line.

PR #1054, opened 2019-10-08
Fix newlines and indentation in toplevel extension points

### LINKED ISSUE #1053: Bug: newlines in string literals in toplevel extension points are not considered for line breaking
**Describe the bug**
Below, the `%expect` has no newline before `{|`, but `%%expect` does put a newline before `{|`. The newline is unnecessary.
```
;;
[%expect {|
___________________________________________________________
|}]

[%%expect
  {|
___________________________________________________________
|}]
```

**How to Reproduce**
I get the above formatting using ocamlformat built from b28bb024fe633fe253b0b5e21fc55d877c1f8ac7 using `--profile=janestreet`.

gpetiot:
Fix #1053 
The linebreaks were incorrect, also we can be more precise in the box structure (no use for `fits` for now), and the option `stritem_extension_indent` should be taken into account to indent the payload of attributes, except for toplevel items. Also more consistency between structure and signature items.

No diff with test_branch for the default, the janestreet and the conventional profiles.

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -562,14 +562,6 @@ let sequence_blank_line c xe1 xe2 =
       b - a - commented_lines > 1
   | `Compact -> false
 
-let fits c x ~f =
-  let fmted = Cmts.preserve f x in
-  3 * String.length fmted < c.conf.margin
-  &&
-  match String.rindex fmted '\n' with
-  | Some i -> i = 0 (* if the first character is '\n' we ignore it *)
-  | None -> true
-
 let rec fmt_extension c ctx key (ext, pld) =
   match (pld, ctx) with
   | ( PStr [({pstr_desc= Pstr_value _ | Pstr_type _; _} as si)]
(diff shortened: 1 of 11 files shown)


src/Fmt_ast.ml
```diff
@@ -562,24 +562,14 @@ let sequence_blank_line c xe1 xe2 =
...
-
 let rec fmt_extension c ctx key (ext, pld) =
   match (pld, ctx) with
   | ( PStr [({pstr_desc= Pstr_value _ | Pstr_type _; _} as si)]
     , (Pld _ | Str _ | Top) ) ->
       fmt_structure_item c ~last:true ~ext (sub_str ~ctx si)
   | PSig [({psig_desc= Psig_type _; _} as si)], (Pld _ | Sig _ | Top) ->
       fmt_signature_item c ~ext (sub_sig ~ctx si)
-  | _ ->
-      let fits = fits c pld ~f:(fmt_payload c (Pld pld)) in
```
Julow: It's great that we can get rid of that.

PR #1056, opened 2019-10-08
Improve: extend option sequence-blank-line to let-bindings

### LINKED ISSUE #1047: Feature request: preserve more blank lines in function body
With option `sequence-blank-line = preserve-one` (added by https://github.com/ocaml-ppx/ocamlformat/pull/814), blank lines in sequences are preserved as expected, but other blank lines in function body are still "eaten".

For example (code adapted from Cohttp example):

```ocaml
let async_main =
  let* resp, raw_body = Client.get (Uri.of_string "https://api.github.com/") in

  let* body = Body.to_string raw_body in
  let code = resp |> Response.status |> Code.code_of_status in

  Printf.printf "Status: %d\n" code;
  Printf.printf "Body: %s\n" body;
  Lwt.return ()
```

becomes

```ocaml
let async_main =
  let* resp, raw_body = Client.get (Uri.of_string "https://api.github.com/") in
  let* body = Body.to_string raw_body in
  let code = resp |> Response.status |> Code.code_of_status in
  Printf.printf "Status: %d\n" code;
  Printf.printf "Body: %s\n" body;
  Lwt.return ()
```

It would be great if such blank lines are left intact, so that some blocks of logic can "stand out" by keeping "distance" from other blocks.

gpetiot:
This is WIP

Fix #1047 

Few issues:
- I suspect the location `body.pexp_loc` of `Pexp_letop {let_; ands; body}` is incorrect, I only checked by printing the location, the definition of letop_binding_body in parser.mly looks suspicious (should reuse more bits of let_binding_body) but I'm not really familiar enough with this code (any ideas @Julow ?)
- The location of the bindings don't include the `in` keyword, not a bug, but as a consequence this feature does not work if there is a `in` then an empty line separating 2 definitions (hence the TO-FIX-tests)


src/Fmt_ast.ml
```diff
@@ -2015,8 +2015,13 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       in
       let fmt_expr = fmt_expression c (sub_exp ~ctx body) in
       let parens = parens || not (List.is_empty pexp_attributes) in
+      let blank_line_after =
+        let fst = (List.last_exn bindings).pvb_loc in
+        blank_line_between c ~fst ~snd:body.pexp_loc
+      in
       fmt_let c ctx ~ext ~rec_flag ~bindings ~parens ~loc:pexp_loc
         ~attributes:pexp_attributes ~fmt_atrs ~fmt_expr ~indent_after_in
+        ~blank_line_after
```
Julow: `fmt_let` has enough context to compute that itself, it should be moved there.

ocamlformat-help.txt
```diff
@@ -201,6 +201,12 @@ OPTIONS (CODE FORMATTING STYLE)
            `=`). loose does. tight-decl is tight for declarations and loose
            for instantiations. The default value is tight.
 
+       --function-blank-line={compact|preserve-one}
```
Julow: Maybe "function" is not the right word, it apply to any expressions. I think "sequence" is great. A sequence in the AST is a `a ; b` expression but I think, from a high level perspective, the word can be also used for `let .. in ..`.

gpetiot: I reverted to keep the old name.

gpetiot:
Replaced by #1075 and #1077

PR #1059, opened 2019-10-09
Improve: remove utility functions from Fmt_ast

gpetiot:
Each commit can (an should) be reviewed separately, only internal API change, no behavioral change.

No diff with test_branch.

### DIFF
--- a/src/Params.ml
+++ b/src/Params.ml
@@ -352,3 +352,13 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
       ; expr_eol= None
       ; break_end_branch= noop
       ; space_between_branches= fmt "@ " }
+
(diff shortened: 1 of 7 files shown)


src/Params.ml
```diff
@@ -13,11 +13,16 @@ module Format = Format_
 open Migrate_ast
 open Fmt
 
-type exp_wrap = Fmt.t -> Fmt.t
```
Julow: Why remove that type ?
It was used to show that `get_exp_wrap` was returning a wrapping function and wasn't one itself.
But it was not implemented that way so it may makes sense.

gpetiot: Yes, and also it was used only once, if we want to abstract that and use a new type, `Fmt.ml` would be a better place to define it.

PR #1022, opened 2019-09-17
[WIP] Fix docked collection expressions in let-bindings

gpetiot:
Replaces #1008 
Fix #967 also fix #1009

PR #1060, opened 2019-10-09
Improve: set conventional as the default profile

gpetiot:
Wait for #1022 before updating the tests about `dock-collection-brackets`

Should we go further and rename the profiles (or add aliases) ?
- `conventional` -> `default`
- `ocamlformat` -> `legible` /  `left-align`


Julow:
A lot of boolean options changed default so on the command line they changed from `--option` to `--no-option`.
This even break our tests.
We should always have both the `--option` and `--no-option` flags available.

gpetiot:
> We should always have both the `--option` and `--no-option` flags available.

I agree

Julow:
Adding a `default` alias seems like a good idea. I don't like the `legible` and `left-align` names for `ocamlformat`, the first one is too opinionated and I don't understand the second, what makes this profile more "left aligned" than `conventional` ?

gpetiot:
> Adding a `default` alias seems like a good idea. I don't like the `legible` and `left-align` names for `ocamlformat`, the first one is too opinionated and I don't understand the second, what makes this profile more "left aligned" than `conventional` ?

It is based on the description of the profile written by Josh in the doc:

> The ocamlformat profile aims to take advantage of the strengths of a parsetree-based auto-formatter, and to limit the consequences of the weaknesses imposed by the current implementation. This is a style which optimizes for what the formatter can do best, rather than to match the style of any existing code. General guidelines that have directed the design include: Legibility, in the sense of making it as hard as possible for quick visual parsing to give the wrong interpretation, is of highest priority; Whenever possible the high-level structure of the code should be obvious by looking only at the left margin, in particular, it should not be necessary to visually jump from left to right hunting for critical keywords, tokens, etc; All else equal compact code is preferred as reading without scrolling is easier, so indentation or white space is avoided unless it helps legibility; Attention has been given to making some syntactic gotchas visually obvious.

gpetiot:
I removed the `legible` alias and updated the README, does it look good to you to merge?

PR #1062, opened 2019-10-09
Improve: allow both values of boolean options

Julow:
Currently, for an option `a`, only one of `--a` or `--no-a` is available on the CLI, depending on its default value.
This means:
- You can't override boolean options from the CLI (eg. if they are set by a `.ocamlformat` file or by a profile)
  Because the CLI flag to do that does not exist.
- The `--a` flag disapear from the CLI and is replaced by `--no-a` if the default value of `a` is changed to `true`
  This is problematic because https://github.com/ocaml-ppx/ocamlformat/pull/1060 changes the default of most options.

This PR makes both `--a` and `--no-a` available.

### DIFF
--- a/ocamlformat-help.txt
+++ b/ocamlformat-help.txt
@@ -525,7 +525,7 @@ OPTIONS
        --no-comment-check
            Unset comment-check.
 
-       --no-quiet, --no-q
+       --no-quiet
            Unset quiet.
 
        --no-version-check
--- a/ocamlformat-help.txt
+++ b/ocamlformat-help.txt
@@ -44,13 +44,14 @@ OPTIONS (CODE FORMATTING STYLE)
        are ignored and can be used as comments.
 
        --align-cases
-           Align match/try cases horizontally.
+           Align match/try cases horizontally. The default value is off.
 
        --align-constructors-decl
-           Align type declarations horizontally.
+           Align type declarations horizontally. The default value is off.
 
        --align-variants-decl
-           Align type variants declarations horizontally.
+           Align type variants declarations horizontally. The default value
+           is off.
 
        --assignment-operator={end-line|begin-line}
            Position of the assignment operator. end-line positions assignment
@@ -105,6 +106,12 @@ OPTIONS (CODE FORMATTING STYLE)
            fit-or-vertical vertically breaks expressions if they do not fit
            on a single line. The default value is wrap.
 
+       --break-infix-before-func
+           Break infix operators whose right arguments are anonymous
+           functions specially: do not break after the operator so that the
+           first line of the function appears docked at the end of line after
+           the operator. The default value is on.
+
        --break-separators={before|after}
            Break before or after separators such as `;` in list or record
            expressions. before breaks the expressions before the separator.
@@ -112,7 +119,8 @@ OPTIONS (CODE FORMATTING STYLE)
            value is before.
 
        --break-sequences
-           Force sequence expressions to break irrespective of margin.
+           Force sequence expressions to break irrespective of margin. The
+           default value is off.
 
        --break-string-literals={auto|never}
            Break string literals. auto mode breaks lines at newlines and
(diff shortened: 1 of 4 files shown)


src/Config_option.ml
```diff
@@ -84,20 +85,23 @@ module Make (C : CONFIG) = struct
       ?(allow_inline = Poly.(section = `Formatting)) ?(deprecated = false)
       update get_value =
     let open Cmdliner in
-    let invert_flag = default in
-    let names_for_cmdline =
-      if invert_flag then
-        List.filter_map names ~f:(fun n ->
-            if String.length n = 1 then None else Some ("no-" ^ n))
```
Julow: I removed this check. This means that there is a `--no-q` flag to disable `--q`. Should I implement it back ?

ocamlformat-help.txt
```diff
@@ -316,22 +327,67 @@ OPTIONS (CODE FORMATTING STYLE)
...
+       --no-break-sequences
+           Disable break-sequences.
+
+       --no-disable
+           Disable disable.
+
+       --no-disambiguate-non-breaking-match
+           Disable disambiguate-non-breaking-match.
+
+       --no-dock-collection-brackets
```
gpetiot: It would be nice to have `--(no-)dock-collection-brackets` (only 1 entry in the doc, but still 2 options), but judging from the Cmdliner API I'm not sure this line can be customized without changing the real name of the option

src/Config_option.ml
```diff
@@ -62,8 +62,9 @@ module Make (C : CONFIG) = struct
   let deprecated_doc ~deprecated =
     if deprecated then " Warning: This option is deprecated." else ""
 
-  let generated_flag_doc ~allow_inline ~doc ~section ~deprecated =
-    Format.sprintf "%s%s%s" doc
+  let generated_flag_doc ~allow_inline ~doc ~section ~default ~deprecated =
+    let default = if default then "on" else "off" in
+    Format.sprintf "%s The default value is $(b,%s).%s%s" doc default
```
gpetiot: I would write `This flag is set by default.` (unset)

src/Config_option.ml
```diff
@@ -84,20 +85,23 @@ module Make (C : CONFIG) = struct
...
-      if invert_flag then
-        List.filter_map names ~f:(fun n ->
-            if String.length n = 1 then None else Some ("no-" ^ n))
-      else names
+    let invert_names = List.map names ~f:(fun n -> "no-" ^ n) in
+    let doc =
+      generated_flag_doc ~allow_inline ~doc ~section ~default ~deprecated
     in
-    let doc = generated_flag_doc ~allow_inline ~doc ~section ~deprecated in
+    let invert_doc = "Disable $(b," ^ List.last_exn names ^ ")." in
```
gpetiot: `Unset the flag ...` or just `Unset ...` ?

ocamlformat-help.txt
```diff
@@ -143,11 +151,12 @@ OPTIONS (CODE FORMATTING STYLE)
        --disable
            Disable ocamlformat. This is used in attributes to locally disable
            automatic code formatting. One can also use [@@@ocamlformat
-           "enable"] instead of [@@@ocamlformat "disable=false"]
+           "enable"] instead of [@@@ocamlformat "disable=false"] The default
```
gpetiot: Punctuation `.` is missing

src/Config_option.ml
gpetiot: Yes I think we should keep it, `--no-q` looks weird IMO.

It was suggested in #299 to use `-Q` to negate `-q` for example, I'm not really a fan of this notation but if it makes sense to you it would be the right time to implement it.

src/Config_option.ml
Julow: Looks better !

ocamlformat-help.txt
Julow: I think that's not possible. I agree it does not look good

src/Config_option.ml
Julow: Did.

PR #1063, opened 2019-10-10
CI: use a script travis-ci.sh to simplify .travis.yml

gpetiot:
Closer to what is done for the ocaml compiler

I will test if the failing builds are correctly reported.


gpetiot:
Tested on #1063, I'm merging to benefit from it right now.

PR #1074, opened 2019-10-14
Internal: API for safe boxes

gpetiot:
I started working on a better way to use boxes:
- no box if there is less than 2 elements (we still need to detect `Fmt.noop` later), but can box 1 element if we use a flag (we should try to eliminate these cases later)
- enforce the use of a separator between 2 elements
- still compatible with the type `Fmt.t`

So far it is used inside a `Safe` module to distinguish with the legacy (soon unsafe) functions, the substitutions in legacy code would be done step by step as there is a lot of code to be modified, but it would be nice to use this API for bugfixes and new developments.

Add box names after #1083 is merged.

### DIFF
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -242,11 +242,23 @@ module Safe = struct
 
   let double_linebreak fs = Format.fprintf fs "\n@;<1000 0>"
 
-  type boxed = One of t | Cons of boxed * sep * t
+  type boxed =
+    | Empty (* this value is not exposed to not allow empty boxes *)
+    | One of t
+    | Cons of boxed * sep * t
 
   let one t = One t
 
   let add b s t = Cons (b, s, t)
+
+  let of_list xs sep ~f =
+    match xs with
+    | [] -> Empty
+    | [x] -> One (f x)
+    | h :: t ->
+        List.fold_left t
+          ~init:(One (f h))
+          ~f:(fun acc x -> Cons (acc, sep, f x))
 
   type box_kind = ?name:string -> int -> t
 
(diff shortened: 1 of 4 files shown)


src/Fmt.mli
```diff
@@ -163,44 +163,101 @@ val with_box_debug : t -> t
...
+  (** Conditionally wrap a format thunk with a vbox with specified
+      indentation. *)
+
+  val hvbox_if : ?name:string -> bool -> int -> box
+  (** Conditionally wrap a format thunk with an hvbox with specified
+      indentation. *)
+
+  val hovbox_if : ?name:string -> bool -> int -> box
+  (** Conditionally wrap a format thunk with an hovbox with specified
+      indentation. *)
```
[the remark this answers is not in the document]
gpetiot: It's better indeed, done.

gpetiot:
> Could you give an example of how it's used ?

I added an example in Fmt_ast (unfortunately it requires an additional parameter `~box_singleton` for this one)

src/Fmt_ast.ml
```diff
@@ -3316,7 +3316,8 @@ and fmt_signature c ctx itms =
     @@ fun c -> fmt_signature_item c (sub_sig ~ctx i)
   in
   let fmt_grp itms = list itms "@\n" fmt_grp in
-  hvbox 0 (list grps "\n@;<1000 0>" fmt_grp)
+  let open Safe in
+  box ~box_singleton:true hvbox 0 (of_list grps double_linebreak ~f:fmt_grp)
```
Julow: I'm not convinced by this example. With `~box_singleton:true`, this is exactly the same as before. This list should never be empty.

src/Fmt.ml
```diff
@@ -185,35 +188,96 @@ let debug_box_close fs =
...
+  let break n o fs = Format.pp_print_break fs n o
+
+  let linebreak o fs = Format.pp_print_break fs 1000 o
+
+  let double_linebreak fs = Format.fprintf fs "\n@;<1000 0>"
+
+  type boxed =
+    | Empty (* this value is not exposed to not allow empty boxes *)
+    | One of t
+    | Cons of boxed * sep * t
```
Julow: This allows `Cons (Empty, sep t)`. `Empty` should be out of this type, eg. `type boxed' = One of t | Cons of ..` and `type boxed = boxed' option`
This causes an `impossible` in the code below.

gpetiot: It should not happen since I didn't export the type `boxed` nor a way of construct `Empty` from the outside.

src/Fmt_ast.ml
gpetiot: Yes this example is not the best, I realized after modifying this part of the code that the box was needed even if there is only 1 element, but it is not the case for 99% of the other uses of boxes, I will write more examples.

src/Fmt.ml
Julow: `add (of_list [] noop ~f) s noop` wouldn't ?
Anyway, I meant that the representation allows it. Which make it less robust and you had to write `| Empty -> impossible "..."`.

gpetiot: Yes I agree noop is problematic, but it has to be removed to not be allowed in these boxes, I will have a look at removing most of them, but maybe in another PR.

gpetiot:
Closing for now

PR #1075, opened 2019-10-14
Improve and cleanup sequence_blank_line (fix empty_line_between)

Julow:
This PR does:
- Change the API of Source a little to avoid creating `Location.t` values.
  I think we should only read those values, we should manipulate pairs of `Lexing.position` instead.
- Use `Lexing` internals to set the starting location to avoid offsetting them.
- Fix `Source.empty_line_between` to be `true` if there is at least one empty line instead of if all lines are empty.
- Change `sequence_blank_line` to use `Source.empty_line_between`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ### (master)
 
+  + Internal: Cleanup sequence_blank_line (#1075) (Jules Aguillon)
   + Improve: set conventional as the default profile (#1060) (Guillaume Petiot)
   + Fix precedence of Dot wrt Hash (#1058) (Guillaume Petiot)
   + Fix break in variant type definition to not exceed the margin (#1064) (Guillaume Petiot)
--- a/src/Ast.ml
+++ b/src/Ast.ml
@@ -414,7 +414,7 @@ module Structure_item : Module_item with type t = structure_item = struct
     ||
     match Conf.(c1.module_item_spacing, c2.module_item_spacing) with
     | `Preserve, `Preserve ->
-        Source.empty_line_between s i1.pstr_loc i2.pstr_loc
+        Source.empty_line_between s i1.pstr_loc.loc_end i2.pstr_loc.loc_start
     | _ ->
         (not (is_simple (i1, c1)))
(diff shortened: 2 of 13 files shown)


gpetiot:
It looks good to me (I detailed the changelog a bit more), but there is a tiny diff with test_branch on js_of_ocaml:

```
diff --git a/examples/graph_viewer/svg.ml b/examples/graph_viewer/svg.ml
index 036517813..d85d3f19c 100644
--- a/examples/graph_viewer/svg.ml
+++ b/examples/graph_viewer/svg.ml
@@ -557,6 +557,7 @@ let _ =
     ();
   let slider = GRange.scale `HORIZONTAL ~draw_value:false ~packing:b#pack () in
   slider#adjustment#set_bounds ~lower:0. ~upper:1. ~step_incr:0.1 ();
+
   (*
   let button = GButton.check_button ~label:"Animate"
       ~packing:b#pack () in
diff --git a/examples/planet/planet.ml b/examples/planet/planet.ml
index 1322aba6c..876a3db6e 100644
--- a/examples/planet/planet.ml
+++ b/examples/planet/planet.ml
@@ -520,6 +520,7 @@ let draw ctx _img shd o _uv normals face_info dir =
         let e = (dy2 *. du3) -. (dy3 *. du2) in
         let f = y1 -. (d *. u1) -. (e *. v1) in
         ctx##transform a d b e c f;
+
         (*
 let (u1, v1) = uv.(v1) in
 let (u2, v2) = uv.(v2) in
```

There is no blank line before the comments in the original files, but there are blank lines inside the comments. Shouldn't èempty_line_between` check that the empty line is at the beginning or at the end of the lines?

Julow:
That's a tricky problem.
Checking only the first and last lines will not work in this case:

```ocaml
  (* This let will wrap *)
  let x =
    1
  in

  (* some comment *)
  next statement;
```

The solution may be to combine the previous algorithm that was looking at comments with `Source.empty_line_between`.

Julow:
I added a fix for that.

PR #1076, opened 2019-10-14
Build: use short form for action in src/dune

emillon:
Hi,

Just a small simplification:
- `(run)` always sets its first argument as dependency
- when the short form is used, targets are inferred from `(with-stdout-to)`

Thanks!


gpetiot:
Looks good to me too, I updated the CHANGES.md, let's wait a bit for the CLA and then we will merge.

PR #1078, opened 2019-10-16
Internal: clean Translation_unit

gpetiot:
No functional change, only refactoring/cleaning/trying to improve the code quality.
Each commit can be reviewed independently.

### DIFF
--- a/src/Translation_unit.ml
+++ b/src/Translation_unit.ml
@@ -91,7 +91,8 @@ let dump_formatted ~input_name ?output_file ~suffix fmted =
   else None
 
 let print_error ?(quiet_unstable = false) ?(quiet_comments = false)
-    ?(quiet_doc_comments = false) conf fmt input_name error =
+    ?(quiet_doc_comments = false) ?(fmt = Format.err_formatter) conf
+    ~input_name error =
   let exe = Filename.basename Sys.argv.(0) in
   let quiet_exn exn =
     match exn with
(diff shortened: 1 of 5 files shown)


src/Translation_unit.ml
```diff
@@ -91,7 +91,8 @@ let dump_formatted ~input_name ?output_file ~suffix fmted =
   else None
 
 let print_error ?(quiet_unstable = false) ?(quiet_comments = false)
-    ?(quiet_doc_comments = false) conf fmt input_name error =
+    ?(quiet_doc_comments = false) ?(fmt = Format.err_formatter) conf
+    ~input_name error =
```
Julow: `quiet_unstable`, `quiet_comments` and `quiet_doc_comments` are never used. They can be removed and `quiet_exn` below can be simplified to `false`.

src/Translation_unit.ml
```diff
@@ -365,23 +361,16 @@ let format xunit (conf : Conf.t) ?output_file ~input_name ~source ~parsed ()
...
-      | Ok t -> (
-        try print_check ~i:1 ~conf t ~source with
-        | Sys_error msg -> Error (User_error msg)
-        | exn -> Error (Ocamlformat_bug {exn}) )
-  in
-  ( match result with
-  | Ok _ -> ()
-  | Error e -> print_error conf Format.err_formatter input_name e ) ;
-  result
+  if conf.disable then Ok source
```
Julow: This function does not seem to do anything when `conf.disable` is `true`. Same when `parsed` is `Error _`. What do you think of moving these checks into `parse_and_format` below ?
Also, we can avoid parsing when `conf.disable` is `true`.

src/Translation_unit.ml
gpetiot: done

src/Translation_unit.ml
gpetiot: indeed we can, done.

src/Translation_unit.ml
```diff
@@ -365,23 +346,14 @@ let format xunit (conf : Conf.t) ?output_file ~input_name ~source ~parsed ()
...
-        try print_check ~i:1 ~conf t ~source with
-        | Sys_error msg -> Error (User_error msg)
-        | exn -> Error (Ocamlformat_bug {exn}) )
-  in
-  ( match result with
-  | Ok _ -> ()
-  | Error e -> print_error conf Format.err_formatter input_name e ) ;
-  result
+  match parsed with
+  | Error exn -> Error (Invalid_source {exn})
```
Julow: This should be checked in `parse_and_format` below.
Does not make sense to have a `result` here.

gpetiot: indeed, done

PR #1079, opened 2019-10-16
CI: factorize tests and check reason build

gpetiot:


tools/travis-ci.sh
```diff
@@ -64,9 +51,12 @@ CheckChangesModified () {
 case $CI_KIND in
 build)
     CheckBuild
+    opam install ocamlformat
```
Julow: Why were we testing `opam install` here and in `build-and-tests` ?

Issue #1072, opened 2019-10-10
Bug: over 90 characters with nested application


ceastlund:
**Describe the bug**
The inner application should be line-wrapped.
```
let _ =
  aa
    (bbbbbbbbb cccccccccccc dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd)
;;
```

**How to Reproduce**
ocamlformat built from 99444f3054e29e2322e56b23422e9984a378170f with `--profile=janestreet`

Issue #1071, opened 2019-10-10
Bug: string literal exceeds 90 columns without line-wrapping


ceastlund:
**Describe the bug**
This string literal exceeds 90 columns yet does not get line-wrapped. If I add one more underscore, it gets wrapped.
```
let _ =
  "_______________________________________________________ _______________________________"
;;
```

**How to Reproduce**
ocamlformat built from 99444f3054e29e2322e56b23422e9984a378170f with `--profile=janestreet`

Issue #1066, opened 2019-10-10
Bug: margin exceeded with nested match


ceastlund:
**Describe the bug**
The last line of this definition should be wrapped, but isn't:
```
let _ =
  match () with
  | _ ->
    (match () with
    | _ -> long_function_name long_argument_name__________________________________________)
;;
```

**How to Reproduce**
Using ocamlformat built from 99444f3054e29e2322e56b23422e9984a378170f:
`ocamlformat --profile=janestreet --inplace --enable-outside-detected-project formatting.ml`

PR #1081, opened 2019-10-16
Fix margin off by one

### LINKED ISSUE #1069: Bug: line over 90 characters with extension point
**Describe the bug**
The nested expression should be wrapped, the line exceeds 90 characters.
```
let _ =
  [%str
    let () = very_long_function_name__________________ very_long_argument_name____________]
;;
```

**How to Reproduce**
ocamlformat built from 99444f3054e29e2322e56b23422e9984a378170f using `--profile=janestreet`.

Julow:
Fix #1069, #1071, #1072 and #1066 

The margin was set to the configured margin + 1.


Julow:
I see. So this is not solving the problem, closing.
What is weird is that the examples in the issues all contains a lot of boxes and breaks and that reducing the margin by 1 make everything break correctly.

PR #1083, opened 2019-10-17
API: optional names for formatting boxes

gpetiot:
Independent from the new API in #1074, it would be nice to benefit from the box names as soon as possible.

### DIFF
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -166,14 +166,14 @@ let box_depth_color () =
 
 let debug_box_open ?name box_kind n fs =
   if !box_debug_enabled then (
-    let openning =
-      if n = 0 then box_kind else Format.sprintf "%s<%d" box_kind n
-    in
     let name =
-      Option.value_map name ~f:(fun s -> Format.sprintf "<%s>" s) ~default:""
+      Option.value_map name
+        ~f:(fun s -> Format.sprintf "%s:%s" box_kind s)
+        ~default:box_kind
     in
+    let openning = if n = 0 then name else Format.sprintf "%s<%d" name n in
     pp_color_k (box_depth_color ())
-      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>%s>" openning name)
+      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>>" openning)
       fs ;
     Int.incr box_depth )
 
--- a/src/Fmt.ml
+++ b/src/Fmt.ml
@@ -167,9 +167,9 @@ let box_depth_color () =
 let debug_box_open ?name box_kind n fs =
   if !box_debug_enabled then (
     let name =
-      Option.value_map name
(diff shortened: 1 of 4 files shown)


src/Fmt.ml
```diff
@@ -164,13 +164,16 @@ let box_depth_colors = [|32; 33; 94; 31; 35; 36|]
...
   if !box_debug_enabled then (
     let openning =
       if n = 0 then box_kind else Format.sprintf "%s<%d" box_kind n
     in
+    let name =
+      Option.value_map name ~f:(fun s -> Format.sprintf "<%s>" s) ~default:""
+    in
     pp_color_k (box_depth_color ())
-      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>>" openning)
+      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>%s>" openning name)
```
Julow:
```suggestion
      (fun fs -> Format.fprintf fs "@<0>[@<0>%s@<0>%s@<0>>" openning name)
```
Otherwise the closing `>` will be counted and formatting may be different due to the margin.

src/Fmt.ml
```diff
@@ -164,13 +164,16 @@ let box_depth_colors = [|32; 33; 94; 31; 35; 36|]
...
   box_depth_colors.(!box_depth % Array.length box_depth_colors)
 
-let debug_box_open box_kind n fs =
+let debug_box_open ?name box_kind n fs =
   if !box_debug_enabled then (
     let openning =
       if n = 0 then box_kind else Format.sprintf "%s<%d" box_kind n
     in
+    let name =
+      Option.value_map name ~f:(fun s -> Format.sprintf "<%s>" s) ~default:""
```
Julow: This will look like `[hv<name>>` or `[hv<2<name>>`.
What do you think of simply using a single char (eg. space or `:`), so it looks like `[hv name>` or `[hv<2 name>` ?

src/Fmt.ml
gpetiot: Fixed

src/Fmt.ml
gpetiot: Fixed

src/Fmt.ml
```diff
@@ -164,11 +164,14 @@ let box_depth_colors = [|32; 33; 94; 31; 35; 36|]
...
 
-let debug_box_open box_kind n fs =
+let debug_box_open ?name box_kind n fs =
   if !box_debug_enabled then (
-    let openning =
-      if n = 0 then box_kind else Format.sprintf "%s<%d" box_kind n
+    let name =
+      Option.value_map name
+        ~f:(fun s -> Format.sprintf "%s:%s" box_kind s)
+        ~default:box_kind
```
Julow: Nitpick: I think using a match instead of `Option.value_map` is both easier to type and to read.

gpetiot: done

PR #1084, opened 2019-10-17
Build: check ocamlformat error codes

emillon:
This ensures that runtest fails even before trying to diff - it also avoids displaying the actual error message as a diff.

Test cases that should fail come with an empty `x.ml.should-fail` file.

### DIFF
--- a/test/passing/gen/gen.ml
+++ b/test/passing/gen/gen.ml
@@ -4,13 +4,20 @@ type setup =
   { mutable has_ref: bool
   ; mutable has_opts: bool
   ; mutable base_file: string option
-  ; mutable extra_deps: string list }
+  ; mutable extra_deps: string list
+  ; mutable should_fail: bool }
 
 let read_lines fn =
   Stdio.In_channel.with_file fn ~f:Stdio.In_channel.input_lines
 
 let add_test ?base_file map src_test_name =
-  let s = {has_ref= false; has_opts= false; base_file; extra_deps= []} in
+  let s =
+    { has_ref= false
+    ; has_opts= false
+    ; base_file
+    ; extra_deps= []
+    ; should_fail= false }
+  in
   map := StringMap.add src_test_name s !map ;
   s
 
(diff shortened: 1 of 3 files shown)


test/passing/gen/gen.ml
```diff
@@ -51,20 +59,24 @@ let emit_test test_name setup =
...
+  in
   Printf.printf
     {|
 (rule
  (targets %s.output)
  (deps .ocamlformat %s)
  (action
    (with-outputs-to %%{targets}
-     (system "%%{bin:ocamlformat}%s %%{dep:%s} || true"))))
+     (system "%s%s"))))
```
Julow: Could switch to `run` when `should_fail` is not set but that would make the code slightly more complicated.

emillon: I tried, but unfortunately the `%{read-lines}` expands to a single string with spaces in between, so ocamlformat gets a large string argument. We could solve that by not doing `%{read-lines}` and generating the actual options in `dune.inc`, but that makes a larger file.

Julow: I see. I think that's OK, we can merge like this,

test/passing/gen/gen.ml
```diff
@@ -51,20 +59,24 @@ let emit_test test_name setup =
     match setup.base_file with Some n -> n | None -> test_name
   in
   let extra_deps = String.concat " " setup.extra_deps in
+  let cmd_prefix = if setup.should_fail then "! " else "" in
+  let cmd =
+    Printf.sprintf "%%{bin:ocamlformat}%s %%{dep:%s}" opts base_test_name
```
Julow: `cmd_prefix` could be printed here instead, so there is less arguments to the printf below.
Also, this could be calling `fprintf` and be a `%a` in the format string below to avoid the intermediate string:
```ocaml
let cmd pp () = ...
```

emillon: This is not performance sensitive, so I did this to move arguments closer to their specification. Anyway, I reverted to how the code was looking originally, that works.

Julow:
I think we can merge this

PR #1089, opened 2019-10-21
Internal: add an API to parse removed configuration values

emillon:
These trigger a custom error, making sure that existing users have an error that is more actionable than "invalid value".

An error message is added for `break-separators=after-and-docked`.

### DIFF
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -402,6 +402,14 @@ module Formatting = struct
            particular, with escape sequences expanded." ) ]
     in
     C.choice ~names ~all ~doc ~section
+      ~removed_values:
+        (Config_option.Removed_value.make_list
+           ~names:["newlines"; "newlines-and-wrap"; "wrap"]
+           ~version:"0.12"
+           ~msg:
+             "It has been replaced by the new default `auto` value, which \
+              breaks lines at newlines and wraps string literals at the \
+              margin.")
       (fun conf x -> {conf with break_string_literals= x})
       (fun conf -> conf.break_string_literals)
 
--- a/src/Conf.ml
+++ b/src/Conf.ml
@@ -369,6 +369,13 @@ module Formatting = struct
         , "$(b,before) breaks the expressions before the separator." ) ]
     in
     C.choice ~names ~all ~doc ~section
+      ~removed_values:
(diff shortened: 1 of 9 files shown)


src/Conf.ml
```diff
@@ -369,6 +369,13 @@ module Formatting = struct
         , "$(b,before) breaks the expressions before the separator." ) ]
     in
     C.choice ~names ~all ~doc ~section
+      ~removed_values:
```
emillon: @Julow is this suggestion accurate?

src/Config_option.ml
```diff
@@ -9,6 +9,26 @@
  *                                                                    *
  **********************************************************************)
 
+module Removed_value = struct
```
Julow: What do you think of moving this module inside `Make` ? So users can access it with same paths as other things of this module.

src/Conf.ml
Julow: You also need `space-around-lists=false`, otherwise there will be a break before the closing `]`.

Julow:
An other option changed, in dd340040, `break-string-literals=auto` replaces `newlines-and-wrap`, `wrap`and `newlines` and we expect diffs for `wrap` and `newlines` users.

src/Config_option.ml
emillon: Functors bodies tend to attract a lot of things, which makes them become harder and harder to understand as the code base grows. I was actually thinking to move this to its own file to keep the interface clear, what do you think about that?

emillon:
Thanks, I'll add these

src/Config_option.ml
Julow: Maybe that would make a lot of files for a single concern ?
I don't think growing the body of functors is a bad thing, why would it be worse than growing a simple module ?
I was suggesting that so the usage is consistent, `C.choice` vs `Config_option.Removed_value.make`. `C.Removed_value` looks better to me because it's obvious from where it come.

src/Conf.ml
emillon: done!

src/Config_option.ml
Julow: Removing the functor would be better indeed :)

PR #1090, opened 2019-10-22
Internal: replace pre_break and if_newline by cbreak

gpetiot:
- export a function `Fmt.cbreak` to have a direct interface to the swiss army knife `custom_break` function from Format, the objective is to use this function instead of the `if_newline`, `or_newline` and `fits_breaks` functions to get closer to the original Format library. If we manage to achieve this, only `max-indent` would have to either be merged into Format or me implemented in another way.
- some calls to `if_newline` are removed
- all calls to `pre_break` are removed (often used in addition to `if_newline` so they can be merged by using `cbreak`)
- the diff in `doc_comments` is a fix as we now reach the 77 col margin

No diff with the `conventional` and `ocamlformat` profiles.

diff with `janestreet` profile: (we now reach the 90 col margin)
```diff
diff --git a/infer/src/integration/Driver.ml b/infer/src/integration/Driver.ml
index 2b6f00462..44161fc96 100644
--- a/infer/src/integration/Driver.ml
+++ b/infer/src/integration/Driver.ml
@@ -158,11 +158,7 @@ let command_error_handling ~always_die ~prog ~args = function
         L.external_error
       else L.die InternalError
     in
-    log
-      "%a:@\n  %s"
-      Pp.cli_args
-      (prog :: args)
-      (Unix.Exit_or_signal.to_string_hum status)
+    log "%a:@\n  %s" Pp.cli_args (prog :: args) (Unix.Exit_or_signal.to_string_hum status)
 ;;
diff --git a/src/jbuild_support/string_with_vars.ml b/src/jbuild_support/string_with_vars.ml
index 67455697..f77d875c 100644
--- a/src/jbuild_support/string_with_vars.ml
+++ b/src/jbuild_support/string_with_vars.ml
@@ -95,9 +95,8 @@ module Upgrade_var = struct
     let static_vars =
       [ ( "<"
         , Deleted
-            "Use a named dependency instead:\n\n\
-            \  (deps (:x <dep>) ...)\n\
-            \   ... %{x} ..." )
+            "Use a named dependency instead:\n\n  (deps (:x <dep>) ...)\n   ... %{x} ..."
+        )
       ; "@", Renamed_to "targets"
```

Regression with test_branch with no option selected:
```diff
diff --git a/infer/src/clang/cFrontend.ml b/infer/src/clang/cFrontend.ml
index 26631c440..38fd771db 100644
--- a/infer/src/clang/cFrontend.ml
+++ b/infer/src/clang/cFrontend.ml
@@ -50,8 +50,7 @@ let do_source_file (translation_unit_context : CFrontend_config.translation_unit
   L.(debug Capture Verbose)
     "@\n Start building call/cfg graph for '%a'....@\n" SourceFile.pp source_file ;
   let cfg = compute_icfg translation_unit_context tenv ast in
-  L.(debug Capture Verbose)
-    "@\n End building call/cfg graph for '%a'.@\n" SourceFile.pp source_file ;
+  L.(debug Capture Verbose) "@\n End building call/cfg graph for '%a'.@\n" SourceFile.pp source_file ;
   (* This part below is a boilerplate in every frontends. *)
   (* This could be moved in the cfg_infer module *)
   NullabilityPreanalysis.analysis cfg tenv ;
```

### DIFF
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -127,8 +127,8 @@ let rec fmt_inline_elements elements =
     | tl -> wrap opn cls (hd $ fmt_inline_elements (space_elt :: tl))
   in
   let rec aux space = function
-    | [] -> space
-    | `Space _ :: t -> aux (fmt "@ ") t
+    | [] -> fmt_opt space
+    | `Space _ :: t -> aux (Some (fmt "@ ")) t
     | `Word w :: t ->
         (* Escape lines starting with '+' or '-' *)
         let escape =
@@ -136,20 +136,20 @@ let rec fmt_inline_elements elements =
             "\\"
           else ""
         in
-        cbreak ~fits:("", 1, "") ~breaks:("", 0, escape)
-        $ str_normalized w $ aux noop t
+        opt space (fun _ -> cbreak ~fits:("", 1, "") ~breaks:("", 0, escape))
+        $ str_normalized w $ aux None t
     | `Code_span s :: t ->
         let s = escape_brackets s in
-        space
+        fmt_opt space
         $ hovbox 0 (wrap "[" "]" (str_normalized ~escape:escape_brackets s))
-        $ aux noop t
+        $ aux None t
     | `Raw_markup (lang, s) :: t ->
         let lang =
           match lang with
           | Some l -> str_normalized l $ str ":"
           | None -> noop
         in
-        space $ wrap "{%%" "%%}" (lang $ str s) $ aux noop t
+        fmt_opt space $ wrap "{%%" "%%}" (lang $ str s) $ aux None t
     | `Styled (style, elems) :: t ->
         let s =
           match style with
@@ -159,17 +159,21 @@ let rec fmt_inline_elements elements =
           | `Superscript -> "^"
           | `Subscript -> "_"
         in
-        space
+        fmt_opt space
         $ wrap_elements "{" "}" ~always_wrap:true (str_normalized s) elems
-        $ aux noop t
+        $ aux None t
     | `Reference (_kind, rf, txt) :: t ->
         let rf = wrap "{!" "}" (fmt_reference rf) in
-        space $ wrap_elements "{" "}" ~always_wrap:false rf txt $ aux noop t
+        fmt_opt space
+        $ wrap_elements "{" "}" ~always_wrap:false rf txt
+        $ aux None t
     | `Link (url, txt) :: t ->
         let url = wrap "{:" "}" (str_normalized url) in
-        space $ wrap_elements "{" "}" ~always_wrap:false url txt $ aux noop t
+        fmt_opt space
+        $ wrap_elements "{" "}" ~always_wrap:false url txt
+        $ aux None t
   in
-  aux noop (List.map elements ~f:(ign_loc ~f:Fn.id))
+  aux None (List.map elements ~f:(ign_loc ~f:Fn.id))
 
 and fmt_nestable_block_element c = function
   | `Paragraph elems -> fmt_inline_elements elems
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -126,30 +126,30 @@ let rec fmt_inline_elements elements =
     | [] -> wrap_if always_wrap opn cls hd
     | tl -> wrap opn cls (hd $ fmt_inline_elements (space_elt :: tl))
   in
-  let rec aux space = function
-    | [] -> fmt_opt space
-    | `Space _ :: t -> aux (Some (fmt "@ ")) t
-    | `Word w :: t ->
+  let rec aux = function
+    | [] -> noop
+    | `Space _ :: `Word w :: t ->
         (* Escape lines starting with '+' or '-' *)
         let escape =
           if String.length w > 0 && Char.(w.[0] = '+' || w.[0] = '-') then
             "\\"
           else ""
         in
-        opt space (fun _ -> cbreak ~fits:("", 1, "") ~breaks:("", 0, escape))
-        $ str_normalized w $ aux None t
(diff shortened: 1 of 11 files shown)


gpetiot:
It turns out this bug in the regression is the same as other margin bugs: the 2 characters ` ;` are printed before we get a chance to break the line.
Do we want to delay this one until this is fixed? For that we would need to move `Fmt.t` elements from one box to another.

test/passing/doc_comments-after.ml.ref
```diff
@@ -256,12 +256,12 @@ exception A of int
 
 module A = struct
   module B = struct
-    (** It does not try to saturate (2) A = B + C /\ B = D + E => A = C + D
-        \+ E Nor combine more than 2 equations (3) A = B + C /\ B = D + E /\
-        F = C + D + E => A = F
+    (** It does not try to saturate (2) A = B + C /\ B = D + E => A = C + D +
```
Julow: This test is broken, the tested feature was to escape the `+` at the beginning of a line.
It fails.

src/Fmt_ast.ml
```diff
@@ -206,10 +206,10 @@ let rec fmt_longident (li : Longident.t) =
   | Lapply (li1, li2) ->
       hvbox 2 (fmt_longident li1 $ wrap "@,(" ")" (fmt_longident li2))
 
-let fmt_longident_loc c ?(pre = noop) {txt; loc} =
-  Cmts.fmt c loc (pre $ fmt_longident txt)
+let fmt_longident_loc c ?pre {txt; loc} =
+  Cmts.fmt c loc (fmt_opt pre $ fmt_longident txt)
```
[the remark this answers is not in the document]
gpetiot: I agree we can move that in another PR.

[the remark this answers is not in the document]
Julow: It's not necessary to move that to an other PR. I just needed confirmation.

test/passing/doc_comments-after.ml.ref
Julow: I didn't manage to fix that.
An other break hint is inserted before the `cbreak` you added, it seems, so it's never breaking.
I'd suggest to revert this changes.

src/Fmt_odoc.ml
```diff
@@ -243,7 +252,8 @@ let fmt_block_element c = function
         | None -> fmt ""
       in
       hovbox 0
-        (wrap "{" "}" (str lvl $ lbl $ fmt "@ " $ fmt_inline_elements elems))
+        (wrap "{" "}"
+           (str lvl $ lbl $ fmt_inline_elements ~space:true elems))
```
Julow: This space is not optional, is this to handle the case when `elems` is empty or only spaces ? In that case, I think that should be handled upfront and not with a flag.

test/passing/doc_comments-after.ml.ref
gpetiot: I will have a look

gpetiot: Somehow the output can get closer to the margin, and since there is a break between each word we just break after `+` instead of before, I just made the part before the original break longer to output the escape character.

lib/Fmt_odoc.ml
```diff
@@ -120,50 +118,62 @@ let list_block_elem elems f =
...
+let space_elt : inline_element with_location =
+  Location_.(at (span []) (`Space ""))
 
-and fmt_inline_elements txt = list txt "" (ign_loc ~f:fmt_inline_element)
+let rec fmt_inline_elements elements =
+  let wrap_elements opn cls ~always_wrap hd = function
+    | [] -> wrap_if always_wrap opn cls hd
+    | tl -> wrap opn cls (hd $ fmt_inline_elements (space_elt :: tl))
+  in
+  let rec aux space = function
```
Julow: This `space` thing feels cumbersome and is only useful for escaping words. What do you think of using the match we already have and match on `` `Space _ :: `Word w :: tl `` to implement the escaping ?

lib/Fmt_odoc.ml
```diff
@@ -120,50 +118,62 @@ let list_block_elem elems f =
...
-      | [] -> url
-      | txt -> wrap "{" "}" (url $ fmt "@ " $ fmt_inline_elements txt) )
+let space_elt : inline_element with_location =
+  Location_.(at (span []) (`Space ""))
 
-and fmt_inline_elements txt = list txt "" (ign_loc ~f:fmt_inline_element)
+let rec fmt_inline_elements elements =
+  let wrap_elements opn cls ~always_wrap hd = function
+    | [] -> wrap_if always_wrap opn cls hd
+    | tl -> wrap opn cls (hd $ fmt_inline_elements (space_elt :: tl))
```
Julow: Can you add a test case where `space_elt` is useful ? Eg. ` ... {b <margin here>+ 2} ...`

lib/Fmt_odoc.ml
gpetiot: It's better indeed, done.

lib/Fmt_odoc.ml
gpetiot: done

test/passing/doc_comments.mli.ref
```diff
@@ -403,3 +403,6 @@ end
 
 (** {e foooooooo oooooooooo ooooooooo ooooooooo} {{!some ref} fooooooooooooo
     oooooooo oooooooooo} {b fooooooooooooo oooooooooooo oooooo ooooooo} *)
+
+(** foooooooooooooooooooooooooooooooooooooooooooooooooo foooooooooooo {b eee
+    \+ eee eee} *)
```
Julow: Could you also add a test with the word to escape (`+`) is the first in the `{}` ? As it involves the `space_elt` hack.

gpetiot: done

Issue #1088, opened 2019-10-18
Bug: comment within assignment, formatting does not stabilize


ceastlund:
**Describe the bug**
This code...
```
let _ =
  aaaaaaa
  (* __________________________________________________________________________________ *)
  := bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
;;
```
...gives this error:
```
ocamlformat: Cannot process "comment_inside_assignment.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: formatting did not stabilize after 10 iterations.
```

**How to Reproduce**
Run ocamlformat with `--profile=janestreet`.

gpetiot:
Fixed with #1093

PR #1093, opened 2019-10-23
Fix unstabilizing comments on assignments

emillon:
Just a test case for #1088 for now.


gpetiot:
No diff with test_branch (default profile and janestreet profile), no issue with the `<-` operator.

src/Fmt_ast.ml
```diff
@@ -1458,10 +1453,19 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       , [(Nolabel, r); (Nolabel, v)] )
     when is_simple c.conf width (sub_exp ~ctx r) ->
       Cmts.relocate c.cmts ~src:pexp_loc ~before:loc ~after:loc ;
+      let break_before, break_after =
+        match c.conf.assignment_operator with
+        | `Begin_line -> (break 1 2, char ' ')
+        | `End_line -> (char ' ', break 1 2)
+      in
```
Julow: This part is unnecessary. It's the same as before.

test/passing/assign_comment.ml.ref
```diff
@@ -0,0 +1,35 @@
+(* see https://github.com/ocaml-ppx/ocamlformat/issues/1088 *)
+
+let _ =
+  r
+    (* _________________________________________________________________ *) :=
```
Julow: This exceed the margin (78/77)

gpetiot: fixed

emillon:
When I started this PR I didn't know about `js_source`, maybe the test file can be merged there? I like small test cases but since all JS examples are already in a single place, it makes sense to keep it there.

In any case, this can be merged I think.

gpetiot:
> When I started this PR I didn't know about `js_source`, maybe the test file can be merged there? I like small test cases but since all JS examples are already in a single place, it makes sense to keep it there.
> 
> In any case, this can be merged I think.

I think it's better here to have a smaller file so we can test both values of the option on a small code snippet.

test/passing/assign_comment-begin_line.ml.ref
```diff
@@ -0,0 +1,35 @@
...
+
+let _ =
+  r
+    (* _________________________________________________________________ *)
+    (* _________________________________________________________________ *)
+    := 1
+
+let _ =
+  r
+    := (* _________________________________________________________________ *)
```
Julow: This comment exceed the margin by 1. Breaking the line won't help so maybe it's ok ?

src/Fmt_ast.ml
```diff
@@ -1466,11 +1461,23 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       , [(Nolabel, r); (Nolabel, v)] )
     when is_simple c.conf (expression_width c) (sub_exp ~ctx r) ->
       Cmts.relocate c.cmts ~src:pexp_loc ~before:loc ~after:loc ;
+      let adj = fmt_if Poly.(c.conf.assignment_operator = `End_line) "@," in
```
Julow: Should move this line inside the `cmts_before` let.

src/Fmt_ast.ml
```diff
@@ -1466,11 +1461,23 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
-           $ Cmts.fmt c loc (fmt_assign_colon c)
-           $ hvbox 2 (fmt_expression c (sub_exp ~ctx v)) ))
+           ( match c.conf.assignment_operator with
+           | `Begin_line ->
+               hvbox 0 (fmt_expression c (sub_exp ~ctx r) $ cmts_before)
+               $ fmt "@;<1 2>:= " $ cmts_after
+               $ hvbox 2 (fmt_expression c (sub_exp ~ctx v))
+           | `End_line ->
+               hvbox 0
+                 (fmt_expression c (sub_exp ~ctx r) $ cmts_before $ fmt " :=")
```
Julow:
```suggestion
                 (fmt_expression c (sub_exp ~ctx r) $ cmts_before $ str " :=")
```

test/passing/assign_comment-begin_line.ml.ref
gpetiot: Yes there is already enough breaks before and after the comment, at some point the user needs to put spaces in comments :D

gpetiot:
I reorganized the tests, I agree `Pexp_setfield` and `Pexp_setinstvar` could be handled in a separate PR.

PR #1095, opened 2019-10-24
Internal: un-vendor Format and remove dead code

gpetiot:
Some simplifications in case we want to redesign the code to better fit our needs, since we know which features are used anyway.


emillon:
I think it'd be better to first move this under `src/`, delete `vendor/` (& update the opam metadata), then do this changes (I think it can be done all in a PR).

gpetiot:
> I think it'd be better to first move this under `src/`, delete `vendor/` (& update the opam metadata), then do this changes (I think it can be done all in a PR).

I rewrote this PR to do this.

src/.ocamlformat-ignore
```diff
@@ -0,0 +1,2 @@
+Format_.ml
+Format_.mli
```
gpetiot: The docstrings are too esoteric to be formatted.

gpetiot:
Because of legal reasons this code and its modifications won't be moved out of the `vendor` directory.
I can still apply these changes (removing dead code) in the vendored directory like it was originally done.

In another PR, I will try to reduce the usage of non-legacy functions (`if_newline`, `or_newline`, `fits_breaks`) to see if it's possible to get closer to the standard Format library. (a solution has to be found for `max-indent`)

PR #1096, opened 2019-10-24
Fix closing quote exceeding the margin

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1071

The closing `"` was in a box with no break.
The first commit fix the bug, the second is only cleanup.


test/passing/js_source.ml.ref
```diff
@@ -9807,7 +9807,8 @@ let _ =
 
 (* FIX: exceed 90 columns *)
```
emillon: Can you remove this line?

Julow: Done.

PR #1097, opened 2019-10-25
WIP: Add examples for each configuration option

emillon:
This will help understand each option value and should be accessible via `odig`.


emillon:
Something like this, but just for profiles (or a demo applet) would be better. Closing.

PR #1099, opened 2019-10-25
Add Box abstraction

emillon:
A box is something that runs before and after a `Fmt.t`. This ensures that the corresponding "open" and "close" part are properly balanced by always maintaining them together.

(This might be superseded by #1074)

### DIFF
--- a/src/Box.ml
+++ b/src/Box.ml
@@ -1,0 +1,30 @@
+(**********************************************************************
+ *                                                                    *
+ *                            OCamlFormat                             *
+ *                                                                    *
+ *  Copyright (c) 2017-present, Facebook, Inc.  All rights reserved.  *
+ *                                                                    *
+ *  This source code is licensed under the MIT license found in the   *
+ *  LICENSE file in the root directory of this source tree.           *
+ *                                                                    *
+ **********************************************************************)
+type t = {opn: Fmt.t; cls: Fmt.t}
+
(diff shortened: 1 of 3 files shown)


Julow:
Note that we already use functional boxes (eg. `hvbox : int -> t -> t`) and the `compose` function can be easily implemented on `t -> t`.

Only the code formatting modules uses imperative open/close functions and this is not making it safe (a little safer but not safe) and requires extra `Box.unsafe_` functions.

emillon:
I think that the main two benefits to this approach are:
- `Box.t` is a smaller type than `t -> t`, so we can introspect these more easily
- the unsafe API is explicit. it's easier to determine where the weird parts are compared to the existing API.

emillon:
(this is not for immediate merge and won't solve all our issues but it's a good starting point to discuss some improvements - the API can be made lighter as well)

Julow:
>     * the unsafe API is explicit. it's easier to determine where the weird parts are compared to the existing API.

I don't agree. The unsafe part of `Box` is used the way as the existing API. If it's about the name, we can more rename change `open_*box` into `unsafe_open_*box`, same for `close_box`.

I don't think we have an abstraction problem. The unsafe open/close functions are only used marginally and will be removed in the future.

Julow:
That makes sense. But I suggest to change this code to not use these functions instead. We tried before and did not succeed, though.

emillon:
We'll find something better, closing.

PR #1102, opened 2019-10-27
Fix: adjust definition of Location.is_single_line to reflect margin

jberdine:
This seems to be needed to be consistent with recent changes to the
handling of the margin. Otherwise, it is possible to have a one-liner
structure item of just the right width that is_single_line does not
hold, resulting in the one-liner being surrounded by open lines.

### DIFF
--- a/test/failing/single_line.mli
+++ b/test/failing/single_line.mli
@@ -1,0 +1,6 @@
+[@@@ocamlformat "module-item-spacing=compact"]
+
+val xx_xxxxxxxx : t -> bool
+val xx_xxxxxxxx : t -> bool
+val xxxxxxxx : t -> [> `Xxxxxxx | `Xxxxxxxxxxx | `Xxxxxxxxxx | `Xxxxxxxxxxxxx]
+val xxxxx : t -> t -> t Xxx.t option
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,7 @@
   + Fix grouping of horizontally aligned comments (#1209) (Guillaume Petiot)
   + Fix dropped comments around module pack expressions (#1214) (Jules Aguillon)
   + Fix regression of comment position in list patterns (#1141) (Josh Berdine)
+  + Fix: adjust definition of Location.is_single_line to reflect margin (#1102) (Josh Berdine)
 
 #### Documentation
 
--- a/lib/Migrate_ast.ml
+++ b/lib/Migrate_ast.ml
@@ -189,6 +189,8 @@ module Location = struct
     Comparable.lexicographic [compare_start; descending compare_end; compare]
 
   let is_single_line x margin =
+    (* The last character of a line can exceed the margin if it is not
+       preceded by a break. Adding 1 here is a workaround for this bug. *)
     width x <= margin + 1 && x.loc_start.pos_lnum = x.loc_end.pos_lnum
 
   let smallest loc stack =
--- a/lib/Migrate_ast.ml
+++ b/lib/Migrate_ast.ml
@@ -189,7 +189,7 @@ module Location = struct
     Comparable.lexicographic [compare_start; descending compare_end; compare]
 
   let is_single_line x margin =
-    width x <= margin && x.loc_start.pos_lnum = x.loc_end.pos_lnum
+    width x <= margin + 1 && x.loc_start.pos_lnum = x.loc_end.pos_lnum
 
   let smallest loc stack =
     let min a b = if width a < width b then a else b in


gpetiot:
> Otherwise, it is possible to have a one-liner
> structure item of just the right width that is_single_line does not
> hold, resulting in the one-liner being surrounded by open lines.

Do you have an example for this?

jberdine:
Yes, IIRC one is:
```
val xx_xxxxxxxx : t -> bool
val xxxxxxxx : t -> [> `Xxxxxx | `Xxxxxxxxxxx | `Xxxxxxxxxx | `Xxxxxxxxxxxxx]
val xxxxx : t -> t -> t Xxx.t option
```
with a margin of 77.

gpetiot:
I tested it on master, the output of `is_single_line` is true for each item of this file.

jberdine:
It seems that `module-item-spacing=compact` is also needed. I added a test file that should be unchanged, and is with this PR, but isn't on master.

gpetiot:
`val xxxxxxxx : t -> [> `Xxxxxxx | `Xxxxxxxxxxx | `Xxxxxxxxxx | `Xxxxxxxxxxxxx]` is 88 characters long but it doesn't break, like #1069 (square brackets), we can start by fixing this.

gpetiot:
I rebased the branch, there is no diff with test_branch except with the `janestreet` profile:

```diff
diff --git a/compiler/num-testsuite/test_big_ints.ml b/compiler/num-testsuite/test_big_ints.ml
index c2390ae3a..2475eb912 100644
--- a/compiler/num-testsuite/test_big_ints.ml
+++ b/compiler/num-testsuite/test_big_ints.ml
@@ -103,10 +103,8 @@ test
 
 ;;
 test 16 eq_big_int (add_big_int (big_int_of_int (-1)) (big_int_of_int 2), big_int_of_int 1)
-
 ;;
 test 17 eq_big_int (add_big_int (big_int_of_int 2) (big_int_of_int (-1)), big_int_of_int 1)
-
 ;;
 testing_function "sub_big_int"
 ;;
@@ -166,7 +164,6 @@ test
 
 ;;
 test 17 eq_big_int (sub_big_int (big_int_of_int 2) (big_int_of_int (-1)), big_int_of_int 3)
-
 ;;
 testing_function "mult_int_big_int"
 ;;
@@ -946,7 +943,6 @@ test
 testing_function "shift_right_big_int"
 ;;
 test 1 eq_big_int (shift_right_big_int unit_big_int 0, unit_big_int)
-
 ;;
 test 2 eq_big_int (shift_right_big_int (big_int_of_int 12345678) 3, big_int_of_int 1543209)
 
diff --git a/compiler/num-testsuite/test_nums.ml b/compiler/num-testsuite/test_nums.ml
index d4fc822a6..469841bcf 100644
--- a/compiler/num-testsuite/test_nums.ml
+++ b/compiler/num-testsuite/test_nums.ml
@@ -97,7 +97,6 @@ test 10 eq_num (sub_num (Int (-1)) (Int biggest_int), pred_num (Int least_int))
 testing_function "mult_num"
 ;;
 test 1 eq_num (mult_num (Int 2) (Int 3), Int 6)
-
 ;;
 test 2 eq_num (mult_num (Int 127) (Int (int_of_string "257")), Int (int_of_string "32639"))
 
diff --git a/compiler/num-testsuite/test_ratios.ml b/compiler/num-testsuite/test_ratios.ml
index 32a8fa7f5..40d53626a 100644
--- a/compiler/num-testsuite/test_ratios.ml
+++ b/compiler/num-testsuite/test_ratios.ml
@@ -74,10 +74,8 @@ test 9 eq_big_int (numerator_ratio r, big_int_of_int 0)
 
 ;;
 testing_function "null_denominator"
-
 ;;
 test 1 eq (null_denominator (create_ratio (big_int_of_int 1) (big_int_of_int (-2))), false)
-
 ;;
 test 2 eq (null_denominator (create_ratio (big_int_of_int 1) zero_big_int), true)
```

The longer lines of this diff exceed the 90 column margin, but it was already the case in master (related to #1066, #1069).
So maybe we can merge it afterall @Julow ?

test/failing/single_line.mli
```diff
@@ -0,0 +1,6 @@
+[@@@ocamlformat "module-item-spacing=compact"]
+
+val xx_xxxxxxxx : t -> bool
+val xx_xxxxxxxx : t -> bool
+val xxxxxxxx : t -> [> `Xxxxxxx | `Xxxxxxxxxxx | `Xxxxxxxxxx | `Xxxxxxxxxxxxx]
```
Julow: This is exceeding the margin, this is why `is_single_line` was `false`.
I think we should comment in the code that the `+ 1` is a workaround a bug.

PR #866, opened 2019-05-29
Fix dropped comments in list patterns and module types 

gpetiot:
On the way to fixing #744 

Diff of test_branch: (the new position of the comments is identical to the source file)
```ocaml
diff --git a/infer/src/IR/Cfg.ml b/infer/src/IR/Cfg.ml
index 112e1e32e..9b1e2d591 100644
--- a/infer/src/IR/Cfg.ml
+++ b/infer/src/IR/Cfg.ml
@@ -68,14 +68,14 @@ let inline_synthetic_method ((ret_id, _) as ret) etl pdesc loc_call : Sil.instr
   in
   let do_instr instr =
     match (instr, etl) with
-    | Sil.Load (_, Exp.Lfield (Exp.Var _, fn, ft), bt, _), [(* getter for fields *) (e1, _)] ->
+    | Sil.Load (_, Exp.Lfield (Exp.Var _, fn, ft), bt, _), (* getter for fields *) [(e1, _)] ->
         let instr' = Sil.Load (ret_id, Exp.Lfield (e1, fn, ft), bt, loc_call) in
         found instr instr'
     | Sil.Load (_, Exp.Lfield (Exp.Lvar pvar, fn, ft), bt, _), [] when Pvar.is_global pvar ->
         (* getter for static fields *)
         let instr' = Sil.Load (ret_id, Exp.Lfield (Exp.Lvar pvar, fn, ft), bt, loc_call) in
         found instr instr'
-    | Sil.Store (Exp.Lfield (_, fn, ft), bt, _, _), [(* setter for fields *) (e1, _); (e2, _)] ->
+    | Sil.Store (Exp.Lfield (_, fn, ft), bt, _, _), (* setter for fields *) [(e1, _); (e2, _)] ->
         let instr' = Sil.Store (Exp.Lfield (e1, fn, ft), bt, e2, loc_call) in
         found instr instr'
     | Sil.Store (Exp.Lfield (Exp.Lvar pvar, fn, ft), bt, _, _), [(e1, _)] when Pvar.is_global pvar
```

PR #1104, opened 2019-10-27
[RFC] regression quick fixes

jberdine:
There are some regressions that I don't know the best fix that fits into the overall plans. I made this PR in order to get something usable, and to be more concrete than opening issues, but the changes themselves are probably not the eventually wanted ones. See the individual commit messages for more info on each.


gpetiot:
Do you have some code samples that are impacted by 108b96e ? I removed this code since it looked redundant with the generic treatment of sequences and test_branch didn't show any changes.

jberdine:
I wonder if you used a version of infer that is reasonably recent when running test_branch. There are a number of changes, for example around code of the form `[%Trace.call ...] ; ...`.

src/Fmt_ast.ml
```diff
@@ -906,8 +906,8 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
           $ fmt_or_k last p.sep_after_final p.sep_after_non_final
         in
         hvbox 0
-          (Cmts.fmt c ppat_loc
-             (p.box
+          (p.box
+             (Cmts.fmt c ppat_loc
```
Julow: This patch cause an unstable comment in `test/passing/comments.ml`.

jberdine: FYI, this commit is separated out in https://github.com/ocaml-ppx/ocamlformat/pull/1141

Do you know if that test was stable before https://github.com/ocaml-ppx/ocamlformat/pull/866 ? I don't know if there is some intended interaction between the other changes in that PR and this one, or if this change was just added to that PR by mistake.

src/Fmt_ast.ml
```diff
@@ -1697,7 +1698,12 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                              | Some i when i <= 2 -> "@ "
                              | _ -> "@;<1 2>" ) )
                        $ fmt_expression c ?box xbody )
-                   $ str ")" $ Cmts.fmt_after c pexp_loc )
+                   $ fmt_or_k
+                       Poly.(c.conf.indicate_multiline_delimiters = `Space)
+                       (fmt_or_k fit (str ")")
+                          (fits_breaks ~force_fit_if:fit ")" ~hint:(1, 0) ")"))
+                       (fits_breaks ~force_fit_if:fit ")" ~hint:(0, 0) ")")
```
Julow: I don't remember why this parentheses are gone. @gpetiot what do you think ?
Could it be that there was no space after the opening ?
eg, it formats like this:
```
   foo (fun () ->
       ...
       abc )

  foo
     (let a = ... in
      a)
```

jberdine: In such cases the opening space was originally taken to be one following `(fun`.

gpetiot: I don't remember either, and I think it was decided it was okay to sometimes have no space at the beginning but to have a space at the end, especially for `(fun` even if `indicate-multiline-delimiters=space`. I didn't manage to find the issue/PR where this have been discussed. There have been a lot of PRs to improve the consistency of parentheses, but maybe some regressions have been introduced along the way. I'm favorable to improve the consistency anyway.

src/Fmt_ast.ml
Julow: I think before that PR, it was a dropped comment. The test is added by this PR.

jberdine: Ok. The high-level question is if, knowing the details of 866, you know of a way to not drop this comment without the regression on the example in https://github.com/ocaml-ppx/ocamlformat/issues/1140 ?

src/Fmt_ast.ml
Julow: I moved that commit to a PR: https://github.com/ocaml-ppx/ocamlformat/pull/1169

gpetiot:
@jberdine can you rebase your PR so we can have a clearer view? I think a lot of things are now fixed in master.

jberdine:
These changes are now covered by fixes in master, https://github.com/ocaml-ppx/ocamlformat/pull/1169, https://github.com/ocaml-ppx/ocamlformat/pull/1342, and https://github.com/ocaml-ppx/ocamlformat/pull/1343.

PR #1105, opened 2019-10-28
Fix the default value documentation for max-indent

### LINKED ISSUE #1101: Bug: `max-indent = none` in .ocamlformat fails
**Describe the bug**
Adding a `max-indent = none` line to .ocamlformat leads to a parse failure:
```
  ("bad value for" (max-indent "invalid value `none', expected an integer")))
Raised at file "src/error.ml" (inlined), line 9, characters 14-30
Called from file "src/error.ml", line 11, characters 19-40
Called from file "src/exn.ml", line 66, characters 8-11
Re-raised at file "src/exn.ml", line 69, characters 6-15
Called from file "src/Conf.ml", line 1904, characters 6-1023
Called from file "list.ml", line 117, characters 24-34
Called from file "src/list0.ml" (inlined), line 22, characters 40-81
Called from file "src/Conf.ml", line 2054, characters 4-45
Called from file "src/Conf.ml", line 99, characters 14-18
Called from file "src/Conf.ml", line 2091, characters 4-61
Called from file "src/Conf.ml", line 2134, characters 20-69
```

gpetiot:
Fix #1101 

This option is defined as an `int option` instead of an `int` on purpose, to avoid calling `set_margin` with an arbitrary value if the user doesn't choose any, and I think it's better to keep it that way.
So the default value displayed in this PR should only be visible in the documentation, and is not used in the predefined profiles (they use `None` since it's an option type).


gpetiot:
> Why not accepting `none` as a value ?

Because that means accepting a string instead of a int (and not having Cmdliner checks), at least now the user knows which value is really used by default.

Julow:
We can easily make a converter that can accept ints and "none".

src/Conf.ml
```diff
@@ -877,8 +877,12 @@ module Formatting = struct
       "Maximum offset ($(docv) columns) added to a new line in addition to \
        the offset of the previous line."
     in
+    (* creating a fresh formatter in case the value of max-indent has been
+       changed for stdout. *)
+    let fs = Format.formatter_of_buffer (Buffer.create 4) in
+    let none = Int.to_string (Format.pp_get_max_indent fs ()) in
```
emillon: What do you think about pulling this into a top-level `default_max_indent_value`? also `4` looks like it's meaningful here - if it's not I suggest passing `0`.

gpetiot: done.

gpetiot:
Can we merge this one?

emillon:
Yes, let's :ship: this one!

Issue #989, opened 2019-08-28
Bug: Unstable comment in variant decl


Julow:
This snippet does not stabilize after 10 iterations:

```ocaml
type a = b

and b =
  | A (* A *)
  | B (* B *)
    (* C ? *)
  | D (* D *)
    (* E *)
```

PR #1108, opened 2019-10-29
Fix unstable comment in variant decl

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/989

There is a regression:

Surprisingly, it appears only once with `test_branch`

```diff
-type blabla = Constr_a (* constr a *) | Constr_b
+type blabla =
+  | Constr_a
+  (* constr a *)
+  | Constr_b
```

The example from the issue formats badly, but that's unrelated:

```diff
 and b =
   | A (* A *)
   | B (* B *)
-    (* C ? *)
-  | D (* D *)
-    (* E *)
+  (* C ? *)
+  | D
+
+(* D *)
+(* E *)
```

### DIFF
--- a/src/Fmt_ast.ml
+++ b/src/Fmt_ast.ml
@@ -3091,6 +3091,8 @@ and fmt_constructor_declaration c ctx ~max_len_name ~first ~last:_ cstr_decl
   in
   let has_cmt_before = Cmts.has_before c.cmts pcd_loc in
   let sparse = Poly.( = ) c.conf.type_decl `Sparse in
+  (* Force break if comment before pcd_loc, it would interfere with an
+     eventual comment placed after the previous constructor *)
   fmt_if_k (not first) (fmt_or (sparse || has_cmt_before) "@;<1000 0>" "@ ")
   $ Cmts.fmt_before ~epi:(break 1000 0) c pcd_loc
   $ Cmts.fmt_before c loc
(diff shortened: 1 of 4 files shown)


gpetiot:
I tried to improve the formatting without success, I think it's acceptable to break whenever there are comments.

emillon:
Yes, breaking seems fine. This does not affect `(** *)` comments, right?

Julow:
> Yes, breaking seems fine. This does not affect `(** *)` comments, right?

It doesn't.

Julow:
Just rebased. Should we merge ?

src/Fmt_ast.ml
```diff
@@ -3089,11 +3089,13 @@ and fmt_constructor_declaration c ctx ~max_len_name ~first ~last:_ cstr_decl
...
          (fits_breaks ~level:3 "" pad))
   in
-  fmt_if (not first)
-    (match c.conf.type_decl with `Sparse -> "@;<1000 0>" | `Compact -> "@ ")
-  $ Cmts.fmt_before c pcd_loc $ Cmts.fmt_before c loc
+  let has_cmt_before = Cmts.has_before c.cmts pcd_loc in
+  let sparse = Poly.( = ) c.conf.type_decl `Sparse in
+  fmt_if_k (not first) (fmt_or (sparse || has_cmt_before) "@;<1000 0>" "@ ")
+  $ Cmts.fmt_before ~epi:(break 1000 0) c pcd_loc
+  $ Cmts.fmt_before c loc
```
gpetiot: It's fine for me but could you add a detailed comment before those changes, and also reference this PR#, in case we need to revert it when we get a more precise comment attachment.

Julow: Done.

PR #1109, opened 2019-10-29
Internal: Make Fmt.t abstract

Julow:
Follow-up on https://github.com/ocaml-ppx/ocamlformat/pull/1106

Added `fmt_opt` to format `Fmt.t option` values, it replaces `Option.call ~f:t`.

Also added `Fmt.eval` to run the formatting and `Fmt.protect` to implement the `protect` function in `Fmt_ast.ml`

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -408,8 +408,9 @@ let init map_ast source asts comments_n_docstrings =
           if not (Location.compare loc Location.none = 0) then
             Hashtbl.set t.remaining ~key:loc ~data:()) ;
     if Conf.debug then (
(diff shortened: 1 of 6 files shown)


src/Cmts.ml
```diff
@@ -108,8 +108,7 @@ end = struct
                    (not (List.is_empty children))
                    "@,{" " }" (dump_ tree children) )))
     in
-    if Conf.debug then set_margin 100000000 $ dump_ tree tree.roots
-    else Fn.const ()
+    if Conf.debug then set_margin 100000000 $ dump_ tree tree.roots else noop
```
emillon: `Fmt.fmt_if_k`?

PR #1110, opened 2019-10-29
Improve: add margin check

### LINKED ISSUE #1107: Feature request: automatic margin check
We could easily implement an automatic check, like after the final formatting iteration when a checkpoint is found we can check the length of each line of the file, and print a warning on stderr.

gpetiot:
Fix #1107
before merging this we would need to redirect stderr to other files in the test, otherwise the warnings may appear during inplace formatting and generate incorrect files.
Edit: actually we don't need it until it is set by default for the tests (that can wait)


emillon:
This is useful but I think it should wait 0.13.

src/Translation_unit.ml
```diff
@@ -230,6 +230,13 @@ let check_all_locations fmt cmts_t =
           "Warning: Some locations have not been considered\n%!" ;
         List.iter ~f:print (List.sort l ~compare:Location.compare)
 
+let check_margin (conf : Conf.t) ~filename ~fmted =
+  if conf.margin_check then
+    List.iteri (String.split_lines fmted) ~f:(fun i line ->
+        if String.length line > conf.margin then
+          Format.fprintf Format.err_formatter
+            "Warning: %s:%i exceeds the margin\n%!" filename i)
```
Julow: This does not print warnings the same way as `Compat.print_warning`. That would require using ocaml's internal warning system, though. Maybe we need our own formatting function for warning/error messages ?

src/Conf.ml
```diff
@@ -844,6 +845,15 @@ module Formatting = struct
       (fun conf x -> {conf with margin= x})
       (fun conf -> conf.margin)
 
+  let margin_check =
+    let doc =
+      "Check that the formatted output does not exceed the margin."
```
Julow: Could you say that it's not an error but a warning?

gpetiot:
Should we "hide" this option like `--check` for example? (not in the `Conf.t` type and not listed in `ocamlformat --help`)

I think it's more of a developer option and it makes no sense of having it in profiles.

Issue #1092, opened 2019-10-22
Bug: regular comments surrounding a doc-comment, ocamlformat cannot process


smiley325:
**Describe the bug**
This code...
````
type t =
  { a : int
  (* oops I wrote a long comment before and after which is going to break the doc-comment *)
  (** [a] is an integer and I like it very much *)
  (* oops I wrote a long comment before and after which is going to break the doc-comment *)
  ; b : int
  }
````

...causes ocamlformat to produce

````
ocamlformat: Cannot process "ocamlformat_woe.mli".
   Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
   BUG: generating invalid ocaml syntax.
````

While the original code arguably _is_ wrong in some (which?) sense, I would rather ocamlformat complain about it as a rule rather than erroring out...

**How to Reproduce**
Run ocamlformat with `--profile=janestreet`.

emillon:
Hi Andrew!

I agree that this error message could be improved a bit.

I'm trying to reproduce this error, but I'm getting a different one ("BUG: misplaced documentation comments."). What version of ocamlformat are you using?

Thanks!

gpetiot:
This error is indeed raised because of the output of ocamlformat and is not in the input.
For some reason it was not allowed to attach comments to docstrings (I don't have the history for that but removing it causes regressions). Hopefully it will be easier to tackle if the comments end up in the AST.

PR #1111, opened 2019-10-29
Improve: give a hint when warning 50 is raised

gpetiot:
Improve the error message in case of warning 50 in the input source (related to #1092 but the warning 50 is produced by the formatting).

### DIFF
--- a/test/passing/warning50.ml.ref
+++ b/test/passing/warning50.ml.ref
@@ -1,0 +1,6 @@
+ocamlformat: Cannot process "warning50.ml".
+  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
+  BUG: misplaced documentation comments.
+File "warning50.ml", line 8, characters 8-56:
+Warning 50: unattached documentation comment (ignored)
+Hint: (Warning 50) disable this check using the `--no-comment-check` option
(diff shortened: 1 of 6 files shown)


test/passing/warning50.ml.ref
```diff
@@ -0,0 +1,6 @@
+ocamlformat: Cannot process "warning50.ml".
+  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
+  BUG: misplaced documentation comments.
+File "warning50.ml", line 8, characters 8-56:
+Warning 50: unattached documentation comment (ignored)
+Hint: (Warning 50) disable this check using the `--no-comment-check` option
```
emillon: I think that this hint will be useful. But I'm not sure whether the whole output is actionable.

When this happens, what should the user do? Should they report the bug, as prompted? It seems to me that ocamlformat is not designed to handle files that emit instances of warning 50. If that is correct, I suggest replacing it with something like that:

> This file contains a documentation comment `(** ... *)` that the OCaml compiler does not know how to attach to the AST. OCamlformat does not support these cases. You can find more information at: <https://github.com/ocaml-ppx/ocamlformat#overview>. If you'd like to disable this check and let ocamlformat make a choice (though it might not be consistent with the ocaml compilers and odoc), you can set the `--no-comment-check` option.

Julow:
This error can be triggered by warning 50 from parsing the input file. In this case, `--no-comment-check` make sense but the part asking to report the bug don't.

This error can also be triggered when parsing again the result of the previous iteration. In this case, we should not hint to use `--no-comment-check` and we should ask to report the bug.

emillon:
That's interesting. I think that we should take that piece of info into account to generate the correct error message and the corresponding outcome (report this as a bug vs pointer to doc and option).

If this info is not available I'd like to take that as an opportunity to test and rework `Translation_unit` so that it's possible!

PR #1117, opened 2019-11-04
Add command line tests

Julow:
So we can avoid regressions.

Also adds a few tests for invalid combinations of options.
It's not complete, there is a lot more to test.


test/cli/gen/gen.ml
```diff
@@ -0,0 +1,61 @@
...
+      update_or_add test_name ~f:(fun e -> { e with should_fail = true })
+  | _ ->
+      entries
+
+let check_test test_name entry ok =
+  let e ok b =
+    if b then Format.kfprintf (fun _ -> false) Format.err_formatter
+    else Format.ikfprintf (fun _ -> ok) Format.err_formatter
+  in
+  let ok = e ok (not entry.has_ref) "@{<error>Error@}: Missing file %s.ref\n" test_name in
```
emillon: I think you can simplify this part by using `invalid_arg` as it's only used to exit with an error message

test/cli/dune
```diff
@@ -0,0 +1,10 @@
+(include dune.inc)
```
emillon: dune file generation is a bit heavy for these kind of tests, I'll try to refactor this in a future PR.

test/cli/gen/gen.ml
Julow: I did that way to print all the warnings before exiting. Do you think I should remove that feature ?

test/cli/dune
Julow: You mean it's slowing dune ? `dune runtest` already take a few seconds.

emillon: I mean that it's a bit heavy to have arguments in a file, output in another file, etc. I'll try to see if that's better with a different kind of test, but this works in the meantime.

Julow: I see. `gen.ml` could read a sexp file, what do you think ?

emillon: let me work something out in another PR, the most important part is that we now have tests!

emillon:
Thanks!

Issue #1112, opened 2019-10-30
Add a test for editor integration


emillon:
Hi,

We (I) should add a test to make sure that the editor integration documented at https://github.com/ocaml-ppx/ocamlformat#vim-setup (and the one shipped in neoformat) works.

PR #1119, opened 2019-11-04
Take file kind from --name when formatting stdin

Julow:
Regression from https://github.com/ocaml-ppx/ocamlformat/pull/1018

This was breaking some editor integration where it's easier to pass the file name than conditionally passing `--intf` or `--impl`, as advertised in the readme (https://github.com/ocaml-ppx/ocamlformat#vim-setup).

With this PR, file kind can be interpreted from the value of `--name`.

On top of https://github.com/ocaml-ppx/ocamlformat/pull/1117


test/cli/stdin_and_name.stdin
```diff
@@ -0,0 +1 @@
+sample/a.ml
```
emillon: can you add "Closes #1112" to this commit?

emillon:
I think you can rebase, add the closing bug message, a changelog entry and we can merge and release 0.12.0.

test/cli/stdin_and_name.stdin
Julow: That's not really testing editor integration but I agree that's enough.

src/Conf.ml
```diff
@@ -1721,11 +1727,14 @@ let validate () =
...
-  else if has_stdin && Option.is_none !kind then
+  else if
+    has_stdin && Option.is_none !kind
+    && Option.is_none (Option.bind ~f:kind_of_ext !name)
+  then
     `Error
       ( false
-      , "Must specify at least one of --impl, --intf or --use-file when \
-         reading from stdin" )
+      , "Must specify at least one of --name, --impl or --intf when reading \
```
Julow: I removed `--use-file` from the message because it's deprecated now and is equivalent to `--impl`

PR #1121, opened 2019-11-04
Some tweaks before release

Julow:
- Commit release in CHANGES.md
- Add `.ocamlformat` file in each test directories
  dune-release run the tests in a way that the root `.ocamlformat` is not reachable
  Anyway, the tests were escaping out of the `_build` directory to find it.

I'm trying `dune-release` and it's working well.
The release procedure in HACKING.org does not look good to me.
Should I try to replace it with a dune-release procedure ?

For now, `--skip-lint` must be passed to dune-release because we don't have a `doc` field in the `.opam`.


emillon:
> Should I try to replace it with a dune-release procedure ?

Let's do that after the release.

Issue #1100, opened 2019-10-25
Bug: expressions with multi-line comments sometimes left as "one line"


ceastlund:
**Describe the bug**
If I write code with a multi-line comment like this, I'd expect `ocamlformat` to leave it split across lines.
```
let _ =
  [ "x"
  ; "y"
  (* first line
     second line *)
  ; "z"
  ]
;;
```
Instead, `ocamlformat` produces this formatting:
```
let _ = [ "x"; "y" (* first line
     second line *); "z" ]
```

**How to Reproduce**
Run with `ocamlformat` built from 953c48a4f63be421f995fb5d4c4168dccd3d4fea and `--profile=janestreet`.

gpetiot:
Duplicate of #912

Issue #1043, opened 2019-10-02
Bug: Alignment of multiline doc comments


Julow:
With `--profile=conventional`:

```ocaml
  (** fooo fooo
    fooo *)
  let fooo = 1
```

The item got moved inside a module (it was previously at toplevel). Second line of the docstring did not get indented.
This is fixed by `--parse-docstrings` but it's not in the `conventional` profile.

gpetiot:
Yes it is due to the `wrap-comments=false` of this profile that keep the comments unchanged (you may not want the second line indented for whatever reason), if you indent it by hand, you get the expected result.

Julow:
In that case, the global indentation changed. The second line was correctly aligned, by hand. Should we adjust the indentation in that case ?

gpetiot:
What I mean is that, as seen below, the comments are not modified in both cases, but if you indent t manually before formatting like in the second case, the output is what is expected:
```ocaml
module M = struct      
  (** fooo fooo
    fooo *)
  let fooo = 1

  (** fooo fooo
      fooo *)
  let fooo = 1
end
```
I don't think ocamlformat does anything wrong here.

Julow:
I mean, if you wrap everything in a module (so you need to add a level of indentation) and use ocamlformat to do that for you, the doc comments will not be indented correctly.

```ocaml
(** fooo fooo
    fooo *)
let fooo = 1
```

then

```ocaml
module M = struct
(** fooo fooo
    fooo *)
let fooo = 1
end
```

gives

```ocaml
module M = struct      
  (** fooo fooo
    fooo *)
  let fooo = 1
end
```

PR #1122, opened 2019-11-04
Fix: break multiline comments

### LINKED ISSUE #912: ocamlformat mishandles multiple-line comments between named function arguments
```
let g = f ~x (* this is a multiple-line-spanning
                comment *) ~y
```

This formats without a closing `;;` and with `~y` on the same line as the end of the comment; clearly ocamlformat is treating this comment like it is part of a single line.

gpetiot:
Fix #912 and #1100, something similar could be done for docstrings to tackle #1043 

The diff of test_branch.sh looks good.

This new behavior is only triggered when `wrap-comments=false` and when the first line of the comment is not empty (we don't want to indent the following lines of the comment if the first one is empty).

### DIFF
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -500,7 +500,7 @@ let fmt_cmt t (conf : Conf.t) ~fmt_code (cmt : Cmt.t) =
     then
       let b_space = String.length s > 0 && Char.equal s.[0] ' ' in
       let e_space =
-        String.length s > 0 && Char.equal s.[String.length s - 1] ' '
+        String.length s > 1 && Char.equal s.[String.length s - 1] ' '
       in
       let fmt_line ~first ~last s =
         if String.(is_empty (strip s)) then
--- a/src/Cmts.ml
+++ b/src/Cmts.ml
@@ -496,8 +496,8 @@ let fmt_cmt t (conf : Conf.t) ~fmt_code (cmt : Cmt.t) =
     match lines with
     | first_line :: _ :: _
       when (not begins_line) && not (String.is_empty first_line) ->
-        let b_space = String.length s > 0 && Char.equal s.[0] ' ' in
-        let e_space =
+        let begin_space = String.length s > 0 && Char.equal s.[0] ' ' in
+        let end_space =
           String.length s > 1 && Char.equal s.[String.length s - 1] ' '
         in
         let fmt_line ~first ~last s =
@@ -507,8 +507,8 @@ let fmt_cmt t (conf : Conf.t) ~fmt_code (cmt : Cmt.t) =
             else str "\n"
           else
             fmt_if (not first) "@;<1000 0>"
-            $ fmt_if b_space " " $ str stripped
-            $ fmt_if (last && e_space) " "
+            $ fmt_if begin_space " " $ str stripped
+            $ fmt_if (last && end_space) " "
         in
         vbox 0 (list_fl lines fmt_line)
     | _ -> str s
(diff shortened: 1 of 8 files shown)


src/Cmts.ml
```diff
@@ -490,10 +490,35 @@ let fmt_cmt (conf : Conf.t) ~fmt_code (cmt : Cmt.t) =
...
             | _, Some _ -> str line $ fmt "@,*") )
   in
-  let fmt_non_code (cmt : Cmt.t) =
+  let fmt_unwrapped_cmt ({txt= s; loc} : Cmt.t) =
+    let lines = String.split_lines s in
+    let first_line_empty =
+      List.length lines > 0 && String.is_empty (List.hd_exn lines)
+    in
+    let begins_line = Source.begins_line t.source loc ~ignore_spaces:false in
+    if String.contains s '\n' && (not begins_line) && not first_line_empty
```
Julow: What is the reason for this condition ?

src/Cmts.ml
```diff
@@ -490,10 +490,35 @@ let fmt_cmt (conf : Conf.t) ~fmt_code (cmt : Cmt.t) =
...
+    then
+      let b_space = String.length s > 0 && Char.equal s.[0] ' ' in
+      let e_space =
+        String.length s > 0 && Char.equal s.[String.length s - 1] ' '
```
Julow:
```suggestion
        String.length s > 1 && Char.equal s.[String.length s - 1] ' '
```
To avoid picking the same character as for `b_space` ?

src/Cmts.ml
gpetiot: - the first one is obviously because we only want to do the special treatment when there are `\n` inside (not using the box and so on)
- the second one: breaking comments that begin at the beginning of the line have no incidence of the associated code (the issue was that the code was not broken when the comment was broken, but if the comment is at column 0 it is very likely that the associated code is already either a one-liner or broken), so no different treatment to avoid a huge diff
- the last one because if the first line of a comment is empty we don't want to indent the following lines relatively to this line

src/Cmts.ml
gpetiot: I agree, but this code (and more) could be simplified since we suppose `s` contains an `\n` anyway (not a fan of hidden assumptions though)

Julow: It's not really an hidden assumption, the `if` is 2 lines above !
I agree this constant deserves a comment.

src/Cmts.ml
Julow: Thanks, that makes sense.

src/Cmts.ml
```diff
@@ -490,10 +490,35 @@ let fmt_cmt (conf : Conf.t) ~fmt_code (cmt : Cmt.t) =
...
+        String.length s > 1 && Char.equal s.[String.length s - 1] ' '
+      in
+      let fmt_line ~first ~last s =
+        if String.(is_empty (strip s)) then
+          if last then if first then str s else fmt "@;<1000 -2>"
+          else str "\n"
+        else
+          fmt_if (not first) "@;<1000 0>"
+          $ fmt_if b_space " "
+          $ str (String.strip s)
```
Julow: You are stripping the same string twice.

gpetiot:
@ceastlund does the new behavior proposed in this PR sound good to you?

ceastlund:
Where should I look to understand the new behavior?

gpetiot:
As I wrote in the initial message, the new behavior (breaking a comment on newlines `\n`) is only triggered when `wrap-comments=false` (it's the case when using the `janestreet` profile) and when the first line of the comment is not empty (we don't want to indent the following lines of the comment if the first one is empty).

So if the first line of a comment is empty, the comment is not modified by this PR, meaning you may want to remove to remove the empty lines at the beginning of comments to have them break properly.

ceastlund:
Okay, yes, between that and reading the changes to JS test files, this change looks good. Thanks very much!

PR #1126, opened 2019-11-05
Use an expect test for cli tests

emillon:
This removes the test generator and replaces it by a driver that runs ocamlformat. The output is compared by dune to `test_cli.expected` which handles promotion.

This removes a generator and many small files while keeping the promotion-based workflow, so I think it's valuable. What do you think?


emillon:
We don't really need to test **all** the combinations though :)

emillon:
Thanks!

PR #1127, opened 2019-11-05
Refactor CLI handling

Julow:
This PR refactors the code in `Conf.ml` that interprets CLI options telling OCamlformat what to do.

This code was very hard to maintain and could contain bugs. Now the code is structured in a way that "impossible" cases are not needed and subtle side effects are removed.

The diffs on the CLI tests are:
```diff
diff --git a/test/cli/test_cli.expected b/test/cli/test_cli.expected
index 06e5a24..62d5afc 100644
--- a/test/cli/test_cli.expected
+++ b/test/cli/test_cli.expected
@@ -82,39 +82,31 @@ let () = print_endline A.x
 err_stdin_name_unknown_ext
 ==========================
 args: --name b.cpp -
-ocamlformat: Must specify at least one of --name, --impl or --intf when reading from stdin
+ocamlformat: Cannot deduce file kind from passed --name. Please specify --impl or --intf
 [1]
 
 err_several_files_and_kind
 ==========================
 args: --impl --check sample/a.mli sample/b.ml
-ocamlformat: ignoring "sample/a.mli" (syntax error)
-File "sample/a.mli", line 5, characters 0-0:
-Error: Syntax error
+ocamlformat: Cannot specify --impl or --intf with multiple inputs
 [1]
 
 err_several_files_and_name
 ==========================
 args: --name foo.ml --check sample/a.mli sample/b.ml
-ocamlformat: ignoring "foo.ml" (syntax error)
-File "foo.ml", line 5, characters 0-0:
-Error: Syntax error
+ocamlformat: Cannot specify --name with multiple inputs
 [1]
 
 err_several_files_and_kind_inplace
 ==================================
 args: --impl --check sample/a.mli sample/b.ml
-ocamlformat: ignoring "sample/a.mli" (syntax error)
-File "sample/a.mli", line 5, characters 0-0:
-Error: Syntax error
+ocamlformat: Cannot specify --impl or --intf with multiple inputs
 [1]
 
 err_several_files_and_name_inplace
 ==================================
 args: --name foo.ml --check sample/a.mli sample/b.ml
-ocamlformat: ignoring "foo.ml" (syntax error)
-File "foo.ml", line 5, characters 0-0:
-Error: Syntax error
+ocamlformat: Cannot specify --name with multiple inputs
 [1]
 
 fmterr_file_and_name
```

It's now an error to pass `--name`, `--impl` or `--intf` with more than 1 input file.

It's also an error to pass `--print-config` with an other "action" (eg. `--output`, `--check`), is it a good idea ?

To go further, this code should be moved to an other module and the `action` type could be changed a little to remove an other "impossible" case in `ocamlformat.ml`.


jberdine:
Re `--name` with multiple inputs, see https://github.com/ocaml-ppx/ocamlformat/pull/740#discussion_r270622306 for a use case.

Julow:
In case of an error, you could not see which file caused it.
Shouldn't `merge-fmt` call ocamlformat once for each file, so it knows which call failed and can pass different names (eg. with a prefix) ?

PR #1134, opened 2019-11-07
Factor out a private lib

emillon:
Hi,

This factors a bit the build by adding an intermediate private library where all the external dependencies are attached, which simplifies a bit the directory structure (the binary entrypoints are split into `bin/`).

This also uses `(public_name)` for executables rather than a manual stanza, which is necessary for `dune-build-info` to work (see #1133).

Note that this library is not really usable even for unit tests since it has some side effects on initialization. I have a branch around that I'll push as a follow-up work after.

Finally, this is completely internal. It does not correspond to what a public API for OCamlformat would look like. It is purely for build system reasons.

Thanks!


bin/dune
```diff
@@ -0,0 +1,23 @@
+(executable
+ (public_name ocamlformat)
+ (package ocamlformat)
+ (modules ocamlformat)
+ (flags (:standard -open Import))
+ ;;INSERT_BISECT_HERE;;
```
Julow: You need to substitute that in `tools/bisect.sh`.
(the `sed` command take only one file so you'll have to call it in a loop)

emillon: `sed` seems to accept multiple files?

emillon:
I renamed to `lib/` and tried to grep for hardcoded paths but I might have missed some.

bin/dune
Julow: It seems yes. My bad.
That's working.

emillon:
Thanks!

Issue #1132, opened 2019-11-07
Bug: verbatim comment formatting not respected


ceastlund:
**Describe the bug**
If I format this code (with a formatted comment, taken from array.ml in Base)...
```
let _ =
  (* Optimal 5-element sorting network:

     {v
        1--o-----o-----o--------------1
           |     |     |
        2--o-----|--o--|-----o--o-----2
                 |  |  |     |  |
        3--------o--o--|--o--|--o-----3
                       |  |  |
        4-----o--------o--o--|-----o--4
              |              |     |
        5-----o--------------o-----o--5
      v} *)
  ()
;;
```
...ocamlformat reindents the formatted comment like this:
```
let _ =
  (* Optimal 5-element sorting network:

     {v
     1--o-----o-----o--------------1
     |     |     |
     2--o-----|--o--|-----o--o-----2
     |  |  |     |  |
     3--------o--o--|--o--|--o-----3
     |  |  |
     4-----o--------o--o--|-----o--4
     |              |     |
     5-----o--------------o-----o--5
     v} *)
  ()
;;
```

**How to Reproduce**
Run ocamlformat built from bcb803eb1867be97483ffe25e8ef0731a03bf19c with `--profile=janestreet`.

gpetiot:
The `{v v}` do not have a special treatment in comments (only in docstrings). What is done with the comments (since the PR merged recently) is that each line of the comment is aligned (to the left) with the first line of the comment, so the number of spaces on the left (the indentation) is not preserved.

Unfortunately, there is not much that can be done, since if we try to preserve the indentation, the number of spaces increases with each iteration of the formatting since we can't know how many spaces Format decides to effectively indent (let's suppose you start at 2 spaces, so we preserve 2 spaces, but Format decides to indent of 2 spaces anyway, so in the next iteration we have 4 spaces to preserve, but Format still decides to indent 2 spaces, and now we have 6 spaces to preserve, and so on.)

A little trick to preserve these spaces would be to insert a character before these spaces like:
```ocaml
let _ =
  (* Optimal 5-element sorting network:

     . 1--o-----o-----o--------------1
     .    |     |     |
     . 2--o-----|--o--|-----o--o-----2
     .          |  |  |     |  |
     . 3--------o--o--|--o--|--o-----3
     .                |  |  |
     . 4-----o--------o--o--|-----o--4
     .       |              |     |
     . 5-----o--------------o-----o--5 *)
  ()
```

[...]

PR #1138, opened 2019-11-12
Preserve comment indentation

Julow:
Closes https://github.com/ocaml-ppx/ocamlformat/issues/1132

Instead of stripping all indentation, only remove the indentation common to all the lines of the comment.

I intentionally add a regression: remove trailing empty lines. This add no diffs with `test_branch`, what do you think ?


gpetiot:
Can you add a test similar to the usecase of Carl with different indentations for different lines?

Julow:
Good idea. Done.

Should I keep removing the empty lines at the end of comments ?

gpetiot:
> Should I keep removing the empty lines at the end of comments ?

It sounds reasonable to me

gpetiot:
Can we merge this one?
Don't hesitate if you have feedback on this @ceastlund

ceastlund:
Looks good to me.

PR #1141, opened 2019-11-13
Fix regression of comment position in list patterns

### LINKED ISSUE #1140: Bug: Regression: comment position in list patterns
**Describe the bug**
https://github.com/ocaml-ppx/ocamlformat/pull/866 introduced a regression in the position of comments in list patterns.

**How to Reproduce**
Executing
```
$ echo "let [(* comment one *) one; (* comment two *) two] = [1; 2]" | _build/default/src/ocamlformat.exe --name "foo.ml" -
```
produces
```
let (* comment one *) [one; (* comment two *) two] = [1; 2]
```
while it should, and used to, produce
```
let [(* comment one *) one; (* comment two *) two] = [1; 2]
```

jberdine:
fix #1140

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,6 +37,7 @@
   + Fix parentheses around optional module parameter (#1212) (Christian Barcenas)
   + Fix grouping of horizontally aligned comments (#1209) (Guillaume Petiot)
   + Fix dropped comments around module pack expressions (#1214) (Jules Aguillon)
+  + Fix regression of comment position in list patterns (#1141) (Josh Berdine)
 
 #### Documentation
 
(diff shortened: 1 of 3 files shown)


Julow:
This is not passing the tests. This is not stabilizing:
```
let _ = function
  | (* AA*) ((*BB*) [(* CC *)x(* DD *) ; (* EE *)y(* FF *)] (* GG *)) (* HH *) -> k
```

gpetiot:
> This is not passing the tests. This is not stabilizing:
> 
> ```
> let _ = function
>   | (* AA*) ((*BB*) [(* CC *)x(* DD *) ; (* EE *)y(* FF *)] (* GG *)) (* HH *) -> k
> ```

Somehow this is stabilizing again after rebasing on master.
No regression with test_branch on any preset profile.

PR #1142, opened 2019-11-13
Remove global ref `Cmts.remove`

emillon:
This replaces a global reference by a new field in `Cmts.t`. It controls whether to remove from `remaining` or not.

This mechanism is only used in `preserve`, which did not have access to `Cmts.t` (as it was referring to the global). So the interface changes a bit so that the correct value is passed.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -21,7 +21,8 @@ type t =
   ; cmts_after: (Location.t, Cmt.t list) Hashtbl.t
   ; cmts_within: (Location.t, Cmt.t list) Hashtbl.t
   ; source: Source.t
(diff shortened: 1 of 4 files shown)


lib/Cmts.ml
```diff
@@ -628,16 +628,19 @@ let drop_inside t loc =
   in
   clear t.cmts_before ; clear t.cmts_within ; clear t.cmts_after
 
+let pop_if_debug t loc =
+  if Conf.debug && t.remove then Hashtbl.remove t.remaining loc
```
gpetiot: Is it really defined twice?

emillon: yes, I moved it but forgot to remove the original one. Thanks!

emillon:
Thanks!

PR #1145, opened 2019-11-13
Internal: sanitize formatting of or-patterns and remove or_newline

gpetiot:
This is the beginning of a rework of the formatting of pattern-matching cases to make it easier to maintain, fix some bugs and only use the vanilla `Format` functions.

- add `Fmt.cbreak` (also used in another PR where I remove some calls to `if_newline`)
- remove `Fmt.or_newline`
- add `Params.get_or_pattern_sep` to manage the configuration of the bar separating or-patterns
- split `fmt_cases` to create a new `fmt_case` function

diff with test_branch on the conventional profile (when `break-cases=fit` or `toplevel`):
```diff
diff --git a/infer/src/clang/cTrans.ml b/infer/src/clang/cTrans.ml
index 2e23911f3..9300f22e2 100644
--- a/infer/src/clang/cTrans.ml
+++ b/infer/src/clang/cTrans.ml
@@ -3366,7 +3366,8 @@ struct
         match lci_capture_kind with
         | `LCK_ByRef (* explicit with [&x] or implicit with [&] *)
         | `LCK_This
-        (* explicit with [this] or implicit with [&] *)  | `LCK_VLAType
+        (* explicit with [this] or implicit with [&] *)
+        | `LCK_VLAType
         (* capture a variable-length array by reference. we probably don't handle
            this correctly elsewhere, but it's definitely not captured by value! *)
```

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1003,12 +1003,6 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
           match i.[0] with '-' | '+' -> true | _ -> false )
         | _ -> false
       in
-      let pro0 =
-        fmt_opt pro
-        $ fits_breaks
-            (if parens then "(" else "")
-            (if nested then "" else "( ")
-      in
       let is_simple {ppat_desc; _} =
         match ppat_desc with
         | Ppat_any | Ppat_constant _ | Ppat_var _
@@ -1032,7 +1026,12 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
                     else open_hovbox
                   in
                   let pro =
-                    if first_grp && first then pro0 $ open_box (-2)
+                    if first_grp && first then
+                      fmt_opt pro
+                      $ fits_breaks
+                          (if parens then "(" else "")
+                          (if nested then "" else "( ")
+                      $ open_box (-2)
                     else if first then
                       Params.get_or_pattern_sep c.conf ~ctx:ctx0
                       $ open_box (-2)
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1025,6 +1025,13 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
         ( list_fl (List.group xpats ~break)
             (fun ~first:first_grp ~last:_ xpat_grp ->
               list_fl xpat_grp (fun ~first ~last xpat ->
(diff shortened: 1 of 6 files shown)


lib/Fmt_ast.ml
```diff
@@ -2794,78 +2782,79 @@ and fmt_cases c ctx cs =
...
-              ( fmt_expression ?eol c ?parens:parens_for_exp xrhs
-              $ fmt_if parens_here
-                  ( match c.conf.indicate_multiline_delimiters with
-                  | `Space -> "@ )"
-                  | `No -> "@,)"
-                  | `Closing_on_separate_line -> "@;<1000 -2>)" ) ) ))
+  let max_len = fold_pattern_len ~f:max cs in
+  list_fl cs (fun ~first ~last case ->
+      let pattern_len = pattern_len case in
+      fmt_case c ctx ~first ~last ~max_len ~pattern_len case)
```
Julow: You could pass the padding here (`max_len - pattern_len`) instead. So there is less argument and we avoid a match expecting 2 `option` to be in sync.

lib/Fmt_ast.ml
```diff
@@ -984,7 +985,6 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
       in
       p.box (list_fl pats fmt_pat)
   | Ppat_or _ ->
-      let has_doc = not (List.is_empty xpat.ast.ppat_attributes) in
```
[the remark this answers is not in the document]
gpetiot: agreed, fixed.

lib/Fmt_ast.ml
gpetiot: agreed, fixed but could not simplify as much as you wanted.
To make it easier to read `break-infix-before-func=false` could be added in our `.ocamlformat`

Julow: Tweaking the `.ocamlformat` is not necessary. My comment was only about separating concerns.
To simplify the code, I think matching on `max_len` and `pattern_len` is easier to read than bind operators. You could also use a `when` clause to disable padding.

lib/Fmt_ast.ml
```diff
@@ -2799,78 +2787,84 @@ and fmt_cases c ctx cs =
...
+        match max_len with
+        | Some max_len when add_padding -> (
+          match pattern_len case with
+          | Some pattern_len ->
+              let pad = String.make (max_len - pattern_len) ' ' in
+              Some
+                (fmt_or_k
+                   Poly.(c.conf.break_cases = `All)
+                   (str pad)
+                   (fits_breaks ~level "" pad))
```
Julow: I agree the binds were looking better than nested matches. That's not important.

PR #1146, opened 2019-11-14
Cmts: Remove symbol strings

emillon:
In `Cmts`, the strings `"before"`, `"after"` and `"within"` are sometimes used as symbols in debugging output. This leaks in the output type of `remaining_comments`. Since these are only used for debugging, they can be replaced by an abstract type that makes the intention clearer.

### DIFF
--- a/lib/Cmts.mli
+++ b/lib/Cmts.mli
@@ -125,7 +125,7 @@ val has_within : t -> Location.t -> bool
 val has_after : t -> Location.t -> bool
 (** [has_after t loc] holds if [t] contains some comment after [loc]. *)
 
-val remaining_comments : t -> (Cmt.t * string * Sexp.t) list
+val remaining_comments : t -> Cmt.t list
 (** Returns comments that have not been formatted yet. *)
 
 val remaining_locs : t -> Location.t list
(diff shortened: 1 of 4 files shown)


lib/Cmts.mli
```diff
@@ -125,7 +125,13 @@ val has_within : t -> Location.t -> bool
...
 (** [has_after t loc] holds if [t] contains some comment after [loc]. *)
 
-val remaining_comments : t -> (Cmt.t * string * Sexp.t) list
+type info
+(** Metadata about a comment. Can not be inspected. *)
+
+val sexp_of_info : info -> Sexp.t
+(** Convert a piece of [info] to a s-expression for debugging purposes. *)
+
+val remaining_comments : t -> (Cmt.t * info) list
```
gpetiot: If this is only for debugging purposes, should we really expose the type `info`?
We could just output it in debug mode when calling `remaining_comments` and there are remaining comments

Julow: This function is used only once and the `info` value is immediately discarded.
Let's remove it.

emillon: yes, I've seen that. but there are several places like that were debugging code is kept "just in case" so I didn't want to be too intrusive. Let's remove this.

Issue #1139, opened 2019-11-12
Bug: fatal errors are swallowed


jberdine:
**Describe the bug**
Since https://github.com/ocaml-ppx/ocamlformat/pull/1127 fatal errors such as `Fatal error: exception ("User Error: malformed .ocamlformat file"` that used to `exit 2` with an uncaught exception now just silently succeed without doing anything.

emillon:
Ah, that's not great. I think we'll need to cut a point release for this. @Julow do you know how to fix this?

emillon:
Ah my bad it's not in the release. All good then. Still needs to be fixed though :)

Julow:
A minor release for that is not needed, #1127 is not released.

The error message changed from

```
Fatal error: exception ("User Error: malformed .ocamlformat file" ...
backtrace
```

to

```
ocamlformat: internal error, uncaught exception:
             ("User Error: malformed .ocamlformat file" ...
             backtrace
```

and the exit status changed from `2` to `1`. That's a regression indeed.

I can't reproduce silently succeed without doing anything. What was the full error message ?

jberdine:
Sorry, the 'silently succeeding' I see is through the emacs integration, but on the command line it does exit with status 1.

I'm not sure what is going on in the emacs integration to do that. Can anyone else repro, or is it just here?

jberdine:
I think the problem is that the error output now is sent to stdout instead of stderr

Julow:
@jberdine https://github.com/ocaml-ppx/ocamlformat/pull/1180 should fix this issue. Is it working for you ?

PR #1147, opened 2019-11-14
Improve user error messages

Julow:
- Add tests for invalid ocamlformat files and env variable.
- Nicer error messages instead of uncaught exceptions with sexp payload and backtrace
- Error messages are printed to stderr


lib/Conf.ml
```diff
@@ -1871,6 +1878,22 @@ let rec collect_files ~segs ~ignores ~enables ~files =
         (ignores, enables, files, Some dir)
       else collect_files ~segs:upper_segs ~ignores ~enables ~files
 
+exception Conf_error of string
+
+let failwith_user_errors ~kind errors =
+  let pp_errors pp = function
```
emillon: I think you can use `Format.pp_print_list` to handle the separator here.

lib/Conf.ml
```diff
@@ -2194,27 +2198,21 @@ let update ?(quiet = false) c {attr_name= {txt; loc}; attr_payload; _} =
                     ; pexp_attributes= []
                     ; _ }
                   , [] )
-            ; _ } ] ->
-          parse_line ~from:`Attribute c str
-      | _ -> Error (`Malformed "string expected") )
+            ; _ } ] -> (
+        match parse_line ~from:`Attribute c str with
+        | Ok c -> Ok c
+        | Error e -> Error (string_of_user_error e) )
```
emillon: You can use `Result.map_error ~f:string_of_user_error` here

lib/Reason.ml
```diff
@@ -52,10 +52,10 @@ module Binary_reason = struct
     let rec loop = function
       | [] ->
           if is_intf_or_impl magic then
-            user_error "Unknown version" [("magic", Sexp.Atom magic)]
+            failwith (Format.sprintf "Unknown version (magic: %s)" magic)
```
emillon: what do you think about defining `failwithf` in `Import`? (using `Format.kasprintf failwith`). I thought it was in `Base`, but it's only in `Core`.

test/cli/test_cli.expected
```diff
@@ -157,3 +157,45 @@ File "foo.ml", line 5, characters 0-0:
 Error: Syntax error
 [1]
 
+conf_bad_version
+================
+args: --impl -
+ocamlformat: Error while parsing .ocamlformat:
+             For option "bad": expecting "0.12-18-gea9fdbb-dirty" but got "bad"
```
emillon: This will always fail as it depends on the commit number, unfortunately. You can postprocess the output but we can leave this test for now I think.

test/cli/test_cli.ml
```diff
@@ -8,19 +8,32 @@ let pp_status ppf = function
...
+      Filename.concat (Unix.getcwd ()) ocamlformat
+    else
+      ocamlformat
+  in
+  let test ?(env = [||]) ?pwd ?(stdin = "") ~name ~args () =
+    let pp_cd pp = function
+      | Some d -> Format.fprintf pp "cd '%s'; " d
+      | None -> ()
+    in
+    let cmd = Format.asprintf "%a%s %s 2>&1" pp_cd pwd ocamlformat args in
```
emillon: This adds a lot of complexity to the test harness just to parse `.ocamlformat` files. We could use `spawn` to simplify that, but actually, does this need to be tested at the end to end level?

What do you think about pulling out the function that parses a `.ocamlformat` file and put that in a test harness instead? It looks like it's the main branch of `Conf.read_config_file`. That way, we can just test a string -> config file and remove all of this.

test/cli/test_cli.ml
```diff
@@ -8,19 +8,32 @@ let pp_status ppf = function
...
     let out = Stdio.In_channel.input_all proc_out in
-    let stat = Unix.close_process proc in
+    let stat = Unix.close_process_full proc in
     Format.printf "%aargs: %s\n%s%a\n\n" pp_banner name args out pp_status stat
   in
   let impl_file = Stdio.In_channel.read_all "sample/b.ml" in
   let intf_file = Stdio.In_channel.read_all "sample/a.mli" in
   let syntax_error_file = Stdio.In_channel.read_all "sample/syntax_error.ml" in
+
+  (* CLI actions *)
```
emillon: I think extracting functions will make this clearer than a comment: `test_cli_actions (); test_user_errors()` etc.

test/cli/test_cli.expected
Julow: I was hoping https://github.com/ocaml-ppx/ocamlformat/pull/1133 would solve this.
I just forgot to mark this PR as "Draft".

test/cli/test_cli.ml
Julow: Yes. That's something I want to do next and it's required to have ocamlformat as a library.

Here I wanted to test end-to-end because I'm checking the error message given to the user.
I think we should both keep this and have the config parsing unit-tested.

test/cli/test_cli.ml
Julow: Yes this would help.

lib/Reason.ml
Julow: This would only be used twice. In the future, this code could be either gone or improved, in both cases this would no longer be useful.

test/cli/test_cli.ml
emillon: It's not too useful to test the same thing at both level. In my experience it's enough to just test the happy path as e2e (and maybe one error to see how it's rendered), but that has not its place in this test harness. We can think how to make an e2e test for file parsing (ocamlformat / ignore / ocp-indent) in a separate PR. (we can probably do it by hand)

emillon: (you'll need #1148 to link a test suite, I can give you a hand with that)

lib/Reason.ml
emillon: I'm fine with a dead `failwithf` in import if you think it's worth it in the meantime.

gpetiot: Yes it would not be in vain, in the future we would need an easy way to emit errors, warnings and hints

test/cli/test_cli.ml
Julow: Yes but here I'm not testing that errors are reported or anything. I'm testing the rendering of messages.
I'm testing a single level, the e2e one.
Unit-tests are for an other PR, this one is about error messages.

lib/Reason.ml
Julow: Yes, that's why I'm against adding this.
We need more than the `Failure` exception (eg. locations, colors, formatting) and I don't want that to be used elsewhere.

test/cli/test_cli.ml
emillon: OK to leave the parsing tests elsewhere. But in that case:
- you don't need that many e2e tests for error rendering. one is enough (do not try to have test coverage from e2e tests)
- since the setup of this test is very different from CLI tests (and that it's not a CLI test), it does not need to be executed in this file.
- this test will be easier expressed directly in a dune file, without the intermediate `.ml` / process handling level.

All these steps should make  your tests in this PR a lot simpler.

Julow: I don't understand what's wrong with the tests.
We need to test messages and their rendering as it can change in the future and we need to know how.

I'm not trying to test the detection of errors, just 4 kinds that are generated from different places.

I think this tests work well with the expect-test output. I tried clean up the code in the last commit, what do you think ? It looks better after splitting.

lib/Reason.ml
gpetiot: I see no problem to have only what is necessary for this PR for now and rework it when we need it

test/cli/test_cli.ml
gpetiot: If dune can do more work for us that would be great, I'm not opposed to having duplicates of tests in many places for now. That's a job we need to do later anyway, determine where and how is tested every usecase, and fill the blanks.

Julow:
As I was arguing, I believe there is no problem with the tests. This new tests are about how error are printed to CLI users and are absolutely not testing the handling of errors.

This cannot be merged yet, the version test is not stable. I'm hoping https://github.com/ocaml-ppx/ocamlformat/pull/1133 will help with that, I think the version check should be relaxed with dev builds.

test/cli/test_cli.ml
emillon: yes, we can live with extra tests. however I don't think we should try to use this test harness for things it'll do worse than dune. to expand a bit:

- in #1117 we added a way to run ocamlformat with some options and stdin and compare the results.
- in #1126 this was simplified since running a binary with arguments is simple enough that it can be done in-process. expect test are a good enough balance for this.

However, if we need to add features like changing directory, setting environment variables, etc, this is more complicated in-process than using `(setenv)` and `(chdir)` in a dune file.

So I suggest that we:
- leave `test_cli` unchanged in this PR and use it only for things that just pass arguments and don't need to set env/read `.ocamlformat` files
- use manually written dune rules for these new tests, in a style similar to the rules generated in #1117. we don't need a generator for this.

As for `test_cli`, maybe it was a bit overreaching, and the right balance is a set of manually written dune rules (without a generator). There are not that many cases, so we can avoid having this `test_cli` and leverage dune's ability to run processes with arguments and diff files. I'm happy to undo this part after this PR if that works for you.

Julow: I agree `test_cli` is becoming too complicated (it was from the beginning) and we will have to maintain it.
Instead of going back to vanilla dune (it still lacks features like allowing failing commands) and is a bit harder to setup when we have new needs (especially with no generators), wouldn't it be better to just use better tools ?
We could use `bos` to cleanup process-handling code or use a cram-like tool (eg. `ppx_expect` if it can do that) ?

emillon: OK, great. We can live with a couple manual rules. I suggest the following plan to move things forward:

- you convert the new tests to plain dune rules (in this PR)
- I undo `test_cli` as plain dune rules (in another PR)
- we try to use craml or mdx later to run the commands (later)

sounds good?

Julow: I'm not sure converting tests to plain dune rules is good idea because it's pretty hard and time consuming.
This PR won't be merged now because of something else. What about:
- Keep this PR open
- We try to use craml or mdx (in another PR) for the tests currently in master
- Rebase this PR on that
?

jberdine:
> This cannot be merged yet, the version test is not stable.

Is the (only) problem that there is a test which includes the current version in the output error message, and that keeps changing? If so, is it possible to change that test to just check the exit code?

Julow:
> Is the (only) problem that there is a test which includes the current version in the output error message, and that keeps changing? If so, is it possible to change that test to just check the exit code?

Yes, #1133 needs more work so I'll workaround that !

Julow:
I rebased on master (cli tests changed) and worked around the version problem.
This is ready for review and merge.

test/cli/dune
```diff
@@ -247,3 +247,102 @@
...
+(rule
+ (targets conf_bad_version.output)
+ (deps
+  (source_tree roots/bad_version))
+ (action
+  (with-outputs-to
+   %{targets}
+   (chdir
+    roots/bad_version
+    (system "! %{bin:ocamlformat} --impl - < %{dep:sample/a.ml} 2>/dev/null")))))
```
jberdine: Note that it is important that the error message is printed on stderr and not stdout, so IIUC redirecting just fd 2 achieves more than just ignoring unstable output, it also checks that the error goes to the right stream.

jberdine:
I haven't managed to find the root cause, but note that with this PR I still see the emacs integration silently swallowing errors.

lib/Conf.ml
```diff
@@ -1765,7 +1771,7 @@ let parse_line config ~from s =
           Error
             (`Bad_value
               ( value
```
jberdine: Shouldn't this `value` be `name`?

jberdine:
Hmm, I don't understand why, but the disagreement between the changes from 1127 and ocamlformat.el still remain after this.

test/cli/dune
Julow: Yes, the test does that. Do you mean I should add that to the comment ?

lib/Conf.ml
Julow: Fixed, thanks !

Julow:
@jberdine What is the error that you expect and don't show up? Do you see the `Could not apply ocamlformat` message?

test/cli/dune
jberdine: I was just thinking to adjust the comment, to avoid future confusion.

jberdine:
When I test with a file whose config specifies a mismatched version, for example, and invoke ocamlformat, I see "Applied ocamlformat" when in fact it errored out. I tried debugging the elisp for a while, and saw that printing the status code returned from call-process is 0. So I don't understand at all.

test/cli/dune
Julow: Did.

Julow:
Could it be that the file passed to ocamlformat is not located where the bad `.ocamlformat` file can be found? (eg. in `/tmp`)

jberdine:
I don't think that is it. I have tested by building the parent commit of pr1127 and see the expected error, and then build pr1127 and the error is silently ignored, and with this pr and see the same as with 1127. I just pick any random infer file, add some extra line, try to reformat and look for the error in the echo area or `*compilation*` buffer depending on the emacs config (I've tried both).

Julow:
This PR wasn't fixing https://github.com/ocaml-ppx/ocamlformat/issues/1139. I opened https://github.com/ocaml-ppx/ocamlformat/pull/1180 for that.

I think this PR is still an improvement, it improves error outputs a lot and adds tests for them.

gpetiot:
If all blocking points have been resolved it's time to merge this one as well.

Julow:
Just rebased. I think we can merge now :)

PR #1148, opened 2019-11-14
Remove load-time argv parsing

emillon:
`Conf.action` had side effects as it immediately parsed `Sys.argv` at load time. This has to be delayed so that we can link the lib in a test suite for example.


emillon:
This is the part that blocks linking, the rest can be improved later.

emillon:
Thanks !

PR #1149, opened 2019-11-14
Internal: set break-infix-before-func=false on ocamlformat codebase

gpetiot:
Customize our `.ocamlformat` file


gpetiot:
I guess Josh downvoted because it is agains the current style applied in ocamlformat that consists in having operators on the left (`ocamlformat` profile)
Should we stick to this rule or is it okay to make this small adjustment if it makes the code (subjectively) prettier?

gpetiot:
Alright so we can close this one for now.

jberdine:
> I guess Josh downvoted because it is agains the current style applied in ocamlformat that consists in having operators on the left (ocamlformat profile)

Yes. I don't have a strong view on this, but have some worry that when the rest of the style encourages focusing on the left margin, putting binders on the right might "hide" them. But this is an abstract worry, I have not seen examples of people getting confused by this particular choice.

PR #1150, opened 2019-11-14
Internal: use opt and fmt_opt to simplify formatting

gpetiot:
split from #1090
No diff with test_branch.sh


lib/Fmt_ast.ml
```diff
@@ -2123,15 +2126,14 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
                ( hvbox 0
                    ( str keyword
                    $ fmt_extension_suffix c ext
                    $ fmt_attributes c ~key:"@" pexp_attributes
                    $ fmt "@;<1 2>"
-                   $ fmt_expression c (sub_exp ~ctx e0)
-                   $ fmt "@," )
-               $ break_unless_newline 1 (-2)
+                   $ fmt_expression c (sub_exp ~ctx e0) )
+               $ break 1 (-2)
```
Julow: This change belongs to #1090 ?

gpetiot: Not necessarily, it's just a "box structure improvement" that is too small to be in it's small PR, and since #1090 was already big enough it ended up in this one.

PR #1151, opened 2019-11-15
Internal: replace inplace formatting by dune staging for make fmt

gpetiot:


Makefile
```diff
@@ -36,7 +36,7 @@ SRCS=$(shell \find bin lib tools -name '[^.]*.ml' -or -name '[^.]*.mli' -or -nam
 
 .PHONY: fmt
 fmt:
-	dune exec -- ocamlformat -i $(SRCS)
+	dune build @bin/fmt @lib/fmt @tools/fmt
```
Julow: If we supported directories or globbing in `.ocamlformat-ignore` files, this would not be needed. I'd like to work with only dune, without magic parameters.
Should we improve that or is Dune going to have that feature ?
What do you think @emillon ?

Julow:
The `fixpoint` rule in the Makefile seems unecessary now since the diffs are staged by dune and dune will exit with an error status.

gpetiot:
> The `fixpoint` rule in the Makefile seems unecessary now since the diffs are staged by dune and dune will exit with an error status.

Yes we can either replace fixpoint by fmt, or keep fixpoint but use ` --check`.

Issue #1154, opened 2019-11-15
Bug: --enable-outside-detected-project flag does not work


ceastlund:
**Describe the bug**
```
$ ocamlformat --enable-outside-detected-project file.ml
File file.ml
Warning: Ocamlformat disabled because [--enable-outside-detected-project] is not set and no project root was found
```

**How to Reproduce**
Run with ocamlformat built from eb187db2361a25caee9d206dca0da6e0e096f093

PR #1156, opened 2019-11-18
Fix --enable-outside-detected-project

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1154
Bug introduced in https://github.com/ocaml-ppx/ocamlformat/pull/1127 (on master: 45f324)

The parsing of option has been moved after that option is checked, at toplevel.
Same for the `--root` option, but I didn't manage to make it work yet while writing tests.


CHANGES.md
```diff
@@ -20,6 +20,7 @@
   + Use opt and fmt_opt to simplify formatting (#1150) (Guillaume Petiot)
   + Replace inplace formatting by dune staging for make fmt (#1151) (Guillaume Petiot)
   + Refactor code that interprets CLI options (#1127) (Jules Aguillon)
+    Fix --enable-outside-project, broken by this PR (#1156) (Jules Aguillon)
```
emillon: You can just update the above line with `(#1127, #1156)`, I think it's enough.

Julow: Agree.

PR #1157, opened 2019-11-19
CLI tests: use dune to run ocamlformat

emillon:
(partial revert of #1126, as discussed in #1147)

Test generators are a bit heavy, but running processes through ocaml is a bit limited when we want to set a working directory or environment.

So, let's use a middle ground solution: run ocamlformat through dune, but write the rules by hand.


Julow:
Why are test generators a bit heavy ?

gpetiot:
For me personally it is easier to maintain OCaml code than dune code but I don't have any other argument against writing rules manually, as long as we don't add new ones too often.

emillon:
As discussed offline with @Julow, the main disadvantage of this approach is that adding new tests involve copy/pasting, so it's possible that we'll forget to add the runtest-diff part for one test someday.

Two things can make this easier in the future:
- a generator whose input data is in a .ml file (not scattered across several files as was the case before)
- some of these e2e tests will become less important as we add ways to test this in a more focused way.

PR #1158, opened 2019-11-19
Fix --debug not working and propagate options functionally

Julow:
`--debug` is broken since https://github.com/ocaml-ppx/ocamlformat/pull/1148
because the parsing of options is done after `debug`, `margin_check` and
`check` are accessed.

This PR fix that by adding a new record `Conf.opts` and passing it
through to formatting functions.
It contains non-formatting options and shouldn't be used by `Fmt_ast`.

Later:
- Move `comment_check`, `max_iters` and `quiet` out of `Conf.t`
  `disable` should stay there
  `margin` and `max_indent` may be moved as well
- Move that part of `Conf.ml` into an other module

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -876,15 +876,19 @@ end = struct
 
   let ( == ) = Base.phys_equal
 
-  let dump fs ctx ast =
+  let dump ctx ast fs =
     Format.fprintf fs "ast: %a@\nctx: %a@\n" T.dump ast T.dump ctx
 
-  let fail ctx ast exc =
-    if Conf.debug then (
-      let bt = Caml.Printexc.get_backtrace () in
-      dump Format.err_formatter ctx ast ;
-      Format.eprintf "%s%!" bt ) ;
-    raise exc
+  let assert_no_raise ~f ~dump x =
+    assert (
+      try
+        ignore (f x) ;
+        true
+      with exc ->
+        let bt = Caml.Printexc.get_backtrace () in
+        dump x Format.err_formatter ;
+        Format.eprintf "%s%!" bt ;
+        raise exc )
 
   (** Predicates to check the claimed sub-term relation. *)
 
@@ -1101,8 +1105,9 @@ end = struct
       | _ -> assert false )
     | Top | Tli _ -> assert false
 
-  let check_typ ({ctx; ast= typ} as xtyp) =
-    try check_typ xtyp with exc -> fail ctx (Typ typ) exc
+  let assert_check_typ xtyp =
+    let dump {ctx; ast= typ} = dump ctx (Typ typ) in
+    assert_no_raise ~f:check_typ ~dump xtyp
 
   let check_cty {ctx; ast= cty} =
     let check_class_type l =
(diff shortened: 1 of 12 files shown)


lib/Ast.ml
```diff
@@ -876,15 +876,19 @@ end = struct
...
+  let assert_no_raise ~f ~dump x =
+    assert (
+      try
+        ignore (f x) ;
+        true
+      with exc ->
+        let bt = Caml.Printexc.get_backtrace () in
+        dump x Format.err_formatter ;
+        Format.eprintf "%s%!" bt ;
+        raise exc )
```
Julow: `Conf.debug` was actually not needed, this code is only reachable when `assert` are enabled. I changed the code to make this clear.

lib/Translation_unit.ml
```diff
@@ -208,30 +209,29 @@ let print_error ?(fmt = Format.err_formatter) conf ~input_name error =
...
       | exn ->
           Format.fprintf fmt
             "  BUG: unhandled exception. Use [--debug] for details.\n%!" ;
-          if Conf.debug then Format.fprintf fmt "%s\n%!" (Exn.to_string exn)
-      )
+          if debug then Format.fprintf fmt "%s\n%!" (Exn.to_string exn) )
 
-let check_all_locations fmt cmts_t =
-  if Conf.debug then
+let check_all_locations ~debug fmt cmts_t =
```
emillon: given that the function is `if <argument> then ...`, I think it would make sense to move the branch to the call site

lib/Translation_unit.mli
```diff
@@ -51,11 +52,18 @@ val parse_and_format :
...
   -> (string, error) Result.t
 (** [parse_and_format xunit conf ?output_file ~input_name ~source] is similar
     to [format] but parses the source according to [xunit]. *)
 
 val print_error :
-  ?fmt:Format.formatter -> Conf.t -> input_name:string -> error -> unit
+     ?fmt:Format.formatter
+  -> debug:bool
+  -> quiet:bool
+  -> check:bool
```
emillon: boolean parameters are usually a code smell, but 3 of them is a bit tough. they were already there hidden in `Conf.t` and a global, though. we can probably do something about this in another PR. do all combinations make sense?

lib/Conf.mli
```diff
@@ -100,17 +101,15 @@ type action =
...
+  ; margin_check: bool
+        (** Check whether the formatted output exceeds the margin. *) }
+(** Options changing the tool's behavior *)
 
-val check : bool
-(** Check whether the input files already are formatted. *)
-
-val margin_check : bool
-(** Check whether the formatted output exceeds the margin. *)
+val action : unit -> action * opts
```
emillon: Rather than returning a tuple, what do you think of adding the booleans to the individual actions? for example, if `action` is `In_out`, `check` is not used.

Julow: `check` is not necessary and *must* not be there !
The `opts` record may contain more fields in the future (taken from `Conf.t`) and makes sense on its own. I don't agree to inline it into actions.

lib/Translation_unit.mli
Julow: The alternative would be to pass `Conf.t`, `Conf.opts` and `check:bool`.
This function is very complicated, there is nothing we can do without changing it. The combinations all make sense.
`check` can be removed easily by catching `Unstable` before and `quiet` can also be handled a step before (harder), that's why I prefer boolean to passing confs.

lib/Translation_unit.ml
Julow: Good idea. That makes the code slightly easier to read.

Julow:
> Good catch. In addition to the other comments, one thing we can do is in a first step, expose `debug` as a ref (and use `!Conf.debug` in the rest of the code), and do the parameter-passing changes in a subsequent PR. That way this get fixed quicker.

Is something blocking this PR ?
I don't think splitting 2 lines of off this PR would make it easier to review.

gpetiot:
Looks good to me, can you add tests that were broken before and are fixed now?

Julow:
With adding tests and to ease review I'll split this PR in two.

PR #1161, opened 2019-11-20
Fix --debug not showing any output

Julow:
Since #1148, some options (`debug`, `check` and `margin_check`) are accessed before options are parsed.
The effect is that `--debug` won't show any debug output and `--margin-check` won't work (I didn't test that). 

This PR adds some tests and quick-fixes that by accessing the option's value only when needed. This *should* always be after options are parsed.
A refactoring PR will follow to improve the code here.


test/cli/test_cli.ml
```diff
@@ -48,6 +48,9 @@ let main ocamlformat =
   test ~name:"fmterr_stdin_bad_kind" ~stdin:intf_file ~args:"--impl -" () ;
   test ~name:"fmterr_file_and_name_bad_kind" ~args:"--name foo.ml sample/a.mli" () ;
   test ~name:"fmterr_stdin_and_name_bad_kind" ~stdin:intf_file ~args:"--name foo.ml -" () ;
+  test ~name:"check_formatted" ~args:"--check sample/a.ml" () ;
+  test ~name:"check_not_formatted" ~args:"--check sample/b.ml" () ;
```
Julow: These 2 checks are finally not necessary but they feel useful. Should I keep them?
(I added them to check if `Conf.check` is useful when combined with `--max-iters=1`).

Julow:
Just rebased

PR #1163, opened 2019-11-20
Sys.argv is deprecated in Base v0.13.0

Julow:
The replacement is `Sys.get_argv ()` but this function is not in Base `v0.12.2`, so I use `Import.Sys.argv` as a workaround.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 
 #### Internal
 
+  + Work with base v0.13.0 (#1163) (Jules Aguillon)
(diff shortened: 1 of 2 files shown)


emillon:
Is this enough to make us compatible with both 0.12.x and 0.13.x?

Julow:
> Is this enough to make us compatible with both 0.12.x and 0.13.x?

Yes

emillon:
That's great, please add a changelog entry and merge ASAP :)

Julow:
It was not working. Fixed, sorry.
I used `Import.Sys` (which is base) instead of `Caml.Sys`

Issue #1164, opened 2019-11-20
Bug: certain match clauses with comments do not stabilize


ceastlund:
**Describe the bug**
I format this code:
```
let _ =
  match x with
  | { y =
        (* _____________________________________________________________________ *)
        ( X _ | Y _ )
    } -> ()
;;
```
I get this error message:
```
ocamlformat: Cannot process "original.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: formatting did not stabilize after 10 iterations.
```

**How to Reproduce**
Run `ocamlformat --profile=janestreet --enable-outside-detected-project original.ml`. Built from a906b5d993a6f321045546778ca8dfb4e45dfd90.

PR #1167, opened 2019-11-26
Fix unstable comment just before or-patterns

Julow:
Fixes https://github.com/ocaml-ppx/ocamlformat/issues/1164


lib/Fmt_ast.ml
```diff
@@ -1030,8 +1030,11 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
                   let loc = xpat.ast.ppat_loc in
                   let force_break = Cmts.has_before c.cmts loc in
                   let leading_cmt =
-                    Cmts.fmt_before ~pro:(Fmt.break 1000 0) ~adj:noop c loc
-                      ~eol:noop
+                    let pro, adj =
+                      if first_grp && first then (noop, fmt "@ ")
+                      else (fmt "@ ", noop)
```
Julow: I didn't find a case where `break 1000 0` was necessary.

lib/Fmt_ast.ml
```diff
@@ -1030,8 +1030,11 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
                   let loc = xpat.ast.ppat_loc in
                   let force_break = Cmts.has_before c.cmts loc in
                   let leading_cmt =
-                    Cmts.fmt_before ~pro:(Fmt.break 1000 0) ~adj:noop c loc
-                      ~eol:noop
+                    let pro, adj =
+                      if first_grp && first then (noop, fmt "@ ")
```
Julow: `pro` must be `noop` when `adj` is not to avoid extra empty lines.

Julow:
The last two indeed look like a fix. I'm not 100% sure for the first two, I added a test for that.

gpetiot:
I doubted at first because it kinda looks weird since the comment is long, no doubt for a shorter comment. But this is what is expected for `break-cases=fit`.

Julow:
I see. I tested too and it's a fix.
Should I put that in the changelog ? If yes, how ?

gpetiot:
If you don't find this too long I would modify your entry: `Fix an unstable comment bug in or-patterns; and fix wrapping of or-patterns in presence of comments when break-cases=fit`

Julow:
That's a bit long for a single line, I tried something.

PR #1169, opened 2019-11-29
Fix regression with `indicate-multiline-delimiters=space`

Julow:
This commit is from https://github.com/ocaml-ppx/ocamlformat/pull/1104

> Various commits have removed some, but not all, of the spaces before
closing parens when indicate_multiline_delimiters = Space. The result
is that the output is very inconsistent about when a closing paren
that matches one on a different line is preceded by a space or
not.

@gpetiot @jberdine

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -163,6 +163,7 @@
   + Fix dropped comments around module pack expressions (#1214) (Jules Aguillon)
   + Fix regression of comment position in list patterns (#1141) (Josh Berdine)
   + Fix: adjust definition of Location.is_single_line to reflect margin (#1102) (Josh Berdine)
+  + Fix regression with `indicate-multiline-delimiters=space` (#1169) (Josh Berdine)
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -122,6 +122,8 @@
   + Fix parenthesizing of optional arguments rebound to non-variables, e.g. `let
     f ?a:(A) = ()` rather than the unparsable `let f ?a:A = ()` (#1305) (Craig Ferguson)
 
+  + Fix regression with `indicate-multiline-delimiters=space` (#1169) (Josh Berdine, Guillaume Petiot)
+
 ### 0.13.0 (2020-01-28)
 
 #### New features
@@ -163,7 +165,6 @@
   + Fix dropped comments around module pack expressions (#1214) (Jules Aguillon)
   + Fix regression of comment position in list patterns (#1141) (Josh Berdine)
   + Fix: adjust definition of Location.is_single_line to reflect margin (#1102) (Josh Berdine)
-  + Fix regression with `indicate-multiline-delimiters=space` (#1169) (Josh Berdine)
 
 #### Documentation
 
--- a/test/passing/wrap_comments_break.ml
+++ b/test/passing/wrap_comments_break.ml
@@ -3,6 +3,6 @@ let _ =
     fffffffffff
       aaaaaaaaaaaaaaaaaaaaaaaaaaaa
       bbbbbbbbbbbbbbbbbbbbbbbbbb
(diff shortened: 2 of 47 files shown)


gpetiot:
I didn't look at the changes for the tests, but if we look at the functions you use:
- `Location.is_single_line` relies on the initial formatting of the file, and that's something we want to avoid in the future (existing usages of this function should progressively be removed if possible). Is there no alternative for what you want to do?
- `Fmt.fits_breaks` is a function I would like to replace in favor of `Fmt.cbreak`, as it is a necessary step towards removing the vendored `Format`, is there a way to use `cbreak` instead in this PR?

Julow:
I'll start maintaining this PR.

jberdine:
I don't have particular reasons for the functions used in this PR, it is just a selective revert of what seemed to be unrelated changes from a previous PR (sorry, I don't remember exactly which one), plus updates to the new labeled-arg interface of fits_breaks.

gpetiot:
I removed the call to `is_single_line` that caused unstable formatting and replaced the associated code with a `fits_breaks` as it is done with other changes in this PR.
Judging from https://github.com/ocaml/ocaml/pull/1229 `fits_breaks` is maybe the only function that we can hope to upstream in Format so we can use it.

test/passing/wrap_comments_break.ml
```diff
@@ -3,6 +3,6 @@ let _ =
     fffffffffff
       aaaaaaaaaaaaaaaaaaaaaaaaaaaa
       bbbbbbbbbbbbbbbbbbbbbbbbbb
-      ~f:(fun x -> return xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx)
+      ~f:(fun x -> return xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx )
```
gpetiot: we should remove this space

gpetiot: fixed

gpetiot:
@jberdine I think my last commit fixed the last `fun` cases, how does it look on your codebase compared to before #801 ?

gpetiot:
@Julow @emillon what should we do about this PR? It fixes a lot of inconsistencies on the closing parentheses.

CHANGES.md
```diff
@@ -22,6 +22,8 @@
   + Fix formatting of empty signature payload `[%a:]` (#1236) (Etienne Millon)
 
   + Fix parenthesizing when accessing field of construct application (#1247) (Guillaume Petiot)
+  
+  + Fix regression with `indicate-multiline-delimiters=space` (#1169) (Josh Berdine)
```
Julow: @gpetiot You could add your name as you fixed big issues.

gpetiot:
> The code looks good to me.
> There is a lot of diffs with `test_branch` (near 1000 lines on Dune and Infer).
> This feature is important, I think, but the change is big. What do you think @gpetiot, @emillon ?

Yes the diff is huge so it's difficult to check everything but from a few samples, I've only seen fixes. So even if users will complain about the diff the output will be more consistent so I think we should merge.

emillon:
let me check!

gpetiot:
> The code looks good to me.
> There is a lot of diffs with `test_branch` (near 1000 lines on Dune and Infer).
> This feature is important, I think, but the change is big. What do you think @gpetiot, @emillon ?

I checked the same thing while setting `indicate-multiline-delimiters=no` as the new default for the conventional profile, and the diff is as big. So one way or another we can't avoid having such a diff.

jberdine:
I added a [test](0ac343ef0f9c3a3ac2b8e9fea2383b9ba789cfca) and [fix](8190c451bdc7bed6deca3b4c6d4aab1a1a69e1ba) for a minor issue with this PR. See the [test-1169](https://github.com/ocaml-ppx/ocamlformat/commits/test-1169) branch.

PR #1171, opened 2019-12-03
Internal: allow tests to be postprocessed with ocp-indent

gpetiot:
I was trying to setup tests using `ocp-indent` to postprocess `ocamlformat` output. These tests require a  `foooo.ml.ocp` file in the test directory.
Some adjustments are necessary in the dune rules, as it currently raises an error after the 1st execution of `make test`.

First step to improve the `ocp-indent-compat` option used by JaneStreet.

### DIFF
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -20,7 +20,6 @@ license: "MIT"
 build: [
   ["ocaml" "tools/gen_version.mlt" "lib/Version.ml" version] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
   "ocaml" {>= "4.06"}
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -29,6 +29,7 @@ depends: [
   "dune" {>= "1.11.1"}
   "fpath"
   "ocaml-migrate-parsetree" {>= "1.3.1"}
+  "ocp-indent" {with-test}
   "odoc" {>= "1.4.2"}
   "re"
   "stdio"
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -20,6 +20,7 @@ license: "MIT"
 build: [
   ["ocaml" "tools/gen_version.mlt" "lib/Version.ml" version] {pinned}
   ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
   "ocaml" {>= "4.06"}
--- a/test/passing/dune.inc
+++ b/test/passing/dune.inc
@@ -1375,6 +1375,17 @@
  (action (diff js_source.ml.ref js_source.ml.output)))
 
 (rule
+ (targets js_source.ml.ocp.output)
+ (deps .ocamlformat )
+ (action
+   (with-outputs-to %{targets}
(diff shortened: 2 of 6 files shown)


test/passing/dune.inc
```diff
@@ -1374,6 +1374,18 @@
  (name runtest)
  (action (diff js_source.ml.ref js_source.ml.output)))
 
+(rule
+ (targets js_source.ml.ocp.output)
+ (deps .ocamlformat )
+ (action
+   (with-outputs-to %{targets}
+     (system "%{bin:ocamlformat} %{read-lines:js_source.ml.opts} %{dep:js_source.ml} | %{bin:ocp-indent} | %{bin:ocamlformat} %{read-lines:js_source.ml.opts} -"))))
```
Julow: Why not taking `js_source.ml.output` or `js_source.ml.ref` as input for ocp-indent ?
Also, what is tested by piping ocamlformat after ocp-indent ? Stability ?
I think we are interested in the output of ocp-indent instead, if ocp-indent-compat was perfect it shouldn't change anything of the output of ocamlformat

test/passing/js_source.ml.opts
```diff
@@ -1,2 +1,3 @@
 --max-iters=3
 --profile=janestreet
+--impl
```
Julow: I think this should be generated by `gen.ml` for every `.ocp` rules.

gpetiot: not necessary if we don't invoke ocamlformat after, removed.

test/passing/dune.inc
gpetiot: I tried to improve that:
- .ref file ->[ocp-indent]-> .ocp.output file
- .ocp.output file ->[promote]-> .ocp file

but dune raises an error: `exception (Invalid_argument js_source.ml.ocp.output)`

Julow: This is not raised by Dune but by `gen.ml` in `register_file` on line 49. You need to handle the `["ocp"; "output"]` case.

gpetiot: indeed, fixed thenks, maybe it's a good time to improve this error.

ocamlformat.opam
```diff
@@ -29,6 +30,7 @@ depends: [
   "dune" {>= "1.11.1"}
   "fpath"
   "ocaml-migrate-parsetree" {>= "1.3.1"}
+  "ocp-indent" {with-test}
```
gpetiot: looks like `ocp-indent` is not installed even with `--with-test`

Julow: This works for me locally with 4.09 and for the CI in 4.07 and 4.08.

Julow: I think I managed to fix this. The issue was that a previous opam pin was cached and re-adding a pin did update the source code but not the opam metadata.

test/passing/gen/gen.ml
```diff
@@ -72,8 +93,9 @@ let emit_test test_name setup =
 (alias
  (name runtest)
  (action (diff %s %s.output)))
-|}
+%s|}
     test_name extra_deps cmd_prefix opts base_test_name ref_name test_name
+    when_ocp
```
Julow: Passing that as a string argument is not necessary. This function (`emit_test`) is printing so this could be directly printed to stdout.

gpetiot: Indeed, fixed.

ocamlformat.opam
gpetiot: Great, thanks!

Issue #1152, opened 2019-11-15
Towards making comment handling more pure


emillon:
Hi,

I've been experimenting with removing mutable state from the formatting parts of ocamlformat. Here's a brain dump of my understanding before I forget.

At the moment, formatting uses mutable state to deal with comments: `Fmt_ast.c` holds a reference to a `Cmts.t`, which is implemented using some hash tables.

This is not a problem in itself, but makes several tasks harder.
As examples:

- unit testing is a bit trickier since one has to set up and check state of this external resource.
- some medium-sized refactoring tasks are made a lot more difficult. I tried to start converting Fmt.t to a an abstract type with an evaluator. One of the first steps was flattening `list_pn` calls to a series of `($)` applications. Turns out that this is not possible, as this changes the order in which comments are consumed. There are some workarounds, such as storing a thunk to call the layout function later, but it is error prone.
- more generally, this means that `f x $ g y` is not the same as `let r = f x in r $ g y`: the latter relies on the compiler's evaluation order
- this can cause subtle bugs; there are cases where a function that calls `Cmts.relocate` is called in a guard (`when` clause). I suspect that this can cause the comments to get dropped in some cases.

Now, how can we solve this?

We can do this by starting from the inside and expanding outwards:

- replace hash tables in `Cmts` by a `Map` reference
- replace the map ref by a mutable field
- replace the record updates by a record ref
- (at this point `Cmts.t` is pure)
- pass values around instead of updating the refs. this changes the signature of e.g. `Sugar` functions
- make `c.cmts` a mutable field
- replace state-modifying functions by wrappers that update `c.cmts`

The remaining parts are a bit less clear. We can either pass `cmts` around in `Fmt_ast`, but that seems heavy. Or we can move state-handling in the evaluation monad, using a state monad over `Cmts` (having `'a Fmt_eval.t = formatter -> Cmts.t -> 'a * Cmts.t`).

In any case, the first steps probably have value in themselves before we can see how to thread the state in the final step.

jberdine:
FWIW, almost from the beginning I have wanted to refactor to combine the format stream monad hiding inside Fmt with the state monad implicit in the comments tables. I wonder if there is a way to perform this combination without making the formatting code overly unreadable. But the interactions are currently very subtle in places, so the benefit would be significant.

gpetiot:
To make testing easier, the `place` and `partition_...` function can be dissociated from the hashtbl handling.
[...]

PR #1172, opened 2019-12-04
Purify Cmts module

emillon:
Hi,

As outlined in #1152, this switches the inner representation of `Cmts.t` so that it uses maps instead of hash tables. More precisely, comments tables are encapsulated as tables and their updates are done using `update_x` functions which should make the next step (turning field updates into functional updates) smoother.

Let me know what you think.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -17,12 +17,29 @@ open Asttypes
 open Parsetree
 
 type t =
-  { cmts_before: (Location.t, Cmt.t list) Hashtbl.t
-  ; cmts_after: (Location.t, Cmt.t list) Hashtbl.t
-  ; cmts_within: (Location.t, Cmt.t list) Hashtbl.t
+  { mutable cmts_before: Cmt.t Location.Multimap.t
+  ; mutable cmts_after: Cmt.t Location.Multimap.t
+  ; mutable cmts_within: Cmt.t Location.Multimap.t
   ; source: Source.t
-  ; remaining: (Location.t, unit) Hashtbl.t
+  ; mutable remaining: Location.Set.t
   ; remove: bool }
+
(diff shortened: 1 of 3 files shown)


emillon:
Also cc @NathanReb who's always in for some surgical refactoring.

lib/Cmts.ml
```diff
@@ -16,14 +16,109 @@ open Migrate_ast
...
+
+  let empty = Set.empty (module Location_comp)
+
+  let add key t = Set.add t key
+
+  let remove key t = Set.remove t key
+
+  let to_list t = Set.to_list t
+end
+
```
gpetiot: Would it make sense to move these modules to `Cmt`, e.g. `Cmt.Set` and `Cmt.Map` to remove some weight from this file?

emillon: yes, I think that would work. Map is a multimap (it returns lists) so the name can reflect that as well. `Location_comp` can also be merged into `Location`.

gpetiot: They can even end up in their own files if we want to keep them independent from comments, or even in a utility file.

emillon: I just did this. Actually, these are *location* set/multimaps, not *cmt* ones, so I moved them into `Location`. It wasn't to clear what `Migrate_ast` was exposing, so I also added an mli file for this module (spoiler: there was minor bits of dead code).

emillon: The mli file I wrote was specific to 4.08, I'll fix it

lib/Migrate_ast.mli
```diff
@@ -0,0 +1,139 @@
+val selected_version :
```
jberdine: This file needs the copyright header.

PR #1178, opened 2019-12-06
Internal: clean error handling of cmdline parsing

### LINKED ISSUE #1139: Bug: fatal errors are swallowed

gpetiot:
Cleaning when investigating #1139

### DIFF
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -75,7 +75,7 @@ let action, opts = Conf.action ()
 
 ;;
 match action with
-| Inplace inputs ->
+| `Ok (Inplace inputs) ->
     let errors =
       List.filter_map inputs
         ~f:(fun {kind; name= input_name; file= input_file; conf} ->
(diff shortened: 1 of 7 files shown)


gpetiot:
Rebased, and grouped together with the other "CLI refactoring" PRs in the changelog.

lib/Conf.mli
```diff
@@ -107,7 +107,7 @@ type opts =
         (** Check whether the formatted output exceeds the margin. *) }
 (** Options changing the tool's behavior *)
 
-val action : unit -> action * opts
+val action : unit -> action Cmdliner.Term.result * opts
```
Julow: This should be `(action * opts) Cmdliner.Term.result`. Eg. constructing `opts` in `validate` in lib/Conf.ml

bin/ocamlformat.ml
```diff
@@ -75,7 +75,7 @@ let action, opts = Conf.action ()
 
 ;;
 match action with
-| Inplace inputs ->
+| `Ok (Inplace inputs) ->
```
Julow: What do you think of moving the action cases to a `action -> int` function, returning the exit status ?

gpetiot: I did something slightly different, returning a result type instead, it's better I think.

lib/Conf.mli
gpetiot: done

PR #1179, opened 2019-12-06
Remove some polymorphic comparisons

emillon:
These can be tricky to remove later on so it is better to isolate them so that the polymorphic comparison is defined only in one place.

I tried to avoid exposing `is_*` function in the spirit of [tell-dont-ask](https://martinfowler.com/bliki/TellDontAsk.html) and preferring pattern matching like in `fmt_direction_flag`, but in some cases it was not possible.

Let me know what you think.

### DIFF
--- a/lib/Migrate_ast.mli
+++ b/lib/Migrate_ast.mli
@@ -3,10 +3,32 @@ val selected_version :
   Migrate_parsetree.Versions.ocaml_version
 
 module Selected_version = Ast_408
-module Parsetree = Selected_version.Parsetree
 module Ast_mapper = Selected_version.Ast_mapper
 module Ast_helper = Selected_version.Ast_helper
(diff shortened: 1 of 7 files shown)


lib/Migrate_ast.mli
```diff
@@ -3,10 +3,34 @@ val selected_version :
...
+end
+
+module Asttypes : sig
+  include module type of Selected_version.Asttypes
+
+  val is_private : private_flag -> bool
+
+  val is_open : closed_flag -> bool
+
+  val is_closed : closed_flag -> bool
```
gpetiot: I'm not sure it's necessary to have both functions here

emillon: I think it simplifies the call site a bit, but it might be overreaching in that case. I inlined it, thanks for the suggestion.

PR #1180, opened 2019-12-09
Fix regression in detecting project's root

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1139

The value of `--name` was used as the starting point when searching for `.ocamlformat` files.
This behavior has been removed in https://github.com/ocaml-ppx/ocamlformat/pull/1127 but this breaks editor integration.
The emacs plugin calls ocamlformat like that: `ocamlformat --name src/my_file.ml /tmp/some_tmp_dir/some_tmp_file.ml`.

This PR restore this behavior and add a test. Editors should probably use the `--root` option but it seems broken, any idea ?


jberdine:
> The value of --name was used as the starting point when searching for .ocamlformat files.
> This behavior has been removed in #1127

Independent of other considerations like editors, this itself seems like a bad idea. What is the motivation for this change?

Julow:
> Independent of other considerations like editors, this itself seems like a bad idea. What is the motivation for this change?

This wasn't intentional. I just found out about this behavior today.
Now there is a test for it and a little bit of doc, so it shouldn't disappear again.

emillon:
> Editors should probably use the --root option but it seems broken, any idea ?

I don't think so, they should either use project detection from ocamlformat or from dune build rules. Otherwise this means that it'll have to be implemented in editors, which doesn't sound great?

Julow:
> I don't think so, they should either use project detection from ocamlformat or from dune build rules. Otherwise this means that it'll have to be implemented in editors, which doesn't sound great?

I agree. Editors shouldn't do that, they might not even know the root of the project.

gpetiot:
The tests look good to me, if you managed to reproduce the bug in emacs before the PR and it is fixed with this PR we can merge.

Julow:
I think no changelog entry is needed as this fix a regression that hasn't been released. The fix works for me but I don't have a complete emacs setup.

PR #1181, opened 2019-12-09
Extract, test and refactor Non_overlapping_interval_tree

emillon:
Hi,

In order:
- this moves `Non_overlapping_interval_tree` to its own file
- some tests are added
- it's refactored so that it does not use mutability internally (that was only during construction, but the result seems better anyway)

Let me know what you think.

### DIFF
--- a/lib/Non_overlapping_interval_tree.ml
+++ b/lib/Non_overlapping_interval_tree.ml
@@ -1,0 +1,100 @@
+module type IN = sig
+  include Comparator.S
+
+  val contains : t -> t -> bool
+
+  val compare_width_decreasing : t -> t -> int
+end
+
+module type S = sig
+  type itv
+
+  type t
+
+  val of_list : itv list -> t
+  (** If there are duplicates in the input list, earlier elements will be
+      ancestors of later elements. *)
+
+  val roots : t -> itv list
+
+  val children : t -> itv -> itv list
+
+  val dump : t -> Fmt.t
+  (** Debug: dump debug representation of tree. *)
+end
+
+module Make (Itv : IN) = struct
+  (* simple but (asymptotically) suboptimal implementation *)
+
+  type itv = Itv.t
+
+  type t = {roots: Itv.t list; map: Itv.t list Map.M(Itv).t}
+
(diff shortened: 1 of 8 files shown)


test/unit/test_unit.ml
```diff
@@ -0,0 +1,100 @@
...
+  in
+  let test_children name l i expected =
+    ( "children: " ^ name
+    , `Quick
+    , fun () ->
+        let t = T.of_list l in
+        let got = T.children t i in
+        Alcotest.check itv_list Caml.__LOC__ expected got )
+  in
+  [ test_dump "empty" [] ""
```
emillon: @NathanReb how do you structure alcotest suites these days? does that look ok or do you think it's better to pull functions and use `List.concat [test_dump; test_roots; test_children]`?

NathanReb: Usually the way I do it is that for each function under test I maintain a list of test-cases `(name * speed * unit -> unit)`.

I then group them by modules or submodules by concatenating them. You can take a look at an example here: https://github.com/ocamllabs/dune-release/blob/master/tests/test_pkg.ml.

It requires a tiny bit of boiler plate but results in nice and helpful output from alcotest and works well with the `-e` option.

From a quick look that's pretty much what you're doing here with one exception, I pass the test case name to `Alcotest.check` instead of the loc as the loc will be the same for every test case and therefore won't be of much help when looking at the error report from Alcotest.

emillon: Thanks, I just did this. Seems better indeed!

emillon:
Thanks. Squashing and merging!

lib/Non_overlapping_interval_tree.ml
```diff
@@ -0,0 +1,100 @@
+module type IN = sig
```
jberdine: This file needs the copyright header.

lib/Non_overlapping_interval_tree.mli
```diff
@@ -0,0 +1,30 @@
+(** A tree of non-overlapping intervals. Intervals are non-overlapping if
```
jberdine: This file needs the copyright header.

PR #1182, opened 2019-12-10
Internal: copy the 4.08 lexer and parser in ocamlformat

gpetiot:
Weird bug (maybe due to menhir?)
```
Error: Module Parser in directory _build/default/lib depends on
Ocamlformat_lib.
This doesn't make sense to me.

Ocamlformat_lib is the main module of the library and is the only module
exposed outside of the library. Consequently, it should be the one depending
on all the other modules in the library.
```
I can fix it be removing the `lib/Migrate_ast.mli` file but it would be better to find another solution.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@
   + Remove global reference `Cmts.remove` (#1142) (Etienne Millon)
   + Remove symbol strings in `Cmts` (#1146) (Etienne Millon)
   + Allow tests to be postprocessed with ocp-indent (#1171) (Guillaume Petiot, Jules Aguillon)
+  + Copy the 4.08 lexer and parser in ocamlformat (#1182) (Guillaume Petiot)
 
 #### Documentation
 
--- a/dune-project
+++ b/dune-project
@@ -13,3 +13,4 @@
 
 (name ocamlformat)
 (using fmt 1.2)
+(using menhir 2.0)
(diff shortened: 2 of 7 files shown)


jberdine:
I'm sorry if this adds overhead, but we need to keep code copied in from elsewhere under `vendor`, and similarly keep code not copied in from elsewhere out of `vendor`.

emillon:
Yes, I agree that this will need to be carefully isolated from the rest of the codebase.

But how does versioning work with an embedded parser? we don't really need OMP anymore, do we?

About the `dune` error, I'm not sure what causes this. `Parser.ml` contains `Ocamlformat_lib.Migrate_ast.Docstrings.info`, which rightfully infers a dependency to `Ocamlformat_lib`, but I don't know where this module name comes from (I'd expect just `Migrate_ast.Docstrings.info`). We can try to write a repro case if that's a dune bug.

emillon:
Also, wouldn't this cause a regression with 4.09 support? I think we should use a feature branch here and clean up this work when it's established that the corresponding feature (formatting invalid files) works.

gpetiot:
I planned to use this customize lexer to improve the attachment of comments with respect to what is already done for docstrings (just an idea, no real plan so far, but needs the lexer to experiment).
It will probably cause regressions with 4.09 indeed though :(

gpetiot:
Closing for now, exploring another solution.

PR #1183, opened 2019-12-10
Remove "Internal" changelog section

emillon:
These will not be useful to end users, and the information is already there in the git log.

See #1166

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,10 @@
   + Add an option `--margin-check` to emit a warning if the formatted output exceeds the margin (#1110) (Guillaume Petiot)
   + Preserve comment indentation when `wrap-comments` is unset (#1138, #1159) (Jules Aguillon)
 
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -17,22 +17,6 @@
   + Remove some calls to if_newline and break_unless_newline and fix break before closing brackets (#1168) (Guillaume Petiot)
   + Fix unstable cmt in or-pattern (#1173) (Guillaume Petiot)
 
-#### Internal
-
-  + Work with base v0.13.0 (#1163) (Jules Aguillon)
```
Julow: Should we keep this one ? It impacts end-users but they can have that informations from opam

gpetiot: Breaking changes can be in the Changes section

emillon: Good point, I added a section for this.

PR #1185, opened 2019-12-11
Remove reason support

emillon:
As discussed in #254, keeping this functionality around is not very useful:

- it has never been released on opam
- it has no known users
- `refmt` can already emit ocaml code

There are limitations with `refmt`, but workarounds should be implemented there, not as an unrelated tool in `ocamlformat`.


README.md
```diff
@@ -152,15 +151,6 @@ OCamlFormat is documented in its man page and through its internal help:
...
 
-## Reason
-
-OCamlFormat is influenced by and follows the same basic design as `refmt` for [Reason](https://github.com/facebook/reason), but outputs OCaml instead of Reason.
-
-Support for converting Reason to OCaml is included in the `ocamlformat_reason` package, which works by using `refmt` to parse Reason into an OCaml parse tree. The Reason converter can be installed using `opam`:
-```
-opam pin add ocamlformat_reason https://github.com/ocaml-ppx/ocamlformat.git
-```
-
```
gpetiot: Maybe we should keep a note there to tell users that we do not interact with Reason code, but redirect them to `refmt`

jberdine:
I'm in favor of removing the unused Reason support. I tried to do it earlier but there was some resistance.

I think it would be good to leave something in the Readme to indicate that tighter cooperation between ocamlformat and refmt would be possible if someone wants to work on it, and that such a contribution would be welcome. In particular, I think that seamless interop would be quite valuable as a tooling solution to reduce the gap between the communities.

It would still seem right to me to acknowledge the past influence of refmt even after removing Reason support though.

emillon:
I left the note acknowledging the influence of `refmt` and added a bit about how to format reason code using ocamlformat.

gpetiot:
Thanks, looks good to be merged now.

PR #1188, opened 2019-12-12
WIP: Format only a line range from a file

emillon:
This adds a `--lines` option to format only part of a file (see #362, #366).

The strategy is to split the file in 3, format the middle part (which should be valid ocaml), and paste the pieces together.

Notes:

- This is very limited in the current state. An extension of this would be to automatically compute the smallest enclosing structure item to format "around" the selection.
- The `Line_range` abstraction doesn't work too well for now. It should evolve to something that partitions the loc space into formattable and verbatim ranges, which in turn should be helpful to implement formatting of invalid files.


gpetiot:
This proposal sounds good to me, it looks similar to what I had in mind for invalid files (and I have the feeling it would work well together --> required when the user selects a line range that delimits invalid code).

lib/Line_range.ml
```diff
@@ -0,0 +1,55 @@
...
+  index_nth source '\n' (low - 1)
+  >>= fun s_pos ->
+  index_nth source '\n' high
+  >>= fun e_pos ->
+  let p1 = s_pos + 1 in
+  let p2 = e_pos + 1 in
+  Some
+    ( String.subo source ~len:p1
+    , String.sub source ~pos:p1 ~len:(p2 - p1)
+    , String.subo source ~pos:p2 )
```
Julow: It's not possible to select the first line.

gpetiot:
What is the state of this PR? Are there some bits that someone else can look at?

lib/Conf.mli
```diff
@@ -89,7 +89,8 @@ type t =
 
 type file = Stdin | File of string
 
-type 'a input = {kind: 'a; name: string; file: file; conf: t}
+type 'a input =
+  {kind: 'a; name: string; file: file; conf: t; line_range: Line_range.t}
```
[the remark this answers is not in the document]
gpetiot: done

lib/Line_range.ml
gpetiot: fixed, not satisfied with the implementation but it allows to move forward with #1207

gpetiot:
Can it be modified to work with cppo input such as:
```ocaml
let opaque_identity x =
  #if OCAML_VERSION < (4, 03, 0)
    x
  #else
    Sys.opaque_identity x
  #endif
```
?

gpetiot:
If the purpose of this PR is only IDE integration it can be replaced by #1609 (the cutting/composing part would not be used)

PR #1189, opened 2019-12-12
Display standard output in the emacs plugin even when ocamlformat does not fail

### LINKED ISSUE #1170: Feature request: show parse-docstrings warnings in emacs
Currently, when there are warnings from `parse-docstrings = true`, they are not shown by the emacs integration. When invoking ocamlformat from a shell, the warnings are sent to stderr, and ocamlformat exits with success `0`. In this case, invoking ocamlformat using the emacs integration silently swallows the warnings. It would be nice to (have an option to) show these warnings.

gpetiot:
Fix #1170

### DIFF
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -261,6 +261,13 @@ function."
                (if ocamlformat--support-replace-buffer-contents
                    (ocamlformat--replace-buffer-contents outputfile)
 		 (ocamlformat--patch-buffer outputfile))
+               (if errbuf
+                 (progn
+                   (with-current-buffer errbuf
+                     (setq buffer-read-only nil)
+                     (erase-buffer))
+                   (ocamlformat--process-errors
+                    (buffer-file-name) bufferfile errorfile errbuf)))
                (message "Applied ocamlformat"))
              (if errbuf
                (progn
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -164,7 +164,7 @@ function."
         (erase-buffer))
       (with-current-buffer errbuf
         (if (eq ocamlformat-show-errors 'echo)
-            (message "%s" (buffer-string))
+          (message "%s" (buffer-string))
           (insert-file-contents errorfile nil nil nil)
           ;; Convert the ocamlformat stderr to something understood by the
           ;; compilation mode.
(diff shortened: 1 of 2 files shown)


emacs/ocamlformat.el
```diff
@@ -261,14 +270,11 @@ function."
...
                (message "Applied ocamlformat"))
-             (if errbuf
-               (progn
-                 (with-current-buffer errbuf
-                   (setq buffer-read-only nil)
-                   (erase-buffer))
-                 (ocamlformat--process-errors
-                   (buffer-file-name) bufferfile errorfile errbuf)))
+             (ocamlformat--process-errors
+               (buffer-file-name) bufferfile errorfile errbuf)
```
Julow: You could also call it from outside this if, after `message`.

emacs/ocamlformat.el
```diff
@@ -157,17 +157,26 @@ function."
...
-      (insert "ocamlformat errors:\n")
-      (while (search-forward-regexp (regexp-quote tmpfile) nil t)
-        (replace-match (file-name-nondirectory filename)))
-      (compilation-mode)
-      (display-buffer errbuf))))
+  (if errbuf
+    (progn
+      (with-current-buffer errbuf
+        (setq buffer-read-only nil)
+        (erase-buffer))
```
Julow: It was already like this but wouldn't make more sense to do that in the "not `'echo`" case ?

emacs/ocamlformat.el
```diff
@@ -157,17 +157,26 @@ function."
...
+      (with-current-buffer errbuf
+        (if (eq ocamlformat-show-errors 'echo)
+          (message "%s" (buffer-string))
+          ;; Checks the size of errorfile
+          (if (> (nth 7 (file-attributes errorfile)) 0)
```
Julow:
```suggestion
          (if (> (file-attribute-size (file-attributes errorfile)) 0)
```
This works for me :)

gpetiot: I added a comment to keep track of why it's not the case

emacs/ocamlformat.el
gpetiot: done

emacs/ocamlformat.el
gpetiot: done

PR #1190, opened 2019-12-12
Add copyright header to Migrate_ast.mli

emillon:


lib/Migrate_ast.mli
```diff
@@ -1,3 +1,14 @@
+(**********************************************************************
+ *                                                                    *
+ *                            OCamlFormat                             *
+ *                                                                    *
+ *  Copyright (c) 2017-present, Facebook, Inc.  All rights reserved.  *
```
gpetiot: should be the current year I guess

jberdine: Ah, actually I haven't gotten around to it, but we don't need years at all now. We could either remove them all in one commit or just not add them or...

emillon: If you can update them that'd be great. thanks!

PR #1191, opened 2019-12-12
Extract & test Cmt.normalized_string

emillon:
This function parses comments as text or code and normalizes them for use in `Cmts.diff`.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -492,6 +492,9 @@ let remaining_before t loc = Location.Multimap.find_multi t.cmts_before loc
 let remaining_locs t = Location.Set.to_list t.remaining
 
 let diff (conf : Conf.t) x y =
-  let norm_cmt = Cmt.normalize ~normalize_impl:(Normalize.impl conf) in
+  let norm_cmt =
+    Cmt.normalized_string ~normalize_comment:Normalize.comment
+      ~normalize_impl:(Normalize.impl conf)
+  in
   let norm l = Set.of_list (module String) (List.map ~f:norm_cmt l) in
   Set.symmetric_diff (norm x) (norm y)
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -492,9 +492,6 @@ let remaining_before t loc = Location.Multimap.find_multi t.cmts_before loc
 let remaining_locs t = Location.Set.to_list t.remaining
 
 let diff (conf : Conf.t) x y =
-  let norm_cmt =
-    Cmt.normalized_string ~normalize_comment:Normalize.comment
-      ~normalize_impl:(Normalize.impl conf)
-  in
+  let norm_cmt = Cmt.normalize ~normalize_impl:(Normalize.impl conf) in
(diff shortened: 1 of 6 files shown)


emillon:
I'm a bit surprised that is uses `Printast.implementation` rather than `Pprintast.structure` to normalize the structure. Do you know why this is the case? it seems that locs do not matter (since it's all built from a string, so `Pprintast` should do (and it's smaller).

lib/Cmts.ml
```diff
@@ -626,25 +626,9 @@ let remaining_before t loc = Location.Multimap.find_multi t.cmts_before loc
...
-              Migrate_ast.Parse.implementation (Lexing.from_string str)
-              |> Normalize.impl conf
-              |> Caml.Format.asprintf "%a" Printast.implementation
-            with _ -> norm_non_code z
-          else norm_non_code z
-    in
-    Set.of_list (module String) (List.map ~f z)
+  let norm_cmt =
+    Cmt.normalized_string ~normalize_comment:Normalize.comment
+      ~normalize_impl:(Normalize.impl conf)
```
gpetiot: I wonder if we should instead:
- move `Normalize.comment` into `Cmt` (like `Cmt.basic_normalize` that does not need to be exported)
- rename `Cmt.normalized_string` to `Cmt.normalize`, that would call `basic_normalize` instead of `normalize_comment`

gpetiot:
> I'm a bit surprised that is uses `Printast.implementation` rather than `Pprintast.structure` to normalize the structure. Do you know why this is the case? it seems that locs do not matter (since it's all built from a string, so `Pprintast` should do (and it's smaller).

It doesn't ring a bell to me, if they both give the same output (no diff with test_branch.sh) we can use the most efficient.

lib/Cmts.ml
emillon: We can't have a `Cmt` -> `Normalize` dependency so `normalize_impl` will have to be passed in, does that sound OK?

gpetiot: Maybe I missed something but `Normalize.comment` does not pull additional Normalize dependencies.

emillon: OK I though you meant doing this step in addition of this PR, not instead of. I'll try something.

jberdine:
I originally found the comparisons of output from `Printast` to be more useful when debugging than the output of `Pprintast`.

gpetiot:
I updated the branch with the change I suggested, so that we don't have to pass normalize_comment to the normalized_string function. Let me know what you think.

emillon:
As you prefer. TBH this was helpful for me to understand this piece of code so I wrote some tests, but we can also close this PR.

gpetiot:
We can also wrap the normalized comments with `C()` directly in the `Cmt.basic_normalize` function if it makes sense and makes it easier to debug. This function should only be used in Normalize and not for formatting anyway.

gpetiot:
I'm closing this one for now, I think I will rewrite the code of Normalize somehow but I would like to avoid having to pass normalization functions to Cmt.ml

PR #1193, opened 2019-12-13
Improve the placement of comments

gpetiot:
still in progress:
- [x] move the algorithm and the abstraction from `Cmts` to `Source`
- [x] fix the regressions when formatting ocamlformat source (make fmt)
- [x] fix the regressions in the tests (make test)
- [ ] test on big projects (dune, mirage, etc.)
- [ ] cleanup and document

The goal is only to improve the maintainability, bugs will be fixed in future pull requests, to keep the diff of this one short enough.


emillon:
I (finally!) had a look at this. If I understand it correctly, the core change is the change from `partition_after_prev_or_before_next` to `Source.cmt_placement`. I have some remarks on the implementation itself but it might a bit early to go in too much detail.

The or-pattern changes seem to be a net improvement to me! And the rules seems much simpler to understand.

It's a bit early to talk integration, but if you can continue improving that on the side and have net wins like the or-patterns I think this can balance the large diff this is going to have. We can also gate that behind a feature flag if the outcome is not super clear.

gpetiot:
I'm closing this one because:
- it diverged from master too much and it's painful to maintain
- I will explore the approach of attaching the comments _only once_, just after the parsing, instead of after each iteration until we reach a fixpoint, for that I will work on having ast passes (and one of these passes will attach the comments in the ast)

hhugo:
> I will explore the approach of attaching the comments only once, just after the parsing, instead of after each iteration until we reach a fixpoint, for that I will work on having ast passes (and one of these passes will attach the comments in the ast)

You'll still need to make sure you reach a fix point with the comment placement.

PR #1198, opened 2019-12-19
gen_version: work when building without a git root

avsm:
In some cases, a development tree doesn't have a .git root
(for example, when a source monorepo is assembled using the
duniverse vendoring tool).  In this situation, instead of
leaking an exception, the gen_version tool with just stamp
the "unknown" version and let the build continue.

Signed-off-by: Anil Madhavapeddy <anil@recoil.org>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 
 #### Bug fixes
 
+  + Fix build version detection when building in the absence of a git root (#1198) (Anil Madhavapeddy)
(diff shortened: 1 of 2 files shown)


avsm:
Thanks, CHANGES entry pushed

emillon:
this'll be superseded by #1133 which I intend to refresh soon but it's nice to have a solution in the meantime.

Note that this solution means that all versioned `.ocamlformat` files will fail because of a version mismatch, but that's already the case in a duniverse.

PR #1199, opened 2019-12-19
Constraint ocaml to 4.08 in opam file when tests are enabled

Julow:
OCamlformat builds for OCaml 4.06 to 4.09 but its tests are only passing between 4.08 and 4.09.

This PR constraints OCaml to `>= 4.08` in Opam when tests are enabled.
This will improve the output of ocaml-ci as tests for these version will not be run, so they don't fail and don't consume power and are instead interpreted as a problem with the dependencies and reported differently.

Currently, that simply disable CI builds of 4.06 to 4.08. In the future, ocaml-ci may split build and test jobs and allow to test builds for these versions.

It seems that this does not prevent from installing OCamlformat on 4.06 and 4.07 but correctly prevent installation with tests (`opam install -t`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: c
 env:
   - CI_KIND=changes
   - CI_KIND=build OCAML_VERSION=4.07.1
-  - CI_KIND=build OCAML_VERSION=4.08.1
+  - CI_KIND=build-and-tests OCAML_VERSION=4.08.1
   - CI_KIND=build-and-tests OCAML_VERSION=4.09.0
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


gpetiot:
The tests are only run for 4.09 and not 4.08 though.

Julow:
True. The tests are passing on 4.08, that's why I allowed them.

An alternative would be to to add the constraint on the runtest invocation: (suggested by @kit-ty-kate)
```
  ["dune" "runtest" "-p" name] {with-test & ocaml:version >= "4.08"}
```
But this will silently not run the tests, what do you think ?

Julow:
I added the constraint to the runtest line too. This way, we can drop (or not) the constraint on OCaml when releasing. This extra constraint doesn't do anything now but will be useful if that happen.
What do you think?

gpetiot:
Looks good, feel free to merge if it's ready.

Issue #1160, opened 2019-11-19
Bug: polymorphic variant type w/ comments does not stabilize


ceastlund:
**Describe the bug**
I format the following code:
```
type t = [
  | `XXXX (* __________________________________________________________________________________ *)
  | `XXXX (* __________________________________________________________________ *)
  | `XXXX (* _____________________________________________________ *)
  | `XXXX (* ___________________________________________________ *)
  | `XXXX (* ___________________________________________________ *)
  | `XXXX (* ________________________________________________ *)
  | `XXXX (* __________________________________________ *)
  | `XXXX (* _________________________________________ *)
  | `XXXX (* ______________________________________ *)
  | `XXXX (* ____________________________________ *)
]
```
I get this error message:
```
ocamlformat: Cannot process "original.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: formatting did not stabilize after 10 iterations.
```

**How to Reproduce**
Run `ocamlformat` built from a906b5d993a6f321045546778ca8dfb4e45dfd90 with `--profile=janestreet`.

hhugo:
Are there any issue from https://github.com/ocaml-ppx/ocamlformat/issues/985 solved by this PR ?

PR #1203, opened 2020-01-02
Fix placement of comments just before a '|'

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/1160

The problem was that when the comment breaks:

```ocaml
type t = [
  | `XXXX
    (* __________________________________________________________________________________ *)
  | `XXXX (* __________________________________________________________________ *)
  | ...
]
```

On the next round, it would be attached to before the second variant instead of after the first one:

```ocaml
type t = [
  | `XXXX
  | (* __________________________________________________________________________________ *)
    `XXXX (* __________________________________________________________________ *)
  | ...
]
```

Which changes the meaning of the comment and force the next one to break.


test/passing/js_source.ml.ref
```diff
@@ -1047,8 +1047,9 @@ let rec variantize : type t. t ty -> t -> variant =
   | Int -> VInt x (* in this branch: t = int *)
   | String -> VString x (* t = string *)
   | List ty1 -> VList (List.map (variantize ty1) x) (* t = 'a list for some 'a *)
-  | Pair (ty1, ty2) -> VPair (variantize ty1 (fst x), variantize ty2 (snd x))
-  (* t = ('a, 'b) for some 'a and 'b *)
+  | Pair (ty1, ty2) ->
+    VPair (variantize ty1 (fst x), variantize ty2 (snd x))
+    (* t = ('a, 'b) for some 'a and 'b *)
```
Julow: This is not a regression, the source is:
```ocaml
    | Pair (ty1, ty2) ->
        VPair (variantize ty1 (fst x), variantize ty2 (snd x))
        (* t = ('a, 'b) for some 'a and 'b *)
```

PR #1204, opened 2020-01-06
Fix align-cases for long patterns

### LINKED ISSUE #1202: Feature request: alignment of "of" and arrow in long match
We recently came across several match expression alignment issues in Owl and related projects. For example, in [this](https://github.com/owlbarn/owl_symbolic/blob/master/src/onnx_specs/onnx_pb.ml#L1003-L1010) code the arrows are not aligned. Initially we thought it's perhaps because one or more of the match lines will be too long if they are aligned, but [this line](https://github.com/owlbarn/owl_symbolic/blob/master/src/onnx_specs/onnx_pb.ml#L947) above in the same file is obviously longer but still got aligned. I'm wondering the reason for this behaviour and if there is any way to fix it.

The other thing is that the word "of" is not aligned in type declaration, such as in this [example](https://github.com/owlbarn/owl/blob/master/src/base/neural/owl_neural_neuron.ml#L3156-L3199). It would be great if the "of"s could also be aligned as the arrows.

gpetiot:
Fix #1202
I removed the check on the size of the formatted result not precise since we obviously don't take the final indentation into account, and overly cautious and prevented to align when it would fit on the same line.
No regression with running test_branch on the preset profiles, but there is a big diff with only `align-cases=true` (most of them are improvements, some cases can be debatable, but since it's still an heuristic it's difficult to balance).

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2836,23 +2836,14 @@ and fmt_cases c ctx cs =
       ~finish:(fun acc -> Some acc)
   in
   let max_len = fold_pattern_len ~f:max cs in
-  let level = match c.conf.break_cases with `Nested -> 2 | _ -> 3 in
   list_fl cs (fun ~first ~last case ->
       let padding =
         let xlhs = sub_pat ~ctx case.pc_lhs in
-        let add_padding =
-          c.conf.align_cases && not (Cmts.has_after c.cmts xlhs.ast.ppat_loc)
-        in
+        let has_cmts = Cmts.has_after c.cmts xlhs.ast.ppat_loc in
         match max_len with
-        | Some max_len when add_padding -> (
+        | Some max_len when c.conf.align_cases && not has_cmts -> (
           match pattern_len case with
-          | Some pattern_len ->
-              let pad = String.make (max_len - pattern_len) ' ' in
-              Some
-                (fmt_or_k
-                   Poly.(c.conf.break_cases = `All)
-                   (str pad)
-                   (fits_breaks ~level "" pad))
+          | Some pattern_len -> Some (break (max_len - pattern_len) 0)
           | _ -> None )
         | _ -> None
       in
@@ -2894,7 +2885,10 @@ and fmt_case c ctx ~first ~last ?padding case =
   in
   let indent = if align_nested_match then 0 else indent in
   let p = Params.get_cases c.conf ~first ~indent ~parens_here in
-  p.leading_space $ leading_cmt
+  fmt_or_k (Option.is_some padding)
+    (break_unless_newline 1000 0)
+    p.leading_space
+  $ leading_cmt
   $ p.box_all
       ( p.box_pattern_arrow
           ( hvbox 0
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2813,32 +2813,18 @@ and fmt_class_type_field c ctx (cf : class_type_field) =
     $ fmt_atrs $ doc_after )
 
 and fmt_cases c ctx cs =
-  let case_breaks case ~padding_len =
-    let fmted_case =
-      let first = false and last = false in
-      let fmt cmts = fmt_case {c with cmts} ctx ~first ~last case in
-      Cmts.preserve fmt c.cmts
-    in
-    let indent =
-      Source.indentation c.source case.pc_rhs.pexp_loc.loc_start
-    in
-    (* The first character is ignored when looking for a linebreak, as it
-       could be a forcedline break in some configuration. *)
-    String.contains fmted_case ~pos:1 '\n'
-    || String.length fmted_case + indent + padding_len >= c.conf.margin
-  in
   let pattern_len {pc_lhs; pc_guard; _} =
     if Option.is_some pc_guard then None
     else
       let xpat = sub_pat ~ctx pc_lhs in
-      let fmted_pat =
+      let fmted =
         Cmts.preserve (fun cmts -> fmt_pattern {c with cmts} xpat) c.cmts
       in
-      let len = String.length fmted_pat in
+      let len = String.length fmted in
       let indent = Source.indentation c.source pc_lhs.ppat_loc.loc_start in
       (* we add the lenght of the 5 following characters '| ' ' ->' *)
-      if len + indent + 5 >= c.conf.margin || String.contains fmted_pat '\n'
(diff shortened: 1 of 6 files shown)


lib/Fmt_ast.ml
```diff
@@ -2783,16 +2783,24 @@ and fmt_class_type_field c ctx (cf : class_type_field) =
...
+      if String.contains fmted_pat '\n' then None
+      else
+        let fmted_case =
+          let first = false and last = false in
+          let fmt cmts = fmt_case {c with cmts} ctx ~first ~last case in
+          Cmts.preserve fmt c.cmts
+        in
+        (* The first character is ignored when looking for a linebreak, as it
+           could be a forcedline break in some configuration. *)
+        if String.contains fmted_case ~pos:1 '\n' then None else Some len
```
Julow: This line break could be from the body of the case and I think we should align in this case too.

The change allowing longer patterns is the removal of `len * 3 >= c.conf.margin` and this won't help with indentation. This hack was added because we don't know the current indentation.

This example with a margin of 22 will result in trailing spaces with `break-cases=all` and exceed the margin without:
```ocaml
let _ =
  let _ =
    let _ =
      match a with
      | Foooo -> 1
      | Fooooooooooo -> 2
    in
    ()
  in
  ()
```

gpetiot:
I improved the situtation regarding the indentation, but it's better not to add the padding when the body of the case breaks, it fixes cases like this one:
```diff
     match provides with
-    | None                      -> always_included :=
-                                     { filename; program = code } :: !always_included
+    | None -> always_included := { filename; program = code } :: !always_included
     | Some (pi, name, kind, ka) ->
```
test_branch.sh testing `align-cases=true` shows a lot of these cases have been fixed.

Julow:
This looks like a different bug. There should be a break just after the `->`, like in the `Some` case.
The indentation issue is indeed fixed (by looking at the source string).

I'm not sure looking if the case breaks is a good way of solving this, this will unalign all the patterns if a single case breaks. It happens sometimes in Owl sources (linked in the issue).
On the other hand, it's slightly better than the previous check (`len * 3 >= c.conf.margin`) at preventing huge paddings while being less arbitrary.

gpetiot:
I feel like to correctly apply the padding without forcing the cases to break we would need to have information such that did some box break (we need introspection on the boxes)

Julow:
What is wrong with breaking every cases ?
```ocaml
  | Small                              ->
      small_but_breaks
  | Big _____________________________a ->
      small_but_breaks
```

gpetiot:
I improved the output by breaking every case before the `|` when a padding is added. But that means we don't have pattern-matching on one line anymore with `--align-cases`.

Julow:
When combined with `break-cases=fit` and a margin of `38`:
```ocaml
let _ =
  match a with
  | Foooooo -> aaaaaaaa && bbbbbbbb
  | Fooooooooooo -> cccccccc
```
formats to:
```ocaml
let _ =
  match a with
  | Foooooo      -> aaaaaaaa
                    && bbbbbbbb
  | Fooooooooooo -> cccccccc
```

gpetiot:
That's surprising, I don't get the same formatting, I get:
```ocaml
let _ =
  match a with
  | Foooooo      ->
      aaaaaaaa && bbbbbbbb
  | Fooooooooooo -> cccccccc
```

Julow:
That bug is gone, I may have not pulled the last version of your PR.

Julow:
This makes a lot of diffs on Owl (that's expected). They usually look good:
```diff
--- a/src/base/algodiff/owl_algodiff_graph_convert.ml
+++ b/src/base/algodiff/owl_algodiff_graph_convert.ml
@@ -23,9 +23,9 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
           let op, prev =
             match hd with
             | DR (_ap, _aa, (_, _, label), _af, _ai, _) -> label
-            | F _a -> Printf.sprintf "Const", []
-            | Arr _a -> Printf.sprintf "Const", []
-            | DF (_, _, _) -> Printf.sprintf "DF", []
+            | F _a                                      -> Printf.sprintf "Const", []
+            | Arr _a                                    -> Printf.sprintf "Const", []
+            | DF (_, _, _)                              -> Printf.sprintf "DF", []
           in
           (* check if the node has been visited before *)
```

But in most cases, the padding is huge:
```diff
--- a/src/base/algodiff/owl_algodiff_reverse.ml
+++ b/src/base/algodiff/owl_algodiff_reverse.ml
@@ -24,7 +24,7 @@ struct
           af := !af + 1;
           tracker := succ !tracker;
           if !af = 1 && !tracker = 1 then reset (register t) else reset t
-        | _ -> reset t)
+        | _                                                -> reset t)
     in
     reset [ x ]

--- a/test/unit_algodiff_matrix_generic.ml
+++ b/test/unit_algodiff_matrix_generic.ml
@@ -315,7 +315,7 @@ module Make (M : Ndarray_Algodiff with type elt = float) = struct
            (fun (b, error_msg) (s, f) ->
              let b', c =
                match f () with
-               | x -> x
+               | x                                         -> x
                | exception Owl_exception.NOT_IMPLEMENTED _ -> true, n
              in
```

There is also a few bug fixes:
```diff
   | Float64   -> owl_float64_ndarray_conv_dilated_cuboid_backward_input_im2col
   | Complex32 -> owl_complex32_ndarray_conv_dilated_cuboid_backward_input_im2col
   | Complex64 -> owl_complex64_ndarray_conv_dilated_cuboid_backward_input_im2col
-  | _         -> failwith
-                   "_owl_dilated_cuboid_conv_backward_input_im2col: unsupported operation"
+  | _         ->
+    failwith "_owl_dilated_cuboid_conv_backward_input_im2col: unsupported operation"
```

jberdine:
This one is good to rebase and merge, right?

gpetiot:
We don't know what to do about the `align-*` options, we are considering removing them, but as they are used in `owl` formatting settings we have to discuss it with them first.

jberdine:
I thought that Jane Street was also in the vertical alignment camp. Is there a desire/plan to have enough vertical alignment in ocamlformat to avoid needing to post-process its output?

gpetiot:
I'm not aware of JaneStreet using these align options, it's not in the profile at least. Our wish is to purely remove these options in ocamlformat and not allow to control the alignment.

gpetiot:
Closing this one for now as the future of these options is uncertain.

PR #1205, opened 2020-01-13
Remove ocamlformat-diff

gpetiot:
Next step is adding a documentation for `merge-fmt`.


JasonGross:
I don't see a way to use `merge-fmt` as a difftool for git (so that `git diff` is only up to ocamlformat).  Is there a way to do this? Or a chance of getting `ocamlformat-diff` back into `ocamlformat`?

Julow:
`merge-fmt` cannot be used as a difftool, indeed.
I thought we could use `diff.*.textconv` for that:

- `.gitconfig`
  ```
  [diff "ocamlformat"]
    textconv = ocamlformat --name "./$1"
  ```

- `.gitattributes`
  ```
  *.ml diff=ocamlformat
  *.mli diff=ocamlformat
  ```

But that's not working correctly. The `--name "./$1"` part is needed so OCamlformat can find its config file and also decide between parsing `.ml` or `.mli` files, this is concatenating `./` with `/tmp/...`. It seems that Git doesn't give the path in the repository to `textconv` programs, so this won't work with nested `.ocamlformat` files.

PR #1206, opened 2020-01-15
Use `bos` instead of `stdio` for file operations

gpetiot:
More precise handling of errors, maybe could still be simplified with a cleaner propagation of errors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 #### Packaging
 
   + Work with base v0.13.0 (#1163) (Jules Aguillon)
+  + Remove dependency to the stdio package (#1206) (Guillaume Petiot)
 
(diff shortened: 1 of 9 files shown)


jberdine:
Why prefer bos over stdio?

gpetiot:
I think the interface of `bos` is more adequate as it interfaces directly with `Fpath` that we already use, I also like the handling of errors better (most of them are currently ignored in this PR, for now, but we should propagate these errors with the errors we already handle, we need to work more on integrating this part of the code with the result type, but it could be done in a later PR).

emillon:
independently of the actual changes, I think this kind of PR should get discussed a bit in advance.

gpetiot:
Closing this one for now as we are not sure to make this change.

PR #1207, opened 2020-01-15
Line/region indentation feature (emacs integration)

gpetiot:
the workflow is:
- `dune build @install && dune install`
- add `(require 'ocamlformat)` to your .emacs

Disclaimer: so far what it does is applying the indentation of the (virtually) formatted output to the unformatted input, so if you break a line like this you can observe cases like:

```ocaml
let foo =
foooo
|> bar
```

because once formatted it is formatted on a single line (indented to 0)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,10 @@
 
   + Consistent positioning of lambda return type annotations when no-break-infix-before-func and pre/post extensions (#1581, @gpetiot)
 
+#### New features
+
+  + Line/region indentation feature (emacs integration) (#1207, @gpetiot)
+
 ### 0.16.0 (2020-11-16)
 
 #### Removed
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -65,6 +65,18 @@ let run_action action opts =
         | Error e -> Error (fun () -> print_error conf opts ~input_name e)
       in
       Result.combine_errors_unit (List.map inputs ~f)
+  | In_out ({kind= Conf.Kind k; file; name= input_name; conf}, None)
+    when Option.is_some opts.numeric -> (
+      let source = source_from_file file in
+      let range = Option.value_exn opts.numeric in
+      match
+        Translation_unit.indentation k ~input_name ~source ~range conf opts
+      with
+      | Ok indents ->
+          List.iter indents ~f:(fun i ->
+              Stdio.print_endline (Int.to_string i) ) ;
+          Ok ()
+      | Error e -> Error [(fun () -> print_error conf opts ~input_name e)] )
   | In_out ({kind; file; name= input_name; conf}, output_file) -> (
       let source = source_from_file file in
       match format ?output_file ~kind ~input_name ~source conf opts with
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -294,6 +294,94 @@ is nil."
     (delete-file bufferfile)
     (delete-file outputfile)))
 
+(defun ocamlformat-args (name start-line end-line)
+  (let*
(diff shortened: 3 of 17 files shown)


gpetiot:
> But the fact that you have to give it line numbers in the formatted output makes it unusable, I think.
> How hard it is to fix that ?

The only assumption we make to have this to work is that the lines before the one specified in option are already formatted (at least the line numbers are the same), this won't be true in 100% cases but it makes sense in a context where it's used in an editor.
Even if the current line (the one specified in this option) would be split be the formatting:
```ocaml
let x =
          let k = 1 in let e = 2 in let z = 3 in let i = 4 in i + i + z + k + k + e (* we want --indent=2 *)
```
the 2nd line would be the 1st line of the new block of lines corresponding to the 2nd line, that would be indented of 2 characters, and it's correct to apply the same indentation to the 2nd line in the original code:
```ocaml
let x =
  let k = 1 in
  let e = 2 in
  let z = 3 in
  let i = 4 in
  i + i + z + k + k + e

let x =
  let k = 1 in let e = 2 in let z = 3 in let i = 4 in i + i + z + k + k + e
```

emillon:
I agree that we'll need something like this, but it's not super clear to me how much needs to be done in ocamlformat and how much will be done by the editor.

Editor integration will be fairly large amount of work, and I think we need to step back a bit from the code and work out a spec first, and see what kind of features we'll need based on that. Then we'll be able to iterate on both. We don't want to ship a new option that will end up not being useful.

gpetiot:
A bit of update on this branch. I chose to copy the behavior of ocp-indent, that means using a `--numeric` option is combination with the `--lines` options, that will print the indentation of each line from that line range. That means the emacs integration could be copy-pasted from the ocp-indent files since we would have the same interface. So I rebased this PR on #1188 and put up a prototype, here is an example:
```ocaml
let foooooooooooo = let foooooooo = foooooooooooo in foooooooooooo

let foooooooooooo = let foooooooooooooo =
let woooooooooooooooo = koooooooooooooooo in
baaaaaar
in
hooohoooo
```

`$ ocamlformat foo.ml --lines 3-7`

```ocaml
let foooooooooooo = let foooooooo = foooooooooooo in foooooooooooo

let foooooooooooo =
  let foooooooooooooo =
    let woooooooooooooooo = koooooooooooooooo in
    baaaaaar
  in
  hooohoooo
```

`$ ocamlformat foo.ml --lines 3-7 --numeric`

```
0
4
4
0   <-- should be 2, the error is made when trying to find where this `in` comes from in the source
2
```

What it would look like in emacs (the integration is not completed, but the interface would be the same as ocp-indent):
```ocaml
let foooooooooooo = let foooooooo = foooooooooooo in foooooooooooo

let foooooooooooo = let foooooooooooooo =
    let woooooooooooooooo = koooooooooooooooo in
    baaaaaar
in
  hooohoooo
```

gpetiot:
This ambiguity with the index corresponding to the line has been fixed, I added a corresponding cli test. There is still an issue when selecting the 1st line but it is due to #1188 (and already reported there).

gpetiot:
For reference there is a PR on opam-user-setup to add ocamlformat: https://github.com/OCamlPro/opam-user-setup/pull/48 I'm not familiar with the way it works but I'm not sure there is an easy way to switch from ocp-indent to ocamlformat in this tool. I'm still digging but if someone has experience with this feel free to help :)

gpetiot:
This PR is in need of a major rewriting.

gpetiot:
I've reused some code from #1188 to reduce the time overhead of this feature, but there is still too much overhead compared from master.

gpetiot:
Replaced by #1609

PR #1208, opened 2020-01-16
Fix location of comment attached to the underscore of an open record

### LINKED ISSUE #1153: Bug: comment placement on "_" part of record patterns
Hi,

Consider the following piece of code where a comment is relative to the `; _` part of a pattern:

```ocaml
type t = {x: int; y: int}

let to_string {x ; _ (* we should print y *)} =
  string_of_int x
```

The pattern gets formatted as `{x (* we should print y *); _}`.

(Note that this is low priority - I stumbled upon this one recently)

gpetiot:
Fix #1153 
This is a bit convoluted because the location of the underscore has first to be inserted in the Loc_tree so that the associated comment would be attached to this location instead of the location of the previous field.
No regression with test_branch on any profile.

### DIFF
--- a/lib/Source.ml
+++ b/lib/Source.ml
@@ -309,5 +309,4 @@ let loc_of_underscore t flds (ppat_loc : Location.t) =
   let loc_underscore = {ppat_loc with loc_start= end_last_field} in
   let filter = function Parser.UNDERSCORE -> true | _ -> false in
   let tokens = tokens_at t ~filter loc_underscore in
-  let token = List.hd_exn tokens in
-  snd token
+  Option.map (List.hd tokens) ~f:snd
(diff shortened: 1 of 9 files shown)


lib/Source.ml
```diff
@@ -299,3 +299,15 @@ let typed_expression (typ : Parsetree.core_type)
...
+  let end_last_field =
+    match List.last flds with
+    | Some (_, p) -> p.Parsetree.ppat_loc.loc_end
+    | None -> ppat_loc.loc_start
+  in
+  let loc_underscore = {ppat_loc with loc_start= end_last_field} in
+  let filter = function Parser.UNDERSCORE -> true | _ -> false in
+  let tokens = tokens_at t ~filter loc_underscore in
+  let token = List.hd_exn tokens in
+  snd token
```
Julow: Nothing in this function guarantee that `tokens` cannot be empty. I think this function should return an option and the unsafe call moved close to the check on the closed_flag or removed.

gpetiot: I changed the return type.

PR #1209, opened 2020-01-17
Fix grouping of horizontally aligned comments

### LINKED ISSUE #1200: Bug: two comments between function and argument do not stabilize
Hi,

The following piece of code does not stabilize with the default profile on 0.12 and master:

```ocaml
let y = f (* a *) (* b *) x
```

It cycles with:

```ocaml
let y = f (* a *)
          (* b *) x
```

Thanks!

gpetiot:
Fix #1200 
The 2 comments were not put in the same group even though they are on consecutive positions on the same line, and should be attached to the same node. This solution avoids removing the linebreak between adjacent comments, and changing the node which the group is attached to, which would cause more regressions.
No regression with test_branch on any profile.


emillon:
Thanks!

PR #1214, opened 2020-01-21
Fix dropped comments around module pack expressions

Julow:
`ocamlformat --debug` was printing a warning on this code (from https://github.com/ocaml-ppx/ocamlformat/pull/1212):
```ocaml
 let helper ?x:((module X) = (module X_add_one : X_typ)) = X.f
```

I couldn't place a comment that would be dropped, so this is not fixing a bug.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1800,17 +1800,21 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         | `Closing_on_separate_line ->
             (str "(", fits_breaks ")" ~hint:(1000, -2) ")")
       in
+      let has_cmts =
+        Cmts.has_before c.cmts pexp_loc || Cmts.has_after c.cmts pexp_loc
+      in
       hovbox 0
         (compose_module
            (fmt_module_expr c (sub_mod ~ctx me))
            ~f:(fun m ->
              hvbox 2
-               (Cmts.fmt c pexp_loc
-                  ( hovbox 0
-                      ( opn_paren $ str "module " $ m $ fmt "@ : "
-                      $ fmt_longident_loc c id )
-                  $ fmt_package_type c ctx cnstrs
-                  $ fmt_atrs $ cls_paren ))))
+               (wrap_if has_cmts "(" ")"
+                  (Cmts.fmt c pexp_loc
+                     ( hovbox 0
+                         ( opn_paren $ str "module " $ m $ fmt "@ : "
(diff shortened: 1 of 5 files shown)


lib/Fmt_ast.ml
```diff
@@ -1789,7 +1789,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
             ?epi const
         $ fmt_atrs )
   | Pexp_constraint
-      ( {pexp_desc= Pexp_pack me; pexp_attributes= []; pexp_loc= _; _}
+      ( {pexp_desc= Pexp_pack me; pexp_attributes= []; pexp_loc; _}
       , { ptyp_desc= Ptyp_package (id, cnstrs)
         ; ptyp_attributes= []
         ; ptyp_loc= _
```
Julow: I didn't find a case where this location is to be considered.

gpetiot: I added some test cases.

lib/Fmt_ast.ml
```diff
@@ -1800,16 +1800,21 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
         (compose_module
            (fmt_module_expr c (sub_mod ~ctx me))
            ~f:(fun m ->
              hvbox 2
-               ( hovbox 0
-                   ( opn_paren $ str "module " $ m $ fmt "@ : "
-                   $ fmt_longident_loc c id )
-               $ fmt_package_type c ctx cnstrs
-               $ fmt_atrs $ cls_paren )))
+               (wrap_if has_cmts "(" ")"
```
Julow: Should we really add these parentheses ? They are only needed to force a test to reproduce a bug that doesn't exist without them !

gpetiot: The bug was there if you have `((* foo *) (module A : T))` as input, after your fix it is printed as `(* foo *) (module A : T)` but the comment is not attached to the same AST node (it may not be problematic now, but it could be in the future)

Julow: The attachment point of the comment is an internal problem. As an user I would understand if a comment move but I would be puzzled if new parentheses are added, as there is already a lot of them in first-class module syntaxes.

gpetiot: It makes sense, alright I reverted the parenthesizing and moved the tests in the `comments.ml` file so we can have a .ref file.
It looks good to me, feel free to merge if it is ready.

Julow:
Thanks for finding the tests, they will be useful to test comment attachment.

PR #1216, opened 2020-01-23
Improve test_branch script

Julow:
This PR improves a lot the `tools/test_branch.sh` script to make it
easier to modify and more robust.

Changes:
- Projects are cloned to `/tmp/ocamlformat_test_branch_cache` instead of
  `test-extra/code` and won't be pulled automatically
- Projects to test are easier to describe inside the script
- Diffs are written into a `test_branch_<project>_<rev_a>_<rev_b>.diff`
  file for each project
- Options `-a <rev_a>` and `-b <rev_b>` on the CLI
  Avoid using the same rev twice by accident and allow to compare with
  something older than master. Not passing any will compare HEAD and its
  merge-base with master.


emillon:
Before going into details I have some questions:
- there's a bunch of other targets in the Makefile. Were they used?
- I think we're past the point where it makes sense to use something other than bash to do this kind of test. sure, it's better than before, but maintaining shell scripts longer than 50 lines is super hard.
- usually, hardcoded paths in /tmp are not a great idea. but we need it to persist so a normal tempfile isn't great either. I think the solution of an in-tree ignored cache was OK in that regard.
- there seems to be some antipatterns in the script like parsing the output of ls. shellcheck should point you to the relevant lines (of course if you decide to use a different language, this concern goes away).

emillon:
(in particular, the fetch deps, cache them, run commands on them and summarize results seems like it would be a nice fit for ocurrent)

Julow:
> there's a bunch of other targets in the Makefile. Were they used?

The only other target was `test_margins`. It's a bit harder but still possible to do now. The script needs to be modified to catch errors the same way.

> I think we're past the point where it makes sense to use something other than bash

I agree. Using ocurrent would make sense, especially with github integration. I'll think about that.

> hardcoded paths in /tmp are not a great idea

Good point. I should revert to caching in-tree.


There is an other thing that this script remove is the ability to keep the last output and debug OCamlformat over it. I'll revert a bit and keep only the useful things.

Julow:
> there seems to be some antipatterns in the script like parsing the output of ls

That's on purpose here.

emillon:
What do you mean? What's the benefit in doing

    for p in `ls "$src_dir"`; do

rather than

    for p in $src_dir/*; do

Julow:
The purpose was to get the name of the directories, to use as strings. Not paths.

Julow:
I reverted most of the changes and kept only the good parts:

- Easier to add projects
- `-a` and `-b` options
- Improved script to build ocamlformat

I kept the Makefile

test-extra/Makefile
```diff
@@ -26,6 +26,7 @@ FIND_ARGS= \
 	-path code/dune/vendor -not -prune -or \
 	-path code/dune/otherlibs -not -prune -or \
 	-path code/dune/example -not -prune -or \
+	-path code/duniverse/duniverse -not -prune -or \
```
gpetiot: not used yet

Julow: I kept it here so uncommenting the projects below works. I changed so this is easier to declare there.

Issue #1226, opened 2020-02-07
Bug: ocamlformat can't process combination of function application, local module opening and attribute


roddyyaga:
ocamlformat doesn't work on the following code (reduced example):
```ocaml
module M = struct
  let f x = x
end

let g=M.f ((let open M in f) [@ocaml.warning "-33"])
```
The error message is `BUG: ast changed` (although I got `BUG: generating invalid ocaml syntax` for the file where this originally occurred).

This is for ocamlformat 0.12 on OCaml 4.07.0 with the default configuration (the same thing also happened on my custom configuration).

Julow:
Thanks for the report. We fixed the issue.

PR #1229, opened 2020-02-10
Fix missing parentheses around let open

Julow:
Fixes https://github.com/ocaml-ppx/ocamlformat/issues/1226

Parentheses around  were removed, changing the AST in some cases.

This adds a regression, it is consistent with other expressions and
should be discussed separately:
```diff
-  [%foo M.(()) [@foo]] ;
+  [%foo (M.(()) [@foo])] ;
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+### (unreleased)
+
+#### Bug fixes
+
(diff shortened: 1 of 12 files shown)


Julow:
This PR is merged into https://github.com/ocaml-ppx/ocamlformat/pull/1230

PR #1235, opened 2020-02-17
CmtSet: Remove the emap part

emillon:
CmtSet.t holds data about a set of comments, accessible either by start
or by end position.

When profiling `ocamlformat`, I noticed that we are spending a lot of
time in the `Map.partitioni_tf` part (10% common, 25% in `js_source.ml`,
can be higher in specially crafted files like `js_source.ml`
concatenated 10 times).

But the `emap` component is never used. It can be safely removed,
yielding a nice performance boost.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -128,49 +128,27 @@ end = struct
     end)
   end
 
-  module Order_by_end = struct
-    type t = Location.t
+  type t = Cmt.t list Map.M(Order_by_start).t
 
-    include Comparator.Make (struct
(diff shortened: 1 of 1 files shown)


emillon:
I have prepared some improvements in that area of the code but I'm less confident about the changes so I'll do that in a separate PR.

emillon:
Allright, my reasoning was that performance was user visible but that's a fine line.

gpetiot:
You can keep it if it makes sense to you, I don't have a strong opinion about this.

PR #1236, opened 2020-02-17
Fix formatting of empty sig payload

emillon:
In `[@a:]`, the payload is an empty signature.

It should be formatted as such; the payload in `[@a]` is an empty structure, which is different.


Julow:
You should add an entry in CHANGES, though.

PR #1230, opened 2020-02-10
Remove unecessary parentheses in extensions

Julow:
As mentioned in #1229

Remove parentheses in:
```diff
 ;;
-(() [@attr])
+() [@attr]

 ;;
-[%ext (() [@attr])]
+[%ext () [@attr]]
```

Rebased on #1229, the bug it fixes would be exacerbated.

PR #1238, opened 2020-02-17
Handle attributes on object overrides ({< >})

emillon:
These were silently dropped.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2399,12 +2399,15 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
             $ fmt_expression c (sub_exp ~ctx f)
       in
       match l with
-      | [] -> wrap "{<" ">}" (Cmts.fmt_within c pexp_loc)
+      | [] ->
+          wrap_if parens "(" ")"
+            (wrap "{<" ">}" (Cmts.fmt_within c pexp_loc) $ fmt_atrs)
       | _ ->
           hvbox 0
             (wrap_if parens "(" ")"
-               (wrap_fits_breaks ~space:false c.conf "{<" ">}"
-                  (list l "@;<0 1>; " fmt_field))) )
+               ( wrap_fits_breaks ~space:false c.conf "{<" ">}"
+                   (list l "@;<0 1>; " fmt_field)
+               $ fmt_atrs )) )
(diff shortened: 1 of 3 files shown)


emillon:
I'm not too sure how to nest the box, the attributes, and the wrap_if.

`Pexp_object`  has the attributes inside the wrap_if, but Pexp_letmodule has it outside, so please let me know what's the best way.

Thanks!

lib/Fmt_ast.ml
```diff
@@ -2390,12 +2390,13 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
       match l with
-      | [] -> wrap "{<" ">}" (Cmts.fmt_within c pexp_loc)
+      | [] -> wrap "{<" ">}" (Cmts.fmt_within c pexp_loc) $ fmt_atrs
       | _ ->
           hvbox 0
             (wrap_if parens "(" ")"
                (wrap_fits_breaks ~space:false c.conf "{<" ">}"
-                  (list l "@;<0 1>; " fmt_field))) )
+                  (list l "@;<0 1>; " fmt_field)))
+          $ fmt_atrs )
```
Julow: > Pexp_object has the attributes inside the wrap_if, but Pexp_letmodule has it outside, so please let me know what's the best way.

This `wrap_if` is used to print parentheses, depending on the context. In this case, the attribute must be inside and `parens` will be `true` if there is attributes.

In this example, the AST changes:
```ocaml
f ({< >} [@a])
```

emillon: OK, I added some test cases.
`parens` does the trick in some cases, but I don't understand why it's true after `let _ =`. This means that we get 

    let _ = ({<>} [@a])

but:

    ;;
    {<>} [@a]

I thought that maybe `let _ = {<>} [@a]` would attach the attribute to the let binding instead of the expression, but that's not the case.

Could it be a bug in the way `parens` is computed?

Julow: Here https://github.com/ocaml-ppx/ocamlformat/blob/master/lib/Ast.ml#L2272
Only in `Pstr_eval` the parentheses are removed when there is an attribute. This is since https://github.com/ocaml-ppx/ocamlformat/pull/1230, to avoid diffs after a bug fix.
It could be nice to lift this a bit, indeed.

emillon: OK, now all should be fixed!

PR #1239, opened 2020-02-17
Fix attributes on coercion

emillon:
It would previously format `(x:>t)[@a]` as `(x:>t[@a])`.


lib/Fmt_ast.ml
```diff
@@ -2328,12 +2328,12 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
-           $ opt t1 (fmt "@ : " >$ (sub_typ ~ctx >> fmt_core_type c))
-           $ fmt "@ :> "
-           $ fmt_core_type c (sub_typ ~ctx t2)
-           $ fmt_atrs ))
+        ( wrap_fits_breaks ~space:false c.conf "(" ")"
+            ( fmt_expression c (sub_exp ~ctx e1)
+            $ opt t1 (fmt "@ : " >$ (sub_typ ~ctx >> fmt_core_type c))
+            $ fmt "@ :> "
+            $ fmt_core_type c (sub_typ ~ctx t2) )
+        $ fmt_atrs )
```
Julow: The whole expression must be wrapped in parentheses if there are attributes (with `wrap_if parens`).

Example:
```ocaml
f ((x :> t) [@a])
```

test/passing/source.ml.ref
```diff
@@ -9223,3 +9223,6 @@ type t = |
 M.(Some x) [@foo]
 
 [@@@foo:]
+
+;;
+(x :> t) [@a]
```
samoht: why is this passing? the source and source.ref are differents!

emillon: when there is a ref file, it checks that formatting of x.ml is equal to x.ml.ref.

emillon: I moved this to attributes.ml which as no ref, seems clearer.

lib/Fmt_ast.ml
emillon: Ah, I see. I added a failing test for this. I don't see how to fix this, though. `wrap_if parens` adds parens in other cases, and using `wrap_fits_break_if` does not seem to do the trick.

Julow: What we need to do in this case is wrap again in parentheses. The `wrap_fits_breaks` stays here, we need an other wrap around `wrap_fits_breaks .. $ fmt_atrs`.

Is it adding parentheses when not needed ? In that case, we should use `wrap_if (parens && has_attr)`. In the cases parentheses are required and there is no attribute (`parens = true`, `has_attr = false`), the inner parentheses are enough.

emillon: `parens && has_attr` did the trick, thanks!

CHANGES.md
```diff
@@ -25,6 +25,9 @@
 
   + Fix parenthesizing when accessing field of construct application (#1247) (Guillaume Petiot)
 
+  + Fix attributes on coercion (#1239) (Etienne Millon)
+    It would previously format `(x:>t)[@a]` as `(x:>t[@a])`.
```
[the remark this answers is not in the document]
emillon: OK, I removed that bit.

PR #1241, opened 2020-02-18
Make test_branch.sh usable on the same branch with different options

gpetiot:
To make it easier to test the future changes of options' default values.

### DIFF
--- a/tools/test_branch.sh
+++ b/tools/test_branch.sh
@@ -43,7 +43,7 @@ opts_b=${2-$opts_a}
 rev_b=$(git rev-parse "${opt_b:-HEAD}")
 rev_a=$(git rev-parse "${opt_a:-$(git merge-base master "$rev_b")}")
 
-if [[ "$rev_a" = "$rev_b" ]]; then
+if [[ "$rev_a" = "$rev_b" && "$opts_a" = "$opts_b" ]]; then
   echo "The base branch is the same as the branch to test ($rev_a)"
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


tools/test_branch.sh
```diff
@@ -43,7 +43,7 @@ opts_b=${2-$opts_a}
 rev_b=$(git rev-parse "${opt_b:-HEAD}")
 rev_a=$(git rev-parse "${opt_a:-$(git merge-base master "$rev_b")}")
 
-if [[ "$rev_a" = "$rev_b" ]]; then
+if [[ "$rev_a" = "$rev_b" && "$opts_a" = "$opts_b" ]]; then
```
Julow: What do you think of a safer check: check that the user typed the same literal rev twice explicitly ?
Eg. `$opt_a = $opt_b` (without "s")

Btw, the naming `opt_a`/`opts_a` that means different things is not nice. Could you change `opt_` to something like `arg_` ?

gpetiot: I agree about renamig the variables, but I don't understand what you are tying to prevent with this check. My idea would be to allow checks with either different revisions or different configurations. But if both revisions are the same and both configs are the same, we don't want to run the checks.

Julow: I was thinking forcing the user to write both revision explicitly (eg. `-a HEAD -b HEAD`). I'm afraid that comparing options will make the check less useful.
Though, this check is mainly useful when running `test_branch` on `master` without any argument, so this new check isn't breaking that.
What do you think ?

gpetiot: As long as it allows me to run `test_branch -a X -b X 'optsA' 'optsB'` I'm fine with it, I don't have a strong opinion against it, I'm letting you rewrite the check and you can merge when it's ready.

Julow: I think your check makes sense. What about moving it to around the `exit 1` line ? So the message is still printed

gpetiot: done

PR #1243, opened 2020-02-18
Fix attribute formatting on Pexp_pack

emillon:
(It was formatted as `(module struct end [@a])`)

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2191,8 +2191,10 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                    | `Closing_on_separate_line -> "@;<1000 -2>)" ) )) )
   | Pexp_pack me ->
       let fmt_mod m =
-        Params.wrap_exp c.conf c.source ~parens:true ~loc:pexp_loc
-          (str "module " $ m $ fmt_atrs)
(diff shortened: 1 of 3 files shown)


lib/Fmt_ast.ml
```diff
@@ -2192,7 +2192,8 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
   | Pexp_pack me ->
       let fmt_mod m =
         Params.wrap_exp c.conf c.source ~parens:true ~loc:pexp_loc
-          (str "module " $ m $ fmt_atrs)
+          (str "module " $ m)
+        $ fmt_atrs
```
Julow: We need an other layer of parentheses for this case:
```
let _ = f ((module struct end) [@a])
```

If `wrap_if parens` adds too much parens, `wrap_if (parens && has_attr)` should be enough because the parens will only guard the attribute.

Julow:
There is an other instance of this bug: 
```ocaml
let _ = ((module M : S) [@a])
```

This is parsed as `Pexp_constraint (Pexp_pack, Ptyp_package)` (the constraint part is in the expression language) and this is handled in an other case, above.

emillon:
Good catch! I fixed this one (also adding parens if necessary).

emillon:
Thanks!

PR #1244, opened 2020-02-19
Remove parens after "let _"

emillon:
This is an extension of #1230 for `let _ = ...` structure items.

As discussed in https://github.com/ocaml-ppx/ocamlformat/pull/1238#discussion_r380832586.

Let me know what you think. Should this be extended to more let bindings?

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -2269,6 +2269,13 @@ end = struct
     ||
     match (ctx, exp) with
     | Str {pstr_desc= Pstr_eval _; _}, _ -> false
+    | ( Str
+          { pstr_desc=
+              Pstr_value
+                (Nonrecursive, [{pvb_pat= {ppat_desc= Ppat_any; _}; _}])
+          ; _ }
+      , _ ) ->
+        false
     | _, exp when has_trailing_attributes_exp exp -> true
     | ( Exp {pexp_desc= Pexp_construct ({txt= Lident id; _}, _); _}
       , {pexp_attributes= _ :: _; _} )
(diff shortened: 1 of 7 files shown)


lib/Ast.ml
```diff
@@ -2269,6 +2269,13 @@ end = struct
     ||
     match (ctx, exp) with
     | Str {pstr_desc= Pstr_eval _; _}, _ -> false
+    | ( Str
+          { pstr_desc=
+              Pstr_value
+                (Nonrecursive, [{pvb_pat= {ppat_desc= Ppat_any; _}; _}])
+          ; _ }
+      , _ ) ->
+        false
```
Julow: I wonder if we can remove parentheses in all `Str _, _` cases ? `test_branch` does not catch any bug.

emillon: Shouldn't it find cases where there were extra parens?

emillon: In any case, I don't have enough experience to determine if it's risky or not, but if it is we can first extend to let _, and all lets.

gpetiot: I don't have any risky case in mind either, but I agree we can start with the current state of the PR and then extend later.

Julow: > Shouldn't it find cases where there were extra parens?

This is to work around the next pattern (`... when has_trailing_attributes_exp ... -> true`). So it only removes parentheses in cases like `let a = (some expr) [@a]`, which are rare.

emillon:
OK! Let's extend that later if required. Thanks for the explanations!

PR #1251, opened 2020-02-26
Internal: extract the comment formatting function as Cmt.fmt

gpetiot:
This is a minor cleaning of Cmts.ml


gpetiot:
I think it's more consistent to handle the formatting of a `Cmt.t` in Cmt.ml, and Cmts.ml really need to be cleaned up.
The `Source` dependency will be added with #1193 anyway, the `Conf` dependency can be removed if we pass a `wrap` parameter instead (unlikely we add more comments-related options in the future)

PR #1252, opened 2020-02-26
Fix parens around binop operations with attributes

### LINKED ISSUE #1250: Bug: attributes attached to certain kinds of binops
Hi,

`ocamlformat` crashes on the following input:

    let _ = ([] @ []) [@a]

Interestingly, this only happens with some kinds of binary operators. For example `("" ^ "") [@a]` crashes as well, but `(0 + 0) [@a]` works. Part of the reason might be that `0 + 0 [@a]` is parsed with the attribute on the apply, not on the second operand.

gpetiot:
Partial fix of #1250 
No regression with test_branch for the preset profiles.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,8 @@
 
   + Fix formatting of attributes on packed modules (#1243) (Etienne Millon)
 
+  + Fix parens around binop operations with attributes (#1252) (Guillaume Petiot)
+
 ### 0.13.0 (2020-01-28)
 
 #### New features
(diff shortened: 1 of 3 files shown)


gpetiot:
There is an issue with `let _ = f (([] @ []) [@a])`

emillon:
FTR, in my tests I've seen issues with the following binops:

- `(x@y)[@a]`
- `(x<y)[@a]`
- `(x&y)[@a]`
- `(x^y)[@a]`
- `(x||y)[@a]`
- `(x=y)[@a]`
- `(x$y)[@a]`
- `(x>y)[@a]`
- `(x|*y)[@a]`

It's probably not worth adding a test case for each one if the handling is uniform. For some reason, it does not happen with `x+y` since parsing is different in this case (without parens, it applies the attribute to pexp_apply).

PR #1276, opened 2020-03-09
Handle OCaml 4.10.0 AST

### LINKED ISSUE #1218: ocaml 4.10 support
Hi,

Ocaml 4.10 [is going to be released soon](https://discuss.ocaml.org/t/ocaml-4-10-0-second-beta/5083). In order to support it, there are some things to wait for like having OMP and base working on 4.10, but also [some new language features to format](https://github.com/ocaml/ocaml/blob/b237282e97095f8ac9dfd8c18a03412406bde6e9/Changes#L202-L220). In particular:
- multi-indices for extended indexing operators
- `module _ = E`

We should make sure these are properly supported.

gpetiot:
Fix #1218
No regression found with test_branch


test/passing/functor.ml
```diff
@@ -55,7 +55,7 @@ module M =
 module type Module_type_fail = sig
   include S
 
-  module F : functor (_ : T) -> sig end
+  module F (_ : T) : sig end
```
Julow: This is a regression. This syntax is new from `4.08`, we should preserve the long form.

gpetiot: done

lib/Sugar.ml
```diff
@@ -291,49 +291,74 @@ let sequence (conf : Conf.t) cmts xexp =
...
+      in
+      (({txt= Some ""; loc= Location.none}, None) :: xargs, xbody)
   | _ -> ([], xmty)
 
 (* The sugar is different when used with the [functor] keyword. The syntax
    M(A : A)(B : B) cannot handle [_] as module name. *)
 let rec functor_ cmts ~for_functor_kw ~source_is_long ({ast= me; _} as xme) =
-  let valid_sugared_name arg =
-    not (String.equal arg.txt "_" && source_is_long me)
+  let is_valid = function
```
Julow: They are all valid. I think the previous name (`valid_sugared_name`) was better.

lib/Sugar.ml
```diff
@@ -291,49 +291,74 @@ let sequence (conf : Conf.t) cmts xexp =
...
+         || (List.is_empty pmod_attributes && not (source_is_long me)) ->
       let ctx = Mod body in
       let body = sub_mod ~ctx body in
       let xargs, xbody_me =
         match pmod_attributes with
         | [] -> functor_ cmts ~for_functor_kw ~source_is_long body
         | _ -> ([], body)
       in
-      ((arg, xarg_mt) :: xargs, xbody_me)
+      (({txt= Some ""; loc= Location.none}, None) :: xargs, xbody_me)
```
Julow: So `None` is for `(_)` and `Some ""` is `()` ? What about having a dedicated type for this ? (instead of `label option loc`, which have too many meanings)

Julow: We shouldn't be afraid to change the whole function and use the types that make it safe instead of the types that make the code shorter.

lib/Sugar.ml
gpetiot: done

lib/Sugar.ml
gpetiot: I packed the 2 values in the same option type, it looks better

Julow: That's a nice improvement !

That wasn't my suggestion but now the type is very similar to the corresponding part of the parsetree, which is enough.

gpetiot:
I fixed this error, it slightly improved the comment placement in a few cases (the cmt relocate is not necessary anymore)

PR #1277, opened 2020-03-09
Support multi-indices extended indexing operators

### LINKED ISSUE #1273: Bug: ocamlformat fails with Bigarray extended indexing in OCaml 4.10.0
ocamlformat fails to format code that uses latest Bigarray extended indexing feature in OCaml 4.10.0

Here is a simple script that breaks when I try and run ocamlformat on it

```ocaml
open Bigarray
let ( .%{ ;.. } ) = Genarray.get
let ( .%{ ;.. } <- ) = Genarray.set
let () = 
   let x = Genarray.create Float64 c_layout [|3;4;5|] in
   x.%{0;0;0} <- 3.;
   Printf.printf "%f\n" x.%{0;0;0} 
```

gpetiot:
Fix #1273


CHANGES.md
```diff
@@ -4,6 +4,8 @@
 
   + Add an option `--format-invalid-files` to print unparsable parts of the input as verbatim text. This feature is still experimental. (#1026) (Guillaume Petiot)
 
+  + Support multi-indices extended indexing operators (#1277) (Guillaume Petiot)
```
Julow: Could you say that it's a new feature from 4.10 ?

Julow:
I talked too fast. This example is a valid program:

```ocaml
let ( .%{} ) () r = Array.length r

;;
let a = () in
a.%{[| 1; 2 |]}
```

The last line formats to
```ocaml
a.%{1; 2}
```

This is a parsing error before 4.10 and an AST change in 4.10.0.

PR #1279, opened 2020-03-10
Support multi indices extended indexing operators

Julow:
There was some bugs in https://github.com/ocaml-ppx/ocamlformat/pull/1277 this PR fixes them.

The module `Ast.Indexing_op` is added and replaces the previous code working with indexing operators.
I add unit tests for `Ast.is_symbol_id`, which indirectly test parsing of indexing operators.


Julow:
> Personally I think it would make sense to make this Index_op module into a new file if it can be isolated from the rest of Ast.ml

I wanted to wait to have a plan before doing this. I agree we should start now !

What do you think of moving this module to `Sugar` ?
`Ast.Indexing_op` would only expose `parse` and the type `raw` (renamed). `Indexing_op` is used later in `Ast`, this is nice for correctness but this is not necessary, we could use unsafe functions, as everything else in this module.

gpetiot:
It would make sense to move it to `Sugar` but `Sugar` already depends on `Ast`, that's why I suggested having a third file (I don't know what kind of unsafe functions you are referring to but unsafe rarely sounds good)

Julow:
> I don't know what kind of unsafe functions you are referring to but unsafe rarely sounds good

Unsafe meant using `Indexing_op.parse` (still better than previous API) instead of `get_sugar`, which does more checks. Those checks and types are not necessary for `Ast` and can be moved to `Sugar`. This makes `Ast` more consistent and defines better how it works. (if there is a "get_sugar" function in `Ast`, that means an abstraction is missing, here it's `Indexing_op.parse`)

Julow:
Unsafe also meant that we'll have to do operations like "get the last argument" that are less sound than what `get_sugar` does. But again, that's how `Ast` works.

gpetiot:
Alright then this split can wait until we cleanup and improve the way `Ast` works. We can merge if it's ready.

PR #1284, opened 2020-03-17
Retain attribute on pexp_setfield

CraigFe:
This attempts to resolve https://github.com/ocaml-ppx/ocamlformat/issues/1259.

PR #1286, opened 2020-03-17
Fix CI script checking changelog

Julow:
The script checking that the changelog is updated in every PRs seems broken.

(work in progress)


Julow:
The script works fine. The problem happened on https://github.com/ocaml-ppx/ocamlformat/pull/1284, restarting the job may have solved it.

Sorry for the noise.

PR #1287, opened 2020-03-17
Remove option doc-comments

Julow:
The first commit implement removed options.
When used, an useful error will be generated (instead of "unknown option") including since which version and a custom message.

When such option is used in a `.ocamlformat` file, the error will be shown everytime OCamlformat is invoked (Dune calls it separately on each file). I opened an issue on Dune to discuss that: https://github.com/ocaml/dune/issues/3275

The second part of this PR is the removal of `doc-comments`.
There is some diffs because OCamlformat was an user of this option.
This part can be reverted or discussed separately.

### DIFF
--- a/lib/Config_option.ml
+++ b/lib/Config_option.ml
@@ -78,6 +78,7 @@ module Make (C : CONFIG) = struct
   let section_name = function
     | `Formatting -> Cmdliner.Manpage.s_options ^ " (CODE FORMATTING STYLE)"
     | `Operational -> Cmdliner.Manpage.s_options
+    | `Removed -> Cmdliner.Manpage.s_options ^ " (REMOVED OPTIONS)"
 
   let from = `Default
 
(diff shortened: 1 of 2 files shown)


gpetiot:
I think it would be better to split this PR in 2:
- improving the deprecating mechanism (and having 2 categories, `deprecated` and `removed`, so an option can be deprecated in 0.14, then removed in 0.15).
- deprecating the `doc-comments` option

emillon:
I agree with @gpetiot, let's deprecate (a plain deprecation warning willdo for now) in 0.14.0 and remove it in 0.15.0.

Julow:
> improving the deprecating mechanism (and having 2 categories, deprecated and removed, so an option can be deprecated in 0.14, then removed in 0.15).

We already have deprecated options. We can improve it with a custom message.

> deprecating the doc-comments option

That part was just demonstration.

Julow:
I removed the demonstration commit for now. Let's not merge this until the release.

gpetiot:
Replaced by #1469 (can't push to this PR)

Issue #1258, opened 2020-03-02
Bug: crash when an attribute is on pexp_setinstvar


emillon:
Hi,

The following program crashes ocamlformat:

    let _ = (a <- b)[@a]

(this is not valid ocaml but this expression is valid in a method)

PR #1288, opened 2020-03-18
Retain attribute on pexp_setinstvar

CraigFe:
This attempts to resolve https://github.com/ocaml-ppx/ocamlformat/issues/1258.

It's unecessary to wrap `(a <- b) [@a]` in parentheses if the expression is the only one in a module definition.

I discovered and tested a few unnecessary parentheses in the process:

- `method a = (1 [@attr])`
- `method a = (0 [@attr]) ; 1`
- `let _ = (0 [@attr]) ; 1` (edit: now tested by https://github.com/ocaml-ppx/ocamlformat/pull/1291 instead)

We may want to keep these for disambiguation purposes, but it seems better to have them tested regardless.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -2279,6 +2279,7 @@ end = struct
           ; _ }
       , _ ) ->
         false
+    (* Object fields do not require parens, even with trailing attributes *)
     | Exp {pexp_desc= Pexp_object _; _}, _ -> false
     | _, exp when has_trailing_attributes_exp exp -> true
     | ( Exp {pexp_desc= Pexp_construct ({txt= Lident id; _}, _); _}
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -2279,6 +2279,7 @@ end = struct
           ; _ }
       , _ ) ->
         false
+    | Exp {pexp_desc= Pexp_object _; _}, _ -> false
     | _, exp when has_trailing_attributes_exp exp -> true
     | ( Exp {pexp_desc= Pexp_construct ({txt= Lident id; _}, _); _}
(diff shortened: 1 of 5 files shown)


CraigFe:
The unnecessary parentheses mentioned here are removed as a side-effect of https://github.com/ocaml-ppx/ocamlformat/pull/1291.

lib/Ast.ml
```diff
@@ -2279,6 +2279,7 @@ end = struct
           ; _ }
       , _ ) ->
         false
+    | Exp {pexp_desc= Pexp_object _; _}, _ -> false
```
Julow: Could you add a small comment explaining why we have this pattern ?

test/passing/object.ml
```diff
@@ -6,6 +6,8 @@ let _ =
     (* some comment *)
     method! x = 2 [@@attr]
 
+    method x = (1 [@attr])
```
Julow: These parentheses are not necessary.

Julow: Is this fixed by https://github.com/ocaml-ppx/ocamlformat/pull/1291 ?

CraigFe: It does not, only those on the LHS of sequences will be, i.e.

```ocaml
method a = (0 [@attr]) ; 1

let _ = (0 [@attr]) ; 1
```

lib/Ast.ml
CraigFe: Done in https://github.com/ocaml-ppx/ocamlformat/pull/1288/commits/ed75834457eb13065794a4806fff5074e36218a9.

Julow:
Let's merge !

PR #1289, opened 2020-03-18
Handle lazy patterns as arguments to class

### LINKED ISSUE #1261: Bug: crash when using a pattern as an argument to class expression
Hi,

    class t (lazy _) = object end

(The output is missing parens around `lazy _`)

emillon:
Fixes #1261

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -1981,6 +1981,7 @@ end = struct
      |Cl {pcl_desc= Pcl_fun _; _}, Ppat_tuple _
      |Cl {pcl_desc= Pcl_fun _; _}, Ppat_construct _
      |Cl {pcl_desc= Pcl_fun _; _}, Ppat_alias _
+     |Cl {pcl_desc= Pcl_fun _; _}, Ppat_lazy _
      |( Exp
           { pexp_desc=
               Pexp_fun _ | Pexp_function _ | Pexp_match _ | Pexp_try _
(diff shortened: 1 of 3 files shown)


lib/Ast.ml
```diff
@@ -1985,6 +1985,7 @@ end = struct
      |Cl {pcl_desc= Pcl_fun _; _}, Ppat_tuple _
      |Cl {pcl_desc= Pcl_fun _; _}, Ppat_construct _
      |Cl {pcl_desc= Pcl_fun _; _}, Ppat_alias _
+     |Cl {pcl_desc= Pcl_fun _; _}, Ppat_lazy _
```
Julow: Some more cases are missing:
- `Ppat_exception`
- `Ppat_variant` with arguments
- `Ppat_or`

I wonder if a refactor of this function is needed. What do you thing of classifying the context argument separately (eg. to `` Patern of pattern | Function | Let | Parameter ``), so we have saner rules that can't forget any cases ?

emillon: I don't think that Ppat_exception can be found here, but yes this looks more complicated than it should. To be fair, I'm not sure I understand why `parenze_x` functions are structured this way, compared to assigning precedence levels and comparing them.

Julow: I think these functions are this way because we did let that happen :( Now it's time to improve them !
Computing precedence would be great but a lot more work.

emillon: OK I'll have a look at this.

emillon: it's likely that we can improve, and @CraigFe has started in this direction, but this is not strictly related to this PR

PR #1290, opened 2020-03-18
Fix formatting of comment with just a newline

### LINKED ISSUE #1268: Bug: index out of bounds when a comment contains just a newline
Hi,

The following program crashes ocamlformat with an index out of bounds exception:

    (*
    *)

(bytes: 0x28 0x2a 0x0a 0x2a 0x29 0x0a)

emillon:
This reworks a bit the Cmts / `fill_text` interface to fix #1268. In particular the existing empty string branch seems to be dead since `(* *)` produces a space character and `(**)` is parsed as a doc comment.

### DIFF
--- a/lib/Cmt.ml
+++ b/lib/Cmt.ml
@@ -134,7 +134,8 @@ let fmt cmt src ~wrap:wrap_comments ~fmt_code =
           fmt_asterisk_prefixed_lines asterisk_prefixed_lines
     else
       match split_asterisk_prefixed cmt with
-      | [""] -> str "(* *)"
+      | [] -> assert false
+      | [""] -> assert false
       | [""; ""] -> str "(* *)"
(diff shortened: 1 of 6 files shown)


lib/Cmt.ml
```diff
@@ -134,7 +134,7 @@ let fmt cmt src ~wrap:wrap_comments ~fmt_code =
           fmt_asterisk_prefixed_lines asterisk_prefixed_lines
     else
       match split_asterisk_prefixed cmt with
-      | [""] -> str "(* *)"
```
Julow: Should we keep an assert here ?
I don't like that but the `asterisk_prefixed_lines` can't handle `[""]`, it would format to `(**)`, which would parse as a documentation comment.

Julow: The `[]` cannot work too.

Issue #1254, opened 2020-03-02
Bug: attribute on sequence causes crash


emillon:
Hi,

The following program causes ocamlformat to crash:

    let _ = (a;b) [@a]

CraigFe:
Currently working on this...

PR #1291, opened 2020-03-18
Retain attribute on pexp_sequence

CraigFe:
This attempts to resolve https://github.com/ocaml-ppx/ocamlformat/issues/1254.

Note that this changes the parens behaviour in two existing cases. I'd appreciate your thoughts on what the true behaviour should be.


emillon:
no sarcasm please :)

CraigFe:
This one is more subtle than the others.

In doing this, I removed some unnecessary parentheses (https://github.com/ocaml-ppx/ocamlformat/pull/1291/commits/b4bed9aca154aee3c88d9c9eb5503b142dbe7b70) in a way that is breaking (see the change in `vendor/`). If we want to keep this, we should probably add a `CHANGES` entry for it.

This PR also made me realise that the parentheses that we're adding due to attributes are not like the ones added by `Params.wrap_exp`. Consider the following:
```ocaml
let _ =
  (aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ;
   bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb) [@a]

let _ =
  f
    ( aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ;
      bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb )
```

The second parentheses are added by `Params.wrap_exp`, and so are spaced according to the `--indicate-multiline-delimiters` option via `Fmt.wrap_fits_breaks_if`, whereas the first pair are not.

I chose not to open this can of worms here, since it's an existing inconsistency.

vendor/parse-wyc/menhir-recover/emitter.ml
```diff
@@ -137,9 +137,9 @@ end = struct
   open Format
 
   let emit_default_value ppf =
-    (fprintf ppf "open %s\n\n"
-       (String.capitalize (Filename.basename Grammar.basename))
-     [@ocaml.warning "-3"]);
+    fprintf ppf "open %s\n\n"
+      (String.capitalize (Filename.basename Grammar.basename))
+    [@ocaml.warning "-3"];
```
Julow: This change is fine !

Julow: `test_branch` found 2 occurrences of this on Dune. Still looks like a fix.

Julow:
No regressions with `test_branch` that are not fixes. I'll merge !

PR #1293, opened 2020-03-18
Deprecate some options before 0.14

Julow:
This PR deprecate the options `doc-comments`, `escape-chars`, `escape-strings` and `extension-sugar`.

The first commit improves the deprecation mechanism by passing a custom message and a version string. This is in the same PR because we don't have deprecated options currently to show diffs.


lib/Conf.ml
```diff
@@ -494,7 +494,12 @@ module Formatting = struct
            change their meaning and on structures, signatures and objects \
            for readability." ) ]
     in
-    C.choice ~names ~all ~doc ~section
+    let deprecated =
+      C.deprecated ~since_version:"0.14"
```
gpetiot: Should we have `0.14.0`instead?

Julow: Indeed ! Fixed.

Julow:
`exp-grouping` is not working correctly, indeed.
What do you think @emillon ?

emillon:
It's a bit less conservative to include this one since it's not the default value in the profiles we ship. What's not working correctly with this option?

Julow:
The problem of `exp-grouping` is that it is applied only to some places and that `begin .. end` can be removed if they are unecessary.
This option is not a lost cause, we can describe every cases where it's useful and remove cases where it's not (currently there is no general rule and no doc). I expect a lot of maintenance and this will remove most `begin .. end` anyway.
An option that always put `begin .. end` when OCamlformat wants would be more useful, I think, to avoid users wondering about inconsistency of a "preserve but only sometimes" option.
Also, it's not used a lot: https://gist.github.com/Julow/110dc94308d6078225e0665e3eccd433#file-output-L137

lib/Conf.ml
```diff
@@ -563,7 +568,10 @@ module Formatting = struct
         , `Hexadecimal
         , "$(b,hexadecimal) mode escapes every character." ) ]
     in
-    C.choice ~names ~all ~doc ~section
+    let deprecated =
+      C.deprecated ~since_version:"0.14.0" "There is no equivalent."
```
gpetiot: Can we have a more precise behavior where the warning (warning only, not the --help manual) depends on the selected value? Like:
- `preserve` : this option is not required anymore, character escaping is now preserved by default, 
- others : this feature has been removed, character escaping is now preserved by default.

Julow: I'm not sure if this is necessary.
The option was already not required when set to the default value, is this something we should warn regardless if the option is deprecated ?
The default didn't change because of the deprecation and we can't predict if it will change after we removed the option.

gpetiot: It make sense. Alright let's merge if it's ready :)

PR #1294, opened 2020-03-18
Remove globals from Ast

emillon:
`Ast.init` would take conf as argument and set 3 globals based on that. This PR replaces them by parameters. This adds a lot of boilerplate but I think this reflect the coupling that exists because of these options.

I think it'd be better if these extra parameters go away (it feels wrong to say that the precedence depends on configuration), but it's hard to say that this is a net improvement. What do you think

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -14,13 +14,10 @@
 open Migrate_ast
 open Parsetree
 
-let extension_sugar = ref `Preserve
-
 let init, register_reset =
   let l = ref [] in
   let register f = l := f :: !l in
-  let init (conf : Conf.t) =
-    extension_sugar := conf.extension_sugar ;
+  let init () =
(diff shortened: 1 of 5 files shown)


jberdine:
(I seem to recall that there used to be some comments in the documentation that these configs could not be set in attributes.)

Overall I'm not sure it is a net gain to remove these refs. Those features are implemented in a way that ends up using config values where they really shouldn't be used. For instance, the `leading-nested-match-parens` and `parens-ite` options were initially implemented by changing the `parenze` functions. But those functions are intended to indicate if there is ambiguity, and discretionary parens ought to be added at a higher point in the code. I'm not as familiar with `extension-sugar`, but hope that it could also be implemented at a higher level instead of munging the precedences.

emillon:
Yeah, it's not clear it's the good way to fix this. Closing this, thanks!

Issue #1295, opened 2020-03-18
Feature request: avoid extra space after certain object types with attributes


CraigFe:
```ocaml
type t = {foo: unit [@a]}

type t = {foo: < .. > [@a] }
```

The extra space after `[@a]` in the second case is added to avoid outputting `>}` which is a keyword. However, it's not necessary to add the protecting space if the object type has an annotation, as is the case here.

The same is also true of:
- inline record types in variants
- polymorphic variants (the token `>]` has the same problem)
- class parameter lists (weird, since parentheses are added in this case)

```ocaml
type t = A of {foo: unit [@a]}

type t = A of {foo: < .. > [@a] }

type t = [`Foo of unit]

type t = [`Foo of < .. >[@a] ]

let _ =
  object
    inherit [b, unit] foo

    inherit [b, (< f: unit >[@a]) ] foo
  end
```

emillon:
Maybe the labels are not precise enough, but I wouldn't call it a bug when we manage to output something equivalent (unlike the related #1267). That's probably worth fixing, but it's more of a style issue.

CraigFe:
Yep, it's a good point. I've renamed the issue accordingly.

I think I just called it a bug because I'd already looked at the code by that point and decided it was unintentional behaviour :slightly_smiling_face:

PR #1296, opened 2020-03-18
Remove trailing spaces after object types inside records and variants

CraigFe:
This fixes https://github.com/ocaml-ppx/ocamlformat/issues/1295.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,7 +47,7 @@
 
   + Retain attributes on field set expressions, e.g. `(a.x <- b) [@a]` (#1284) (Craig Ferguson)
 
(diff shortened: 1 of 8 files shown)


CraigFe:
Rebased and added a changelog entry.

CraigFe:
> Do you think it could be easily fixed in this PR?

As far as implementation goes, the two bugs are distinct, but it's not difficult to extend this PR to cover both cases. I've now done that (and changed the PR title accordingly).

Julow:
Thanks !

> As far as implementation goes, the two bugs are distinct, but it's not difficult to extend this PR to cover both cases. I've now done that (and changed the PR title accordingly).

Thanks for that too :)

Issue #1257, opened 2020-03-02
Bug: attribute on get operation causes crash


emillon:
Hi,

The following program crashes ocamlformat:

    let _ = (a.(b))[@a]

(Might be related to #1250.)

PR #1300, opened 2020-03-19
Retain attribute on indexing operators

CraigFe:
This resolves https://github.com/ocaml-ppx/ocamlformat/issues/1257.

The same bug applied to extended infix operators, which makes sense since the `Array.get` sugar is handled in the same place.

Both are now fixed:

```ocaml
let _ = (a.(b)) [@a]

let _ = (a.*?!@{b} <- c) [@a]
```


test/passing/attributes.ml
```diff
@@ -216,3 +216,11 @@ let _ = (0 + 0) [@a]
 let _ = (a.x <- 1) [@a]
 
 let _ = f ((a.x <- 1) [@a])
+
+let _ = (a.(b)) [@a]
```
Julow: Parentheses are not necessary here. This can be fixed by modifying `Ast.parenze_exp` and calling `Indexing_op.parse`. (this is not ideal because we call parse again and again but improving this would be a lot of work)

CraigFe: Implemented an alternative solution in https://github.com/ocaml-ppx/ocamlformat/pull/1300/commits/2e7280aa1726b0a7b7451de1f02412098f559f46.

Julow: That works !

Issue #1256, opened 2020-03-02
Bug: attributes on tuples can cause ocamlformat to crash


emillon:
Hi,

The following programs cause ocamlformat to crash (due to dropped parens):

    ;;(0, 0)[@a]

and:

    let _ = f ((0, 0)[@a])

CraigFe:
The first case is probematic even without an attribute:
```ocaml
;; (0, 0)
```
→
```ocaml
;;
0, 0
```

CraigFe:
Scratch that, `;; 0, 0` is actually valid. News to me :slightly_smiling_face:

PR #1301, opened 2020-03-19
Fix missing parentheses around tuples with attributes

CraigFe:
This fixes https://github.com/ocaml-ppx/ocamlformat/issues/1256.


CraigFe:
Nice catch on the left-alignment bug. Should be fixed by https://github.com/ocaml-ppx/ocamlformat/pull/1301/commits/c3f82a195bd8cfd4bb76a3d367333f31e350c166.

CraigFe:
> The fix to the alignment could also be a fix for the `indicate-multiline-delimiter` bug in #1291

True. I wanted to save that sort of fix for a single pass to remove non-config-dependent wrapping, but it seems unavoidable in this case due to the weird tuple-parens behaviour.

CraigFe:
This can probably be considered for merge.

Issue #1269, opened 2020-03-03
Bug: comment changed with newline in cinaps-like comment


emillon:
Hi,

The following program crashes ocamlformat because the comments changed:

    (*$(**)"
    "*)

hexdump:

    00000000: 282a 2428 2a2a 2922 0a22 2a29 0a         (*$(**)"."*).

Interestingly, the problem goes away if any of the dollar sign, nested comment, or quotes are removed.

PR #1303, opened 2020-03-19
Preserve cinaps comments containing code

Julow:
Inserting a syntax error in a cinaps comment would make OCamlformat fallback to wrapping them, resulting in a mess:

```diff
-(*$
-  print_newline () ;
-  <SYNTAX ERROR>
-  List.iter
-    (fun s -> Printf.printf "let ( %s ) = Pervasives.( %s )\n" s s)
-    ["+"; "-"; "*"; "/"]
-*)
+(*$ print_newline () ; <SYNTAX ERROR> List.iter (fun s -> Printf.printf "let
+  ( %s ) = Pervasives.( %s )\n" s s) ["+"; "-"; "*"; "/"] *)
```

This patch fixes that by keeping the comment as it is.

This comment changes meaning when wrapped because OCaml treat quotes in comments specially: (fixes https://github.com/ocaml-ppx/ocamlformat/issues/1269)

```ocaml
(*$(**)"
"*)
```

### DIFF
--- a/test/passing/cinaps.ml
+++ b/test/passing/cinaps.ml
@@ -42,3 +42,18 @@ let foo = foo
 (*$QR foo Q.small_int (fun i-> foo i (+) [1;2;3] = List.fold_left (+) i
   [1;2;3] ) *)
 let foo = foo
+
+(* Cinaps comment should not wrap if they don't parse. The first one would
+   crash and the second become a mess *)
+
+(*$(**)"
+"*)
+
+(*$
+  print_newline () ;
+  <SYNTAX ERROR>
+  List.iter
+    (fun s -> Printf.printf "let ( %s ) = Pervasives.( %s )\n" s s)
+    ["+"; "-"; "*"; "/"]
+*)
+(*$*)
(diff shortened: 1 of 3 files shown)


test/passing/cinaps.ml
```diff
@@ -1,3 +1,7 @@
+File "cinaps.ml", line 1, characters 0-4:
+Warning 50: unattached documentation comment (ignored)
+File "cinaps.ml", line 1, characters 0-4:
+Warning 50: unattached documentation comment (ignored)
```
CraigFe: What creates these warnings?

emillon: They are generated by the ocaml parser itself. It is responsible for turning doc-comments into attributes and attaching them to the right place.

Julow: I overlooked them but they make the test fail obviously !
This can be fixed for this test by adding a `.ref` file.

The locations are not right.

Julow: It seems they weren't needed, I just removed them. I have no idea what happened.

Issue #1255, opened 2020-03-02
Bug: syntax with dot collapses when argument is an integer


emillon:
Hi,

This kind of input causes a crash in ocamlformat:

    let _ = (0).*(0)

The reason is that it tries to format as something that starts with `0.`, which is lexed as a float.

The "canonical" example is `(0).(0)` which cannot happen in a well-typed way, but other similar ones might happen when redefining `( .*() )` or `( .*()<- )` and similar operators.

PR #1304, opened 2020-03-19
Surround indexing operators lhs when it's an integer

Julow:
Fixes https://github.com/ocaml-ppx/ocamlformat/issues/1255

This is fixed by adding yet an other case to the `parenze_ast` function.

Fun fact: Intergers with a suffix or floats don't need parentheses.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -2311,7 +2311,8 @@ end = struct
       , {pexp_desc= Pexp_construct _ | Pexp_variant _; _} )
       when e == exp ->
         true
-    (* Integers without suffixes must be parenthesised on the lhs of an indexing operator *)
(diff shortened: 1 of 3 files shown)


lib/Ast.ml
```diff
@@ -2311,6 +2311,11 @@ end = struct
       , {pexp_desc= Pexp_construct _ | Pexp_variant _; _} )
       when e == exp ->
         true
+    (* Integers must be parenthesed on the lhs of an indexing operator *)
```
CraigFe:
```suggestion
    (* Integers without suffixes must be parenthesised on the lhs of an indexing operator *)
```

Julow: Thanks :)

Julow:
Let's merge this.

Issue #1260, opened 2020-03-02
Bug: crash on optional argument rebound to non-variable


emillon:
Hi,

The following program causes ocamlformat to crash:

    let f ?a:(A) = ()

(it tries to format it as `let ?a:A = ()`)

I had to check [the exact parsing rules](https://github.com/ocaml/ocaml/blob/1fd6dc077d63fc8e0b36e2a684dae55a0f3383d4/parsing/parser.mly#L2088-L2091). What they say is that only ppat_var and ppat_any are recognized without parens.

I have to admit it's difficult to find a reasonable use case for this combination of features, but that's valid. A similar situation hapens with `?a:(())` and `?a:(0)`.

CraigFe:
I'm on it.

PR #1305, opened 2020-03-20
Ensure optional argument rebindings use sufficiently many parens

CraigFe:
This resolves https://github.com/ocaml-ppx/ocamlformat/issues/1260.

I chose to do this by adding the logic in `Fmt_ast.fmt_fun_args`. It's probably possible to do this by instead adding special cases to `Ast.parenze_pat`. Unfortunately `fmt_pattern` (via which `parenze_pat` is called) doesn't respect its `parens` argument in many cases, making that route difficult. I removed an existing special case for records in `Ast.parenze_pat` which is superceded by the new logic.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -100,6 +100,9 @@
   + Fix invalid (unparsable) docstrings (#1315) (Guillaume Petiot)
     When parsing a comment raises an error in odoc, it is printed as-is.
 
+  + Fix parenthesizing of optional arguments rebound to non-variables, e.g. `let
+    f ?a:(A) = ()` rather than the unparsable `let f ?a:A = ()` (#1305) (Craig Ferguson)
+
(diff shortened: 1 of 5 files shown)


CraigFe:
Nice catch on the attributes interaction. 

I've added a fix for that case now, although I think it's an unpleasant solutionit's a bit of a hack.

gpetiot:
The test related to #1310 is not fixed on master despite merging #1317, I feel like we need more precise contexts to decide when parentheses can be added/skipped.

Julow:
The bug I mentioned above is introduced by this PR. We should fix it before merging.

I agree, we need to improve the handling of parentheses (not in this PR), it's getting harder to maintain.

gpetiot:
I fixed the last bugs, it looks good to me now.

PR #1306, opened 2020-03-20
Fix parentheses around infix operators with attributes

CraigFe:
This resolves https://github.com/ocaml-ppx/ocamlformat/issues/1250.

As a side-effect, there are now unnecessary parentheses around infix application chains with attributes attached, e.g.:

```diff
-;; a + b + c [@attr]
+;; (a + b + c) [@attr]
```

This is a breaking change. I chose to propose the change anyway, since I find the precedence binding of attributes to infix application confusing. For instance:

```ocaml
(* Here the attribute binds to [(a::b)] *)
let _ = (a :: b [@attr])

(* Here the attribute binds to [b] only *)
let _ = (a @ b [@attr])
```

If we decide to keep only the parentheses necessary, I think it's necessary to add attributes to the infix precedence logic. In that case, we might want to consider what the interaction with the `infix-precedence` option should be.

Let me know what you think.


gpetiot:
I removed the unnecessary parentheses, no regression with test_branch. I added an ugly test of the context that should be merged with the parenze functions in Ast but it's such a mess right now that I didn't manage to do it.

test/passing/attributes.ml
```diff
@@ -209,30 +209,29 @@ let _ = (module M : S) [@a]
...
 let _ = f ((module M : S) [@a])
 
-let _ = (([] @ []) [@a])
+let _ = ([] @ []) [@a]
 
 (* Infix operator should left-align with the inner parens *)
 let _ =
-  (( a_____________________________________
-   @ b_____________________________________ ) [@a])
+  ( a_____________________________________
```
CraigFe: This test now must be changed to remain useful.

CraigFe: Done.

Julow:
Thanks for the work !

PR #1311, opened 2020-03-23
Preserve empty doc-comments syntax

### LINKED ISSUE #1282: Feature request: consistent handling of the empty doc-comment

The OCaml manual [states](https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#ss:ocamldoc-syntax) that _the_ empty doc-comment is `(**)`. The OCaml lexer [claims](https://github.com/ocaml/ocaml/blob/trunk/parsing/lexer.mll#L484) only that `(**)` is _a_ empty doc-comment. OCamlformat formats it as `(***)`, which I was surprised to learn has the same AST.

I suggest that the two be unified, either by adding mention of `(***)` to the OCaml manual or by switching OCamlformat's chosen representation to `(**)`. I'm happy to make the PR to either `ocaml/ocaml` or OCamlformat depending on which solution we choose.

**Additional context**

- OCaml syntax highlighting in Tuareg mode (emacs), highlights `(**)` as a doc-comment and `(***)` as a regular comment (see https://github.com/ocaml/tuareg/blob/master/tuareg.el#L705).

- The OCaml compiler itself contains occurrences of both `(**)` and `(***)`.

gpetiot:
Fix #1282 
This fix is ugly but it looks like the parser is at fault, the empty comment is not considered a docstring so we have to check what the source looked like when formatting comments.

### DIFF
--- a/lib/Cmt.ml
+++ b/lib/Cmt.ml
@@ -150,7 +150,17 @@ let fmt cmt src ~wrap:wrap_comments ~fmt_code =
         hvbox 2 (wrap "(*$" cls (fmt "@;" $ formatted $ fmt "@;<1 -2>"))
     | Error () -> fmt_non_code cmt
   in
-  match cmt.txt with
-  | "" | "$" -> fmt_non_code cmt
-  | str when Char.equal str.[0] '$' -> fmt_code cmt
-  | _ -> fmt_non_code cmt
+  let src_prefix =
+    let pos_cnum = cmt.loc.loc_start.pos_cnum + 4 in
+    let loc_end = {cmt.loc.loc_start with pos_cnum} in
+    Source.string_at src cmt.loc.loc_start loc_end
+  in
+  match src_prefix with
+  (* "(**)" is not parsed as a docstring but as a regular comment containing
+     '*' and would be rewritten as "(***)" *)
+  | "(**)" -> str "(**)"
+  | _ -> (
+    match cmt.txt with
+    | "" | "$" -> fmt_non_code cmt
+    | str when Char.equal str.[0] '$' -> fmt_code cmt
+    | _ -> fmt_non_code cmt )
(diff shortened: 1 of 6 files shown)


lib/Cmt.ml
```diff
@@ -150,7 +150,17 @@ let fmt cmt src ~wrap:wrap_comments ~fmt_code =
...
     | Error () -> fmt_non_code cmt
   in
-  match cmt.txt with
-  | "" | "$" -> fmt_non_code cmt
-  | str when Char.equal str.[0] '$' -> fmt_code cmt
-  | _ -> fmt_non_code cmt
+  let src_prefix =
+    let pos_cnum = cmt.loc.loc_start.pos_cnum + 4 in
+    let loc_end = {cmt.loc.loc_start with pos_cnum} in
+    Source.string_at src cmt.loc.loc_start loc_end
```
Julow: I don't like creating `Lexing.position` objects. Looking at the implementation of `string_at`, this is not necessary at all.

What do you think of adding an a `Source.string_at' : t -> int -> len:int -> string` ?
`Source.string_at` would be a thin wrapper around it and we won't need to build complicated objects.

lib/Cmt.ml
```diff
@@ -150,7 +150,17 @@ let fmt cmt src ~wrap:wrap_comments ~fmt_code =
...
+  in
+  match src_prefix with
+  (* "(**)" is not parsed as a docstring but as a regular comment containing
+     '*' and would be rewritten as "(***)" *)
+  | "(**)" -> str "(**)"
```
Julow: The computation and matching of `src_prefix` should be done in the match below in a `| "*" -> ` case. Making it contained for readability and avoiding to compute it when it's not needed.

lib/Cmt.ml
gpetiot: good idea, done.

lib/Cmt.ml
gpetiot: done

lib/Cmt.ml
Julow: `Source.sub` is a nice name :)

PR #1313, opened 2020-03-24
Fix travis script checking the changelog

Julow:
It isn't working since a few weeks. This PR is to debug it.

### DIFF
--- a/tools/travis-ci.sh
+++ b/tools/travis-ci.sh
@@ -53,8 +53,10 @@ CheckChangesModified () {
         if HasNoChangelogNeededLabel ; then
             echo skipped
         else
-            git diff --exit-code "$TRAVIS_COMMIT_RANGE" -- CHANGES.md \
-                > /dev/null && exit 1 || echo pass
+            if git diff --exit-code "$TRAVIS_COMMIT_RANGE" -- CHANGES.md > /dev/null
+            then echo "The changelog is not uptodate"; exit 1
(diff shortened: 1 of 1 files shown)


gpetiot:
I feel like the time debugging this script over and over again is not worth it, now we all have the habit to update the CHANGES.md file for each PR, maybe it's time to remove this check?

Julow:
I debugged and found two problems:
- `$TRAVIS_COMMIT` don't point to a valid commit sometimes, fixed.
- Github rate limiting prevent checking for the label, this was silently skipping the check. Now it's doing the changelog check anyway.

I agree the time to debug this script is not well spent.
I don't agree with your second point, I'm not used to it and will probably fail to update the changelog if we remove it. We should keep it.

emillon:
Yeah, we had the same kind of problem on dune.

Julow:
To summarize, with this PR the changelog check will be performed if we hit rate limit (eg. this PR) instead of silently not checking.
I think this is fine because most PRs requires changes and I prefer to explicitly ignore a red CI sometimes than not trusting it.

Julow:
Here is what would happen:

| | Before | This PR |
|---|---|---|
| no label, no changelog | fails only sometimes | failure |
| no label, changelog entry | success | success |
| label, _ | success | fails sometimes |

gpetiot:
Alright then you can merge if it's ready.

PR #1315, opened 2020-03-24
Fix invalid (unparsable) docstrings

### LINKED ISSUE #1265: Bug: index out of bounds with unterminated odoc comment
Hi,

    (**{v*)

There might be a bug in odoc, but ocamlformat can also catch the error.

gpetiot:
Fix #1265


test/passing/invalid_docstring.ml.ref
```diff
@@ -0,0 +1,3 @@
+Warning: Invalid documentation comment:
+comment could not be parsed
```
emillon: I think that because it's being fixed in odoc, this error message will depend on the version of odoc that's used. If that's the case it's better to just swallow the error.

gpetiot: Do you suggest we return the comment as-is and treat it as a success?

emillon: That's what I was suggesting, yes. We already do this when cinaps comment don't parse, for example.

But thinking a bit more about this, it's better to be consistent with the normal odoc warning case.

However, this means that the test will be flaky. Testing at the unit level instead would work, but I don't think we have a way to inject a doc comment parser at the moment. So I'm fine with removing the test as well.

gpetiot: done

test/passing/invalid_docstring.ml
```diff
@@ -0,0 +1 @@
+(**{v*)
```
emillon: Is this going to display a warning if a fixed version of odoc is used?

gpetiot: I forgot your comment about removing the test, I'll do that

lib/Fmt_ast.ml
```diff
@@ -369,15 +369,17 @@ let fmt_direction_flag = function
...
   let location = loc.Location.loc_start in
   let location =
     { location with
       pos_cnum= location.pos_cnum + 3 (* Length of comment opening *) }
   in
-  let parsed = Odoc_parser.parse_comment_raw ~location ~text in
-  match parsed with
-  | {value; warnings= []} -> Ok value
+  match Odoc_parser.parse_comment_raw ~location ~text with
+  | exception _ -> Ok Verbatim
```
Julow: I think we should return an `Error` here.
What's wrong with printing "Warning: Invalid documentation comment" ?

The error message will change depending on the current Odoc version but we already expect that, each versions of Odoc could change error messages and error conditions.

An unstable test is an issue but we don't need to make OCamlformat's stderr output always stable (it's not).

gpetiot: I tend to agree with you, @emillon what do you think?

emillon: Yes, that seems a good point to me as well. To clarify, I didn't mean to imply that stderr should be stable, just that it can't be tested in a stable way.

PR #1316, opened 2020-03-24
Fix spacing inside parens for symbols when the spacing was handled by the englobing exp

### LINKED ISSUE #1249: Bug: (*) is formatted without spaces in some contexts
Hi,

In most cases `( * )` is formatted with spaces (that's necessary because otherwise it clashes with comment syntax).

But here's a case where it's not:

    let op = if b then ( * ) else ( + ) in ()

It gets formatted as `if true then (*) else (+)`, which crashes ocamlformat.

This seems to be related to the `~parens:false` param.

This also happens in the following contrived example (which might indicate a larger problem):

    assert ( * )

### LINKED ISSUE #1307: Bug: crash with attribute on multiplication operator
Hi,

The following program causes ocamlformat to crash:

    ;;( * )[@a]

This might related to #1249.

gpetiot:
Fix #1249
Fix #1307 
This fix is a bit too ad-hoc to my liking but it's the only solution I've found so far. Since an expression only composed of a symbol identifier is very simple I think it's okay to override the parenthesizing mechanism of if-then-else and assert in this case.


test/passing/symbol.ml
```diff
@@ -0,0 +1,9 @@
+;;
+let op = if b then ( * ) else ( + ) in
+()
+
+;;
+assert ( * )
+
+;;
+(( * ) [@a])
```
Julow: These parentheses are not necessary.
This could be fixed by checking `List.is_empty exp.pexp_attributes` in `Ast.is_displaced_infix_op`. Testing is needed to make sure this is sound.

gpetiot: It improves this case indeed, thanks.

gpetiot:
> ```
> assert ( ( * ) [@a] );
> assert (abc [@a])
> ```

I didn't manage to fix this inconsistency. I think there is a lot lot lot more work to clean the parenthesizing mechanism until we get something that is maintainable Should we merge this PR as-is or postpone it until we clean the paren mechanism?

PR #1317, opened 2020-03-25
Add missing parentheses around OR-patterns with attributes

### LINKED ISSUE #1310: Bug: missing parentheses on `Ppat_or` with attributes
The following program crashes OCamlformat:

```ocaml
let (Foo ((A | B) [@attr])) = ()
```

OCamlformat tries to format the pattern as "`Foo (A | B)[@attr]`", which causes the attribute to change locations in the AST.

gpetiot:
Fix #1310

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -872,11 +872,19 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
         @@ Cmts.fmt c ~pro:(break 1 0) loc (fmt_opt pro $ k)
   | _ -> fun k -> Cmts.fmt c ppat_loc (fmt_opt pro $ k) )
   @@ ( if List.is_empty ppat_attributes then Fn.id
-     else
-       let maybe_wrap =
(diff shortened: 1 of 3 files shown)


lib/Fmt_ast.ml
```diff
@@ -872,11 +872,19 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
...
+           match ctx0 with
+           | Pat {ppat_desc= Ppat_construct _; _}
+            |Pat {ppat_desc= Ppat_variant _; _} ->
+               true
+           | _ -> false )
+         | _ -> true
        in
-       fun k -> maybe_wrap (k $ fmt_attributes c ~key:"@" ppat_attributes) )
+       wrap_if parens_attr "(" ")"
+         (k $ fmt_attributes c ~key:"@" ppat_attributes) )
```
Julow: What do you think of moving the whole expression to its own function ?

gpetiot: done, let me know what you think

Julow: That's nice :)

PR #1319, opened 2020-03-27
Refactor Cmts (followup for #1172)

emillon:
This is a followup for #1172.

While updating #1152 I had a look at the next steps and split a couple functions to make the flow a bit clearer. This makes `fmt_cmts` pure for example, and moves some functions that do not need to use the inner of the module directly in the `Cmts` wrapper in `Fmt_ast`, which simplifies parameter passing.

No change in behaviour is expected.

Let me know what you think.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -31,6 +31,8 @@ module Cmts = struct
 
   let fmt_before c = fmt_before c.cmts c.conf ~fmt_code:c.fmt_code
 
+  let fmt_within c = fmt_within c.cmts c.conf ~fmt_code:c.fmt_code
+
   let fmt_after c = fmt_after c.cmts c.conf ~fmt_code:c.fmt_code
 
   let fmt c ?pro ?epi ?eol ?adj loc =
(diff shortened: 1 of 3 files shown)


lib/Fmt_ast.ml
```diff
@@ -29,15 +29,25 @@ type c =
...
+      (* delay the after comments until the within comments have been removed *)
+      let after = fmt_after c ?pro ?epi loc in
+      let open Fmt in
+      before $ inner $ after
+
+  let fmt_list ?pro ?epi ?eol c locs init =
+    List.fold locs ~init ~f:(fun k loc ->
+        fmt ?pro ?epi ?eol ?adj:None c loc k)
+
+  let fmt_within c = fmt_within c.cmts c.conf ~fmt_code:c.fmt_code
```
gpetiot: I liked having `fmt_within` between `fmt_before` and `fmt_after` (also smaller diff)

emillon: Nice catch, I forgot to reorder them after the refactor was done

PR #1320, opened 2020-03-27
Extract fmt_elements_collection

emillon:
All `elements_collection` are formatted the same way. This extracts a function that implements this behavior.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -76,28 +76,26 @@ let protect =
 let update_config ?quiet c l =
   {c with conf= List.fold ~init:c.conf l ~f:(Conf.update ?quiet)}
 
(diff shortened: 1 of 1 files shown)


lib/Fmt_ast.ml
```diff
@@ -76,28 +76,25 @@ let protect =
 let update_config ?quiet c l =
   {c with conf= List.fold ~init:c.conf l ~f:(Conf.update ?quiet)}
 
+let fmt_elements_collection ?(first_grp = true) ?(last_grp = true)
```
Julow: Shouldn't we call these `first_sep` and `last_sep` ?

emillon: We can, I've been struggling to find a good name for these parameters.

PR #1321, opened 2020-03-27
Add a 'How to locally disable ocamlformat' section in the doc

### LINKED ISSUE #1318: Bug: disabling through attribute doesn't fully work
**Describe the bug**
On the following file:
~~~ocaml
[@@@ocamlformat "disable"]

module M = struct
end
include M
~~~
with empty `.ocamlformat`, ocamlformat inserts an empty line between `end` and `include` and `ocamlformat --check` exits with `1`.

gpetiot:
Fix #1318 
As #1318 can't be fixed programmatically, the right way to disable ocamlformat is now explained in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,23 @@ To quote (and sed) [prettier's page on option philosophy](https://prettier.io/do
 
 ### How to locally disable ocamlformat?
 
-To disable the formatting of a specific expression you must attach an `[@@ocamlformat "option=VAL"]` attribute to this expression in the processed file, such as:
+To disable the formatting of a specific toplevel item you must attach an `[@@ocamlformat "option=VAL"]` attribute to this item in the processed file, such as:
 
 ```ocaml
-let do_not_touch = [
-  x; y; z ] [@@ocamlformat "disable"]
+let do_not_touch
+    (x : t)
+      (y : t)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +79,18 @@ To quote (and sed) [prettier's page on option philosophy](https://prettier.io/do
 >
 > The more options ocamlformat has, the further from the above goal it gets. The debates over styles just turn into debates over which ocamlformat options to use.
 
+### How to locally disable ocamlformat?
+
+To disable the formatting of a specific expression you must attach an `[@@ocamlformat "option=VAL"]` attribute to this expression in the processed file, such as:
```
hhugo: This is technically attaching the attribute to the item. For an expression, it would be `[@ocamlformat ..]`

gpetiot: Oh right, thanks

gpetiot:
> Looks good.
> Should we mention `[@@@ocamlformat "disable"]` too ?
> It has quirks but exists. It can be useful to disable formatting for every items in a `struct` without attaching `[@@ocamlformat]` to each one.

I think in this case it's better to attach `[@@ocamlformat "disable"]` to the struct, otherwise people might still complain that the spacing between items changed inside the struct.

PR #1322, opened 2020-03-30
Fix disabling with attributes on < 4.08

emillon:
With OCaml < 4.08, attributes do not have a location: they only have a name (with a loc) and a payload (with a loc). Starting with >= 4.08, they do, as `attr_loc`.
    
We use OMP, so we never see the former representation. However, OMP does not create a new loc or use `Location.none`, it copies the name's loc into the attribute's loc.
    
This means that we have two situations:
    
- `>= 4.08`: parser provides precise locs:

```
                        [@@@name "payload"]
    attr.attr_name.loc      ^^^^
    attr.attr_loc       ^^^^^^^^^^^^^^^^^^^
```

- `< 4.08`: parser provides loc for attr_name, and OMP copies it:

```
                        [@@@name "payload"]
    attr.attr_name.loc      ^^^^
    attr.attr_loc           ^^^^
```

`Source.extend_loc_to_include_attributes`, used to disable formatting on parts of the AST, tries to determine in which case we are to recover a better loc in the < 4.08 case.

The existing test would check for `Location.none`, which is never the case. So the corresponding test was disabled, but with this fix it works fine.


emillon:
@gpetiot let me know what you think. I tried reenabling other tests on < 4.08 in case this fixes them, but that's not the case. But maybe we can use that technique to fix tests on older versions as well. Not sure it's worth it.

emillon:
TBH I'm not completely sure what `extend_loc_to_include_attributes` is used for. It's not used in all code paths, so it seems OK to only fix attributes on demand; if we try to fix them all it means re-running the lexer after all attributes, which seems costly (I haven't measured).

I wonder if it would work to package the 4.10 parser (with OMP types) and use that instead of the one from compiler-libs + migrating the AST.

gpetiot:
Alright we can look that up in another PR, is this one ready to be merged?

jberdine:
No problem with this PR, but it makes me wonder if we shouldn't just require ocaml >= 4.08.

emillon:
Yes, agree once it's easier to distribute ocamlformat independently of the current switch, >= 4.08 seems good.

PR #1324, opened 2020-03-30
Encapsulate token filtering in Source

emillon:
This removes functions that return tokens from `Source`, so that filtering is done in that module. This allows a small refactoring to simplify the common use case:

Instead of filtering, then matching (pseudo-code)

```ocaml
    tokens_between start end ~f:(function
      | (Char _ | Int _ | String _) -> true
      | Dotdot -> true
      | _ -> false)
    |> function
      | [ ((Char _ | Int _ | String _), loc1); (* note repetition with above *)
          (Dotdot, _) ;
          ((Char _ | Int _ | String _), loc2) ] ->
          (loc1, loc2)
```

We use a `filter_map` to capture both steps:

```ocaml
    tokens_between start end ~f:(function
      | (Char _ | Int _ | String _) -> Some `Constant
      | Dotdot -> Some `Dotdot
      | _ -> None)
    |> function
      | [ (`Constant, loc1); (Dotdot, _) ; (`Constant, loc2) ] ->
          (loc1, loc2)
```

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -915,26 +915,9 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
            ))
   | Ppat_constant const ->
       fmt_constant c ~loc:(Location.smallest ppat_loc ppat_loc_stack) const
-  | Ppat_interval (l, u) -> (
-      (* we need to reconstruct locations for both side of the interval *)
-      let toks =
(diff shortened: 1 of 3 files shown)


gpetiot:
This simplifies the usage in Fmt_ast but the code of Source has become more difficult to read, I feel like having an additional layer of type variants doesn't help, but I don't know how to rewrite this better.

emillon:
If that part is not helpful, let's remove it. How does it look now?

PR #1328, opened 2020-04-02
opam file: Remove weird url field

kit-ty-kate:
I'm not sure what's the goal of having this but it feels very much out of place.
I noticed this in https://github.com/ocaml/opam-repository/pull/16133 because the `url` field is not in its usual place, and i just ignored it in previous releases.

I believe that without this PR it is not possible to pin ocamlformat locally without it being pushed upstream. Also it fetches an unnecessary archive everytime it is updated.


emillon:
Thanks! Yes, indeed it does not seem to be useful.

PR #1329, opened 2020-04-03
Cleaning the Translation_unit API

gpetiot:
Cleaning the Translation_unit API a bit in case we release a public API for ocamlformat
(didn't manage to make it compile with a separate opam file as private/public libraries don't mix well together)

### DIFF
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -24,8 +24,8 @@ let format ?output_file ~kind ~input_name ~source conf opts =
   else
     let f =
       match kind with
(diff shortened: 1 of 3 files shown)


gpetiot:
> `impl` and `intf` are only used as first argument of `parse_and_format`, each once. What do you think of exposing this function partially applied on `impl` and `intf` and completely remove them (and the type `'a t`) from the interface ? eg. Exposing `parse_and_format_impl` and `parse_and_format_intf`.

That's a good idea, done.

emillon:
Thanks for this. I should add that any API deserves to be clean, not only public ones :)

PR #1331, opened 2020-04-10
Revert deprecation of doc-comments

Julow:
This was done too fast, alternatives weren't tested on real projects and users.

This caused some problems, Owl is explicitly using the deprecated option and Irmin won't upgrade. This may also cause huge diffs on Infer.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### unreleased
+
+#### Changes
+
+  + Revert depreciation of the `doc-comments` option (#1331) (Jules Aguillon)
+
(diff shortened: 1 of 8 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+### unreleased
+
+#### Changes
+
+  + Revert depreciation of the `doc-comments` option (#1331) (Jules Aguillon)
```
gpetiot, line 5: I think it should be "deprecation", can you also link #1293 so we can easily jump back and forth? thanks

Julow: :+1:

PR #1332, opened 2020-04-10
Make the formatting API public

gpetiot:
Follow-up on #1329


gpetiot:
As Etienne pointed out exporting a library is not really in our plans right now, closing this one for now.

PR #1335, opened 2020-04-14
Revert doc-comments change to the default profile

Julow:
This surprised users.

We should release 0.14.1 with this patch.

cc @samoht

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 #### Changes
 
   + The default for `doc-comments` is changed to `after` (#1335) (Jules Aguillon)
-    This revert a change introduced in 0.14.0.
+    This reverts a change introduced in 0.14.0 (#1012).
 
   + Revert deprecation of the `doc-comments` option (#1331) (Jules Aguillon)
     This reverts a change introduced in 0.14.0 (#1293).
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 #### Changes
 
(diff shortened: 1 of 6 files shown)


samoht:
Why does it change the formatting of vendored code?

appart that LGTM!

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+### unreleased
+
+#### Changes
+
+  + The default for `doc-comments` is changed to `after` (#1335) (Jules Aguillon)
+    This revert a change introduced in 0.14.0.
```
gpetiot, line 6: Can you mention #1325 here so we can easily jump back and forth between the PRs ? thanks

gpetiot:
> Why does it change the formatting of vendored code?

That's because the vendored code uses the `conventional` profile (as I made it a separate library before), as opposed to the rest of ocamlformat code using the `ocamlformat` profile (that has not been rolledback)

CHANGES.md
Julow: The change is from https://github.com/ocaml-ppx/ocamlformat/pull/1012. The change to the `ocamlformat` profile is not reverted by this PR.

Julow:
We should probably disable formatting of vendored code. This makes using `git subtree` impossible and unecessary diffs.

gpetiot:
> We should probably disable formatting of vendored code. This makes using `git subtree` impossible and unecessary diffs.

I was not aware of that, sure we can disable it.

Issue #1165, opened 2019-11-20
Feature request: interpret normal comments like doc comments (as ocp-indent does)


ceastlund:
**Is your feature request related to a problem? Please describe.**
There is currently a round-trip problem when running code through ocamlformat and then ocp-indent. Consider code like this:
```
type t =
  { field : ty
  (* Here is some verbatim formatted text:
     {v
       starting at column 7
     v}*)
  }
```
If I run it through ocamlformat as of a906b5d993a6f321045546778ca8dfb4e45dfd90, I get the following:
```
type t =
  { field : ty
        (* Here is some verbatim formatted text:
           {v
             starting at column 7
           v}*)
  }
```
The whole comment has been indented uniformly, as ocamlformat puts comments on fields farther to the right than that.
Now if I run that result through ocp-indent, I get this:
```
type t =
  { field : ty
  (* Here is some verbatim formatted text:
     {v
             starting at column 7
           v}*)
  }
```
Now ocp-indent has moved the comment back to the left. However, because it treates `{v ... v}` as a verbatim block, it left it indented further to the right.
If I then save this code, and run it through ocamlformat and ocp-indent again, that comment will move several spaces further to the right, and on and on. The process does not have a fixed point.

For our current workflow at Jane Street of using ocamlformat with ocp-indent, I think we need one of (a) consistent comment handling between ocp-indent and ocamlformat, (b) ocamlformat not to move any whitespace within comments, or (c) ocamlformat to offer editor integration so we can drop ocp-indent. Pretty clearly (a) is preferable to (b), and I expect (c) is a much bigger lift, but any of the above will alleviate the immediate problem for us.

gpetiot:
I had a look at the `js_source.ml` test file postprocessed by `ocp-indent` and didn't find more cases like this. Did you find more that would require a generic fix from us or a specific one for this case would be enough?

ceastlund:
Well, I'd want this to work for other doc-comment forms like `{[` ... `]}`, if that's what you mean.

gpetiot:
I meant if this issue appeared in other ast nodes than in comments attached after a record field.
But I'm not sure adding a special treatment for this when `ocp-indent-compat` is set is a good solution.
On the other hand, this problem would disappear when ocamlformat is able to reformat invalid files and you don't have to rely on your `ocp-indent` post-process step (if you consider the input and output of ocp-indent in this case, the issue lies what ocp-indent does in my opinion).

[...]

PR #1336, opened 2020-04-14
Parse comments as docstrings if they contain odoc constructs

gpetiot:
This was suggested in the last message of #1165 (but it is not relevant to the first part of the issue)

The approach is to check if a comment contains odoc constructs (and not only checks if it can be formatted as a docstring, since regular comments can), if so: format it as a docstring.
The code is a bit ugly as we already have a dependency from Fmt_odoc to Cmt so we can't have one in the other direction (and it's not clear what code should be moved where).

### DIFF
--- a/lib/Cmt.ml
+++ b/lib/Cmt.ml
@@ -93,6 +93,21 @@ let fmt_multiline_cmt ?epi ~opn_pos ~starts_with_sp first_line tl_lines =
   in
   vbox 0 (list_fl unindented fmt_line $ fmt_opt epi)
 
+(* has to be duplicated to avoid dependency cycle: cannot have dependency on
+   Fmt_odoc *)
+let parse_docstring ~loc text =
+  let location = loc.Location.loc_start in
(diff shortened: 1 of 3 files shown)


emillon:
I think we need to discuss a bit where we want to go with plain comments. In general, we can't say much about them since there are so many different ways to use them.

Independently of the correctness of this PR, maybe the right direction is the other one, ie never touch these comments?

Julow:
About the dependencies, `Cmt` shouldn't contain formatting functions and `Fmt_odoc` shouldn't depend on `Cmt`. The fix for the second is simple, `Fmt_odoc.diff` shouldn't be here but in `Translation_unit` instead (it should be factored with `Cmt.diff`, leaving a more meaningful function in `Cmt` too).

gpetiot:
Probably replaced by #1352

gpetiot:
This approach does not fix the issue reported by Carl in #1165 by the way, it does not make it worse either but it would have to be reverted to fix #1165 anyway so I'm closing it.

PR #1337, opened 2020-04-14
Internal: fix compiler warnings

gpetiot:
Got some brand new compiler warnings (and 1 error) after upgrading.

### DIFF
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -36,15 +36,15 @@ let to_output_file output_file data =
 
 let source_from_file = function
   | Conf.Stdin -> In_channel.input_all In_channel.stdin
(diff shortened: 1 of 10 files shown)


gpetiot:
I'm using 4.10.0, but I did a full opam upgrade, I think at least dune (2.5.0) has been upgraded, probably more, I didn't copy the log. But the CI is alright anyway.

Julow:
I don't get any warning on `master` (except one by Dune because of `(using menhir 2.1)`) with OCaml 4.10 and Dune 2.5.0. Which version of OCaml are you using ?

gpetiot:
As Jules found out it was due to a rogue dune-workspace file. Closing this PR then.

PR #1339, opened 2020-04-15
Mention test_branch in HACKING.org

Julow:
This can be useful for external contributors or to have a support for discussion.

There is a workflow I have with this script that I didn't document yet:
It is possible to modify the projects sources and run again `test_branch` but it is necessary to comment that line: `make -C test-extra test_setup test_unstage test_clean test_pull`
This should be made easier, I'll improve that.

What do you think?

### DIFF
--- a/HACKING.org
+++ b/HACKING.org
@@ -35,3 +35,5 @@ Git is used to manage the script's state:
 - Release on Opam-repository
   =dune-release opam pkg=
   =dune-release opam submit=
+
+- Each release should be announced on [[https://discuss.ocaml.org/]]
--- a/HACKING.org
+++ b/HACKING.org
@@ -15,6 +15,8 @@ Git is used to manage the script's state:
 - Review the impact of the release on real source code:
   In =test-extra/Makefile=, uncomment the extended list of projects to test
   and run =tools/test_branch.sh -a "previous_release"=.
+  Diffs should be discussed with the maintainers of these projects to gather
+  feedback before continuing the release.
 
 - In =CHANGES.md=, change =(unreleased)= with the current date
   Commit
--- a/HACKING.org
+++ b/HACKING.org
@@ -1,4 +1,20 @@
+* Testing on other projects
+
+The =tools/test_branch.sh= script builds 2 revisions of OCamlformat and run them
+on the sources of some projects.
+The list of tested project is (and can be edited) in =test-extra/Makefile=.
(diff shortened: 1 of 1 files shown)


gpetiot:
That's a good idea.
Since we are on improving the process, once we internally valid the changes with test_branch, we need external validation from the main users.
What about using release candidates, something along the lines of:
- create a tag vX.Y.Z.rcN
- open a migrating PR on important projects testing the rcN branch, wait for people to validate the changes (not merge), if not, modify and increment the rc number
- wait for all projects to agree on some rcN

Julow:
I think we should test such process before committing it. We usually aims at having as few changes as possible so there is chances that breaking-change validation procedures are always different.

HACKING.org
```diff
@@ -1,5 +1,21 @@
...
+Git is used to manage the script's state:
+- staged changes are the diffs introduced by formatting with the first revision
+- unstaged changes are the diffs introduced by upgrading OCamlformat to the
+  second rev
+
 * Release procedure
 
+- Review the impact of the release on real source code:
+  In =test-extra/Makefile=, uncomment the extended list of projects to test
+  and run =tools/test_branch.sh -a "previous_release"=.
```
samoht, line 17: It's *very* important that this review is made externally visible to the maintainers of these projects and that a discussion is started with them to decide whether the changes are acceptable. 

Please add an item to reflect this in the release procedure.

Julow: I clarified that

samoht:
In the release checklist we also need to announce properly to our users that the release is done, by creating a new discuss thread on https://discuss.ocaml.org/

I think it would also be nice to announce it on other relevant channels too (e.g in slack/discourse).

Julow:
I added that too. I didn't add details as I wasn't sure how to write that. What do you think?

emillon:
RC versions are an improvement to our current workflow, but for some projects I think we can be even more continuous:

- at the beginning of the release cycle, open an empty PR against target project
- every week, refresh the PR branch by rebasing it against the project's master and ocamlformat's master
- when ocamlformat is released, this becomes the upgrade PR.

This can be refined even more, by opening a branch corresponding to a feature branch in ocamlformat.

Julow:
Let's test that workflow first before writing it down because we will have adjustment to make.
I'll merge this PR.

Issue #1338, opened 2020-04-15
Confusing interaction between options doc-comments and doc-comments-val


Julow:
When we added `doc-comments-val` to control placement of comments around `val` items, `doc-comments`'s scope was reduced.
This made sense at the time because we planned to remove `doc-comments` but this have been reverted.

Given the name of the options, users will expect that `doc-comments-val` takes precedence over `doc-comments` but that its default doesn't matter. If `doc-comments-val` is not set, `doc-comments` should be used for `val` items too.

This could also be fixed simply by improving the documentation.

samoht:
Having a priority between these options seems like a good idea. It will help reduce the confusion to our users, especially since these new options suddenly pop-ins the profiles and change their default behavior.

CraigFe:
@samoht: I'm far from the most informed on this issue, but as I had some part in motivating the original change I'll try to give some context. (For the record, I find it difficult to see `doc-comments-val` as useful too :slightly_smiling_face:)

When the issue of consistent placement of doc comments was [first raised](https://github.com/ocaml-ppx/ocamlformat/issues/1010), I was thinking about doc-comments around `module` and `module type` definitions only, since these were previously placed according to a heuristic. It wasn't ideal to always place the comments after (since this is difficult to read for large `struct`s), so it was decided to co-opt the existing `doc-comments` option for this purpose, carrying over this decision to other types of structure item. 

In the [ensuing discussion](https://github.com/ocaml-ppx/ocamlformat/issues/1010), it seems that `doc-comments-val` was proposed as an orthogonal option for "_everything else_": types of structure items where many people will want the comments to go below (because Odoc renders the comments below).

I think this was confusing for two reasons:
1. the names `doc-comments` and `doc-comments-val` do not suggest orthogonality;
1. `doc-comments-val` did not include `type` declarations, for which many people want doc-comments to go below to match the Odoc rendering (I certainly just forgot about them at the time, which was a shame). I think this was the major cause of large undesirable diffs with `0.14.0`.

While changing `doc-comments-val` to be applied in priority order _does_ fix (1), I think the scope of `doc-comment-val` is incorrect: I can't really imagine why someone would want different `val` and `type` placement choices, and I don't think the choice of placement around `struct`s should be affecting `type`s.

Once again, I'm not particularly well-informed on this issue, but I suspect that the 'restricted scope' option shouldn't be applying to `val`s, but `module`s instead.

Julow:
[...]

PR #1340, opened 2020-04-15
Add doc-comments-val=unset

Julow:
When `doc-comments-val` was added, the scope of `doc-comments` was reduced and
didn't overlap with the new option.
This is confusing because the name of the option suggest that `doc-comments-val`
overlap with `doc-comments` but not by default.

This also added unecessary maintainance to users of `doc-comments`, they had to
understand and add a new option to their config.

This new value is set by default. This may impact users who kept 0.14.0
defaults when upgrading to 0.14.1 (eg. setting `doc-comments=before` and
not setting `doc-comments-val`).

Closes https://github.com/ocaml-ppx/ocamlformat/issues/1338

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -33,7 +33,7 @@ type t =
   ; disable: bool
   ; disambiguate_non_breaking_match: bool
   ; doc_comments: [`Before | `After]
-  ; doc_comments_val: [`Before | `After]
+  ; doc_comments_val: [`Before | `After | `Unset]
   ; doc_comments_padding: int
   ; doc_comments_tag_only: [`Fit | `Default]
   ; dock_collection_brackets: bool
@@ -505,7 +505,10 @@ module Formatting = struct
     in
     let names = ["doc-comments-val"] in
     let all =
-      [ ( "after"
+      [ ( "unset"
+        , `Unset
+        , "$(b,unset) lets $(b,doc-comments) set the position" )
+      ; ( "after"
         , `After
         , "$(b,after) puts documentation comments after their corresponding \
            declarations." )
@@ -1660,7 +1663,7 @@ let janestreet_profile =
   ; disable= false
   ; disambiguate_non_breaking_match= false
   ; doc_comments= `Before
-  ; doc_comments_val= `Before
+  ; doc_comments_val= `Unset
   ; doc_comments_padding= 1
   ; doc_comments_tag_only= `Fit
   ; dock_collection_brackets= false
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -507,7 +507,7 @@ module Formatting = struct
     let all =
       [ ( "unset"
         , `Unset
-        , "$(b,unset) lets $(b,doc-comments) set the position" )
+        , "$(b,unset) lets $(b,doc-comments) set the position." )
       ; ( "after"
         , `After
         , "$(b,after) puts documentation comments after their corresponding \
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+### unreleased
+
+#### Changes
+
+  + Set `doc-comments-val` to `unset` by default (#1340) (Jules Aguillon)
+    Placement of documentation comments on `val` and `external` items is now
+    controled by `doc-comments` unless this option is set.
+
 ### 0.14.1 (2020-04-14)
 
 #### Changes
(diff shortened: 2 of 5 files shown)


lib/Conf.ml
```diff
@@ -1660,7 +1663,7 @@ let janestreet_profile =
   ; disable= false
   ; disambiguate_non_breaking_match= false
   ; doc_comments= `Before
-  ; doc_comments_val= `Before
+  ; doc_comments_val= `Unset
```
Julow, line 1666: This is the `janestreet` profile, I don't expect this to cause regressions.

gpetiot:
It looks good to me (didn't made any tests yet, we need to check the diff compared to what we had on 0.13.0 this time).
Not important: how about we make the name of the variant value a bit more explicit: (the more obvious it is, the better)
doc-comments-val=
- doc-comments
- same-as-doc-comments
- inherit-doc-comments

Julow:
I think it is already good enough. Options that overlap that way are intuitive and the name "unset" shows it clearly. A more complicated name would get in the way I think.

gpetiot:
Here is the diff on `dune`, which looks better:

```diff
diff --git a/src/dune/colors.ml b/src/dune/.formatted/colors.ml
index 061f54d2..59714de3 100644
--- a/src/dune/colors.ml
+++ b/src/dune/.formatted/colors.ml
@@ -47,8 +47,8 @@ let setup_err_formatter_colors () =
     List.iter [ err_formatter; Dune_util.Report_error.ppf ] ~f:(fun ppf ->
         let funcs = (pp_get_formatter_tag_functions ppf () [@warning "-3"]) in
         pp_set_mark_tags ppf true;
-        (pp_set_formatter_tag_functions ppf
-           { funcs with
-             mark_close_tag = (fun _ -> Ansi_color.Style.escape_sequence [])
-           ; mark_open_tag
-           } [@warning "-3"]))
+        pp_set_formatter_tag_functions ppf
+          { funcs with
+            mark_close_tag = (fun _ -> Ansi_color.Style.escape_sequence [])
+          ; mark_open_tag
+          } [@warning "-3"])
Done: 4312/4779 (jobs: 4)File "src/dune_lang/t.ml", line 1, characters 0-0:
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git diff --no-index --color=always -u src/dune_lang/t.ml src/dune_lang/.formatted/t.ml)
diff --git a/src/dune_lang/t.ml b/src/dune_lang/.formatted/t.ml
index b5abde53..f19cdcd4 100644
--- a/src/dune_lang/t.ml
+++ b/src/dune_lang/.formatted/t.ml
@@ -80,29 +80,29 @@ module Deprecated = struct
     let tfuncs =
       (Format.pp_get_formatter_tag_functions ppf () [@warning "-3"])
     in
-    (Format.pp_set_formatter_tag_functions ppf
-       { tfuncs with
-         mark_open_tag =
-           (function
-           | "atom" ->
-             state := In_atom :: !state;
-             ""
-           | "makefile-action" ->
-             state := In_makefile_action :: !state;
-             ""
-           | "makefile-stuff" ->
-             state := In_makefile_stuff :: !state;
-             ""
-           | s -> tfuncs.mark_open_tag s)
-       ; mark_close_tag =
-           (function
-           | "atom"
-           | "makefile-action"
-           | "makefile-stuff" ->
-             state := List.tl !state;
-             ""
-           | s -> tfuncs.mark_close_tag s)
-       } [@warning "-3"]);
+    Format.pp_set_formatter_tag_functions ppf
+      { tfuncs with
+        mark_open_tag =
+          (function
+          | "atom" ->
+            state := In_atom :: !state;
+            ""
+          | "makefile-action" ->
+            state := In_makefile_action :: !state;
+            ""
+          | "makefile-stuff" ->
+            state := In_makefile_stuff :: !state;
+            ""
+          | s -> tfuncs.mark_open_tag s)
+      ; mark_close_tag =
+          (function
+          | "atom"
+          | "makefile-action"
+          | "makefile-stuff" ->
+            state := List.tl !state;
+            ""
+          | s -> tfuncs.mark_close_tag s)
+      } [@warning "-3"];
```

lib/Conf.ml
gpetiot: I didn't see regressions with the doc-comments on core_kernel, so it looks safe to me

lib/Conf.ml
```diff
@@ -505,7 +505,10 @@ module Formatting = struct
     in
     let names = ["doc-comments-val"] in
     let all =
-      [ ( "after"
+      [ ( "unset"
+        , `Unset
+        , "$(b,unset) lets $(b,doc-comments) set the position" )
```
gpetiot, line 510: A `.` is missing at the end.

PR #1343, opened 2020-04-18
Fix extra break before function body of fun

jberdine:


jberdine:
This is a partial revert of 35d0c21 from https://github.com/ocaml-ppx/ocamlformat/pull/801 but does not seem like related/intended change.

jberdine:
Any feedback on this one?

jberdine:
Any feedback on this one?

PR #1344, opened 2020-04-18
[RFC] Only consider non-constant constructors for align-constructors-…

jberdine:
…decl

Essentially the idea is to align the `of` so produce e.g.:
```
type t =
  | An  of a
  | The of b
  | Other
```
instead of:
```
type t =
  | An    of a
  | The   of b
  | Other
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 #### Changes
 
+  + In `align-constructors-decl` mode, only consider non-constant constructors (#1344) (Josh Berdine)
+
   + Set `doc-comments-val` to `unset` by default (#1340) (Jules Aguillon)
     Placement of documentation comments on `val` and `external` items is now
     controled by `doc-comments` unless this option is set.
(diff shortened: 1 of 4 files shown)


gpetiot:
Thanks for the patch, although we are considering deprecating then removing the `align-*` options. Is this a style you are using at Facebook and you would need to be able to upgrade to the last ocamlformat version?

jberdine:
I'm not aware of any of the `align-` options being used here, and it's not something blocking updates. I was just going through and testing the impact of all the options, and this jumped out at me as something easily changed that might be interesting to others.

jberdine:
If the `align-*` options are on borrowed time, it makes sense to close this.

jberdine:
Also, regarding the `align-*` options, it might be worth considering two sets of them separately:
1. the "simple" cases where what comes to the left of the alignment point has fixed width, like constructor names; versus
2. the more complex cases where the alignment point comes after something of unknown width, such as a pattern of a match case.

It seems much more likely to find good solutions for the cases in 1 than those in 2.

gpetiot:
That's a good point, what is motivating us to remove of these options is that a change of 1 line in a set of aligned items forces the update of all other aligned items, and that generates excessive diffs.
Another minor point (as it's only implementation) is that we have to iter on every item before we can align them (even if we have less computation to do in your case 1), that makes the code more complicated than it should be, for a benefit that is not clear (see my first point).

PR #1350, opened 2020-04-19
Do not clear the emacs *compilation* buffer on successful reformat

### LINKED ISSUE #1348: Bug: `*compilation*` buffer cleared on success
When ocamlformat is invoked via the emacs integration, the `*compilation*` buffer is changed even if ocamlformat succeeded. It should leave the buffer unchanged if reformatting was accomplished successfully. FWIW, this is a regression wrt 0.12.

jberdine:
fix #1348
reopen #1170

Showing warnings from parsing docstrings would be good, but it isn't
workable to clobber the compilation buffer and force a rebuild on
every reformat.

This reverts commit 73eded97c944da8daf8d96f3ddbc6ba5611297f1.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -150,7 +150,6 @@
   + Add an option `--margin-check` to emit a warning if the formatted output exceeds the margin (#1110) (Guillaume Petiot)
   + Preserve comment indentation when `wrap-comments` is unset (#1138, #1159) (Jules Aguillon)
   + Improve error messages (#1147) (Jules Aguillon)
-  + Display standard output in the emacs plugin even when ocamlformat does not fail (#1189) (Guillaume Petiot)
 
 #### Removed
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -129,7 +133,6 @@
   + Add an option `--margin-check` to emit a warning if the formatted output exceeds the margin (#1110) (Guillaume Petiot)
   + Preserve comment indentation when `wrap-comments` is unset (#1138, #1159) (Jules Aguillon)
   + Improve error messages (#1147) (Jules Aguillon)
-  + Display standard output in the emacs plugin even when ocamlformat does not fail (#1189) (Guillaume Petiot)
```
gpetiot, line 132: I'm not sure we should remove entries from the changelog unless the code has been removed in the same release. Here it can be useful for people running ocamlformat 0.13.0.

jberdine: Oh yes, sorry, that was not intentional, just what happened with `git revert`. I'll fix.

jberdine:
I fixed the mistaken change to the changelog. Should I merge now, or is 0.14.2 intended to have only changes related to docstrings?

gpetiot:
> I fixed the mistaken change to the changelog. Should I merge now, or is 0.14.2 intended to have only changes related to docstrings?

Sorry about the delay on the reviews. My idea was to indeed release a 0.14.2 that would only change the doc-comments option (I still didn't have any feedback from other maintainers from the solution that have been discussed in #1354 ), and then to merge all of your pull requests at the beginning of the 0.15 cycle.

PR #1358, opened 2020-05-04
Merge doc-comments-val with doc-comments, new value before-except-val; after becomes after-when-possible

gpetiot:
See #1354 for the discussion that lead to this option.

Currently doing usual tests on big projects, no diff for:
- mirage
- irmin
- index
- mdx
- dune
- owl
- tezos


lib/Fmt_ast.ml
```diff
@@ -465,11 +465,11 @@ let fmt_docstring_around_item' ?(is_val = false) ?(force_before = false)
...
-            if is_val then
-              match c.conf.doc_comments_val with
-              | (`Before | `After) as conf -> conf
-              | `Unset -> c.conf.doc_comments
-            else c.conf.doc_comments
+            match c.conf.doc_comments with
+            | `Before -> `Before
+            | `Before_except_val when is_val -> `After
+            | `Before_except_val -> `Before
+            | `After_when_possible -> `After
```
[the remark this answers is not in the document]

jberdine:
Nice! So I think that this is good to go, and then there isn't anything else blocking 0.14.2, right?

gpetiot:
Well I would have liked to have the opinion of @emillon @Julow @samoht on the option before emerging it, as we only discussed it the both of us. Just to be sure there is no objection, as the topic was very controversial some weeks ago.

jberdine:
Yes, good point, it would be good to hear from the others.

samoht:
That looks good to me!

What happens to people using `after` in their config file? Will they get a warning that the option has been renamed?

gpetiot:
> What happens to people using `after` in their config file? Will they get a warning that the option has been renamed?

I forgot about the warnings. I need to add warnings indeed (and same for doc-comments-val for the people that got accustomed to it)

gpetiot:
Here are the warnings now displayed:
```
$ ocamlformat lib/Ast.ml --doc-comments-val=unset
ocamlformat: option `--doc-comments-val': value `unset` has been removed in
             version 0.14.2. The same behavior can be achieved by setting
             `doc-comments` only.

$ ocamlformat lib/Ast.ml --doc-comments-val=before
ocamlformat: option `--doc-comments-val': value `before` has been removed in
             version 0.14.2. If you are using `doc-comments-val=before` in
             combination with `doc-comments=before` then only
             `doc-comments=before` is now required to achive the same
             behavior. If you are using `doc-comments-val=before` in
             combination with `doc-comments=after` this behavior is not
             available anymore.

$ ocamlformat lib/Ast.ml --doc-comments-val=after
ocamlformat: option `--doc-comments-val': value `after` has been removed in
             version 0.14.2. If you are using `doc-comments-val=after` in
             combination with `doc-comments=before` please now use
             `doc-comments=before-except-val`. If you are using
             `doc-comments-val=after` in combination with
             `doc-comments=after` then only `doc-comments=after` is now
             required to achieve the same behavior.

$ ocamlformat lib/Ast.ml --doc-comments=after
ocamlformat: option `--doc-comments': value `after` has been removed in
             version 0.14.2. This value has been renamed
             `after-when-possible` to take into account the technical
             limitations of ocamlformat, the behavior is unchanged.
```

lib/Conf.ml
```diff
@@ -514,7 +527,43 @@ module Formatting = struct
...
+         in OCamlFormat v0.15.0, please use `doc-comments` instead. If you \
+         are using `doc-comments-val=before` in combination with \
+         `doc-comments=before` then only `doc-comments=before` is now \
+         required to achive the same behavior. If you are using \
+         `doc-comments-val=before` in combination with `doc-comments=after` \
+         this behavior is not available anymore. If you are using \
+         `doc-comments-val=after` in combination with `doc-comments=before` \
+         please now use `doc-comments=before-except-val`. If you are using \
+         `doc-comments-val=after` in combination with `doc-comments=after` \
+         then only `doc-comments=after` is now required to achieve the same \
```
jberdine, line 542:
```suggestion
         then only `doc-comments=after-when-possible` is now required to achieve the same \
```

lib/Conf.ml
```diff
@@ -514,7 +527,43 @@ module Formatting = struct
...
+               is now required to achive the same behavior. If you are \
+               using `doc-comments-val=before` in combination with \
+               `doc-comments=after` this behavior is not available anymore."
+        ; C.removed_value ~name:"after" ~version:"0.14.2"
+            ~msg:
+              "If you are using `doc-comments-val=after` in combination \
+               with `doc-comments=before` please now use \
+               `doc-comments=before-except-val`. If you are using \
+               `doc-comments-val=after` in combination with \
+               `doc-comments=after` then only `doc-comments=after` is now \
```
jberdine, line 561:
```suggestion
               `doc-comments=after` then only `doc-comments=after-when-possible` is now \
```

Issue #993, opened 2019-08-29
Style suggestion: support infix operators at end of line


CraigFe:
**What is the reason for this new style?**
This is a follow-on to the discussion in https://github.com/ocaml-ppx/ocamlformat/issues/961. The outcome of that is that OCamlformat is now somewhat opinionated about the use of binds: it tries to put them at the start of the line. Given the following input:

```ocaml
let configure i =
  configure_myocamlbuild () >>=
  (* comment *)
  configure_opam ~name:opam_name i >>= fun () ->
  let no_depext = Key.(get ctx no_depext) in
  configure_makefile ~no_depext ~opam_name >>= fun result ->
  foo
```

Ocamlformat will now output this:

```ocaml
let configure i =
  configure_myocamlbuild ()
  >>= (* comment *)
  configure_opam ~name:opam_name i
  >>= fun () ->
  let no_depext = Key.(get ctx no_depext) in
  configure_makefile ~no_depext ~opam_name >>= fun result -> foo
```

Personally, I like the operators being at the start; but there are good reasons for wanting them at the end instead. In particular: `let+` compatibility. The original input snippet (somewhat taken from the [Mirage codebase](https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L568) could easily be converted to `let+` syntax since the result of a function is bound on the same line as the function, e.g:

```ocaml
let configure i =
  let+ () = configure_myocamlbuild in
  (* comment *)
  let+ () = configure_opam ~name:opam_name i () in
  let no_depext = Key.(get ctx no_depext) in
  let+ result = configure_makefile ~no_depext ~opam_name in
  foo
```

Note that the conversion to `let+` syntax from the postfix form touches only single lines; the same is not true of converting from OCamlformat's preferred form.

**Describe the solution you'd like**
I think this is perhaps a case for an option. Something along the lines of `bind=[start|end]` that allows choosing between the various behaviours. This is tricky, because there are uses of infix operators that _always_ want to be at the start of the line, for instance when using 'combinator pattern matching' in the Irmin source code: https://github.com/mirage/irmin/blob/master/src/irmin/node.ml#L52. I think it's fairly justifiable to treat bind operators differently, since they have an agreed-upon semantics.

There is a related question about always putting binds on newlines, rather than trying to cram many of them into the same line, but perhaps this is already enough to be going on with :slightly_smiling_face: 

CC: @Julow

jberdine:
[...]

PR #1359, opened 2020-05-05
Break after arrow of fun args of infix ops

jberdine:
when "break-infix-before-func = true" and "break-sequences = true"

Related to #993

### DIFF
--- a/test/passing/infix_bind.ml
+++ b/test/passing/infix_bind.ml
@@ -211,3 +211,15 @@ let _ = match () with A -> ( x |> function y -> y (*  *) ) | B -> ()
 let _ = x |> function y -> ( function _ -> y (* A *) ) (* B *)
 
 let _ = () (* This is needed here to avoid the comment above from moving *)
+
+include struct
+  [@@@ocamlformat "break-sequences = true"]
+
+  let () =
+    a____________________________________
+    >>= b____________________________________
+    >>= fun () ->
+    c____________________________________
+    >>= d____________________________________
+    >>= fun () -> e
+end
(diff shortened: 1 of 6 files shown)


test/passing/infix_bind.ml
```diff
@@ -211,3 +211,15 @@ let _ = match () with A -> ( x |> function y -> y (*  *) ) | B -> ()
 let _ = x |> function y -> ( function _ -> y (* A *) ) (* B *)
 
 let _ = () (* This is needed here to avoid the comment above from moving *)
+
+include struct
+  [@@@ocamlformat "break-sequences = true"]
```
jberdine, line 216: Is there a better way to add a test with this config?

gpetiot:
I'm not a fan of adding a new behavior based on the combination of 2 options. I didn't see it suggested in #993 is it something that you would like for your codebase at FB ?

Besides I agree that the `break-infix-before-func` option works only on simple code and is broken otherwise, when we have time we can work on a replacement for that.

jberdine:
I don't have strong views on the config interface, but FWIW the reasoning behind the current setup is to view these bind-like infix ops as fancy sequential compositions, and so to use the break-sequences config. That analogy is pretty loose though. Was just trying not to add more options. :-)

There have been a few minor requests for this sort of thing, so we might set it, not sure yet. There seems to be some divergence though, as `break-infix-before-func = false` has seemed mostly not usable. So I don't know.

jberdine:
After further discussion and consideration, it seems better to change code to use monadic let binders. With that context, it does not seem like a good idea to make the implementation of ocamlformat more complex and its behavior less regular and harder to predict.

PR #1352, opened 2020-04-20
Preserve unwrapped comments

gpetiot:
The goal is to:
- fix #1165 
- not break #912 and #1100 

We have to insert a break to preserve #912 and #1100 but we have to do the least possible work (to trying to be too smart), as the wrapping is disabled it's okay not to be optimal and we should be able to preserve as much as possible from the comment shape of the input.

I'm not very satisfied with using `break_unless_newline` here as I aimed at removing every use of this function but didn't find any alternative so far.

Diff of test_branch with the `janestreet` profile (due to having a break):
```diff
diff --git a/vendor/ocamlformat_support/format_.ml b/vendor/ocamlformat_support/format_.ml
index 02269ab..aed7b32 100644
--- a/vendor/ocamlformat_support/format_.ml
+++ b/vendor/ocamlformat_support/format_.ml
@@ -85,7 +85,8 @@ type pp_token =
   | Pp_tbegin of tbox (* beginning of a tabulation box *)
   | Pp_tend (* end of a tabulation box *)
   | Pp_newline (* to force a newline inside a box *)
-  | Pp_if_newline (* to do something only if this very
+  | Pp_if_newline
+  (* to do something only if this very
                      line has been broken *)
   | Pp_string_if_newline of string
   (* print a string only if this very
diff --git a/infer/src/base/TaskBar.ml b/infer/src/base/TaskBar.ml
index 273d2ee22..0de25c495 100644
--- a/infer/src/base/TaskBar.ml
+++ b/infer/src/base/TaskBar.ml
@@ -72,7 +72,8 @@ let draw_top_bar fmt ~term_width ~total ~finished ~elapsed =
     ++ ( "%s"
@@ -72,7 +72,8 @@ let draw_top_bar fmt ~term_width ~total ~finished ~elapsed =
     ++ ( "%s"
        , max (String.length elapsed_string) 9
          (* leave some room for elapsed_string to avoid flicker. 9 characters is "XXhXXmXXs" so it
-            gives some reasonable margin. *) )
+            gives some reasonable margin. *)
+       )
   in
   let top_bar_size = min term_width top_bar_size_default in
   let progress_bar_size = top_bar_size - size_around_progress_bar in
diff --git a/infer/src/clang/cTrans.ml b/infer/src/clang/cTrans.ml
index 6d77c6f3b..912d8da20 100644
--- a/infer/src/clang/cTrans.ml
+++ b/infer/src/clang/cTrans.ml
[...]

PR #1360, opened 2020-05-05
Break after a multiline argument in an argument list

gpetiot:
Found while working on #1352

### DIFF
--- a/test/passing/args_grouped.ml
+++ b/test/passing/args_grouped.ml
@@ -61,3 +61,13 @@ let test_file_renamings_from_json =
         assert_raises exc (fun () -> test_output test_input)
   in
   foooooooooooooooo
+
+let gen_with_record_deps ~expand t resolved_forms ~dep_kind =
+  let foooooooooooooooooooooo =
+    expand
+      (* we keep the dir constant here to replicate the old behavior of:
+         (chdir foo %{exe:bar}). This should lookup ./bar rather than
+         ./foo/bar *)
+      resolved_forms ~dir:t.dir ~dep_kind ~expand_var:t.expand_var
+  in
+  {t with expand_var}
--- a/test/passing/args_grouped.ml
+++ b/test/passing/args_grouped.ml
@@ -1,0 +1,63 @@
+let nullsafe_optimistic_third_party_params_in_non_strict =
+  CLOpt.mk_bool
+    ~long:
+      "nullsafe-optimistic-third-party-params-in-non-strict"
+      (* Turned on for compatibility reasons. Historically this is because
+         there was no actionable way to change third party annotations. Now
+         that we have such a support, this behavior should be reconsidered,
+         provided our tooling and error reporting is friendly enough to be
+         smoothly used by developers. *)
+    ~default:true
+    "Nullsafe: in this mode we treat non annotated third party method params as if they were \
+     annotated as nullable."
+
+let test_file_renamings_from_json =
+  let create_test test_input expected_output _ =
+    let test_output input =
+      DifferentialFilters.FileRenamings.VISIBLE_FOR_TESTING_DO_NOT_USE_DIRECTLY.from_json input
+    in
+    foo
+  in
+  fooooooooooooooo
+
+let eval location exp0 astate =
+  let rec eval exp astate =
+    match (exp : Exp.t) with
+    | Var id ->
+        Ok (eval_var (* error in case of missing history? *) [] (Var.of_id id) astate)
+    | Lvar pvar ->
+        Ok (eval_var [ValueHistory.VariableAccessed (pvar, location)] (Var.of_pvar pvar) astate)
+    | Lfield (exp', field, _) ->
+        goooooooo
+  in
+  fooooooooooooooooooooo
+
+let declare_locals_and_ret tenv pdesc (prop_ : Prop.normal Prop.t) =
+  let foooooooooooooo =
+    BiabductionConfig.run_in_re_execution_mode
+      (* no footprint vars for locals *)
+      sigma_locals_and_ret ()
+  in
+  fooooooooooooooooooooooooooo
+
+let bottom_up fooooooooooo =
+  let empty = Int.equal 0 !scheduled && Queue.is_empty pending in
+  if empty then (
+    remaining := 0 ;
+    L.progress "Finished call graph scheduling, %d procs remaining (in, or reaching, cycles).@."
+      (CallGraph.n_procs syntactic_call_graph) ;
+    if Config.debug_level_analysis > 0 then CallGraph.to_dotty syntactic_call_graph "cycles.dot" ;
+    foooooooooooooooooo )
+  else fooooooooooooooooo
+
+let test_file_renamings_from_json =
+  let fooooooooooooo =
+    match expected_output with
+    | Return exp ->
+        assert_equal ~pp_diff
+          ~cmp:DifferentialFilters.FileRenamings.VISIBLE_FOR_TESTING_DO_NOT_USE_DIRECTLY.equal exp
+          (test_output test_input)
+    | Raise exc ->
+        assert_raises exc (fun () -> test_output test_input)
+  in
+  foooooooooooooooo
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,8 @@
 
   + Break long literal strings at the margin (#1367) (Guillaume Petiot)
 
+  + Break after a multiline argument in an argument list (#1360) (Guillaume Petiot)
+
 ### 0.14.2 (2020-05-11)
 
 #### Changes
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -468,6 +468,10 @@ let drop_inside t loc =
   clear `Within ;
   clear `After
 
+let drop_before t loc =
+  update_cmts t `Before ~f:(fun m -> Location.Multimap.remove m loc) ;
+  t
+
 let has_before t loc =
   pop_if_debug t loc ;
   Location.Multimap.mem t.cmts_before loc
--- a/lib/Cmts.mli
+++ b/lib/Cmts.mli
@@ -92,6 +92,8 @@ val fmt_within :
 
 val drop_inside : t -> Location.t -> unit
 
+val drop_before : t -> Location.t -> t
+
 val has_before : t -> Location.t -> bool
 (** [has_before t loc] holds if [t] contains some comment before [loc]. *)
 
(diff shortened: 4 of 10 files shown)


test/passing/comments_args.ml.ref
```diff
@@ -5,7 +5,9 @@ let emit_wrapper_function =
     (Hhas_pos.pos_to_span ast_fun.Ast.f_span)
     false (* is_async *) false (* is_generator *) false
     (* is_pair_generator *) hoisted true (* no_injection *) true
-    (* inout_wrapper *) is_interceptable false (* is_memoize_impl *) Rx.NonRx
+    (* inout_wrapper *) is_interceptable false
+    (* is_memoize_impl *)
+    Rx.NonRx
```
gpetiot, line 10: closer to the input, probably partially due to the comment attachment (will be improved to another PR)

gpetiot:
I don't know why it s breaking in these cases, the multiline check should be enough. But it's an ugly patch mixing the `break_unless_newline` with the `fits_breaks` that were already in the equation, I had low expectations on this one.

gpetiot:
I managed to have something cleaner by only fixing the `Ast.is_simple` function (for exprs). The `:=` changes in the tests look like a fix to me. Can you check on your codebase, and then I will check on other big projects.

test/passing/args_grouped.ml
```diff
@@ -0,0 +1,42 @@
...
+        goooooooo
+  in
+  fooooooooooooooooooooo
+
+let declare_locals_and_ret tenv pdesc (prop_ : Prop.normal Prop.t) =
+  let foooooooooooooo =
+    BiabductionConfig.run_in_re_execution_mode
+      (* no footprint vars for locals *)
+      sigma_locals_and_ret
+      ()
```
gpetiot, line 40: the comment is only attached to `sigma_locals_and_ret` so this is a complex expression, causing a break before `()`

gpetiot:
Even if it's not a big regression I would like to fix it before merging this PR.
But I didn't find the solution yet, do you have many occurrences of this issue? do they all follow the same pattern as this one?

jberdine:
From what I could see, they all seemed to have that pattern, yes. It's hard to be sure though.

There were not a large number of changes. Note that infer is now using the 0.14.1-13-g000da99 (aka master) version of ocamlformat, so testing should be much easier than before.

gpetiot:
I think I fixed it, there was escaped `\n` that didn't make the formatted string to break but lead the heuristic to guess the string was breaking.

jberdine:
I tested the current version and found 2 cases that are maybe interesting:
```diff
modified   infer/src/biabduction/SymExec.ml
@@ -1161,7 +1161,8 @@ let declare_locals_and_ret tenv pdesc (prop_ : Prop.normal Prop.t) =
     in
     BiabductionConfig.run_in_re_execution_mode
       (* no footprint vars for locals *)
-      sigma_locals_and_ret ()
+      sigma_locals_and_ret
+      ()
   in
   let sigma' = prop_.Prop.sigma @ sigma_locals_and_ret in
   let prop' = Prop.normalize tenv (Prop.set prop_ ~sigma:sigma') in
```
Here my understanding is that the comment is attached to the `sigma_locals_and_ret` arg, and it is therefore considered to be a multi-line arg, and so there is a break after. This could go either way. I expect that it would be more understandable to users if the decision to break was not as dependent on internal implementation details like the particulars of comment attachment. Not a big deal though.

The other is:
```diff
modified   infer/src/unit/DifferentialFiltersTests.ml
@@ -23,8 +23,8 @@ let test_file_renamings_from_json =
     match expected_output with
     | Return exp ->
         assert_equal ~pp_diff
-          ~cmp:DifferentialFilters.FileRenamings.VISIBLE_FOR_TESTING_DO_NOT_USE_DIRECTLY.equal exp
-          (test_output test_input)
+          ~cmp:DifferentialFilters.FileRenamings.VISIBLE_FOR_TESTING_DO_NOT_USE_DIRECTLY.equal
+          exp (test_output test_input)
     | Raise exc ->
         assert_raises exc (fun () -> test_output test_input)
   in
```
This one I expect is a case of a bit of noise when getting very close to the margin due to conflicting with Format's max_indent. Note that the previous version reaches column 98 with a margin configured to 100.

Those are not blockers.

gpetiot:
Great, before merging this one I would like to test the combination of #1352, #1360 and #1367; if the output of test_branch looks good I will integrate them in the preview `janestreet` branch to check we are not causing regressions to avoid a rollback later.

gpetiot:
Minor update on the PR, here is now the diff of test_branch:

```diff
diff --git a/infer/src/base/Config.ml b/infer/src/base/Config.ml
index 29ce48283..583da2d45 100644
--- a/infer/src/base/Config.ml
+++ b/infer/src/base/Config.ml
@@ -1596,7 +1596,8 @@ and nullsafe_optimistic_third_party_params_in_non_strict =
          Historically this is because there was no actionable way to change third party annotations.
          Now that we have such a support, this behavior should be reconsidered, provided
          our tooling and error reporting is friendly enough to be smoothly used by developers.
-      *) ~default:true
+      *)
+    ~default:true
     "Nullsafe: in this mode we treat non annotated third party method params as if they were \
      annotated as nullable."
 
diff --git a/src/dune/expander.ml b/src/dune/expander.ml
index 70eae332..2f634af2 100644
--- a/src/dune/expander.ml
+++ b/src/dune/expander.ml
@@ -463,7 +463,8 @@ let gen_with_record_deps ~expand t resolved_forms ~dep_kind =
     expand
       (* we keep the dir constant here to replicate the old behavior of: (chdir
          foo %{exe:bar}). This should lookup ./bar rather than ./foo/bar *)
-      resolved_forms ~dir:t.dir ~dep_kind ~expand_var:t.expand_var
+      resolved_forms
+      ~dir:t.dir ~dep_kind ~expand_var:t.expand_var
   in
   { t with expand_var }
 
diff --git a/src/dune/foreign_rules.ml b/src/dune/foreign_rules.ml
index 4987a6f8..3b250147 100644
--- a/src/dune/foreign_rules.ml
+++ b/src/dune/foreign_rules.ml
@@ -88,8 +88,8 @@ let build_c_file ~sctx ~dir ~expander ~include_flags (loc, src, dst) =
     (let src = Path.build (Foreign.Source.path src) in
      Command.run
      (* We have to execute the rule in the library directory as the .o is
-        produced in the current directory *) ~dir:(Path.build dir)
-       (Ok ctx.ocamlc)
+        produced in the current directory *)
+       ~dir:(Path.build dir) (Ok ctx.ocamlc)
        [ A "-g"
        ; include_flags
        ; Dyn (Build.map c_flags ~f:(fun x -> Command.quote_args "-ccopt" x))
@@ -121,7 +121,8 @@ let build_cxx_file ~sctx ~dir ~expander ~include_flags (loc, src, dst) =
      let c_compiler = Ocaml_config.c_compiler ctx.ocaml_config in
      Command.run
      (* We have to execute the rule in the library directory as the .o is
-        produced in the current directory *) ~dir:(Path.build dir)
+        produced in the current directory *)
+       ~dir:(Path.build dir)
        (Super_context.resolve_program ~loc:None ~dir sctx c_compiler)
        ( [ Command.Args.S [ A "-I"; Path ctx.stdlib_dir ]
          ; include_flags
```

jberdine:
The one in dune/expander.ml looks like the same behavior I saw where an arg with an attached comment gets considered to be nonsimple due to the existence/length of the comment. The other changes are bug fixes.

gpetiot:
> This could go either way. I expect that it would be more understandable to users if the decision to break was not as dependent on internal implementation details like the particulars of comment attachment. Not a big deal though.

I also think it's not a big deal but this may cause unwanted diff the next time users upgrade ocamlformat. But I have no idea how to fix this, so if there is no objection I will merge it soon.

jberdine:
Just wondering, but is leaving out preceding and trailing comments when pre-formatting in is_simple an option?

gpetiot:
> Just wondering, but is leaving out preceding and trailing comments when pre-formatting in is_simple an option?

I tried something in this direction in my last commit. @jberdine can you test on your codebase there is no regression? I'm dropping the comments before, to not prevent wrapping in this case:
```
   (* fooooooooooooo *)
   fooo foooooooooooo fooooooo
```

but not after, so we have a break here, and the parameters are not hidden after the comment:
```
   fooooooo
   (* fooooooooooooooo *)
   fooooooooooo
```

The test_branch diff looks good:
```diff
diff --git a/infer/src/base/Config.ml b/infer/src/base/Config.ml
index bd9252ee2..9dbffe6a1 100644
--- a/infer/src/base/Config.ml
+++ b/infer/src/base/Config.ml
@@ -1646,7 +1646,8 @@ and nullsafe_optimistic_third_party_params_in_non_strict =
          Historically this is because there was no actionable way to change third party annotations.
          Now that we have such a support, this behavior should be reconsidered, provided
          our tooling and error reporting is friendly enough to be smoothly used by developers.
-      *) ~default:true
+      *)
+    ~default:true
     "Nullsafe: in this mode we treat non annotated third party method params \
      as if they were annotated as nullable."
 
diff --git a/src/dune/foreign_rules.ml b/src/dune/foreign_rules.ml
index a3fe2af7..92c14441 100644
--- a/src/dune/foreign_rules.ml
+++ b/src/dune/foreign_rules.ml
@@ -90,8 +90,8 @@ let build_c_file ~sctx ~dir ~expander ~include_flags (loc, src, dst) =
     (let src = Path.build (Foreign.Source.path src) in
      Command.run
      (* We have to execute the rule in the library directory as the .o is
-        produced in the current directory *) ~dir:(Path.build dir)
-       (Ok ctx.ocamlc)
+        produced in the current directory *)
+       ~dir:(Path.build dir) (Ok ctx.ocamlc)
        [
          A "-g";
          include_flags;
@@ -124,7 +124,8 @@ let build_cxx_file ~sctx ~dir ~expander ~include_flags (loc, src, dst) =
      let c_compiler = Ocaml_config.c_compiler ctx.ocaml_config in
      Command.run
      (* We have to execute the rule in the library directory as the .o is
-        produced in the current directory *) ~dir:(Path.build dir)
+        produced in the current directory *)
+       ~dir:(Path.build dir)
        (Super_context.resolve_program ~loc:None ~dir sctx c_compiler)
        ( [
            Command.Args.S [ A "-I"; Path ctx.stdlib_dir ];
```

gpetiot:
Can we merge this one?

jberdine:
Sorry to be slow to test, I plan to, just need a chance to context switch.

Issue #900, opened 2019-06-26
Wishlist for the 1.0 release


gpetiot:
- [ ] make the `conventional` profile the default profile (currently it is the `ocamlformat` profile)
- [ ] reduce the number of options
   * [ ] `break-string-literals=newlines-and-wrap` should be the default
- [ ] reduce the number of values for options
- [ ] organize the options better (group them together by purpose: ocp-indent, indentation, readability, etc.)
- [ ] make the options discoverable in the terminal when hitting `<TAB>`
- [ ] dunify the tests
- [ ] dunify the release process (`dune-release` ?)

vphantom:
Not sure if it's appropriate here, but I'm waiting for a `break-separators = before-and-docked` option for records and lists, so that we could make opening and closing on separate lines from the contents.  The (single!) thing I miss from Perl is easier editing and a cleaner diff when adding/removing an item in a list, regardless of position (first and last included, as you could trail a useless comma after the last item in Perl, unlike say JS).

It's actually the only thing which I feel is still not perfect in my "janestreet + 11 tweaks" configuration now that 0.10 arrived. :smiley:

gpetiot:
Ocamlformat 0.11 has only been released this morning, I didn't have time to publish an official announcement yet.

Wilfred:
I've left feedback on #932 regarding removal of options: I'm not sure if you just want to change the default or remove the option entirely.

gpetiot:
> I've left feedback on #932 regarding removal of options: I'm not sure if you just want to change the default or remove the option entirely.

I submitted #1057 to address this.

rbjorklin:
I'm all for reducing the number of options but do want to leave my 2 cents before things get cemented. I'm not sure what's the right way of leaving feedback so I'm leaving a note here.

In a professional setting where one or more people review your code I've found it very helpful to have a more vertical style as it helps a lot with making it easier to spot the relevant part of a changed line during review.

I'm currently using this as my starting point:
```
break-cases = fit-or-vertical
break-infix = fit-or-vertical
break-fun-decl = fit-or-vertical
type-decl = sparse
```

gpetiot:
The roadmap is being rewritten, instead of having this meta-issue I will open issues with the remaining points for the 1.0 release and add them to the 1.0 milestone.
I'm closing this one.

PR #1349, opened 2020-04-19
Update headers

jberdine:
- Do not need year anymore

- Added script to automate updating the headers using headache

PR #1361, opened 2020-05-08
Dunify release

Et7f3:
In reference of https://github.com/ocaml-ppx/ocamlformat/issues/900#issue-460827331 (opened by @gpetiot) Here metadata are stored with dune who is responsible to build opam metadata (less metadata to update when releasing)

Some notes:
- header is removed because now it is generated by dune. I can use `ocamlformat.template` to inject header but it will be at the end.
- opam file differ
```diff
-["ocaml" "tools/gen_version.mlt" "lib/Version.ml" version] {pinned}
```
`%%VERSION%%` = `git describe --always --dirty`
the script = `git describe --tags --dirty --always`
I don't know how should I handle this. I don't get how we can print 0.14.2 for instance and not the commit I have tried with manual `git tag` my repo but no luck :shrug:
maybe https://github.com/ocaml-ppx/ocamlformat/pull/1133 handle correctly ?
- The dune-project is formatted with dune (it is sub-optimal because of https://github.com/ocaml/dune/issues/2291)

I open in draft because I might have broken version field.


jberdine:
It should not be necessary to have a header in a file generated from another file that does have a header. I make no guarantees that the bot is capable of handling that though. But a "header" at the end would not be seen either. Let's not worry about the header until the bot yells at me. It is complaining already and will until #1349 is merged anyhow.

But I expect that this will have to wait for at least #1133. I also could not figure out how to specify the build field of the dune-generated opam file, and getting the correct version when pinning is necessary.

Et7f3:
Thanks for your quick reply. I have tested to manually pin master and my branch with [esy](https://esy.sh) and I get `unknown` as version. I will test with opam to see if we got the same version or not.

> But I expect that this will have to wait for at least #1133. I also could not figure out how to specify the build field of the dune-generated opam file, and getting the correct version when pinning is necessary.

I can force dune to add a specific build command (with the template) but it will not update as dune upgrade and change his cli.

ocamlformat.opam
```diff
@@ -1,33 +1,20 @@
...
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
-]
 depends: [
   "ocaml" {>= "4.06"}
   "alcotest" {with-test}
   "base" {>= "v0.11.0"}
   "base-unix"
   "cmdliner"
-  "dune" {>= "2.2.0"}
+  "dune" {>= "2.1.0"}
```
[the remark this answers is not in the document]
Et7f3: oops my bad. I tested a bug in dune (when we manually add constraint on dune) and forget to remove.

ocamlformat.opam
```diff
@@ -1,33 +1,20 @@
...
+license: "MIT"
 homepage: "https://github.com/ocaml-ppx/ocamlformat"
 bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
-dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
-license: "MIT"
-build: [
-  ["ocaml" "tools/gen_version.mlt" "lib/Version.ml" version] {pinned}
-  ["dune" "build" "-p" name "-j" jobs]
-  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
-]
```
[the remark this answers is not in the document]
Et7f3: Ok I have mixed the original build command (for versionning) and the other command generated by dune.

ocamlformat.opam
```diff
@@ -39,5 +26,17 @@ depends: [
...
+  ["ocaml" "tools/gen_version.mlt" "lib/Version.ml" version] {pinned}
+  [
+    "dune"
+    "build"
+    "-p"
+    name
+    "-j"
+    jobs
+    "@install"
+    "@runtest" {with-test}
```
jberdine, line 40: Is it obvious that these have the same behavior as the version before this PR?

jberdine: Ok, I hadn't appreciated that `{with-test}` could be placed on individual arguments instead of whole commands.

Ok, so `@install` only builds the files to be installed, but does not actually copy them into some other directories, etc. Correct?

Et7f3: I will ask on discord for confirmation.

mseri: Yes they are doing almost the same thing. As written the command build all installable libraries executables (Even if not needed for the tests) and runs the tests. You can add different filters on each string in the list In the opam files, it’s quite convenient at times

gpetiot:
I rebased and updated the constraint on `base` according to #1396. I don't think there is a way to format the dune-project file though but it's not an issue.

Et7f3:
`dune format-dune-file FILE` This format any dune file.

gpetiot:
> `dune format-dune-file FILE` This format any dune file.

Yes but it's not integrated the same way dune files are when running `dune @fmt` with the promote feature. We would need to make an inplace transformation in the Makefile. I wouldn't mind waiting until they are automatically formatted by dune.

jberdine:
FWIW it could be `tmp=$(mktemp -t 'dune-format'); dune format-dune-file dune-project > tmp; mv -f tmp dune-project`

emillon:
@jberdine a motivation for this for example is that when you change `(lang dune ...)` and use opam file generation, the bound in the opam file will be updated automatically.

ocamlformat.opam.template
```diff
@@ -0,0 +1,13 @@
+build: [
```
emillon, line 1: If we merge #1133 before this, we can remove this file before merging

gpetiot:
Rebased onto master, should be compatible with the dune-build-info changes.

PR #1367, opened 2020-05-19
Break long literal strings at the margin

### LINKED ISSUE #1366: Bug: newline escapes exceed margin
**Describe the bug**
The newline escapes in this string literal are past the 90-character margin:
```
let _ =
  "_ _____________________ ___________ ________ _____________ ________ _____________ _____\n\n\
  \ ___________________ ______ _____ ___________ _______ _____ _____ __ ___ ______ ___"
;;
```

**How to Reproduce**
Style the above file with `ocamlformat --profile=janestreet`, using `ocamlformat` built from f6bc80a4bd46eab0ba37cb9308f22dd04cf86858 (the `gpetiot/janestreet` branch).

gpetiot:
Fix #1366 
there was a bug in the printing of escaped newlines, no break was printed before a linebreak (and doing it causes way more regressions so this is not a good fix), as Carl suggested in #1366 I've merged the linebreaks with the preceding words to avoid having the linebreaks alone on their line.

As a side-effect it made some long argument list non-break anymore, but this but it looks like this break was a fortunate side-effect of the original code. It will be fixed with #1360 hopefully. I will check if this is really the case.

Here are the changes of test_branch (some are fixes as we are now correctly breaking at the margin, some will be fixed by #1360):
```diff
diff --git a/vendor/parse-wyc/menhir-recover/emitter.ml b/vendor/parse-wyc/menhir-recover/emitter.ml
index e6a664f..63f6cb1 100644
--- a/vendor/parse-wyc/menhir-recover/emitter.ml
+++ b/vendor/parse-wyc/menhir-recover/emitter.ml
@@ -294,8 +294,7 @@ end = struct
     in
     fprintf ppf
       "let token_of_terminal (type a) (t : a %s.terminal) (v : a) : token =\n\
-      \  match t with\n"
-      menhir;
+      \  match t with\n" menhir;
     Terminal.iter case
 
   let emit_nullable ppf =
diff --git a/infer/src/clang/cArithmetic_trans.ml b/infer/src/clang/cArithmetic_trans.ml
index 4a98df832..0f844757a 100644
--- a/infer/src/clang/cArithmetic_trans.ml
+++ b/infer/src/clang/cArithmetic_trans.ml
@@ -178,7 +178,8 @@ let unary_operation_instruction translation_unit_context uoi e typ loc =
       let uok = Clang_ast_j.string_of_unary_operator_kind uoi.Clang_ast_t.uoi_kind in
       L.(debug Capture Medium)
         "@\n\
-         WARNING: Missing translation for Unary Operator Kind %s. The construct has been ignored...@\n"
+         WARNING: Missing translation for Unary Operator Kind %s. The construct has been \
+         ignored...@\n"
         uok ;
       (e, [])
 
diff --git a/infer/src/clang/cTrans.ml b/infer/src/clang/cTrans.ml
index b3a61788b..5db534ec6 100644
--- a/infer/src/clang/cTrans.ml
+++ b/infer/src/clang/cTrans.ml
@@ -2267,8 +2267,7 @@ module CTrans_funct (F : CModule_type.CFrontend) : CModule_type.CTranslation = s
           L.debug Capture Medium
             "couldn't translate initListExpr properly: list lengths do not match:@\n\
             \  field_exps is %d: [%a]@\n\
-            \  stmts      is %d: [%a]@\n"
-            (List.length field_exps)
+            \  stmts      is %d: [%a]@\n" (List.length field_exps)
             (Pp.seq ~sep:"," (Pp.pair ~fst:Exp.pp ~snd:(Typ.pp Pp.text)))
             field_exps (List.length stmts)
             (Pp.seq ~sep:"," (Pp.of_string ~f:Clang_ast_proj.get_stmt_kind_string))
diff --git a/infer/src/concurrency/RacerDDomain.ml b/infer/src/concurrency/RacerDDomain.ml
index c97c4b5e0..6efde26d1 100644
--- a/infer/src/concurrency/RacerDDomain.ml
+++ b/infer/src/concurrency/RacerDDomain.ml
@@ -532,9 +532,9 @@ let pp_summary fmt {threads; locks; accesses; return_ownership; return_attribute
      Accesses %a @\n\
      Ownership: %a @\n\
      Return Attribute: %a @\n\
-     Attributes: %a @\n"
-    ThreadsDomain.pp threads LockDomain.pp locks AccessDomain.pp accesses OwnershipAbstractValue.pp
-    return_ownership Attribute.pp return_attribute AttributeMapDomain.pp attributes
+     Attributes: %a @\n" ThreadsDomain.pp threads LockDomain.pp locks AccessDomain.pp accesses
+    OwnershipAbstractValue.pp return_ownership Attribute.pp return_attribute AttributeMapDomain.pp
+    attributes
 
 
 let pp fmt {threads; locks; accesses; ownership; attribute_map} =
~
diff --git a/compiler/lib-runtime/gen/gen.ml b/compiler/lib-runtime/gen/gen.ml
index 9959ec8d2..31eb474d5 100644
--- a/compiler/lib-runtime/gen/gen.ml
+++ b/compiler/lib-runtime/gen/gen.ml
@@ -56,7 +56,8 @@ let () =
           let name = Filename.basename f in
           let content = read_file f in
           Printf.printf
-            "let %s = Js_of_ocaml_compiler.Builtins.register ~name:\"%s\" ~content:\"%s\"\n"
+            "let %s = Js_of_ocaml_compiler.Builtins.register ~name:\"%s\" \
+             ~content:\"%s\"\n"
             (to_ident (Filename.chop_extension name))
             (String.escaped name)
             (String.escaped content))
```

edit: #1360 is not enough to fix this regression
edit2: now fixed in this PR
edit3: no diff at all with test_branch now


lib/Fmt_ast.ml
```diff
@@ -332,6 +332,17 @@ let fmt_constant c ~loc ?epi const =
           fmt_words ~epi mode curr $ opt next fmt_next
         in
         let lines = String.split ~on:'\n' s in
+        let lines =
+          if break_on_newlines then lines
+          else
+            (* linebreaks are merged with the preceding line when possible
+               instead of having a blank line in the list *)
+            List.foldi lines ~init:[] ~f:(fun i acc -> function
+              | "" when i < List.length lines - 1 -> (
```
jberdine, line 341: Probably best to precompute `List.length lines` to avoid complexity potentially quadratic in number of lines.

gpetiot: Oh right, I will fix this

gpetiot:
Great, before merging this one I would like to test the combination of #1352, #1360 and #1367; if the output of test_branch looks good I will integrate them in the preview `janestreet` branch to check we are not causing regressions to avoid a rollback later.

jberdine:
> I would like to test the combination of #1352, #1360 and #1367

It is times like this when I think we should add changes entries only when doing the final merge into master.

PR #1372, opened 2020-05-22
Fix missing open line between multi-line let-binding with poly-typexpr

jberdine:
The is_simple check for let bindings only considers the location of
the body exp. This location happens to usually be large enough. But
for one of a group of let bindings which has an explicit polymorphic
type constraint, the location of the body exp contains only the body,
and does not include the location of the pattern. This means that
cases such as:
```ocaml
[@@@ocamlformat "let-binding-spacing=compact"]

let f x = x

and g : 'a. (_ -> _ -> _ -> 'a) -> _ -> _ -> _ -> 'a =
 fun h a b -> h (i a) (i b) (i c)
```
get misformatted since `g` is misidentified as being a one-liner, and
 the open line between `f` and `g` gets removed.

This patch fixes this by using the location of the entire let-binding
instead of the body expression.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -4191,8 +4191,8 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
         || (not (List.is_empty itmJ.pvb_attributes))
         || Ast.break_between c.source ~cmts:c.cmts
              ~has_cmts_before:Cmts.has_before ~has_cmts_after:Cmts.has_after
-             (Exp itmI.pvb_expr, cI.conf)
-             (Exp itmJ.pvb_expr, cJ.conf)
+             (Exp {itmI.pvb_expr with pexp_loc= itmI.pvb_loc}, cI.conf)
+             (Exp {itmJ.pvb_expr with pexp_loc= itmJ.pvb_loc}, cJ.conf)
       in
       let grps = List.group bindings ~break in
       let fmt_grp ~first:first_grp ~last:last_grp bindings =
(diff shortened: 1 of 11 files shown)


lib/Fmt_ast.ml
```diff
@@ -4126,8 +4126,8 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
         || (not (List.is_empty itmJ.pvb_attributes))
         || Ast.break_between c.source ~cmts:c.cmts
              ~has_cmts_before:Cmts.has_before ~has_cmts_after:Cmts.has_after
-             (Exp itmI.pvb_expr, cI.conf)
-             (Exp itmJ.pvb_expr, cJ.conf)
+             (Exp {itmI.pvb_expr with pexp_loc= itmI.pvb_loc}, cI.conf)
+             (Exp {itmJ.pvb_expr with pexp_loc= itmJ.pvb_loc}, cJ.conf)
```
jberdine, line 4130: I didn't see a case for value_binding in place of using `Exp` here, but that would be a less-hacky fix. I'm not sure how interconnected with other things adding such a constructor would be.

gpetiot: I think it would be better to add a constructor in the `Ast.t` type for each kind of ast element, I don't think the constructors of this type are interconnected.

jberdine: Done.

jberdine:
There is a test in the PR description that I tried and failed to understand how to add to the test suite as it currently is set up. Where / how should that be added, considering that `let-binding-spacing=compact` is needed to see the difference before vs after?

gpetiot:
I think you can do the same thing as for `test/passing/module_item_spacing.ml` and its variants

jberdine:
@gpetiot could you check the added test?

PR #1375, opened 2020-05-26
More precise ocamlformat version comparison

gpetiot:
The goal is to have more precise and user-friendly error messages, and to allow `X.Y.Z-fooo` versions when the `X.Y.Z` version is set in the .ocamlformat file.

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -1864,20 +1864,56 @@ let string_of_user_error = function
   | `Unknown (name, _) -> Format.sprintf "Unknown option %S" name
   | `Bad_value (name, msg) -> Format.sprintf "For option %S: %s" name msg
 
+(** strips [exe] version of the attributes that are not used in [expected]. *)
+let strip_version ~expected ~exe =
+  let open Semver in
+  let Semver.{major; minor; patch; prerelease; build} = exe in
+  let prerelease, build =
+    match expected with
+    | {prerelease= []; build= []; _} -> ([], [])
+    | {prerelease= []; _} -> ([], build)
+    | {build= []; _} -> (prerelease, [])
+    | _ -> (prerelease, build)
+  in
+  Option.value_exn (from_parts major minor patch prerelease build)
+
+let check_version ~expected ~exe =
+  match Semver.of_string expected with
+  | Some expected_v -> (
+    match Semver.of_string exe with
+    | Some exe_v -> (
+        let exe_v = strip_version ~expected:expected_v ~exe:exe_v in
+        match Semver.compare exe_v expected_v with
+        | 0 -> Ok ()
+        | x when x < 0 ->
+            Error
+              (Format.sprintf
+                 "expected ocamlformat version to be at least %S but got \
+                  %S. Please upgrade."
+                 expected exe)
+        | _ ->
+            Error
+              (Format.sprintf
+                 "expected ocamlformat version to be at most %S but got %S. \
+                  Please downgrade."
+                 expected exe) )
+    | None ->
+        Error
+          (Format.sprintf "expected ocamlformat version to be %S but got %S."
+             expected exe) )
+  | None -> Error (Format.sprintf "malformed version number %S." expected)
+
 let parse_line config ~from s =
   let update ~config ~from ~name ~value =
     let name = String.strip name in
     let value = String.strip value in
     match (name, from) with
-    | "version", `File _ ->
-        if String.equal Version.version value || !no_version_check then
-          Ok config
+    | "version", `File _ -> (
+        if !no_version_check then Ok config
         else
-          Error
-            (`Bad_value
-              ( name
-              , Format.sprintf "expecting %S but got %S" Version.version
-                  value ))
+          match check_version ~expected:value ~exe:Version.version with
+          | Ok () -> Ok config
+          | Error msg -> Error (`Bad_value ("version", msg)) )
     | name, `File x ->
         C.update ~config ~from:(`Parsed (`File x)) ~name ~value ~inline:false
     | name, `Attribute ->
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -1888,7 +1888,12 @@ let check_version ~config ~exe =
           | 0 -> Ok ()
           | x when x < 0 -> Error (err_msg ^ " Please upgrade.")
           | _ -> Error (err_msg ^ " Please downgrade.") )
-      | None -> Error err_msg )
+      | None -> (
+        match exe with
+        | "unknown" ->
+            Error
+              (err_msg ^ " Could not determine current ocamlformat version.")
+        | _ -> Error err_msg ) )
   | None -> Result.failf "malformed version number %S." config
 
 let parse_line config ~from s =
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -1880,27 +1880,18 @@ let strip_version ~expected ~exe =
 let check_version ~expected ~exe =
   match Semver.of_string expected with
   | Some expected_v -> (
-    match Semver.of_string exe with
-    | Some exe_v -> (
-        let exe_v = strip_version ~expected:expected_v ~exe:exe_v in
-        match Semver.compare exe_v expected_v with
-        | 0 -> Ok ()
-        | x when x < 0 ->
-            Error
-              (Format.sprintf
-                 "expected ocamlformat version to be at least %S but got \
-                  %S. Please upgrade."
-                 expected exe)
-        | _ ->
-            Error
-              (Format.sprintf
-                 "expected ocamlformat version to be at most %S but got %S. \
-                  Please downgrade."
-                 expected exe) )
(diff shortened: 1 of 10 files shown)


ocamlformat.opam
```diff
@@ -35,6 +35,7 @@ depends: [
   "ocp-indent" {with-test}
   "odoc" {>= "1.4.2"}
   "re"
+  "semver2"
```
Julow, line 38: A note for later: Most users are installing OCamlformat with Opam in the same switch as their other dependencies. This may be inconvenient when one of OCamlformat's dependency has a version constraint (which can be added afterward).
This library has a dependency on angstrom < 0.14.0, which is already incompatible with some projects, for example, mrmime, colombe, h2.

test/unit/test_unit.ml
```diff
@@ -27,6 +27,51 @@ module Test_location = struct
...
+    ; test "unknown 'exe'" ~expected:"0.14.2" ~exe:"unknown"
+        (Error
+           "expected ocamlformat version to be \"0.14.2\" but got \
+            \"unknown\".")
+    ; test "equal" ~expected:"0.14.1" ~exe:"0.14.1" (Ok ())
+    ; test "accepted" ~expected:"0.14.1" ~exe:"0.14.1-15-g273f6f6-dirty"
+        (Ok ())
+    ; test "too old" ~expected:"0.14.1" ~exe:"0.14.0"
+        (Error
+           "expected ocamlformat version to be at least \"0.14.1\" but got \
```
Julow, line 57: When receiving this message I would be tempted to upgrade to 0.14.2. "at least" is not true here.
I think the fact that `0.14.0-something` is between `0.14.0` and `0.14.1` is not necessary for users and should be removed.

ocamlformat.opam
gpetiot: I'm experimenting something to improve this workflow, so we can have ocamlformat installed independently of what the user has in its current switch. The goal is to not worry about ocamlformat dependencies

test/unit/test_unit.ml
gpetiot: I simplified the error message

test/unit/test_conf.ml
```diff
@@ -0,0 +1,41 @@
...
+      (Error "malformed version number \"0.33..foo\".")
+  ; test "invalid 'exe'" ~expected:"0.14.2" ~exe:"0.33..foo"
+      (Error
+         "expected ocamlformat version to be \"0.14.2\" but got \
+          \"0.33..foo\".")
+  ; test "unknown 'exe'" ~expected:"0.14.2" ~exe:"unknown"
+      (Error
+         "expected ocamlformat version to be \"0.14.2\" but got \"unknown\".")
+  ; test "equal" ~expected:"0.14.1" ~exe:"0.14.1" (Ok ())
+  ; test "accepted" ~expected:"0.14.1" ~exe:"0.14.1-15-g273f6f6-dirty"
```
emillon, line 22: It makes sense to accept that but I think that the new semantics should be documented somewhere. Maybe a new section in the manpage (since only CLI options are documented, the fact that you can put `version=` in a config file is actually not directly documented)

test/unit/test_conf.ml
```diff
@@ -0,0 +1,41 @@
...
+         "expected ocamlformat version to be \"0.14.1\" but got \
+          \"0.14.0-15-g273f6f6-dirty\". Please upgrade.")
+  ; test "too recent" ~expected:"0.14.0" ~exe:"0.14.1"
+      (Error
+         "expected ocamlformat version to be \"0.14.0\" but got \"0.14.1\". \
+          Please downgrade.")
+  ; test "too recent 2" ~expected:"0.14.0" ~exe:"0.14.1-15-g273f6f6-dirty"
+      (Error
+         "expected ocamlformat version to be \"0.14.0\" but got \
+          \"0.14.1-15-g273f6f6-dirty\". Please downgrade.") ]
```
emillon, line 39: when version number subsitution does not work for some reason (say, `git` is not installed), `gen_version.mlt` will put `"unknown"` in the executable. What do you think about adding a test for this case? And maybe having a special error message in that case as well?

test/unit/test_conf.ml
```diff
@@ -0,0 +1,41 @@
+let test_check_version =
+  let test name ~expected ~exe res =
```
emillon, line 2: It's a bit surprising that `~expected` is not the expected test output here. What do you think about using `~config` to mean that it's the value found in the config file ("specification" might work as well)? That way `res` can be renamed to a more standard `expected`.

Actually this applies to `check_version` as well, it seems that `config:string -> exe:string -> ...` describes better what this function does.

ocamlformat.opam
emillon: I wouldn't worry too much, this dependency is easy to delete later if need be.

lib/Conf.ml
```diff
@@ -1864,20 +1864,47 @@ let string_of_user_error = function
   | `Unknown (name, _) -> Format.sprintf "Unknown option %S" name
   | `Bad_value (name, msg) -> Format.sprintf "For option %S: %s" name msg
 
+(** strips [exe] version of the attributes that are not used in [expected]. *)
+let strip_version ~expected ~exe =
+  let open Semver in
+  let Semver.{major; minor; patch; prerelease; build} = exe in
+  let prerelease, build =
```
emillon, line 1871: I think that can be simplified to

```
let prelease = if List.is_empty expected.prelease then [] else prerelease in
let build  = if List.is_empty expected.build then [] else build in
```

(or with pattern matching; the point is that the two decisions are independent)

lib/Conf.ml
```diff
@@ -1864,20 +1864,47 @@ let string_of_user_error = function
...
+      in
+      match Semver.of_string exe with
+      | Some exe_v -> (
+          let exe_v = strip_version ~expected:expected_v ~exe:exe_v in
+          match Semver.compare exe_v expected_v with
+          | 0 -> Ok ()
+          | x when x < 0 -> Error (err_msg ^ " Please upgrade.")
+          | _ -> Error (err_msg ^ " Please downgrade.") )
+      | None -> Error err_msg )
+  | None -> Error (Format.sprintf "malformed version number %S." expected)
```
emillon, line 1895: If that's your thing:

```suggestion
  | None -> Result.failf "malformed version number %S." expected
```

(this also works above)

jberdine:
Note that, as discussed elsewhere before, it is necessary to have a precise version check that fails if the running binary was not built from the commit specified in the config file. If you want a loose check for:

> allow X.Y.Z-fooo versions when the X.Y.Z version is set in the .ocamlformat file

then it needs to be some other check than the current one.

gpetiot:
I improved the check (the one in Semver returned incorrect results when there are build infos) and wrote a documentation, let me know what you think.

test/unit/test_conf.ml
```diff
@@ -0,0 +1,53 @@
...
+  ; test "invalid exe" ~config:"0.14.2" ~exe:"0.33..foo"
+      (Error
+         "expected ocamlformat version to be \"0.14.2\" but got \
+          \"0.33..foo\".")
+  ; test "unknown 'exe'" ~config:"0.14.2" ~exe:"unknown"
+      (Error
+         "expected ocamlformat version to be \"0.14.2\" but got \
+          \"unknown\". Could not determine current ocamlformat version.")
+  ; test "equal" ~config:"0.14.1" ~exe:"0.14.1" (Ok ())
+  ; test "accepted" ~config:"0.14.1" ~exe:"0.14.1-15-g273f6f6-dirty" (Ok ())
```
jberdine, line 24: This case is worrying. This is the situation that would arise when someone has a checkout around and mistakenly ends up using it. This is one of the scenarios that the version check is intended to protect against.

What is the desired use case for this new loose check? I ask because I don't understand why the existing `-no-version-check` option is insufficient. Also, I don't know why it would not work to allow some wildcards in the specified versions in order to indicate that a loose check is desired.

gpetiot: My idea is that `x.y.z-foo` should be close enough to `x.y.z` to be accepted, otherwise the version number would have been `x.y.z+1` or more. Basically:
- if .ocamlformat sets `version=x.y.z-fooo`, we check that we run the exact commit build
- if .ocamlformat sets `version=x.y.z`, we check that we are running a build of the `x.y.z` "family" that might be the exact version, or some commits after

jberdine: Yes, that is the situation I'm worried about. It is not unexpected for a config file to have a constraint of the form `version=x.y.z` and failing if the binary version is something like `x.y.z-d` is precisely a goal of the version check.

I still don't understand where this is needed that `--no-version-check` is insufficient for?

gpetiot: I thought this new behavior (and not requiring --no-version-check) was a good thing (I can't find anything about a need for this, I thought I read that somewhere though), but if it's not I will close this PR .

PR #1376, opened 2020-05-26
Distinguish hash-getter from hash-comparison infix operators

### LINKED ISSUE #1371: Bug: Space around operator isn't working
Hi !

**Describe the bug**

I have a case where ocamlformat removes the space around a custom infix operator.

**How to Reproduce**

Version: 0.14.2

Original file
```ocaml
let formula_base x =
    let open Formula.Infix in
    (Expr.typeof x) #== (Lit (Type IntType))
    #&& (x #<= (Expr.int 4))
    #&& ((Expr.int 0) #< x)
```

Produced diff:
```ocaml
- #&& ((Expr.int 0) #< x)
+ #&& ((Expr.int 0)#<x)
```

Config:
```
profile=default
quiet=false
max-iters=10
comment-check=true
wrap-fun-args=true
wrap-comments=false
type-decl-indent=2
type-decl=compact
stritem-extension-indent=0
space-around-variants=true
space-around-records=true
space-around-lists=true
space-around-arrays=true
single-case=compact
sequence-style=terminator
parse-docstrings=false
parens-tuple=always
parens-ite=false
ocp-indent-compat=false
nested-match=wrap
module-item-spacing=sparse
max-indent=68
match-indent-nested=never
match-indent=0
margin=80
let-open=preserve
let-module=compact
let-binding-spacing=compact
let-binding-indent=2
let-and=compact
infix-precedence=indent
indicate-multiline-delimiters=space
indent-after-in=0
if-then-else=compact
function-indent-nested=never
function-indent=2
field-space=loose
extension-sugar=preserve
extension-indent=2
exp-grouping=parens
escape-strings=preserve
escape-chars=preserve
dock-collection-brackets=true
doc-comments-tag-only=default
doc-comments-padding=2
doc-comments-val=unset
doc-comments=before (file .ocamlformat:11)
disable=false
cases-exp-indent=4
break-struct=force
break-string-literals=auto
break-sequences=true
break-separators=after
break-infix-before-func=false
break-infix=wrap
break-fun-sig=smart (file .ocamlformat:10)
break-fun-decl=smart (file .ocamlformat:9)
break-collection-expressions=fit-or-vertical (file .ocamlformat:8)
break-cases=fit-or-vertical (file .ocamlformat:7)
break-before-in=fit-or-vertical (file .ocamlformat:6)
assignment-operator=end-line (file .ocamlformat:5)
align-variants-decl=true (file .ocamlformat:4)
align-constructors-decl=true (file .ocamlformat:3)
align-cases=true (file .ocamlformat:2)
```

gpetiot:
fix #1371

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -77,6 +77,17 @@ let is_infix e =
   | Pexp_ident {txt= Lident i; _} -> is_infix_id i
   | _ -> false
 
+let is_hash_getter_id i =
+  let is_infix_char = function
+    | '$' | '%' | '*' | '+' | '-' | '/' | '<' | '=' | '>' | '|' | '&' | '@'
+     |'^' | '#' | '.' ->
+        true
+    | _ -> false
+  in
+  match (i.[0], i.[String.length i - 1]) with
+  | '#', ('#' | '.') when String.for_all i ~f:is_infix_char -> true
+  | _ -> false
+
 module Indexing_op = struct
   type brackets = Round | Square | Curly
 
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -61,16 +61,20 @@ let is_monadic_binding e =
   | Pexp_ident {txt= Lident i; _} -> is_monadic_binding_id i
   | _ -> false
 
-let is_infix_id i =
-  match (i.[0], i) with
-  | ( ( '$' | '%' | '*' | '+' | '-' | '/' | '<' | '=' | '>' | '|' | '&' | '@'
-      | '^' | '#' )
-    , _ )
-   |( _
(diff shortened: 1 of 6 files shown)


lib/Ast.ml
```diff
@@ -77,6 +77,17 @@ let is_infix e =
   | Pexp_ident {txt= Lident i; _} -> is_infix_id i
   | _ -> false
 
+let is_hash_getter_id i =
+  let is_infix_char = function
+    | '$' | '%' | '*' | '+' | '-' | '/' | '<' | '=' | '>' | '|' | '&' | '@'
+     |'^' | '#' | '.' ->
```
Julow, line 83: This list is already in `is_infix_id`. Could you factorize the code ?

PR #1377, opened 2020-05-26
Fix inconsistencies of the closing parentheses with indicate-multiline-delimiters

### LINKED ISSUE #1174: Bug: indicate-multiline-delimiters = closing-on-separate-line doesn't affect anonymous functions
**Describe the bug**
When ocamlformat 0.12 from opam and `indicate-multiline-delimiters = closing-on-separate-line` is set I would expect an anonymous function or local module open to have its closing paren on a separate line.  For example:
```ocaml
(* ocamlformat --enable-outside-detected-project test.ml --indicate-multiline-delimiters=closing-on-separate-line -m 20 *)
let _ =
  List.map
    (fun x ->
      some code here) (* <- no newline before the closing delimiter *)
    some_list
```

**How to Reproduce**
Using the `ocamlformat` invocation in the comment above and ocamlformat 0.12.  This was run under `/tmp` with no `.ocamlformat` present and appears consistent with the results in the ocamlformat test suite.

It makes it editing paren-wrapped multi-line expressions more difficult than it should be because the last line needs special treatment.

gpetiot:
fix #1174 no regression with test_branch on the main profiles

@hcarty how does it look on your codebase? feel free to give your feedback on this patch

~TODO: revert #1418  in this PR~ done


hcarty:
Thank you so much for working on this!

I probably can't test this until Friday but I'll get you feedback as soon as I can.

hcarty:
Ok, I tried this branch with some of our code at work.  It's a nice improvement, thank you!  I have a few questions about some inconsistencies I noticed.

1. The closing paren in an anonymous function is indented from the indentation level of the opening line.  For example, [here](https://github.com/ocaml-ppx/ocamlformat/pull/1377/files#diff-223942c11f900e48435ac3c9b242e1e2R126).  Can that be lined up with the opening line?
1. In looking through the change, I noticed that the option doesn't seem to affect parens around [tuples](https://github.com/ocaml-ppx/ocamlformat/blob/5c9e6ef391e4a6d943399ebff675a91cd5a7f590/test/passing/ite-fit_or_vertical_closing.ml.ref#L103) or when they're used to delimit expression bounds:
    ```
    let _ =
      Some
        (Constructor
           {
             looooooooooooooooong = loooooooooooooooooooooooooooong;
             loooooong = looooooooooooooooooooooong;
           })
    ```
    Can these be affected as well by `closing-on-separate-line`?

gpetiot:
I extended the coverage of this option.
The weird side-effect is that we are exceeding the margin sometimes, but it was also possible before this change, so I don't consider this to be blocking.
@jberdine do you have any thoughts?

hcarty:
Apologies for streaming these in. I found another case, local open syntax, where the closing delimiter isn't on a separate line:
```ocaml
let _ =
  M.(
    loooooooooooooooooooooong + loooooooooooooooooong
    + llllllllllloooooooooooooooooonnnnnnnnnnnnnggggggggggg)
```
Tested against 13fe7e2

Thank you for these improvements!

jberdine:
While reviewing a test diff, I noticed a missed case:
```ocaml
modified   sledge/src/domain_sh.ml
@@ -53,7 +53,7 @@ let exec_inst pre inst =
       Some
         (Exec.move pre
            (IArray.map reg_exps ~f:(fun (r, e) ->
-                (Var.of_reg r, Term.of_exp e) )))
+                (Var.of_reg r, Term.of_exp e) ) ))
   | Load {reg; ptr; len; _} ->
       Exec.load pre ~reg:(Var.of_reg reg) ~ptr:(Term.of_exp ptr)
         ~len:(Term.of_exp len)
```

There is also a rather strange one:
```ocaml
modified   sledge/bin/domain_itv.ml
@@ -83,7 +83,8 @@ let rec texpr_of_nary_term subtms typ q op =
     match op with
     | Texpr1.Add ->
         Some
-          (Texpr1.Binop (Texpr1.Mul, base, apron_of_q coeff, typ, Texpr0.Rnd))
+          (Texpr1.Binop (Texpr1.Mul, base, apron_of_q coeff, typ, Texpr0.Rnd)
+          )
     | Texpr1.Mul
     (* only handle positive integer exponents *)
       when Z.equal Z.one (Q.den coeff) && Q.sign coeff = 1 ->
```
This change is perhaps technically correct, but it seems like it would be better to break the box than to place the closing paren by itself on a line.

By and large, I think that this patch improves the consistency of spacing before closing parens. That's very good, thanks!

On the other hand, the situation when there are multiple closing parens one after another on a line is more highlighted as suboptimal now. In particular, once there is one closing paren whose match is on a different line, then all succeeding closing parens are preceded by a space. This isn't so bad, but it is not informative since only the first one actually means anything. I have not investigated how complex the implementation would be, but just to note that it would be better to produce e.g.
```ocaml
       Some
         (Exec.move pre
            (IArray.map reg_exps ~f:(fun (r, e) ->
                (Var.of_reg r, Term.of_exp e) )) )
```
(where there is a space before a closing paren whose matching open paren is on a different line from the preceding closing paren)
or
```ocaml
       Some
         (Exec.move pre
            (IArray.map reg_exps ~f:(fun (r, e) ->
                (Var.of_reg r, Term.of_exp e) )))
```
(where only the first successive closing paren on a line gets a space).

Regarding the margin issue, have you tried to place an empty string with a non-zero length inside the box that closes just before the closing paren, and then format the closing paren as width zero? If that could be done, I expect that you could get the boxes to break and leave room for the closing parens.

gpetiot:
I made some improvements with the unnecessary spaces between closing parentheses, and the missing case of Pexp_apply should now be supported. Please let me know how it looks on your codebase.

hcarty:
@gpetiot This is a wonderful improvement.  I found two more inconsistencies in eabc210 when running :
```
dune exec ocamlformat -- --disable-conf-files --indicate-multiline-delimiters=closing-on-separate-line --enable-outside-detected-project parens.ml
```
I get two inconsistent cases.  The first doesn't put the closing paren on a new line when delimiting an expression in function application:
```ocaml
let _ =
  i'm_a_function loooooooooooong
    (loooooooooooong looooooooooooooong loooooooooooooong
       [ loooooooooong; loooooooooooong; loooooooooooooooooooooong ])
```
The second, which is the same basic idea but with a longer list expression in the `(` `)` delimited expression.  The closing paren is on a new line but it isn't vertically aligned with its opening delimiter:
```ocaml
let _ =
  i'm_a_function loooooooooooong
    (loooooooooooong looooooooooooooong loooooooooooooong
       [
         loooooooooong;
         loooooooooooong;
         loooooooooooooooooooooong;
         looooooooooong;
         looooooooong;
         long;
       ]
       )
```

jberdine:
I see what looks like an instance of the same issue as hcarty's first one:
```diff
modified   sledge/src/domain_sh.ml
@@ -54,7 +54,7 @@ let exec_inst pre inst =
       Some
         (Exec.move pre
            (IArray.map reg_exps ~f:(fun (r, e) ->
-                (Var_of_Llair.reg r, Term_of_Llair.exp e) )))
+                (Var_of_Llair.reg r, Term_of_Llair.exp e) )) )
   | Load {reg; ptr; len; _} ->
       Exec.load pre ~reg:(Var_of_Llair.reg reg) ~ptr:(Term_of_Llair.exp ptr)
         ~len:(Term_of_Llair.exp len)
```

I'm still interested what you think about breaking boxes instead of leaving parens alone on lines, and what you make of the larger question of spacing repeated closing parens.

gpetiot:
I improved the Pexp_apply again in my last commit. Regarding the repeated closing parens I tried several things without success, the only improvements in this PR are the low hanging fruits when several parens are closed at the same time (I mean in the same part of a function of Fmt_ast, didn't find a solution when parens are printed at different parts of the code).
I think we are close to what is achievable with what we have at hand.

hcarty:
Thank you so much @gpetiot!

gpetiot:
> I wonder if without a general treatment of multiple closing parens, maybe it would be better for now to leave out special casing the low-hanging fruit too?

I reverted this change to be more consistent.
There are still some regressions regarding the margin, I'm trying to fix this now or it will be a friction point for the next release.

jberdine:
@gpetiot, I'm not sure what the situation is for this one. Are you currently unhappy with the margin behavior? I just rechecked the current state of this branch and it looked pretty good. There are some cases of the closing parens getting unlucky with the margin and getting broken just before. Is trying to make those cases break the parenthesized box rather than just before the paren what you referred to as the friction point? It would be nice to improve there.

Depending on the timing and synchronization constraints, I think that we would not absolutely have to have this in 0.15, but it would be nice to.

hcarty:
I would personally love to have this in 0.15.! I also understand if it isn't ready in some cases and would have to wait for a later release.

gpetiot:
Yes I would prefer fixing the margin issues before merging this PR to avoid having a diff for this in 0.15 and in 0.16 later. I don't have time to look at the improvements that you @jberdine suggested before we make the release, that's why I didn't merge yet, but if it's not necessary I don't mind merging.
What do you think @Julow @emillon ?

emillon:
> What do you think @Julow @emillon ?

I agree that it'd be better to leave that aside for 0.15 and iron it out on the side to avoid huge diffs twice.

lib/Ast.ml
```diff
@@ -1475,7 +1475,8 @@ end = struct
       | Pstr_eval (e0, _) -> assert (e0 == exp)
       | Pstr_value (_, bindings) ->
           assert (
-            List.exists bindings ~f:(fun {pvb_expr; _} -> pvb_expr == exp) )
+            List.exists bindings ~f:(fun {pvb_expr; _} -> pvb_expr == exp )
```
gpetiot, line 1478: regression

lib/Cmts.ml
```diff
@@ -67,7 +67,7 @@ end = struct
       ( match p.ppat_desc with
       | Ppat_record (flds, Open) ->
           Option.iter (Source.loc_of_underscore src flds p.ppat_loc)
-            ~f:(fun loc -> locs := loc :: !locs)
+            ~f:(fun loc -> locs := loc :: !locs )
```
gpetiot, line 70: regression

lib/Cmts.ml
```diff
@@ -368,7 +368,7 @@ let break_comment_group source margin {Cmt.loc= a; _} {Cmt.loc= b; _} =
   let horizontal_align =
     line_dist a b = 0
     && Option.value_map (Source.string_between source a.loc_end b.loc_start)
-         ~default:true ~f:(fun x -> String.(strip x |> is_empty))
+         ~default:true ~f:(fun x -> String.(strip x |> is_empty) )
```
gpetiot, line 371: regression

test/passing/issue289.ml
```diff
@@ -3,93 +3,93 @@
...
         | AAAAAAAAAAAAAAAAAAAa -> x.idddddddddddddddddddddddddd
-        | BBBBBBBBBBBBBBBB -> ccccccccccccccccccccccc)
+        | BBBBBBBBBBBBBBBB -> ccccccccccccccccccccccc )
   ; field
       "id"
       ~doc:"Toy ID."
       ~args:[]
       ~typ:(non_null guid)
-      ~resolve:(fun _ctx x -> x.id)
+      ~resolve:(fun _ctx x -> x.id )
```
gpetiot, line 36: regression

test/passing/sig_value.mli.ref
```diff
@@ -1,7 +1,7 @@
 val f : f:(string[@att]) (** doc *) -> unit
 
 val f :
-     f:(string[@att])
+     f:(string[@att] )
```
gpetiot, line 4: regression

test/passing/wrap_comments_break.ml
```diff
@@ -3,6 +3,6 @@ let _ =
     fffffffffff
       aaaaaaaaaaaaaaaaaaaaaaaaaaaa
       bbbbbbbbbbbbbbbbbbbbbbbbbb
-      ~f:(fun x -> return xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx)
+      ~f:(fun x -> return xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx )
```
gpetiot, line 6: regression

test/passing/issue289.ml
```diff
@@ -64,23 +64,23 @@ let foo =
 let foo =
   let open Gql in
   [ field "id" ~doc:"Toy ID." ~args:[] ~typ:(non_null guid)
-        ~resolve:(function _ctx -> x.id)
+        ~resolve:(function _ctx -> x.id )
```
gpetiot, line 67: regression

gpetiot:
I think I've fixed all regressions and inconsistencies now, should we have a discussion about changing the value of this option for the conventional default profile? (from `indicate-multiline-delimiters=space` to `...=no`). Would it ease the transition for projects that use the conventional profile like mirage and irmin? @emillon @samoht @Julow

samoht:
We have started to unset the option in the default mirage/irmin profiles (https://github.com/mirage/irmin/blob/master/.ocamlformat#L4) so yes I think changing that value in the default profile makes sense.

gpetiot:
Done in #1452, that should be merged before this one to reduce the diff in the vendored libs.

lib/Fmt_ast.ml
```diff
@@ -1847,36 +1866,55 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
+            if Location.is_single_line pexp_loc c.conf.margin then Fit
+            else Break
+          in
           fmt_if parens "("
           $ hvbox 2
-              (fmt_args_grouped ~epi:(fmt_atrs $ fmt_if parens ")") e0 e1N1)
-      )
+              ( fmt_args_grouped ~epi:fmt_atrs e0 e1N1
+              $ fmt_if_k parens
+                  ( match c.conf.indicate_multiline_delimiters with
```
emillon, line 1913: This pattern is present in many places. Is there a way to abstract it a bit?

emillon:
I was intimidated by the diffstat, but ignoring the whitespace changes it's manageable and understandable.

It terms of release management, if this is merged this means that everybody is going to get diffs: as a user of the default profile you get diffs because the default got changed to `no`. Your 2 options are to keep it as is, or to switch to `spaces` and get diffs in the opposite way. The end result is more consistent for everyone but it could be a bit tough.

lib/Fmt_ast.ml
gpetiot: I factorized in places where it was feasible, in some other places it could still be improved but it does not only depend on `indicate-multiline-delimiters`, parenthesizing as a whole can still be improved

PR #1378, opened 2020-05-26
Externalize Prec, Assoc and better precedence ambiguity type

gpetiot:
No behavioral change, just some Spring cleaning in the Ast.ml file:
- externalizing precedence type in Prec.t
- externalizing associativity type in Assoc.t
- use a friendlier type for precedence ambiguity

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -1721,22 +1721,27 @@ end = struct
       binary operators, otherwise returns [None] if [ctx] has no precedence
       or [Some None] if [ctx] does but [ast] does not. *)
   let ambig_prec ({ast; _} as xast) =
-    prec_ctx xast
-    >>| fun (prec_ctx, which_child) ->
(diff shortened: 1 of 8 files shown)


lib/Ast.ml
```diff
@@ -1801,22 +1721,27 @@ end = struct
...
+      match prec_ast ast with
+      | Some prec_ast ->
+          let ambiguous =
+            match Prec.compare prec_ctx prec_ast with
+            | 0 ->
+                (* which child and associativity match: no parens *)
+                (* which child and assoc conflict: add parens *)
+                Assoc.equal which_child Non
+                || not (Assoc.equal (Assoc.of_prec prec_ast) which_child)
+            | cmp when cmp < 0 ->
```
Julow, line 1735: This and the next case can be written `| cmp -> cmp >= 0` (and one of the two comment can be dropped).

PR #1379, opened 2020-05-26
Remove unnecessary parens around object

gpetiot:
no diff with test_branch


lib/Fmt_ast.ml
```diff
@@ -2535,13 +2535,15 @@ and fmt_class_structure c ~ctx ?ext self_ fields =
...
+  let no_attr p = List.is_empty p.ppat_attributes in
   hvbox 2
     ( hvbox 0
         ( str "object"
         $ fmt_extension_suffix c ext
         $ opt self_ (fun self_ ->
-              fmt "@;" $ wrap "(" ")" (fmt_pattern c (sub_pat ~ctx self_)) )
-        )
+              fmt "@;"
+              $ wrap_if (no_attr self_) "(" ")"
```
Julow, line 2545: This assumes that `fmt_pattern` always adds parentheses when there is attributes. We do this everywhere but I think we should stop.
What do you think of passing `~parens:false` to `fmt_pattern` ? The logic in `fmt_pattern_attributes` need to be tweaked but I think this will give better results and saner code.

gpetiot: I agree, I aimed at the smallest diff, but this would be better

JeffersonW13 (commented):
Okay thanks stop him

PR #1380, opened 2020-05-27
Preserve fun keyword

### LINKED ISSUE #1373: Feature request: Provide option to keep f x = fun a b ->
Currently expressions like this:
```
let make_compare op = fun lhs rhs -> ...
```
will always get rewritten like this
```
let make_compare op lhs rhs = ...
```
This makes code look more uniform but it loses information. Splitting a function declaration up in two like this can serve as documentation about the intended staging of the function. Thus losing this by formatting can decrease readability.

There are more [involved techniques](https://blog.janestreet.com/making-staging-explicit/) to make this explicit but they require refactoring and can be overkill for small projects or if a dependency to JaneStreet's libraries is not an option. 

An option to preserve this would be nice.

gpetiot:
Fix #1373, the reason for this change is that I agree using the `fun` keyword has meaning and should be preserved.
No diff with test_branch

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
   + Do not break inline elements such as `{i blah}` in docstrings (#1346) (Josh Berdine)
 
+  + Preserve `fun` keyword (#1380) (Guillaume Petiot)
+
 #### Bug fixes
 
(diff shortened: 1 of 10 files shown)


gpetiot:
That's a good point, closing this PR.

PR #1382, opened 2020-05-27
Remove trailing space after expression when followed by an attribute and break before attributes attached to multi-line phrases

### LINKED ISSUE #1356: Bug: trailing spaces introduced for expressions with attributes
There are many cases where OCamlformat will introduce trailing spaces before wrapped attributes:

```ocaml
;;
(a_______________________________________________________________________________ 
[@attr]) ()

;;
(a_______________________________________, b____________________________________) 
[@attr]

;;
{ a____________________________________ = b___________________________________ } 
[@attr]
```

gpetiot:
fix #1356
I had to change the kind of boxes in a lot of places to maintain the current behavior of wrapping the attribute at the end of the previous line when possible.
But `fmt_attributes c ~pre:(str " ")` is still used in a lot of places so there must be more issues like this, @jberdine was this `pre` intended? Should we preserve the current behavior or is it okay to put the attribute in the next line instead of wrapping at the end of the previous line?


jberdine:
Just for the record: I do not consider trailing spaces to be a "bug". It's nice not to generate them, but it is really not something that I would want to complicate the code too much over. But that is just my opinion.

I do not recall any particular need to place attributes on the same line instead of letting a newline appear before them. But IIRC @hhugo added most of the handling of attributes since early on attributes were just dropped. Any recollection Hugo?

hhugo:
I do not recall any reason

test/passing/js_source.ml.ocp
```diff
@@ -95,11 +96,13 @@ let () =
...
       ()
-    done [@foo]];
+    done
+    [@foo]];
   [%foo
     for x = () to () do
       ()
-    done [@foo]];
+    done
+               [@foo]];
```
gpetiot, line 105: due to ocp-indent

test/passing/js_source.ml.ocp
```diff
@@ -132,22 +135,29 @@ class x =
...
-  end [@foo]
+  end
+  [@foo]
 
 (* Class type expressions *)
 class type t =
   object
-    inherit t [@@foo]
+    inherit t
+  [@@foo]
```
gpetiot, line 145: same

test/passing/js_source.ml.ocp
```diff
@@ -132,22 +135,29 @@ class x =
...
+    method x : t
+    [@@foo]
 
-    method private x : t [@@foo]
+    method private x : t
+    [@@foo]
 
-    constraint t = t' [@@foo]
+    constraint t = t'
+  [@@foo]
```
gpetiot, line 160: same

gpetiot:
I extended the change to more constructs to have an idea of what it will look like. Most of the time having the attributes at the beginning of the line make it easier to read and see to what it is attached to.

test/passing/attributes.ml
```diff
@@ -215,17 +215,19 @@ let _ = ([] @ []) [@a]
...
-     @ b_____________________________________ ) [@a])
+     @ b_____________________________________ )
+     [@a])
 
 (* Attribute should wrap as a block *)
 let _ =
   ( a_______________________________________________________________________
-  @ b________________________________________________________________ ) [@a]
+  @ b________________________________________________________________ )
+  [@a]
```
CraigFe, line 225: Looks like this attribute should fit. If not, this test isn't checking the property it was intended for anymore and should be modified.

test/passing/attributes.ml
```diff
@@ -275,12 +277,14 @@ let _ = f ((0, 0) [@a])
 
 ;;
 ( a________________________________________
-, b________________________________________ ) [@a]
+, b________________________________________ )
+[@a]
```
CraigFe, line 281: This attribute, and the two below it, definitely look like they should fit.

jberdine: The phrase that the attribute attaches to itself breaks. So breaking in this case is consistent with analogous situations.

CHANGES.md
```diff
@@ -30,6 +30,8 @@
 
   + Fix placement of comments on constants (#1383) (Guillaume Petiot)
 
+  + Remove trailing space after expression when followed by an attribute and make attributes more obvious at the beginning of a line (#1382) (Guillaume Petiot)
```
jberdine, line 33: Would it be accurate to write something like:
```suggestion
  + Remove trailing space after expression when followed by an attribute and break before attributes attached to multi-line phrases (#1382) (Guillaume Petiot)
```
and leave the more obvious part to the interpretation of readers?

test/passing/js_source.ml.ocp
gpetiot: Before I merge this issue @ceastlund @hhugo is this kind of ocp-indent quirks an issue for JaneStreet? I didn't see any regression using test_branch so I guess this code pattern isn't that common

ceastlund: We run into this sometimes and yeah, it's really annoying. But it's not super common.

lib/Fmt_ast.ml
```diff
@@ -853,7 +853,7 @@ and fmt_core_type c ?(box = true) ?(in_type_declaration = false) ?pro
         @@ hvbox 4
              ( hvbox 2 fmt_field
              $ fmt_docstring_padded c doc
-             $ fmt_attributes c ~pre:(str " ") ~key:"@" atrs )
+             $ fmt_attributes c ~pre:(fmt "@ ") ~key:"@" atrs )
```
gpetiot, line 856: I guess we can have the `pre` depending on the `ocp-indent-compat` option so we keep the previous behavior when this option is set and JaneStreet will not impacted by the `ocp-indent` bugs

gpetiot:
This change is now only applied if `ocp-indent-compat` is not set so there won't be regressions on JaneStreet codebase.

lib/Fmt_ast.ml
```diff
@@ -541,6 +541,10 @@ let sequence_blank_line c (l1 : Location.t) (l2 : Location.t) =
       loop l1.loc_end (Cmts.remaining_before c.cmts l2)
   | `Compact -> false
 
+(* Breaking before an attribute can confuse ocp-indent that will produce a
+   suboptimal indentation. *)
+let pre_attributes c = fmt_or c.conf.ocp_indent_compat " " "@ "
```
jberdine, line 546: I'm sorry for not thinking of this suggestion before, but I am inclined to try a different implementation, where instead of adding the `pre_attributes` function, `fmt_attributes` is changed to test if `pre` is "@ " and if so change it to " ". That will more tightly localize the mess made by ocp-indent compatibility.

test/passing/attributes.ml
CraigFe: I think this test still wants either modified or removed. I introduced it as part of https://github.com/ocaml-ppx/ocamlformat/pull/1306 to guard against a bug where the attribute was split across multiple lines on wrap.

lib/Fmt_ast.ml
gpetiot: Good point, it looks better indeed. No regression with test_branch.

test/passing/attributes.ml
jberdine: Ok. I don't have a strong position either way. @gpetiot feel free to include a change to this test in this PR as you see fit.

Issue #1323, opened 2020-03-30
Bug: comments inside attributes are misplaced


CraigFe:
**Describe the bug**

The following example,

```ocaml
type foo = Alpha | Beta [@@ocaml.warning "-37" (* Explanation of warning *)]

type foo =
  | Alpha______________________________
  | Beta_______________________________ 
  [@@ocaml.warning "-37" (* Explanation of warning *)]
```

gets OCamlformat-ted to:

```ocaml
type foo = Alpha | Beta [@@ocaml.warning "-37"] (* Explanation of warning *)

type foo =
  | Alpha______________________________
  | Beta_______________________________
[@@ocaml.warning "-37"]

(* Explanation of warning *)
```

The comments have been placed outside the attribute. IMO, this is 'fine but not ideal' in the first case and a bug in the second.

PR #1383, opened 2020-05-28
Fix placement of comments on constants

gpetiot:
Related to #1323 but insufficient to fix it somehow. No diff with test_branch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,8 @@
 
   + Remove unnecessary parens around object (#1379) (Guillaume Petiot)
 
+  + Fix placement of comments on constants (#1383) (Guillaume Petiot)
+
 ### 0.14.2 (2020-05-11)
 
 #### Changes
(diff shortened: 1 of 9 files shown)


gpetiot:
I'm waiting a bit before merging it, as #1387 can make it obsolete (having the loc in the ast instead of inferring them)

hhugo:
Note that the loc in the ast will only be valid with ocaml >= 4.10.

hhugo:
I wonder if we should lex the full source once when calling Source.create instead of lexing bits and pieces multiple times..

emillon:
FYI, in the read/format/output part, the profile is about:

- parsing: 13%
- comment handling (Cmts.init): 33%
- formatting + box layout: 46%

And calls to `Source` functions that reparse the file are barely visible.

jberdine:
Interesting, thanks @emillon! The interval tree data structure was a quick and slow implementation, and it shows. Regarding the effect that queries to `Source` have though, I think that the story is more involved. In particular, since we have to keep reformatting in a loop until a fixed point is reached, the cost of a use of Source is not immediately clear. In particular, if it can lead to another entire iteration (or more) of the format-parse loop, then the cost is much higher than what raw `perf` data would indicate. It is this sort of cost that I am more worried about, rather than the efficiency of the token parsing itself, which is not free but also not bad. Basically, the uses of Source are the point where we introduce backtracking that `Format` does not natively support, which is what takes the algorithm from polynomial time (`Format`) to exponential time (like pretty and friends).

PR #1387, opened 2020-06-03
Use 4.11 AST

kit-ty-kate:
This PR should add support for OCaml 4.11 to ocamlformat.
At the moment the tests fail during testing (`reformat_string.ml` gets its comments removed), however it seems to be fixed by https://github.com/ocaml-ppx/ocamlformat/pull/1383.

I'm not sure how critical the location fields are to ocamlformat so I'll let you decide whether the `loc` and `Location.none` are correctly placed here.

Concerning the deprecation of `Longident.parse` I can try to use `Parse.longident` instead, however the semantics are different between the two functions, also I'm not sure how to get it without accessing to `compiler-libs` directly as `ocaml-migrate-parsetree` does not have the function.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -59,7 +59,8 @@ end = struct
         , PStr
             [ { pstr_desc=
                   Pstr_eval
-                    ( { pexp_desc= Pexp_constant (Pconst_string (doc, None))
+                    ( { pexp_desc=
+                          Pexp_constant (Pconst_string (doc, _, None))
                       ; pexp_attributes
                       ; _ }
                     , [] )
@@ -73,7 +74,8 @@ end = struct
                    [ { pstr_desc=
                          Pstr_eval
                            ( { pexp_desc=
-                                 Pexp_constant (Pconst_string (doc, None))
+                                 Pexp_constant
+                                   (Pconst_string (doc, Location.none, None))
                              ; pexp_loc= Location.none
                              ; pexp_attributes=
                                  m.attributes m pexp_attributes
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -296,9 +296,9 @@ let fmt_constant c ~loc ?epi const =
   | Pconst_integer (lit, suf) | Pconst_float (lit, suf) ->
       str lit $ opt suf char
   | Pconst_char x -> wrap "'" "'" @@ str (char_escaped ~loc c x)
-  | Pconst_string (s, Some delim) ->
+  | Pconst_string (s, _, Some delim) ->
       wrap_k (str ("{" ^ delim ^ "|")) (str ("|" ^ delim ^ "}")) (str s)
-  | Pconst_string (s, None) -> (
+  | Pconst_string (s, _, None) -> (
       let delim = ["@,"; "@;"] in
       let contains_pp_commands s =
         let is_substring substring = String.is_substring s ~substring in
@@ -549,7 +549,7 @@ let rec fmt_extension c ctx key (ext, pld) =
     , PStr
         [ { pstr_desc=
               Pstr_eval
-                ({pexp_desc= Pexp_constant (Pconst_string (s, _)); _}, [])
+                ({pexp_desc= Pexp_constant (Pconst_string (s, _, _)); _}, [])
           ; _ } ]
     , _ ) ->
       str s
@@ -587,7 +587,7 @@ and fmt_attributes c ?pre ?suf ~key attrs =
       , PStr
           [ { pstr_desc=
                 Pstr_eval
-                  ( { pexp_desc= Pexp_constant (Pconst_string (doc, None))
+                  ( { pexp_desc= Pexp_constant (Pconst_string (doc, _, None))
                     ; pexp_attributes= []
                     ; _ }
                   , [] )
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -22,7 +22,7 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
-  "ocaml" {>= "4.06" & < "4.11"}
+  "ocaml" {>= "4.06" & < "4.12"}
   "alcotest" {with-test}
   "base" {>= "v0.12.0" & < "v0.14"}
   "base-unix"
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -31,7 +31,7 @@ depends: [
   "fix"
   "fpath"
   "menhir"
-  "ocaml-migrate-parsetree" {>= "1.5.0"}
+  "ocaml-migrate-parsetree" {>= "1.7.3"}
   "ocp-indent" {with-test}
   "odoc" {>= "1.4.2"}
   "re"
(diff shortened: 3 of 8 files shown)


lib/Source.ml
```diff
@@ -231,15 +231,15 @@ let string_literal t mode (l : Location.t) =
...
         | Parser.LBRACKETAT | Parser.LBRACKETATAT | Parser.LBRACKETATATAT ->
             true
         | _ -> false )
       l
   in
   match toks with
-  | [(Parser.STRING (_, None), loc)]
-   |(Parser.STRING (_, None), loc)
+  | [(Parser.STRING (_, _, None), loc)]
+   |(Parser.STRING (_, _, None), loc)
```
kit-ty-kate, line 242: It seems like these datatypes are taken from `compiler-libs` and not from `ocaml-migrate-parsetree`. Is there any chance we could use the parsetree instead?

Otherwise, would it make sense to simply allow `Parser.STRING _`? (i.e. `"str"`, `{|str|}`, `{delim|str|delim}`, … and not just `"str"`)

emillon: That's a good point. OMP does not embed the different parsers, and when we reparse bits of the input like here, it's always using the parser from compiler libs. We've talked in the past of packaging a fixed recent version of the parser instead of migrating the AST from the compiler-libs version, that's another point in favor of this.

emillon: as discussed with @kit-ty-kate, this function is used to determine the string's precise loc, which is now available in the 4.11 AST, so it might be possible to get rid of most of it!

lib/Cmts.ml
```diff
@@ -73,7 +73,7 @@ end = struct
                    [ { pstr_desc=
                          Pstr_eval
                            ( { pexp_desc=
-                                 Pexp_constant (Pconst_string (doc, None))
+                                 Pexp_constant (Pconst_string (doc, Location.none, None))
```
emillon, line 76: this is already the `pexp_loc` below and the `attr_loc` above so I think this is the right solution

lib/Fmt_ast.ml
```diff
@@ -996,11 +996,11 @@ and fmt_pattern c ?pro ?parens ({ctx= ctx0; ast= pat} as xpat) =
           @@
           match ppat_desc with
           | Ppat_var {txt= txt'; _}
-            when field_alias ~field:lid1.txt (Longident.parse txt')
+            when field_alias ~field:lid1.txt (Longident.parse txt' [@ocaml.warning "-3"])
```
emillon, line 999: Can you replace the `Longindent.parse` calls to a new function `Migrate_ast.Parse.longident`? that way there's only one place where we suppress the warning. You can try to replace it with `Parse.longindent` (the one from compiler-libs) or `unflatten` after splitting with `String.split ~on:"."` but that can be done in a separate PR, after or before this one.

ocamlformat.opam
```diff
@@ -22,7 +22,7 @@ build: [
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 depends: [
-  "ocaml" {>= "4.06" & < "4.11"}
+  "ocaml" {>= "4.06" & < "4.12"}
```
gpetiot, line 25: we also need `ocaml-migrate-parsetree` >= 1.7.3

jberdine:
It would be good to check that `Cmts.Loc_tree.of_ast` is implemented in such a way that the newly added locations in the 4.11 parsetree make into the location tree. Maybe nothing needs to be done, but I'm not sure and it would be good to check.

lib/Fmt_ast.ml
```diff
@@ -294,9 +294,9 @@ let fmt_constant c ~loc ?epi const =
   | Pconst_integer (lit, suf) | Pconst_float (lit, suf) ->
       str lit $ opt suf char
   | Pconst_char x -> wrap "'" "'" @@ str (char_escaped ~loc c x)
-  | Pconst_string (s, Some delim) ->
+  | Pconst_string (s, _, Some delim) ->
       wrap_k (str ("{" ^ delim ^ "|")) (str ("|" ^ delim ^ "}")) (str s)
-  | Pconst_string (s, None) -> (
+  | Pconst_string (s, _, None) -> (
```
hhugo, line 299: One need to look for comments using the currently discarded locations

lib/Source.ml
emillon: I opened #1392 for this part

emillon:
FTR, I suggested that the new location added to Pconst_string could replace the `tokens_at` part in `Source.string_literal`. That's true, but with a twist: the location in the token refers to the string with delimiters (always double quotes, since delimited strings are raw), while the location in the AST only refers to the payload. So if we want to use this location (which requires a 4.11 parser), we'll need to adjust the literal lexer to skip the enclosing quotes.

emillon:
About comments attached to the new location: my understanding is that there is always a strict inclusion between the two locs:

Consider a string literal as an expression, let's destructure it as `{pexp_loc = loc_expr; pexp_descr = Pconst_string (_, loc_const, _)}`.

The locs look like this:

### double quoted string

    "hello"
     ^^^^^  loc_const
    ^^^^^^^ loc_expr

### raw string 

    {delim|hello|delim}
           ^^^^^        loc_const
    ^^^^^^^^^^^^^^^^^^^ loc_expr

### quoted extension

    {%f|hello|}
        ^^^^^   loc_const
    ^^^^^^^^^^^ loc_expr

### quoted extension with delimiter


    {%f delim|hello|delim}
              ^^^^^        loc_const
    ^^^^^^^^^^^^^^^^^^^^^^ loc_expr

There is no lexical way that I know of to put comments within `loc_expr` (in the `%f delim` case, it's possible to insert whitespace, but no comments). I'm not super familiar with the comment attachment algorithm, but I think that means that nothing will get attached to `loc_const`. So an assertion that the loc tree is empty at the end of formatting (which is already in place?) should be enough.

Does that seem correct @hhugo @gpetiot @jberdine ?

jberdine:
I expect that by and large the comment placement behavior won't be affected due to the strict nesting. However, uses of `Cmts.is_adjacent` could care. It would be good to check those and see whether that will matter, and decide whether it is better to adjust `is_adjacent` to take delimiters into account, or to ensure that the string locations include them.

emillon:
I think this should be OK, we can refine that if necessary when we add tests for 4.11 features. So this seems ready to get merged in my eyes. What do you think @gpetiot ?

emillon:
(we'll add changelog entries for explicit 4.11 features)

gpetiot:
Is there a way to integrate #1383 in this ? i.e. use the loc of constants from the ast for ocaml >= 4.11 and use the functions of #1383 for ocaml < 4.11 ?

emillon:
I don't think so. These new locs are not the location of the constants themselves, so no comments should get attached to them. The only place we can leverage the new locs is for re-lexing the payload.

jberdine:
> These new locs are not the location of the constants themselves

I'm curious: what are they the loc of then?

emillon:
They are the loc for string contents (within delimiters), highlighted as `loc_const` in https://github.com/ocaml-ppx/ocamlformat/pull/1387#issuecomment-649406257.

PR #1392, opened 2020-06-17
Build on OCaml 4.11+alpha3

emillon:
This PR makes ocamlformat build on OCaml 4.11+alpha3. This is independent from changing the base OMP version as done in #1387. To test this, one needs to pin odoc to a fairly recent version, which changes the output of some tests. This can be fixed independently.

The main change here is that when reparsing source code, instead of using the `Parser.token` type for `compiler-libs`, we use an embedded copy of the 4.11 type. When using a previous version of ocaml, an upwards migration is performed (here, `Location.none` is added). This is less invasive (and better from a licensing perspective) than vendoring the whole parser from 4.11.

Let me know what you think!


emillon:
There's no direct support for 4.11 in ocaml-ci for now (see ocurrent/ocaml-ci#195), but in the meantime this can be tested with the following setup:

Setup global switches;

```
opam switch create ocamlformat-4.09.1 4.09.1 
opam switch create ocamlformat-4.10.0 ocaml-base-compiler.4.10.0      
opam switch create ocamlformat-4.11.0-alpha2 4.11.0+alpha2 --repositories=default,beta
```

Create the following dune-workspace.dev file:
```
(lang dune 1.0)
(context (opam (switch ocamlformat-4.09.1)))
(context (opam (switch ocamlformat-4.10.0)))
(context (opam (switch ocamlformat-4.11.0-alpha2)))
```

Install dependencies in switches (`opam install --switch ocamlformat-$v $packages`)

And one can run tests on all versions at once using `dune runtest --workspace dune-workspace.dev`.

jberdine:
Do you see why I keep saying that dune-workspace files and the scripts to set up the corresponding switches should be in the repo?

lib/token-latest/lt_411/token_latest.ml
```diff
@@ -0,0 +1,249 @@
...
+  | MUTABLE -> MUTABLE
+  | MODULE -> MODULE
+  | MINUSGREATER -> MINUSGREATER
+  | MINUSDOT -> MINUSDOT
+  | MINUS -> MINUS
+  | METHOD -> METHOD
+  | MATCH -> MATCH
+  | LPAREN -> LPAREN
+  | LIDENT x -> LIDENT x
+  | LETOP x -> LETOP x
```
gpetiot, line 176: There is no `LETOP` before 4.11

lib/token-latest/lt_411/token_latest.ml
```diff
@@ -0,0 +1,249 @@
...
+  | CHAR x -> CHAR x
+  | BEGIN -> BEGIN
+  | BARRBRACKET -> BARRBRACKET
+  | BARBAR -> BARBAR
+  | BAR -> BAR
+  | BANG -> BANG
+  | BACKQUOTE -> BACKQUOTE
+  | ASSERT -> ASSERT
+  | AS -> AS
+  | ANDOP x -> ANDOP x
```
gpetiot, line 246: There is no `ANDOP` before 4.08

emillon: ah, right. I'll fix this down to 4.06

emillon: done!

jberdine:
@emillon, The token-latest library is vendored in right? Could it be moved into the `vendor` directory to keep the copyright and license distinction clear?

emillon:
I put it in its own library because it's cleaner that way (like we do for `import`), but it could as well be a single module as part of the main `ocamlformat_lib` library (it has no licensing concerns). I wouldn't call it vendored, but if you prefer it that way, it works.

jberdine:
Ok thanks. I don't have any objection to it being in `lib` except that the code outside `vendor` all has the uniform copyright and license header. If it is appropriate to put that header on the `token-last` files, then by all means, put it under `lib` and run `make headers`.

PR #1396, opened 2020-06-25
Define a vendored utils lib to be compatible with base.v0.14.0

gpetiot:


jberdine:
I realize that this is a small patch, but I have to say that I don't think that the basic approach is a good long-term solution. In particular, I do not think that we should complicate the code (other than a clearly designated compatibility module/library). Things like needing to remember which `List` functions need to be prefixed with `Utils.` is really not great. It is also not great to end up reimplementing everything that changes across versions of dependencies. And once there are incompatibilities in type definitions, things get much worse. Instead, I think that there are two options:

1. Admit that current versions of ocamlformat need current versions of dependencies.

2. Define a compatibility shim library/module, `Compat` say, that is opened in all source files by adding `-open Compat` to the build flags. (Or just make `Compat` part of `Import`.) In that module we can override the items that are version-sensitive. The hope would be to use conditional compilation here, to avoid reimplementing things that change in dependencies. We should be implementing the old interfaces in terms of the new ones, not the other way around. We want a situation where from the perspective of the users of the `Compat` module, they see the current dependency interfaces and get the actual current implementations. In cases where older dependencies are used, non-standard and non-passthrough implementations can be used. We need to have this set up in such a way that when there are problems, we can just drop old version support by deleting some code. We do not want to be in the situation where we need to reimplement things to drop old version support. At a high level, the current version's interface and implementation should be primary, and the compatibility shims should only adapt to the older versions, not the other way around.

vendor/utils/list.mli
```diff
@@ -0,0 +1,7 @@
+val partition_map :
+  'a list -> f:('a -> [`Fst of 'b | `Snd of 'c]) -> 'b list * 'c list
```
emillon, line 2: This is using the base 0.13 API. What do you think about using the new API (based on `Either`) and updating the call sites so that this function is easier to delete when we move the support window?

gpetiot: done

vbmithr:
Why not avoid depending on base/stdio?

gpetiot:
Because we already are deeply depending on this (kinda too late), removing it would require a huge amount of work that would better be directed towards fixing bugs and improving the integration of ocamlformat with the rest of the ocaml tools.

PR #1405, opened 2020-07-01
Support quoted extensions (since ocaml 4.11)

### LINKED ISSUE #1401: 4.11 support: quoted extensions
Quoted extensions are lighter syntax for delimited strings in extension points: `{%foo|...|}` is lighter syntax for `[%foo {||}]`, and `{%foo bar|...|bar}` for `[%foo {bar|...|bar}]`.

In order not to output code that's invalid on 4.10, I suggest that we output the sugar version only when it was already present in the input source.

Relevant ocaml PRs: ocaml/ocaml#8820, ocaml/ocaml#9166.

gpetiot:
Fix #1401


lib/Fmt_ast.ml
```diff
@@ -541,6 +541,14 @@ let sequence_blank_line c (l1 : Location.t) (l2 : Location.t) =
       loop l1.loc_end (Cmts.remaining_before c.cmts l2)
   | `Compact -> false
 
+let fmt_quoted_string c key ext s strloc delim =
+  let delim = Option.value delim ~default:"" in
+  Cmts.fmt c strloc
```
emillon, line 546: Are there some comments that get attached here?

test/passing/quoted_strings.ml
```diff
@@ -0,0 +1,65 @@
...
+
+let foo = ({%foo | foooooooooooooo |} (* A *) [@attr]) (* B *) [@@attr]
+
+let foo = (* A *) ({%foo | foooooooooooooo |} [@attr]) (* B *) [@@attr]
+
+let foo = (* A *) {%foo sep| foooooooooooooo |sep} (* B *) [@@attr]
+
+{%%foo | foooooooooooooo |}
+
+{%%foo | foooooooooooooo | (* A *) } (* B *)
```
emillon, line 30: Ah! I'm surprised this parses. What does A get attached to? the payload?

gpetiot: Oh indeed, I just realized that the string is not closed at this line, it's closed at the `|}` in the next item, so the comments are not allowed here, I will clean up the tests.

lib/Fmt_ast.ml
gpetiot: It looks like there is none, I simplified the code.

lib/Fmt_ast.ml
```diff
@@ -557,6 +564,26 @@ let rec fmt_extension c ctx key (ext, pld) =
...
+                  ; pexp_loc
+                  ; pexp_loc_stack= _
+                  ; pexp_attributes }
+                , attrs )
+          ; pstr_loc } ]
+    , _ )
+    when Source.is_quoted_string c.source pstr_loc ->
+      let doc, atrs = doc_atrs attrs in
+      fmt_docstring c doc
+      $ Cmts.fmt c pstr_loc @@ hvbox 0 @@ Cmts.fmt c pexp_loc @@ hvbox 0
```
Julow, line 583: None of the tests are triggering this code but are instead testing attributes and comments around expressions and structure items.
I suggest to remove the tests with attributes and comments (or keep a single one) to avoid confusion.

Looking at the parser I couldn't find any input that would add attributes or comments here. This may be added in the future but this code would most probably move them (which is a fatal error) because it clashes with the syntax for comments/attributes around expressions.
I suggest to remove this code and to add an assert that there is no comments (necessary to avoid the warning "Some locations have not been considered").

CHANGES.md
```diff
@@ -38,6 +38,10 @@
 
   + Do not add a space to minimal comments `(* *)`, `(** *)` and `(*$ *)` (#1407) (Guillaume Petiot)
 
+#### New features
+
+  + Support quoted extensions, new ocaml 4.11 feature (#1405) (Guillaume Petiot)
```
Julow, line 43: Shouldn't this be "a new ..." ?

lib/Fmt_ast.ml
gpetiot: That's a good point, I added the assertions. (please nobody suggest the compiler team to allow this in the parser)
I cleared up some lines in the tests, but I like to keep a good number with attributes and comments, to have them updated in case we change something else (box, attrs, comment attachment algorithm)

test/passing/quoted_strings.ml
```diff
@@ -0,0 +1,54 @@
+let foo = {%foo | foooooooooooooo |}
+
+let foo = (* A *) ({%foo | foooooooooooooo |} (* B *) [@attr]) (* C *)
+
+let foo = (* A *) {%foo sep| foooooooooooooo |sep}
+
+let foo = {%foo | foooooooooooooo |} [@@attr]
```
emillon, line 7: I think it's expected that there's no space after the extension, like `{%foo|payload|}`. That's how they are described in the PR and in the manual, and that's consistent with delimiters.

lib/Fmt_ast.ml
Julow: More tests don't harm.

test/passing/quoted_strings.ml
gpetiot: Oh yeah that's right, it's fixed now.

lib/Fmt_ast.ml
```diff
@@ -557,6 +566,26 @@ let rec fmt_extension c ctx key (ext, pld) =
...
     ->
       fmt_signature_item c ~ext (sub_sig ~ctx si)
+  (* Quoted extensions (since ocaml 4.11) {%ext delim|...|delim}. Comments
+     and attributes are not allowed by the parser. *)
+  | ( ("%" | "%%")
+    , ext
+    , PStr
+        [ { pstr_desc=
+              Pstr_eval
+                ( { pexp_desc= Pexp_constant (Pconst_string (str, _, delim))
```
Julow, line 576: The second argument of `Pconst_string` is a location that has to be considered for comments (with an assert like the others)

lib/Fmt_ast.ml
```diff
@@ -537,6 +537,15 @@ let sequence_blank_line c (l1 : Location.t) (l2 : Location.t) =
       loop l1.loc_end (Cmts.remaining_before c.cmts l2)
   | `Compact -> false
 
+let fmt_quoted_string key ext s = function
+  | None | Some "" ->
```
Julow, line 541: Is `Some ""` possible ?

gpetiot: Looks like it's not possible indeed, I will simplify

lib/Fmt_ast.ml

PR #1407, opened 2020-07-01
Do not add a space to minimal comments

gpetiot:
Minimal comments as `(* *)`, `(** *)` or `(*$ *)` (containing only 1 space) were added another space after being formatted.

### DIFF
--- a/lib/Fmt.ml
+++ b/lib/Fmt.ml
@@ -268,7 +268,7 @@ let fill_text ?epi text =
       ~equal:(fun x y -> String.is_empty x && String.is_empty y)
       (String.split (String.rstrip text) ~on:'\n')
   in
-  fmt_if (Char.is_whitespace text.[0]) " "
+  fmt_if (String.starts_with_whitespace text) " "
   $ hovbox 0
       ( hvbox 0
           (hovbox 0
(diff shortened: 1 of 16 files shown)


lib/Fmt.ml
```diff
@@ -280,5 +280,8 @@ let fill_text ?epi text =
                       close_box $ fmt "\n@," $ open_hovbox 0
                   | Some _ when not (String.is_empty curr) -> fmt "@ "
                   | _ -> noop )))
-      $ fmt_if (Char.is_whitespace text.[String.length text - 1]) " "
+      $ fmt_if
```
emillon, line 283: What do you think about creating `String.starts_with_whitespace` and `String.ends_with_whitespace` ? that would replace the `space_i` function in `Fmt_ast` as well. This can go in `Import.String` for example.

gpetiot: Good idea, done.

vendor/compat/string.ml
```diff
@@ -0,0 +1,10 @@
+open Base
+
+let starts_with_whitespace s =
+  (not (String.is_empty s)) && Char.is_whitespace s.[0]
+
+let ends_with_whitespace s =
+  let len = String.length s in
+  len > 1 && Char.is_whitespace s.[len - 1]
```
emillon, line 8: The asymmetry between the two functions is a bit surprising. In particular, this means that `ends_with_whitespace " " = false`.

Would it work to define `ends_with_whitespace` as "not empty and last char is whitespace" (like `starts_with_whitespace`) and move the `len > 1` check to call sites?

gpetiot: Yes we can, I wanted to remove the `len > 1` test from the callsites but I can put them back.

gpetiot: done

PR #1412, opened 2020-07-08
Start re-organizing Ast utility functions

gpetiot:
More PRs like this are coming (if this one is accepted)

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -200,10 +200,7 @@ module Longident = struct
 
   let is_hash_getter = test ~f:String_id.is_hash_getter
 
-  let is_index_op i =
-    match List.last (Longident.flatten i) with
-    | Some ident -> String_id.is_index_op ident
-    | None -> false
+  let is_index_op i = Longident.last i |> String_id.is_index_op
 
   let is_symbol i = is_prefix i || is_infix i || is_index_op i
 end
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -34,9 +34,7 @@ let ( init
 
 (** Predicates recognizing special symbol identifiers. *)
 
-module Char = struct
-  include Char
-
+module Char_id = struct
   let is_kwdop = function
     | '$' | '&' | '*' | '+' | '-' | '/' | '<' | '=' | '>' | '@' | '^' | '|'
      |'!' | '%' | ':' | '?' ->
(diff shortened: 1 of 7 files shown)


lib/Ast.ml
```diff
@@ -194,19 +157,74 @@ module Indexing_op = struct
...
+  let is_monadic_binding = test ~f:String_id.is_monadic_binding
+
+  let is_infix = test ~f:String_id.is_infix
 
-let is_symbol_id i = is_prefix_id i || is_infix_id i || is_index_op_id i
+  let is_hash_getter = test ~f:String_id.is_hash_getter
 
-let is_symbol e = is_prefix e || is_infix e || is_index_op e
+  let is_index_op i =
+    match List.last (Longident.flatten i) with
```
Julow, line 206: There is `Longident.last`, which has the advantage of not having the `None` case.

lib/Ast.ml
```diff
@@ -34,58 +34,21 @@ let ( init
 
 (** Predicates recognizing special symbol identifiers. *)
 
-let is_prefix_id i =
-  match i with
-  | "!=" -> false
-  | _ -> ( match i.[0] with '!' | '?' | '~' -> true | _ -> false )
+module Char = struct
+  include Char
```
Julow, line 38: I'm not a fan of "improving" modules like this, especially if the added functions wouldn't fit in the original module.
What do you think of naming it `Char_id`, like you did with `String_id` ?
I think the same for `Longident` and `Exp` (it would be the only module without `_id`) below.

gpetiot: I agree for `Char`; for `Longident` and `Exp` I aim to squeeze in the other related functions defined in this file (is_simple, etc.) and merge it with `Expression`, so they won't only contain id functions hopefully. That's the original plan but if it ends up being too messy I might do what you suggest

lib/Fmt_ast.ml
```diff
@@ -3288,7 +3289,7 @@ and fmt_constructor_declaration c ctx ~max_len_name ~first ~last:_ cstr_decl
       | Pcstr_tuple x -> List.is_empty x
       | Pcstr_record x -> List.is_empty x
     in
-    let len_around = if is_symbol_id txt then 4 else 0 in
+    let len_around = if String_id.is_symbol txt then 4 else 0 in
```
emillon, line 3292: It's already useful to organize `Ast` into better-scoped modules, thanks for this. I agree with @Julow is that it's a good occasion to add some tests if you can.

If you want to try to take it a step further (not in this PR!), I noticed that this API is returning a lot of booleans. In the spirit of tell-dont-ask, you can try to build an abstraction that encapsulates the boolean logic. In this example, something like: `String_id.of_name txt |> String_id.padding ~max_len_name`. If this goes well, `is_symbol` might not even need to be exposed. This kind of things. There's a balance to be found, it can be tricky.

gpetiot: That's a good idea, I will see better what we can do with it when the file has been cleaned up (more like purified)

lib/Ast.ml
gpetiot: done

PR #1413, opened 2020-07-08
Re-organizing Ast utility functions

gpetiot:
Follow-up of #1412, each commit can be reviewed separately (not finished cleaning up the file but the diff is already long enough)

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -222,6 +222,12 @@ module Longident = struct
     fit_margin c (length x)
 end
 
+module Attr = struct
+  let is_doc = function
+    | {attr_name= {Location.txt= "ocaml.doc" | "ocaml.text"; _}; _} -> false
+    | _ -> true
+end
+
 module Exp = struct
   let test_id ~f = function
     | {pexp_desc= Pexp_ident {txt= i; _}; _} -> f i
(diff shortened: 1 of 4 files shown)


lib/Ast.ml
```diff
@@ -203,6 +210,25 @@ module Longident = struct
...
+      | Ldot (x, y) -> length x + 1 + String.length y
+      | Lapply (x, y) -> length x + length y + 3
+    in
+    fit_margin c (length x)
+end
+
+module Attr = struct
+  let is_doc = function
+    | {attr_name= {Location.txt= "ocaml.doc" | "ocaml.text"; _}; _} -> false
+    | _ -> true
```
Julow, line 231: You meant to name this `is_not_doc` ?

emillon: Unfortunately it's already present in the original version like this :/

I'd rather not have negatives in function names, so I suggest either naming it `is_normal_attribute` (`is_attribute` / `has_attribute_syntax` maybe?) or keeping `is_doc_comment`, inverting the booleans and fixing the call sites (`List.for_all ~f:is_doc_comment xxx_attributes`).

Sidenote, I remember extracting this function in `Cmts.ml` so you can remove it and convert it to use this one if applicable.

lib/Ast.ml
```diff
@@ -219,49 +245,112 @@ module Exp = struct
...
+            false
+        | Ok l ->
+            List.iter ~f:(fun e -> Hashtbl.set memo ~key:e ~data:true) l ;
+            true )
+
+  let is_simple (i, c) =
+    Poly.(c.Conf.module_item_spacing = `Compact)
+    && Location.is_single_line i.pexp_loc c.Conf.margin
+
+  let break_between _s ~cmts ~has_cmts_before ~has_cmts_after (i1, c1)
```
emillon, line 293: it might be worth checking if this unused argument can be removed?

gpetiot: good catch, done.

lib/Ast.ml
gpetiot: I can't believe we missed this one all this time, fixed thanks!

gpetiot:
As a followup I was thinking of moving the "ast in context" functions in a separate file In_ctx.ml (or Ctx.ml) where we can also group the functions in modules named after the ast types (Exp, Pat, Mod, Mty, etc). What do you think?

PR #1414, opened 2020-07-10
Preserve comment position around match keyword

### LINKED ISSUE #341: Preserve position of comments in `match (*cmt*) expr with`.

gpetiot:
Fix #341


lib/Fmt_ast.ml
```diff
@@ -2298,17 +2298,11 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       in
       match compact with
       | None ->
-          (* TODO: remove leading_cmt once we figure out how to not introduce
-             regression with 4.07 Without the line below {[ let () = ( (*
-             before *) match (* after *) x with _ -> x) ]} Gets reformatted
-             into {[ let () = match (* before *) (* after *) x with _ -> x ]} *)
-          let leading_cmt = Cmts.fmt_before c e0.pexp_loc in
```
Julow, line 2305: This is still a concern. On 4.06.1, I get this diff on the test added in this PR:
```diff
-let _ =
-  (* before match *)
-  match (* after match *) x with _ -> 1
+let _ = match (* before match *)
+              (* after match *) x with _ -> 1
```

gpetiot: I added a check on the compiler version, it's a bit too ad-hoc to my taste but we can think of a better design if the problem arises again somewhere else

jberdine:
There is a change that I don't think is intended:
```diff
modified   sledge/bin/frontend.ml
@@ -353,7 +353,8 @@ let should_inline : Llvm.llvalue -> bool =
           | FPTrunc | FPExt | PtrToInt | IntToPtr | BitCast | AddrSpaceCast
             ) ->
           true (* inline casts *)
-      | _ -> false (* do not inline if >= 2 uses *) )
+      | _ -> false
+      (* do not inline if >= 2 uses *) )
     | None -> true )
   | None -> true
```
This can be worked around, but seems to be orthogonal to the goal of this PR. Do you know if it is an inescapable ripple effect of the change?

gpetiot:
I think it is due to the location from loc_stack added to the global list of locations (maybe taking the first one is wrong), I will investigate that next week.

gpetiot:
Wrong approach, closing it.

hhugo:
to discriminate comments before and after the `match` keyword, I think you would need to take the last location of loc_stack, not the first one

gpetiot:
I tried this, along with other things, but there is always some regression. I'm looking for the root of the issue, since it seems the comments locations are correct, I don't know where it goes wrong.

PR #1419, opened 2020-07-17
Fix break between keyword and expr with if-then-else=keyword-first

### LINKED ISSUE #1417: Bug: if expression with extension node and sufficiently long comment in condition does not stabilize
**Describe the bug**
I have a small test case that results in the error message "BUG: formatting did not stabilize after 10 iterations."

**How to Reproduce**

On a file with the following contents:

```
let foo () =
  if%bind
    (* this is a medium length comment of some sort *)
    this is a medium length expression of_some sort
  then x
  else y
;;
```

Run:

ocamlformat --profile=janestreet --enable-outside-detected-project --disable-conf-files the_file.ml

gpetiot:
Fix #1417

Not breaking used to cause excessive indentation:
```diff
diff --git a/infer/src/base/FileDiff.ml b/infer/src/base/FileDiff.ml
index c39d9b75c..147b5fddc 100644
--- a/infer/src/base/FileDiff.ml
+++ b/infer/src/base/FileDiff.ml
@@ -67,8 +67,9 @@ let parse_directives directives =
   if List.is_empty directives
   then (* handle the case where both files are empty *)
     []
-  else if (* handle the case where the new-file is empty *)
-          List.for_all ~f:(UnixDiff.equal UnixDiff.Old) directives
+  else if
+    (* handle the case where the new-file is empty *)
+    List.for_all ~f:(UnixDiff.equal UnixDiff.Old) directives
```

Some cases are debatable:
```diff
diff --git a/infer/src/IR/ProcAttributes.ml b/infer/src/IR/ProcAttributes.ml
index e7e999c0a..452156773 100644
--- a/infer/src/IR/ProcAttributes.ml
+++ b/infer/src/IR/ProcAttributes.ml
@@ -202,10 +202,11 @@ let pp
     translation_unit;
   if not (PredSymb.equal_access default.access access)
   then F.fprintf f "; access= %a@," (Pp.of_string ~f:PredSymb.string_of_access)
 access;
-  if not
-       ([%compare.equal: (Mangled.t * Typ.t * Pvar.capture_mode) list]
-          default.captured
-          captured)
+  if
+    not
+      ([%compare.equal: (Mangled.t * Typ.t * Pvar.capture_mode) list]
+         default.captured
+         captured)
```

but we can't have have specific treatments without making it even more difficult to maintain.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 #### Bug fixes
 
+  + Fix break between keyword and expr with `if-then-else=keyword-first` (#1419, @gpetiot)
+
 #### New features
 
 ### 0.15.0 (2020-08-06)
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -369,7 +369,7 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
                 ( fmt_or_k first
                     (str "if" $ fmt_extension_suffix)
                     (str "else if")
-                $ fmt_attributes $ str " " $ fmt_cond xcnd )
+                $ fmt_attributes $ fmt "@ " $ fmt_cond xcnd )
               $ fmt "@ ")
       ; box_keyword_and_expr=
           (fun k ->
--- a/test/passing/ite-kw_first.ml.ref
+++ b/test/passing/ite-kw_first.ml.ref
@@ -45,8 +45,9 @@ f
 
 ;;
 f
-  ( if and_ even
-         loooooooooooooooooooooooooooooooooooooooooooooooooooooooooonger
+  ( if
+      and_ even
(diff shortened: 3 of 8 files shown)


hhugo:
@ceastlund, just making sure this is a behavior we are happy with.
That is, newline after `if`, if the condition doesn't hold in one line.

Before:
```ocaml
if long_function_call 
     long_argument
then ..
else ..
```
After:
```ocaml
if 
  long_function_call 
     long_argument
then ..
else ..
```

ceastlund:
@hhugo, it looks like this solves enough issues, e.g. indentation after `else if`, that it's not worth pushing back on it just due to style preference. It's the sort of thing we'll adapt to easily enough, whether it's anyone's "favorite" behavior or not.

PR #1421, opened 2020-07-17
Fix alignment of 'then' keyword in parenthesised expression

### LINKED ISSUE #1415: Bug: misalignment of parenthesised `if ... then ... else` expressions
The following excerpt is OCamlFormat-ted:

```ocaml
let () =
  ( if a then
    b_____________________________________________________________________
  else c )
  |> f

let () =
  ( if a______________________________________________________________________
  then b )
  >>= fun () -> c

let () =
  [
    ( if a_______________________________________________ then
      b_____________________
    else c_________________________ );
  ]

let () =
  f
    ( if a___________________________________________________________________
    then b_________________________________________________________________
    else c_________________________________________________________________ )

[@@@ocamlformat "indicate-multiline-delimiters=no"]

let () =
  f
    (if a___________________________________________________________________
   then b_________________________________________________________________
    else c_________________________________________________________________)

```

In each case, the `if ... then ... else` keywords are not left-aligned, whereas in contexts with no surrounding parens they would be left-aligned.

I suspect this is a bug, but feel free to recategorise as a feature request if you think it's more appropriate :slightly_smiling_face:

gpetiot:
Fix #1415, no diff with test_branch.


lib/Params.ml
```diff
@@ -294,7 +294,9 @@ let get_if_then_else (c : Conf.t) ~first ~last ~parens ~parens_bch
     match xcond with
     | Some xcnd ->
         hvbox
-          (if parens then -2 else 0)
+          ( match imd with
+          | `No -> if parens then -1 else 0
```
emillon, line 298: What do you think about matching with parens to simplify the logic?

```
match parens, imd with
| false, _ -> 0
| true, `No -> -1
| true, ... -> -2
```

(extracting a local variable for level might improve this, too)

gpetiot: indeed, done! The whole function would need rewriting anyway, we need to simplify the formatting of if-then-else expressions and remove some of the supported styles!

PR #1422, opened 2020-07-17
Clean fit/break behavior

gpetiot:
Cleaning of the `Fmt.fits_breaks` api, no diff on the output. (no diff with test_branch)

### DIFF
--- a/lib/Fmt.mli
+++ b/lib/Fmt.mli
@@ -124,22 +124,17 @@ val break_unless_newline : int -> int -> t
 
 (** Conditional on breaking of enclosing box ----------------------------*)
 
+type behavior = Fit | Break
+
 val fits_breaks :
-     ?force_fit_if:bool
-  -> ?force_break_if:bool
-  -> ?hint:int * int
-  -> ?level:int
-  -> string
-  -> string
-  -> t
+  ?force:behavior -> ?hint:int * int -> ?level:int -> string -> string -> t
 (** [fits_breaks fits nspaces offset breaks] prints [fits] if the enclosing
     box fits on one line, and otherwise prints [breaks], which is a string
     that optionally follows a break [hint] (that is a pair
     [(nspaces, offset)] equivalent to the break hint ["@;<nspaces offset>"]). *)
 
 val fits_breaks_if :
-     ?force_fit_if:bool
-  -> ?force_break_if:bool
+     ?force:behavior
   -> ?hint:int * int
   -> ?level:int
   -> bool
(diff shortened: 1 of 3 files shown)


lib/Fmt_ast.ml
```diff
@@ -4239,9 +4246,8 @@ and fmt_structure_item c ~last:last_item ?ext {ctx; ast= si} =
             let epi =
               match c.conf.let_binding_spacing with
               | `Compact -> None
-              | `Sparse ->
-                  let force_fit_if = last && last_grp && last_item in
-                  Some (fits_breaks ~force_fit_if "" "\n")
+              | `Sparse when last && last_grp && last_item -> None
```
emillon, line 4249: with Sparse, if the condition is true, this used to return `Some (fits_breaks ~force_fit_if:false "" "\n")`, but now returns `None`. Is this equivalent? (just checking)

lib/Fmt.mli
```diff
@@ -124,22 +124,17 @@ val break_unless_newline : int -> int -> t
...
+
 val fits_breaks :
-     ?force_fit_if:bool
-  -> ?force_break_if:bool
-  -> ?hint:int * int
-  -> ?level:int
-  -> string
-  -> string
-  -> t
+  ?force:behavior -> ?hint:int * int -> ?level:int -> string -> string -> t
```
emillon, line 130: Just to check: is this common enough that `?force` is not passed? in other word, does it make sense to keep the parameter optional?

lib/Fmt.mli
```diff
@@ -124,22 +124,17 @@ val break_unless_newline : int -> int -> t
...
 (** [fits_breaks fits nspaces offset breaks] prints [fits] if the enclosing
     box fits on one line, and otherwise prints [breaks], which is a string
     that optionally follows a break [hint] (that is a pair
     [(nspaces, offset)] equivalent to the break hint ["@;<nspaces offset>"]). *)
 
 val fits_breaks_if :
-     ?force_fit_if:bool
-  -> ?force_break_if:bool
```
emillon, line 142: were there cases where both optional parameters were passed? I suppose not?

gpetiot: There was, but it didn't make sense and it is the purpose of this PR

lib/Fmt.mli
gpetiot: Yes most of the times we don't pass this parameter, I will try to reduce the usage of this parameter in the future whenever possible

lib/Fmt_ast.ml
gpetiot: When `last && last_grp & &last_itm` was true it just returned `Some (fmt "")`(since the force_fit mode means we just format the fits string which is empty) that was used for epi, so we can just have None as epi instead.

PR #1426, opened 2020-07-23
Add a github action to check changelog

emillon:
The semantics are the same as the Travis check:

- check pull requests against master
- if pull request has label "no-changelog-needed", do nothing
- display an error if there are no diffs on `CHANGES.md`

Compared to the Travis check:

- it retriggers automatically when a label is added or removed
- it will not be subject to API rate limiting
- it is likely to be a bit faster

A couple comments on the implementation:

- the types of pull request events correspond to the default ones plus the ones related to labels (so that it triggers a check when labels change)
- this uses `actions/checkout@v1` because there is a bug in `v2` that can check out the wrong commit sometimes. See actions/checkout#237.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,20 @@
+name: Changelog check
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+
+jobs:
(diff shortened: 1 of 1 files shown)


emillon:
Closing in favor of #1427 (we can't see the changes immediately if PR comes from a fork)

PR #1427, opened 2020-07-23
Add a github action to check changelog

emillon:
The semantics are the same as the Travis check:

- check pull requests against master
- if pull request has label "no-changelog-needed", do nothing
- display an error if there are no diffs on `CHANGES.md`

Compared to the Travis check:

- it retriggers automatically when a label is added or removed
- it will not be subject to API rate limiting
- it is likely to be a bit faster

A couple comments on the implementation:

- the types of pull request events correspond to the default ones plus the ones related to labels (so that it triggers a check when labels change)
- this uses `actions/checkout@v1` because there is a bug in `v2` that can check out the wrong commit sometimes. See actions/checkout#237.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,20 @@
+name: Changelog check
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: git diff
+      if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changelog-needed') }}
+      env:
+        BASE_REF: ${{ github.event.pull_request.base.ref }}
+      run: |
+        ! git diff --exit-code origin/$BASE_REF -- CHANGES.md


emillon:
What should we do for integration? should we give this a test ride for a couple weeks before disabling the travis check? Also, what about the travis 4.07 build? should we keep it?

gpetiot:
> What should we do for integration? should we give this a test ride for a couple weeks before disabling the travis check? Also, what about the travis 4.07 build? should we keep it?

Since the 4.07 build is already checked with ocaml-ci I think we can remove the travis check (and the changelog check as well).

.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
+name: Changelog check
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
```
Julow, line 10: Ubuntu 20 is a LTS

emillon: it's only in preview in github actions and `ubuntu-latest` is still 18.04, so that's the most stable option for now.

emillon:
Thanks for the feedback, I'll activate this one and prepare a PR to remove travis in a few weeks once we're sure this one works.

PR #1430, opened 2020-07-24
Recognise eliom file extensions

jrochel:


gpetiot:
Hi, thanks for the contribution. I've never used eliom so far, does anyone have more experience with that? (@emillon @Julow @jberdine ). I couldn't find any differences with the ocaml syntax from a few examples so far, I wonder if there is more to it.

hhugo:
I think the extension is just used to drive the build system. The syntax is the plain ocaml one, the build rules are however more complicated.

hhugo:
I suspect this change alone is enough. I would be nice to include some eliom  file in the testsuite thought. With real life examples, to see how code look in theses cases.

gpetiot:
I rebased, updated the changelog and added a test from the user manual.

jrochel:
@gpetiot Marvellous, thanks!

test/passing/eliom_ext.eliom
```diff
@@ -0,0 +1,34 @@
...
+  Eliom_client.onload
+    (* NB The service underlying the server_function isn't available on the
+       client before loading the page. *)
+    (fun () ->
+      Lwt.async (fun () -> log "Hello from the client to the server!") )
+
+[%%shared
+type some_type = int * string list [@@deriving json]
+
+type another_type = A of some_type | B of another_type [@@deriving json]]
```
hhugo, line 15: The trailing `]` is not very visible. Would one introduce a break before ?

gpetiot: I think we can think about making it affected by `indicate-multiline-delimiters` but not in this PR indeeed

PR #1431, opened 2020-07-24
Add missing break in module statement

### LINKED ISSUE #1410: Bug: poor indentation when breaking include of a functor application
```ocaml
include F
          (M)
          (struct
            type t

            val unit : unit
          end)
```
reformats to itself. The indentation starts from the beginning of the functor name `F`. Probably we should break after `include`.

gpetiot:
Fix #1410

Diff of test_branch:
```diff
diff --git a/infer/src/concurrency/starvationDomain.ml b/infer/src/concurrency/starvationDomain.ml
index 687749fa8..0d58643ce 100644
--- a/infer/src/concurrency/starvationDomain.ml
+++ b/infer/src/concurrency/starvationDomain.ml
@@ -371,9 +371,8 @@ let is_recursive_lock event tenv =
 ;;
 
 module CriticalPair = struct
-  include ExplicitTrace.MakeTraceElem
-            (CriticalPairElement)
-            (ExplicitTrace.DefaultCallPrinter)
+  include
+    ExplicitTrace.MakeTraceElem (CriticalPairElement) (ExplicitTrace.DefaultCallPrinter)
 
   let make ~loc acquisitions event thread = make { acquisitions; event; thread } loc
 
diff --git a/src/dune/action.ml b/src/dune/action.ml
index 8906f8ec..418ba68e 100644
--- a/src/dune/action.ml
+++ b/src/dune/action.ml
@@ -83,9 +83,10 @@ module For_shell = struct
 
   module rec Ast : Ast = Ast
 
-  include Action_ast.Make (String_with_sexp) (String_with_sexp) (String_with_sexp)
-            (String_with_sexp)
-            (Ast)
+  include
+    Action_ast.Make (String_with_sexp) (String_with_sexp) (String_with_sexp)
+      (String_with_sexp)
+      (Ast)
 end
 
 module Relativise = Action_mapper.Make (Ast) (For_shell.Ast)
diff --git a/src/dune/action_dune_lang.ml b/src/dune/action_dune_lang.ml
index 4c49004e..86fb00ec 100644
--- a/src/dune/action_dune_lang.ml
+++ b/src/dune/action_dune_lang.ml
@@ -20,9 +20,10 @@ module type Uast =
 
 module rec Uast : Uast = Uast
 
-include Action_ast.Make (String_with_vars) (String_with_vars) (String_with_vars)
-          (String_with_vars)
-          (Uast)
+include
+  Action_ast.Make (String_with_vars) (String_with_vars) (String_with_vars)
+    (String_with_vars)
+    (Uast)
 
 module Mapper = Action_mapper.Make (Uast) (Uast)
 
diff --git a/src/dune/action_unexpanded.ml b/src/dune/action_unexpanded.ml
index bd8841c6..0be7f927 100644
--- a/src/dune/action_unexpanded.ml
+++ b/src/dune/action_unexpanded.ml
@@ -451,28 +451,20 @@ end = struct
     end
   end
 
-  include Make (Action.Ast) (Sets) (Outcome)
-            (struct
-              let ( +@+ ) acc fn =
-                { acc with targets = Path.Build.Set.add acc.targets fn }
-              ;;
-
-              let ( +< ) acc fn = { acc with deps = Path.Set.add acc.deps fn }
-
-              let ( +<+ ) acc fn =
-                { acc with deps = Path.Set.add acc.deps (Path.build fn) }
-              ;;
-
-              let ( +<- ) acc fn =
-                { acc with targets = Path.Build.Set.remove acc.targets fn }
-              ;;
-
-              let ( +<! ) acc prog =
-                match prog with
-                | Ok p -> acc +< p
-                | Error _ -> acc
-              ;;
-            end)
+  include
+    Make (Action.Ast) (Sets) (Outcome)
+      (struct
+        let ( +@+ ) acc fn = { acc with targets = Path.Build.Set.add acc.targets fn }
+        let ( +< ) acc fn = { acc with deps = Path.Set.add acc.deps fn }
+        let ( +<+ ) acc fn = { acc with deps = Path.Set.add acc.deps (Path.build fn) }
+        let ( +<- ) acc fn = { acc with targets = Path.Build.Set.remove acc.targets fn }
+
+        let ( +<! ) acc prog =
+          match prog with
+          | Ok p -> acc +< p
+          | Error _ -> acc
+        ;;
+      end)
 
   module Partial_with_all_targets =
     Make (Partial.Past) (Sets) (Outcome)
diff --git a/src/dune/sub_system_name.ml b/src/dune/sub_system_name.ml
index f94344d8..ec37bb96 100644
--- a/src/dune/sub_system_name.ml
+++ b/src/dune/sub_system_name.ml
@@ -1,9 +1,10 @@
 open! Stdune
 
-include Interned.Make
-          (struct
-            let initial_size = 16
-            let resize_policy = Interned.Conservative
-            let order = Interned.Natural
-          end)
-          ()
+include
+  Interned.Make
+    (struct
+      let initial_size = 16
+      let resize_policy = Interned.Conservative
+      let order = Interned.Natural
+    end)
+    ()
diff --git a/src/dune/variant.ml b/src/dune/variant.ml
index c87d82f5..053c3c60 100644
--- a/src/dune/variant.ml
+++ b/src/dune/variant.ml
@@ -1,12 +1,13 @@
 open! Stdune
 
-include Interned.Make
-          (struct
-            let initial_size = 256
-            let resize_policy = Interned.Conservative
-            let order = Interned.Fast
-          end)
-          ()
+include
+  Interned.Make
+    (struct
+      let initial_size = 256
+      let resize_policy = Interned.Conservative
+      let order = Interned.Fast
+    end)
+    ()
```

### DIFF
--- a/test/passing/open-auto.ml.ref
+++ b/test/passing/open-auto.ml.ref
@@ -31,7 +31,7 @@ let _ =
   y
 
 let () =
-  ( Term.(
+  ( (let open Term in
     term_result
       ( const Phases.phase1 $ arch $ hub_id $ build_dir $ logs_dir
       $ setup_logs ))
--- a/test/passing/open-auto.ml.ref
+++ b/test/passing/open-auto.ml.ref
@@ -31,7 +31,7 @@ let _ =
   y
 
 let () =
-  ( (let open Term in
+  ( Term.(
     term_result
       ( const Phases.phase1 $ arch $ hub_id $ build_dir $ logs_dir
       $ setup_logs ))
@@ -176,16 +176,13 @@ class a =
 
 let _ =
   (* a *)
-  let (* b *)
-  open (* c *) struct
+  let (* b *) open (* c *) struct
     type t
   end
   (* d *) in
   (* e *)
   (* f *)
-  let (* g *)
-  open (* h *) A (* i *) (B)
-  (* j *) in
+  let (* g *) open (* h *) A (* i *) (B) (* j *) in
   (* k *)
   ()
 
@@ -204,19 +201,21 @@ open struct
   type t
 end
 
-open functor
-       (A : S)
-       ->
-       struct
-         type t
-       end
+open
+  functor
+    (A : S)
+    ->
+    struct
+      type t
+    end
 
-open functor
-       (_ : S)
-       ->
-       struct
-         type t
-       end
+open
+  functor
+    (_ : S)
+    ->
+    struct
+      type t
+    end
 
 open A (B)
 
@@ -232,18 +231,20 @@ let _ =
   let open struct
     type t
   end in
-  let open functor
-             (A : S)
-             ->
-             struct
-               type t
-             end in
-  let open functor
-             (_ : S)
-             ->
-             struct
-               type t
-             end in
+  let open
+    functor
+      (A : S)
+      ->
+      struct
+        type t
+      end in
+  let open
+    functor
+      (_ : S)
+      ->
+      struct
+        type t
+      end in
   let open A (B) in
   let open (A : S) in
   let open (val x) in
@@ -257,19 +258,21 @@ open struct
   type t
 end [@@attr]
 
-open functor
-       (A : S)
-       ->
-       struct
-         type t
-       end [@@attr]
+open
+  functor
+    (A : S)
+    ->
+    struct
+      type t
+    end [@@attr]
 
-open functor
-       (_ : S)
-       ->
-       struct
-         type t
-       end [@@attr]
+open
+  functor
+    (_ : S)
+    ->
+    struct
+      type t
+    end [@@attr]
 
 open A (B) [@@attr]
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@
 
   + Preserve comment position around `match` and `try` keywords (#1458, @gpetiot)
 
+  + Add missing break in module statement (#1431, @gpetiot)
+
 #### New features
 
 ### 0.15.0 (2020-08-06)
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2256,14 +2256,20 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       hovbox 0
         (wrap_if outer_parens "(" ")"
            ( hvbox 0
-               ( fits_breaks_if ?force inner_parens "" "("
-               $ fits_breaks ?force "" "let "
-               $ Cmts.fmt c popen_loc
-                   ( fits_breaks ?force ""
-                       (if override then "open! " else "open ")
-                   $ fmt_module_statement c ~attributes noop
-                       (sub_mod ~ctx popen_expr) )
-               $ fits_breaks ?force opn " in"
+               ( hvbox 0
+                   ( fits_breaks_if ?force inner_parens "" "("
+                   $ fmt_module_statement c ~attributes
+                       ~keyword:
+                         ( hvbox 0
+                             ( fits_breaks ?force "" "let"
+                             $ fits_breaks ?force "" ~hint:(1, 0) ""
+                             $ Cmts.fmt_before c popen_loc
+                             $ fits_breaks ?force ""
+                                 (if override then "open!" else "open") )
+                         $ fits_breaks ?force "" ~hint:(1, 0) "" )
+                       (sub_mod ~ctx popen_expr)
+                   $ Cmts.fmt_after c popen_loc
+                   $ fits_breaks ?force opn " in" )
                $ fmt_or_k force_fit (fmt "@;<0 2>")
                    (fits_breaks ?force "" ~hint:(1000, 0) "")
                $ fmt_expression c (sub_exp ~ctx e0)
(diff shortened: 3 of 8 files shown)


test/passing/open-auto.ml.ref
```diff
@@ -232,18 +234,20 @@ let _ =
...
-               type t
-             end in
-  let open functor
-             (_ : S)
-             ->
-             struct
-               type t
-             end in
+  let open
+        functor
```
Julow, line 238: Is this expected ? (not related to the issue linked in this PR)
I'd expect the `let open` to not break but the functor indented only by two spaces.

test/passing/open-auto.ml.ref
```diff
@@ -31,10 +31,10 @@ let _ =
...
 
 let () =
-  ( (let open Term in
-    term_result
-      ( const Phases.phase1 $ arch $ hub_id $ build_dir $ logs_dir
-      $ setup_logs ))
+  ( Term.(
+      term_result
+        ( const Phases.phase1 $ arch $ hub_id $ build_dir $ logs_dir
+        $ setup_logs ))
```
Julow, line 37: This is correct, I think. I wonder which change caused this ?

test/passing/open-auto.ml.ref
gpetiot: Yes for now "let open" is considered being in single keyword. It didn't shock anyone so far regarding the lack of break but we can definitely change that in a future PR.

gpetiot: Good catch, I didn't notice the indentation, I will have a look.

gpetiot:
I rewrote the fits/breaks matching to better distinguish the 2 different styles of output to ease the maintenance. I also moved the comments that could sit between `let` and `open` to avoid weird outputs and it would be too difficult to always indent the expression from `let open` instead of `open` otherwise. What do you think?

test/passing/open-auto.ml.ref
gpetiot: Before there were some cases where the syntax changed depended on whether it fitted on a single line or not (if it didn't fit we used the `let open` syntax in this case). Now it only depends on the shape of the expression so the `.` syntax is used for Pexp_apply.

Personally I found it weird to have the syntax change when the line became longer.

test/passing/open-auto.ml.ref
```diff
@@ -176,15 +176,15 @@ class a =
 
 let _ =
   (* a *)
-  let (* b *)
-  open (* c *) struct
+  (* b *)
+  let open (* c *) struct
```
Julow, line 180: This is a regression.

test/passing/open-auto.ml.ref
Julow: I agree.

test/passing/open-auto.ml.ref
gpetiot: fixed

test/passing/open-auto.ml.ref
Julow, line 37: I like that we get rid of this "feature". I hope this was not used in the wild.

jberdine: Note that using the `let open` form in case the body breaks is entirely intentional. Note in particular that this change does not save any lines and increases the indentation uselessly. So FTR this looks like a regression to me. I agree that the parenthesization in the diff is odd though.

Julow: I ran test_branch with the wrong arguments and misunderstood the change.
Indeed this seems to remove the whole feature.

gpetiot: The thing is that the box layout and the formatting now depends on whether it fits or breaks, in a way that is not feasible with the fits_breaks function, so I need to distinguish the 2 different implementations like I did in this PR

gpetiot: The issue with `let-open=auto` is that sometimes it doesn't stabilize, eg:
```ocaml
let rec pp_hpred_env pe0 envo f hpred =
  color_wrapper pe0 f hpred ~f:(fun pe f hpred ->
      match hpred with
      | Hpointsto (e, se, te) ->
          let pe' =
            match (e, se) with
            | Lvar pvar, Eexp (Var _, _) when not (Pvar.is_global pvar) ->
                Pp.{pe with obj_sub= None} (* dont use obj sub on the var defining it *)
            | _ ->
                pe
          in
          fooooooooooooooooo
    )
```

gpetiot:
Can we merge this one? I maintained the global behavior with `let-open=auto`, despite personally being in favor of removing this behavior.

This example:
```ocaml
let rec pp_hpred_env pe0 envo f hpred =
  color_wrapper pe0 f hpred ~f:(fun pe f hpred ->
      match hpred with
      | Hpointsto (e, se, te) ->
          let pe' =
            match (e, se) with
            | Lvar pvar, Eexp (Var _, _) when not (Pvar.is_global pvar) ->
                Pp.{pe with obj_sub= None} (* dont use obj sub on the var defining it *)
            | _ ->
                pe
          in
          fooooooooooooooooo
    )
```
does not converge on master or on this PR and is inherent to the automatic syntactic change of the `auto` behavior. That's why I would like to remove this behavior in the future. But putting that aside this PR should be a good improvement.

jberdine:
I planned to have another look at this, if you can give me a couple more days.

I see the problem with using the single-line detection to determine the short versus long form. I would like to think again about ways to resolve that. It might be that automatically detecting this isn't feasible due to interactions with comment attachment, but that would be unfortunate. If so, I think that it makes more sense to go the route of preserving the form used in the source, rather than rewriting multi-line long-form `let open M in` to `M.(`.

gpetiot:
Would it really be a big loss to only keep the `preserve` behavior and get rid of `short/long/auto`? The `ocamlformat`, `janestreet` and `conventional` profiles all use the preserve behavior. Personally I think if the user chooses one syntax and not the other he must have some reasons for that and we should just preserve the user syntax.

gpetiot:
Maybe we can just merge it as is, keeping the non-converging behavior that as already here anyway, so at least this issue can be closed.
And we can discuss whether or not we should deprecate this option in another issue?

PR #1438, opened 2020-08-05
Release process: use dune-release automatic mode

emillon:


emillon:
@Julow I know that you have a custom git config that prevents this from working, but I did it like that last time to test `dune-release` and it worked very well. What do you think about making this the recommended process?

PR #1445, opened 2020-08-07
Specify package version in opam file

clembu:
A lot of projects will tend to either pin ocamlformat to a certain commit or add constraints on its version in their dependency to have project-wide formatting consistency and sometimes specific features (like the recent `.eliom` support for example).

When these pins or constraint change, pulling the changes doesn't change the version of the package because it doesn't come from the opam versioned repo but from a git repo somewhere, and this file here will be read to determine the version and check it against the constraint.

I propose to maintain this version number here for every package version published, to avoid problems with pins and constraints in the future.
Unless you're alright with projects having to uninstall and reinstall ocamlformat everytime they need to change their version, of course.

Another possibility is to specify `"dev"` as the version long as it's not an actual version published on opam, and change it for the release candidates on their own branch. I'm not sure which is better.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 #### Bug fixes
 
   + Fix break between keyword and expr with `if-then-else=keyword-first` (#1419, @gpetiot)
+  + Specify version in opam file to help with pins and opam constraints (#1445, @facelesspanda)
 
 #### New features
 
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -10,6 +10,7 @@
 ##########################################################################
 
 opam-version: "2.0"
+version: "0.15"
 maintainer: "OCamlFormat Team <ocamlformat-team@fb.com>"
 authors: "Josh Berdine <jjb@fb.com>"
 homepage: "https://github.com/ocaml-ppx/ocamlformat"


facelesspanda:
check failed so I added a line in the changelog, before seeing that a `changelog not needed` label existed. Should I keep the line in the log or remove it?

emillon:
Hi,

We're trying to move to a simpler system where version is sourced from the git repo (#1133) and opam files are generated (#1361), so this feel a bit like a step backwards to add the version number in the opam file.

But I agree that there's no good way to set a version number when ocamlformat is part of a monorepo or more generally when its sources are not pulled from a release tarball or this git repository.

One thing we can do is add an escape hatch - a dune rule that checks if a special file is present and use that as a version number if it exists when the package is built (or detect it if this file is not present - that would be what happens for most users). Would that work for your use case?

@jberdine @ceastlund I think that would also help you vendor ocamlformat in your organizations - let me know if that's not the case.

gpetiot:
Sorry if I fail to see the root of the issue, would having `x.y.z+dev` instead of `x.y.z-commit-id` fix the issue and be easier to implement with what is provided by dune-build-info?

gpetiot:
The more I look at this issue, the more I see something that should get implemented by the build system, I'm closing this issue since the discussion didn't progress since it was opened!

PR #1451, opened 2020-08-27
Improve formatting of groups of items

gpetiot:
The goal is to remove the over-factorization that is done by using make_groups, and handle `'a list` instead of `'a list list` which would make it easier to maintain and e.g. optimize the printing of `;;`

No diff with test_branch.sh, the diff in the tests is a slight fix and is not worth mentioning in the changelog.

### DIFF
--- a/test/passing/tests/js_source.ml.ref
+++ b/test/passing/tests/js_source.ml.ref
@@ -169,7 +169,6 @@ end
 
 (* Structure items *)
 let%foo[@foo] x = 4
-
 and[@foo] y = x
 
 type%foo t = int [@@foo]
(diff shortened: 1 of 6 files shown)


test/passing/tests/js_source.ml.ref
```diff
@@ -169,7 +169,6 @@ end
 
 (* Structure items *)
 let%foo[@foo] x = 4
-
 and[@foo] y = x
```
gpetiot, line 172: This is more consistent regarding the formatting of other items, the break should be doubled if there is a docstring and not an attribute like this.

PR #1452, opened 2020-08-28
Set 'indicate-multiline-delimiters=no' on default profile

gpetiot:
Setting the more "vanilla" value of this option for the default profile.
And also disabling the formatting of vendored libraries (should have been done long ago).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 #### Changes
 
+  + Set 'indicate-multiline-delimiters=no' on default profile (#1452, @gpetiot)
+
 #### Bug fixes
 
   + Fix break between keyword and expr with `if-then-else=keyword-first` (#1419, @gpetiot)
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -773,14 +773,14 @@ module Formatting = struct
     in
     let names = ["indicate-multiline-delimiters"] in
     let all =
-      [ ( "space"
-        , `Space
-        , "$(b,space) prints a space inside the delimiter to indicate the \
-           matching one is on a different line." )
-      ; ( "no"
+      [ ( "no"
         , `No
         , "$(b, no) doesn't do anything special to indicate the closing \
            delimiter." )
+      ; ( "space"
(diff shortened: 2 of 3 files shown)


Julow:
Why is this more "vanilla" ? Is there more discussion about this than https://github.com/ocaml-ppx/ocamlformat/pull/1377#issuecomment-682003304 ?

.ocamlformat-ignore
```diff
@@ -0,0 +1,3 @@
+vendor/*/*
```
emillon, line 1: Can you pull this change in a different PR?

emillon:
@Julow The two main arguments I would make for this feature to be opt-in:

- is that it's a style that's only possible with an automatic formatter. It's not the way most humans would write OCaml code. As such, it can be surprising seeing `_) )` somewhere.
- it is not super stable. that's not at the point of `doc-comments=after`, but we yanked changes to this feature from 0.15.0 because it was too impacting.

Both of these are at odds with the goals of the default profile, which is meant to be boring and stable in the long run.

.ocamlformat-ignore
gpetiot: done in #1453

Julow:
Seeing `) )` is indeed very surprising but I think this feature is useful for readability and while editing.
Setting it to `no` makes very compact code and makes it harder to select code that is deeply nested into parentheses, for example.

The most annoying example is nested matches:

```ocaml
let _ =
  match Foo.bar x with
  | This y -> (
    match Foo.baz y with
    | This z -> z
    | That w ->
        print something ;
        a (b (c w)))
  | That z -> z
```

I think the default profile has an other job, which is to ensure readability and avoid ambiguous syntaxes (to humans).
We should consider setting this option to `closing-on-separate-line`:

```ocaml
let _ =
  match Foo.bar x with
  | This y -> (
    match Foo.baz y with
    | This z -> z
    | That w ->
        print something ;
        a (b (c w))
  )
  | That z -> z
```

I agree that this option has and had a lot of bugs.

PR #1453, opened 2020-08-28
Do not format vendored libraries

gpetiot:


emillon (approved):
:+1: it seems that the proper way to do that would be to make these different dune-projects but in the meantime that works

PR #1457, opened 2020-09-02
Add tests from the ocp-indent testbase

gpetiot:
- This mainly adds tests that cause ocamlformat to fail (and we should aim at making pass with `--format-invalid-files=auto` some day), a few passing tests are also added in the mix.
- The generation of the dune.inc file has been automatized for failing tests just like for the passing tests.
- A few non-failing tests were in the failing dir, they have been moved to the passing directory.


Julow (approved):
Looks good.

PR #1458, opened 2020-09-04
Preserve comment position around `match` and `try` keywords

### LINKED ISSUE #341: Preserve position of comments in `match (*cmt*) expr with`.

gpetiot:
Reboot of #1414
Fix #341 

The diff of test_branch looks like a fix, the comments were previously englobed with the pattern matching inside parens, now the parens only englobe the pattern matching.


lib/Cmts.ml
```diff
@@ -280,6 +280,24 @@ let relocate (t : t) ~src ~before ~after =
           r |> Location.Set.remove src |> Location.Set.add after
           |> Location.Set.add before) )
 
+let relocate_match_cmts (t : t) src tok ~whole_loc ~matched_loc =
+  let f map =
+    let f Cmt.{loc= cmt_loc; _} =
+      Source.is_before_match_keyword src tok ~exp_loc:whole_loc ~cmt_loc
```
Julow, line 286: I suggest to break this function. `tok` and `whole_loc` are constant, which mean that the `tokens_at` call could be made only once outside of the loop.
The function `f` would then only need to do the comparison, which is as explicit and more efficient.

Julow (approved):
Looks good.

lib/Cmts.mli
```diff
@@ -55,6 +55,19 @@ val relocate :
     locations) comments before [src] to [before] and comments after [src] to
     [after]. *)
 
+val relocate_match_cmts :
```
hhugo, line 58: This is not only  used for `match`, can we find a better name ? (and update the doc comment)

lib/Source.ml
```diff
@@ -367,3 +367,8 @@ let is_quoted_string t loc =
       | _ -> false)
   in
   not (List.is_empty toks)
+
+let is_before_match_keyword t kwd ~exp_loc ~cmt_loc =
```
hhugo: This is not only used for `match`, can we find a better name ?

lib/Cmts.ml
gpetiot: Good point, I modified the api of source.mli and simplified the relocating function

lib/Cmts.ml
```diff
@@ -280,6 +280,24 @@ let relocate (t : t) ~src ~before ~after =
...
+      List.partition_tf (Location.Multimap.find_multi map matched_loc) ~f
+    in
+    let map =
+      List.fold_left ~init:map (List.rev before) ~f:(fun map ->
+          Location.Multimap.add_multi map whole_loc)
+    in
+    Location.Multimap.change_multi map matched_loc after
+  in
+  update_cmts t `Before ~f ;
+  update_cmts t `After ~f ;
```
hhugo, line 297: Why do we call `update_cmts` with `After` ?

hhugo (changes requested):

lib/Source.ml
gpetiot: I've changed the function, the new one is now called `loc_of_keyword`

lib/Cmts.ml
gpetiot: it's unnecessary indeed

lib/Cmts.mli
gpetiot: I've renamed the function `relocate_pattern_matching_cmts` and improved the doc comment

lib/Source.ml
```diff
@@ -367,3 +367,8 @@ let is_quoted_string t loc =
       | _ -> false)
   in
   not (List.is_empty toks)
+
+let loc_of_keyword t loc kwd =
+  match tokens_at t loc ~filter:(Poly.( = ) kwd) with
+  | [] -> None
+  | (_, loc) :: _ -> Some loc
```
emillon, line 374: Just reading this function in isolation, it's not super clear whether this is intended to find one token or just return the first one. If it's just for one token, can you add an `assert false` in the non-singleton case; or if you're interested in the first one (I think it's that?), can you rename it to something like `loc_of_first_token_at`?

lib/Fmt_ast.ml
```diff
@@ -1561,6 +1561,14 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
   let {pexp_desc; pexp_loc; pexp_attributes; _} = exp in
   update_config_maybe_disabled c pexp_loc pexp_attributes
   @@ fun c ->
+  ( match pexp_desc with
```
emillon, line 1564: If the numbers of free variables is reasonable I think that this block can be extracted to a function

lib/Source.ml
gpetiot: done

lib/Fmt_ast.ml
gpetiot: done, I've moved that code in Cmts.ml and didn't export the other function

gpetiot:
@hhugo do you need to review it one last time before I merge it?

emillon (approved):

PR #1461, opened 2020-09-09
Remove the deprecated 'doc-comments-val' option

gpetiot:


lib/Conf.ml
```diff
@@ -501,74 +500,9 @@ module Formatting = struct
         , "$(b,before) puts comments before the corresponding code." ) ]
     in
     C.choice ~names ~all ~doc ~section
-      ~removed_values:
-        [ C.removed_value ~name:"after" ~version:"0.14.2"
-            ~msg:
-              "This value has been renamed `after-when-possible` to take \
-               into account the technical limitations of ocamlformat, the \
-               behavior is unchanged." ]
```
Julow, line 509: Why remove this ?

Julow (commented):
We should probably keep a list of removed options and a short message to help users migrate. Not very important for this option but it may be useful for https://github.com/ocaml-ppx/ocamlformat/pull/1462.
I started to work on this a while ago: https://github.com/ocaml-ppx/ocamlformat/compare/master...Julow:removed-options

lib/Conf.ml
gpetiot: oh right, reverting this part!

emillon:
Yes, I agree a quick CLI test for these would be useful.

emillon:
Also it seems a bit early to remove this, I suggest we wait until 0.17.0 for this one.

gpetiot:
Since this option was already effectively removed from the formatting, and only listed in the list of options from the manpage, I don't think we should hold this PR until 0.17.0, it is now listed as a removed option.

PR #1463, opened 2020-09-09
Remove the deprecated 'escape-strings' option

gpetiot:


ocamlformat-help.txt
```diff
@@ -202,12 +202,6 @@ OPTIONS (CODE FORMATTING STYLE)
            on the preceding line and closed on the following line. The flag
            is set by default.
 
-       --escape-strings={preserve|decimal|hexadecimal}
```
emillon, line 205: Can you add a quick CLI test to make sure we're displaying a nice error message?

gpetiot: I added a test, there is no nice message for now but Jules already has a branch to work on this (https://github.com/ocaml-ppx/ocamlformat/compare/master...Julow:removed-options)

gpetiot: Rebased, there is now already a test for option escape-chars.

PR #1464, opened 2020-09-10
Fix parentheses around infix applications having attributes

### LINKED ISSUE #1455: Bug: AST changed for attribute on an application operator
The following program crashes OCamlFormat:
```ocaml
(f @@ a) [@attr]
```

OCamlFormat tries to shift the attribute inside the parens with `(f @@ a [@attr])`, changing the AST.

gpetiot:
fix #1455, no diff with test_branch


emillon (approved):

Julow (approved):
Looks good.

PR #1465, opened 2020-09-10
Fix parentheses around the index arg of a non-sugared index operation

### LINKED ISSUE #1454: Bug: "ast changed" with valid code
Ocamlformat version: 0.15.0

**Describe the bug**
Ocamlformat fails to format some valid code.

**How to Reproduce**
Input:
```
let foo = (String.get [@bar]) filename (len - 1) = 'i'
```
Ocamlformat reports :
```
  BUG: ast changed.

  Code: -32603 
```

Edit: please ignore this part below 😅 
~~Also note something else, if the attribute is removed, ocamlformat succeeds, but the output is missing the `String.get` call:
Input:~~
```
let foo = String.get filename (len - 1) = 'i'
```
~~Output:~~
```
let foo = filename.[len - 1] = 'i'
```
----

The selected options are just `m=120` but can reproduce without that as well.

gpetiot:
fix #1454, no diff with test_branch

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -156,10 +156,13 @@ module Indexing_op = struct
     match all_args_unlabeled [] args with
     | None | Some [] -> None
     | Some (lhs :: args) -> (
-        let {Location.txt= ident; loc} = ident in
+      match ident with
+      | {pexp_desc= Pexp_ident {txt= ident; loc}; pexp_attributes= []; _}
+        -> (
         match get_sugar_ident ident args with
         | None -> None
         | Some (op, rhs) -> Some {lhs; op; rhs; loc} )
+      | _ -> None )
 end
 
 module String_id = struct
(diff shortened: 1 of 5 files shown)


lib/Ast.ml
```diff
@@ -1644,7 +1644,8 @@ end = struct
           match i.[0] with
           | '!' | '?' | '~' -> Some (High, Non)
           | _ -> Some (Apply, Non) ) )
-      | Pexp_apply ({pexp_desc= Pexp_ident ident; _}, args)
+      | Pexp_apply
+          ({pexp_desc= Pexp_ident ident; pexp_attributes= []; _}, args)
         when Option.is_some (Indexing_op.get_sugar ident args) -> (
```
emillon, line 1649: Just to make sure, is there still a test that covers this branch now?

gpetiot: Yes this is the default case, when there is `String.get s i` that will automatically get sugared into the bracket form, we have some tests already.

emillon: allright!

emillon (approved):

lib/Ast.ml
Julow: Perhaps we should directly pass the first argument of `Pexp_apply` to `get_sugar` and move such details where they make sense ?

gpetiot: Good idea, done!

PR #1466, opened 2020-09-16
Use base instead of Location.Set and Location.Multimap

emillon:
When I added `Location.Set` and `Location.Multimap` I was not too familiar with some `base` APIs. This PR replaces these monomorphic containers by generic versions:

- `Location.Set` is replaced by plain `Set` calls
- `Location.Multimap` is replaced by a new `Multimap` module that contains extensions to `Map` for maps which hold a list as value.

Let me know how that looks.


gpetiot (approved):
Looks good to me, thanks!

PR #1469, opened 2020-09-22
Document removed options

gpetiot:
Revival of #1287


Julow (approved):
Looks good.

Here is how the message looks when the option is set in an `.ocamlformat` file, which is nice:
```
ocamlformat: Error while parsing .ocamlformat:
             For option "escape-chars": This option has been removed in version 0.16.0. Concrete syntax will now always be preserved.
```

It might be useful to add it as a test.

PR #1470, opened 2020-09-22
Printast: do not pass indent level

emillon:
This removes the `int` argument (always set to 0) in order to decrease the diff for the ppxlib port.


gpetiot (approved):
LGTM

PR #1471, opened 2020-09-22
Use Ast_helper in Ast.ml

emillon:
This uses Ast_helper in more places in order to ease the transition to ppxlib.


gpetiot (approved):
Looks good

PR #1474, opened 2020-09-24
Add Emacs package metadata to ocamlformat.el

bcc32:
Tested by running `M-x package-lint-current-buffer` in Emacs.


gpetiot (approved):
Looks good to me, thanks!

PR #1475, opened 2020-09-24
Rewrite CLI tests as cram tests

gpetiot:


Julow (approved):
Looks good !

It should be possible to merge some tests which are variation of the same test in a single cram file. For example:
- `stdin_*`
- `max_iter_*`
- `fmterr_file_*`
- `err_*`
- `env_*` and `conf_*`
- `check_*`

PR #1477, opened 2020-09-25
Use plain Format from OCaml for debug messages

emillon:
These are never put in the formatted output, so the plain Format works.

### DIFF
--- a/lib/Cmt.ml
+++ b/lib/Cmt.ml
@@ -9,7 +9,6 @@
 (*                                                                        *)
 (**************************************************************************)
 
-module Format = Format_
 open Migrate_ast
 
 module T = struct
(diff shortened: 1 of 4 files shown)


lib/Cmt.ml
```diff
@@ -25,7 +25,7 @@ module T = struct
       ; Comparable.lift Location.compare ~f:loc ]
 
   let sexp_of_t {txt; loc} =
-    Sexp.Atom (Format.asprintf "%s %a" txt Migrate_ast.Location.fmt loc)
+    Sexp.Atom (Caml.Format.asprintf "%s %a" txt Migrate_ast.Location.fmt loc)
```
gpetiot, line 28: This can be left and `module Format = Format_` can be removed instead.

PR #1478, opened 2020-09-25
Abstract types of AST iterations in terms of a "fragment" type

emillon:
In several places, we are doing the same kind of processing but on different bits of the AST. For example normalizing structures/signatures/use_files, or collecting comments in structure/signatures/use_file. The current abstraction is to pass mapping functions to the bits that do the processing, but this has several drawbacks: this is not very regular (sometimes we need to pass `mapper -> 'a -> 'a`, sometimes `'a -> 'a -> bool`, etc), and it leaks `mapper` in the interface.

This PR adds a type `'a fragment` which represents operations on `'a`, and replaces completely `Translation_unit.t` (which took the dual approach of putting all operations in a record).

This simplifies many places by only exposing 1 function that operates on fragments rather than the 3 variants, and in the context of ppxlib migration this means that we will only have to change the type of `map_ast` so that it takes a ppxlib-based mapper instead of an OMP mapper.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -45,14 +45,13 @@ let find_at_position t loc pos =
 module Loc_tree : sig
   include Non_overlapping_interval_tree.S with type itv = Location.t
 
-  val of_ast :
-    (Ast_mapper.mapper -> 'a -> _) -> 'a -> Source.t -> t * Location.t list
+  val of_ast : 'a Mapper.fragment -> 'a -> Source.t -> t * Location.t list
 end = struct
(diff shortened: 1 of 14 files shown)


emillon:
Notes for review:
- I kept that split in several commits but I'll squash that at the end
- I'm not sure what the best name is for this - in a previous version I used "domain" instead of "fragment".

gpetiot (approved):
Really good improvement, thanks! `fragment` fits very well to my opinion.

emillon:
Thanks! Let's merge this!

PR #1479, opened 2020-09-25
Parse-wyc: vendor Ast_helper and Docstrings

emillon:
These modules are not provided anymore in OMP2, so we copy them from 4.08 and adapt the imports.


emillon:
This is my least favorite of all the ppxlib PRs, the only one with a huge positive diffstat, but we don't have a ton of alternatives.

gpetiot (approved):
I'm not a fan of the huge diff either, especially since we don't know how long the parse-wyc library will be necessary, but if we have to let's do it

emillon:
Yes, somehow I think this represents better the cost of vendoring a parser.

PR #1481, opened 2020-09-26
Use file-truename for temp files in ocamlformat.el

juxd:
On MacOS, `make-temp-file` returns a file name that is a symlink. This returns `nil` for `get-file-buffer`, and doesn't delete the file when `delete-file` uses it.


gpetiot (approved):
Looks good to me, thanks!

PR #1483, opened 2020-09-28
Add missing parentheses around a pattern matching that is the left-hand part of a sequence when an attribute is attached

### LINKED ISSUE #1480: Bug: missing parentheses around a pattern matching when attribute is attached
This pattern-matching
```ocaml
let _ =
  (match bar with _ -> ())[@ocaml.warning "-4"];
  foo
```
gets formatted as
```ocaml
let _ =
  match[@ocaml.warning "-4"] bar with _ -> () ;
  foo
```
which doesn't have the same Ast.

The parentheses are properly used when there is no attribute:
```ocaml
let _ =
  (match bar with _ -> ()) ;
  foo
```

gpetiot:
fix #1480 (allows to format lwt codebase, test/core/test_lwt.ml and src/unix/lwt_io.ml were unable to be formatted), no diff with test_branch.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -2204,31 +2204,22 @@ end = struct
       | Pexp_let _ | Pexp_match _ | Pexp_try _ -> true
       | _ -> false
     in
-    let sequence ctx_pexp_desc exp =
-      match (ctx_pexp_desc, exp.pexp_attributes) with
-      | ( Pexp_sequence (({pexp_desc= Pexp_match _ | Pexp_try _; _} as e), _)
-        , _ :: _ )
-        when e == exp ->
-          true
-      | Pexp_sequence _, _ :: _ -> false
-      | ( Pexp_sequence
-            ( ( { pexp_desc=
-                    Pexp_extension
-                      ( _
-                      , PStr
-                          [ { pstr_desc=
-                                Pstr_eval
-                                  ({pexp_desc= Pexp_sequence _; _}, [])
-                            ; _ } ] )
-                ; _ } as lhs )
-            , _ )
+    let exp_in_sequence lhs rhs exp =
+      match (lhs.pexp_desc, rhs, exp.pexp_attributes) with
+      | (Pexp_match _ | Pexp_try _), _, _ :: _ when lhs == exp -> true
+      | _, _, _ :: _ -> false
+      | ( Pexp_extension
+            ( _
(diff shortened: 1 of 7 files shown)


emillon (approved):

lib/Ast.ml
```diff
@@ -2228,6 +2228,14 @@ end = struct
         false
     (* Object fields do not require parens, even with trailing attributes *)
     | Exp {pexp_desc= Pexp_object _; _}, _ -> false
+    | ( Exp
+          { pexp_desc=
+              Pexp_sequence
+                (({pexp_desc= Pexp_match _ | Pexp_try _; _} as e), _)
+          ; _ }
+      , {pexp_attributes= _ :: _; _} )
```
Julow, line 2236: Could you add a comment explaining this case ?
It is quite confusing, especially the `pexp_attributes` pattern.

This is a special case of the next pattern, which seems to be a special case of the following pattern and `Pexp_sequence` near the end.
This is quite hard to understand. What about moving everything related to sequence to an other function and keeping only one `Exp {pexp_desc = Pexp_sequence}` here.
There is only 2 generic patterns and they would need to be duplicated (I think this is a win in readability, especially since there is so much special handling): `Exp.has_trailing_attributes` and `_` in `Exp` near the end.

gpetiot: I improved the situation a bit, but the whole mechanism needs to be rewritten some day, it's painful to maintain.

emillon (approved):

lib/Ast.ml
```diff
@@ -2332,22 +2345,9 @@ end = struct
...
-          , _ )
-        when lhs == exp ->
-          true
-      | Pexp_sequence (lhs, _) when lhs == exp ->
-          exposed_right_exp Let_match exp
-      | Pexp_sequence (_, rhs) when rhs == exp -> false
-      | _ -> parenze () )
+      | Pexp_sequence (lhs, rhs) -> exp_in_sequence lhs rhs exp
+      | _ -> Exp.has_trailing_attributes exp || parenze () )
+    | _, exp when Exp.has_trailing_attributes exp -> true
```
Julow, line 2350: I didn't expect that this case can be moved. I found no diffs with `test_branch`.
This simplifies the code quite a bit.

Julow (approved):
Looks good.

PR #1484, opened 2020-09-28
Split invalid files to separately check if fragments can be parsed

### LINKED ISSUE #1473: Bug: ast changed when using --format-invalid-files
Hi,

On ocamlformat 0.15.0, trying to format the following file with `--format-invalid-files=auto` exits with an error saying that the AST changed:

```
let f () =
  begin
    print_endline "hello"

let f x = x + x
```

gpetiot:
~fix #1473~ edit: already fixed on master


hhugo:
I don't think you should split on newlines. It will not behave correctly in the presence of string literals and comments. I think doing the split on tokens makes more sense.

(My hunch is that we should teach parse_wyc about the heuristic you're implementing here)

gpetiot:
> I don't think you should split on newlines. It will not behave correctly in the presence of string literals and comments. I think doing the split on tokens makes more sense.
> 
> (My hunch is that we should teach parse_wyc about the heuristic you're implementing here)

I agree, right now this is mostly still an experiment, based with the assumption that the starting state of the file is mostly already formatted or at least structured (with double linebreaks between items and so on) and that the user is adding something invalid to this state. I think this is a sane assumption for a degraded mode of ocamlformat.

But yes in the future we could think of a way to remove this assumption, though I'm afraid it would be much less efficient in practice, and we might end up adding most of the file as verbatim most of the time

gpetiot:
I improved the handling of comments, removing all the warnings and weird behaviors we previously had. The code is ready to be reviewed now.

The goal of this PR is to improve the `format-invalid-files = auto` feature that we currently have, but in the long run I'm doubting about the usefulness of such feature, as we thought we needed it for the editor integration (re-indent a range of lines, or indent when pressing <tab>) while in practice with the experiments I've done in https://github.com/ocaml-ppx/ocamlformat/pull/1207 I couldn't find a way to reuse the recovered file to deduce the indentation of invalid files (as it's printed verbatim anyway), and we need to infer it from the keywords we can find in the context anyway.
cc @emillon @hhugo @Julow do you have any thoughts?

ceastlund:
The only place I know of where we want invalid file formatting is in editors.

gpetiot:
My bad, #1207 is using this feature afterall! But we could still remove the public option `--format-invalid-files` and only keep the api.

gpetiot:
Replaced by #1609

PR #1485, opened 2020-10-01
Fix invalid fragment delimiters of format-invalid-files recovery mode

hhugo:
- do not use double quoted string to store invalid source as it would emit invalid source in the presence of double quotes.
- simplify the logic wrapping invalid source with extension point.


hhugo:
I'm not sure where to open this PR (here or in parse-wyc)

gpetiot:
It's the right place for this PR, as I archived the parse-wyc repo. By the way I tried to find a testcase that exhibit the bug on master and would be fixed by this PR but I couldn't find one, do you have one?

hhugo:
You just need to have double quote inside an invalid code fragment.
`let x = "123" + ` should do it I think

gpetiot (approved):
Looks good to me, thanks!

PR #1488, opened 2020-10-06
Expose the fragment arg in Translation_unit API

gpetiot:
Minor change to reduce the diff in another PR

### DIFF
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -24,8 +24,8 @@ let format ?output_file ~kind ~input_name ~source conf opts =
   else
     let f =
       match kind with
-      | `Impl -> Translation_unit.parse_and_format_impl
-      | `Intf -> Translation_unit.parse_and_format_intf
+      | `Impl -> Translation_unit.parse_and_format Use_file
+      | `Intf -> Translation_unit.parse_and_format Signature
     in
     f ?output_file ~input_name ~source conf opts
 
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -19,15 +19,11 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter)
 ;;
 Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 
-let format ?output_file ~kind ~input_name ~source conf opts =
+let format ?output_file ~kind:(Conf.Kind k) ~input_name ~source conf opts =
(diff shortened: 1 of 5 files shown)


lib/Translation_unit.mli
```diff
@@ -11,25 +11,17 @@
 
 type error
 
-val parse_and_format_impl :
-     ?output_file:string
+val parse_and_format :
+     'a list Migrate_ast.Mapper.fragment
```
emillon, line 15:
```suggestion
     _ list Migrate_ast.Mapper.fragment
```

To show that we're not using the result

emillon: it's possible to increase the polymorphism to `_ fragment` by replacing the `List.is_empty` call somewhere in `Translation_unit` by a `is_empty : _ fragment -> bool` call

bin/ocamlformat.ml
```diff
@@ -24,8 +24,8 @@ let format ?output_file ~kind ~input_name ~source conf opts =
   else
     let f =
       match kind with
-      | `Impl -> Translation_unit.parse_and_format_impl
-      | `Intf -> Translation_unit.parse_and_format_intf
+      | `Impl -> Translation_unit.parse_and_format Use_file
```
emillon, line 27: you can get rid of `Impl/Intf` and simplify this bit, but you need to erase the type argument from the GADT by defining something like:

```
type kind :
  | Kind of _ fragment
```

And having the cmdliner term return that. This can also be used in `parse_and_format` instead of the `_` wildcard.

lib/Translation_unit.mli
gpetiot: I simplified the type but didn't find the is_empty call you are speaking about.

bin/ocamlformat.ml
gpetiot: Good idea, done.

lib/Translation_unit.mli
emillon: come to think of it, `_ list` is good enough. if fragments are extended to something that's not a list, we'll need to change this area anyway. The call site I had in mind is:

https://github.com/ocaml-ppx/ocamlformat/blob/f0aa54b429f2a8d6b1b84686f6e2c6d6161fa89a/lib/Fmt_ast.ml#L4538

(but it's already exposing a `'a list fragment`)

emillon (approved):

PR #1490, opened 2020-10-09
Add tests for Literal_lexer

emillon:
I was not too sure what the "mode" parameter was used for, so I added some tests. The tests make it explicit, but the behavior for "normalize" is that:

- hard tabs are converted to escapes
- escaped newlines are converted to hard newlines
- escaped spaces are converted to spaces

This mechanism is used for `break_string_literals` only.

### DIFF
--- a/test/unit/test_literal_lexer.ml
+++ b/test/unit/test_literal_lexer.ml
@@ -1,0 +1,80 @@
+open Base
+open Ocamlformat_lib
+
+let single_quote = '\''
+
+let double_quote = '"'
+
+let backslash = '\\'
+
+let space = ' '
+
+let tab = '\t'
+
+let newline = '\n'
+
+let tests_string =
+  let test_opt name s mode ~expected =
+    ( "string: " ^ name
+    , `Quick
+    , fun () ->
+        let got = Literal_lexer.string mode s in
+        Alcotest.check Alcotest.(option string) Caml.__LOC__ expected got )
+  in
+  let test_one name s mode ~expected =
+    test_opt name s mode ~expected:(Some expected)
+  in
+  let test name s ~expected_preserve ~expected_normalize =
+    [ test_one (name ^ " (preserve)") s `Preserve ~expected:expected_preserve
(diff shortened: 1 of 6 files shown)


test/unit/test_literal_lexer.ml
```diff
@@ -0,0 +1,90 @@
...
+        ~expected_normalize:(String.of_char_list [space])
+    ; test "backslash n"
+        (String.of_char_list [double_quote; backslash; 'n'; double_quote])
+        ~expected_preserve:(String.of_char_list [backslash; 'n'])
+        ~expected_normalize:(String.of_char_list [newline]) ]
+
+let tests_char =
+  let parse s =
+    let lexbuf = Lexing.from_string s in
+    match Literal_lexer.char lexbuf with r -> Some r | exception _ -> None
```
gpetiot, line 61: Maybe we could replace the lexer exception by a result type? (not necessarily in this PR)

gpetiot (approved):
Looks good to me, thanks!

test/unit/test_literal_lexer.ml
emillon: it's defined in a mll file so I don't think we can change the API.

gpetiot: Yes we can, we did it in mdx: https://github.com/realworldocaml/mdx/blob/master/lib/lexer_mdx.mll

emillon: Ah nice, I'll try that!

emillon: Done. I moved the lexbuf creation inside that function as well since it's always built from a string. how do that look?

gpetiot: It looks better, thanks!

gpetiot (approved):

test/unit/test_literal_lexer.ml
emillon: thanks for the tip!

PR #1491, opened 2020-10-09
Cleanup string and char escapes

emillon:
OCamlformat has to deal with escaping in strings in two cases

- when `escape-strings` is set to anything other than `preserve` (this has been removed in #1462)
- when `Source.string_literal` cannot recover the original string in the source code. In that case, a decimal escape is used.

This PR is based on the observation that the latter case is never used. It seems to never actually be the case (in the test suite and on `test_branch`, including with older compilers, ). So we can make `Source.string_literal` total and delete the corresponding dead code.

In technical terms, the new assumption is that the tokens found in `Pconst_string(_, _, None)` (double quoted strings) start by a `STRING` token.

Once we require a OCaml >= 4.11 parser we can use the payload loc directly.

The char situation is similar but with less dead code associated.


hhugo:
`Source.string_literal` will not work in cases where you don't have the source. I think we had such a case with the reason support (now removed)

hhugo (approved):

emillon:
Great, thanks for the explanation. I was puzzled as to why we wouldn't get the source code.

PR #1497, opened 2020-10-12
Testing emacs integration

gpetiot:
It would be nice to have some kind of tests for that considering the many changes this file is getting these days.
So far I'm having some troubles writing the `test.el` file since I'm not particularly versed in the emacs shenanigans, any help is welcomed!
cc @bcc32 @juxd @hhugo @jberdine

### DIFF
--- a/test/emacs/dune
+++ b/test/emacs/dune
@@ -1,0 +1,2 @@
+(cram
+ (deps %{bin:ocamlformat} %{bin:emacs} test.el ../../emacs/ocamlformat.el))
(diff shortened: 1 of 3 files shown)


bcc32:
Do you know about [Elisp Regression Tests](https://www.gnu.org/software/emacs/manual/html_node/ert/index.html)?  It might be useful to use that library to define the tests.  Then you can use a simple [command](https://www.gnu.org/software/emacs/manual/html_node/ert/Running-Tests-in-Batch-Mode.html#Running-Tests-in-Batch-Mode) to run the tests from a dune `test` caluse.

gpetiot:
I didn't know about this, thanks for the pointers, I will have a look!

PR #1499, opened 2020-10-12
Normalization of code spans in docstrings: do not change the spaces within the code spans

### LINKED ISSUE #1494: Bug: double spaces in document code blocks are not preserved
The following diff is suggested by `ocamlformat.0.15.0`, with the `parse-docstrings` option:

```diff
-(** [trim "  "] is [""] *)
+(** [trim " "] is [""] *)
```
Here, Odoc changes the string inside the code block, which changes the semantics of the example being documented. I suspect that code segments in documentation comments should never have their AST changed by formatting (just as is true of the file AST itself).

This seems likely to be an issue with Odoc itself, but I'm reporting here first in case there are any subtleties to OCamlformat's usage of the parser.

gpetiot:
fix #1494, no diff with test_branch.sh


hhugo:
I don't understand how this is supposed to work. What would happened with the following ?
```ocaml
(** [abc '"'     '"'] *)
```

emillon:
I agree with @hhugo . It seems that we have two choices: either we don't touch what's inside `[ ]` (maybe trimming whitespace around), or we interpret that as code, and we'll need a parser. I'm not super familiar with odoc semantics but I don't think code are particularly meant to be interpreted as ocaml code. In particular, `[ ... ]` is documented as "source code style" while `{[ ... ]}` is "preformatted source code style", so I don't think we should try to interpret these.

gpetiot:
Yes we can expect anything in `[...]` so we shouldn't parse it as ocaml code. FTR the function I tried in this PR just splits a string in fragments that are 'in' or out' of double quotes, to distinguish string literals from the rest of the contents, but indeed it doesn't work with Hugo's example.
Let's only make a simpler trimming then.

test/passing/doc_comments.mli.ref
```diff
@@ -327,12 +327,12 @@ end
...
-(** [Multi Line]
+(** [Multi
+     Line]
 
-    [A lot of spaces]
+    [A lot of    spaces]
 
-    [Very
-    looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong] *)
+    [Very looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong] *)
```
emillon, line 335: we now can't break these, but that seems more correct to me

emillon:
I think this kind of comment is interesting to consider (I suppose it's a normal comment at the moment because of this normalization)

https://github.com/ocaml-ppx/ocamlformat/blob/138d2c67151d5f88816e6e2707aabb5c75099fd2/lib/Fmt.mli#L22-L30

With this branch, `[ ]` is still rewritten to `[]` and `[@ ]` to `[@]`. How do we expect these should be written? I'm not sure adding quotes should be necessary. Maybe we should handle this the same way we would handle backticks in markdown, and do nothing with the contents.

gpetiot:
I think in this case it would be acceptable to use double quotes around these strings like `[" "]` since this is the ocaml syntax anyway ,but I have no strong stance I'm fine with not trimming the contents at all as well.

emillon:
Interesting, I just checked the OCaml documentation and it's not using any escapes at all:

https://github.com/ocaml-ppx/ocamlformat/blob/138d2c67151d5f88816e6e2707aabb5c75099fd2/vendor/ocamlformat_support/format_.mli#L1115

https://github.com/ocaml-ppx/ocamlformat/blob/138d2c67151d5f88816e6e2707aabb5c75099fd2/vendor/ocamlformat_support/format_.mli#L1150

So to support that we probably should keep the original string.

emillon (approved):
Seems to be right!

PR #1502, opened 2020-10-14
Preserve short-form extensions for structure item extensions

### LINKED ISSUE #1501: Preserving short-form structure item extensions
Hi,

Structure items can benefit from extension sugar as well, but at the moment the long syntax is always used.

For example, `open%ext M` is always rewritten to `[%%ext open M]`. For consistency with what we did in #1293 for `extension-sugar`, I think that we should emit the concrete syntax from the source file.

gpetiot:
Fix #1501 (since this issue focuses on structure items)

it seems that the deprecated `extension-sugar` extension is taken into account only for expressions, for structure items we always use the sugared syntax, and never for signature items.
The option is now taken into account for structure items.

No diff with test_branch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,7 @@
   + Fix misalignment of cases in docked `function` match (#1498, @gpetiot)
 
   + Preserve short-form extensions for structure item extensions (#1502, @gpetiot)
+    For example `open%ext M` will not get rewritten to `[%%ext open M]`.
 
 #### New features
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,8 @@
 
   + Fix misalignment of cases in docked `function` match (#1498, @gpetiot)
 
+  + Preserve short-form extensions for structure item extensions (#1502, @gpetiot)
+
 #### New features
 
 ### 0.15.0 (2020-08-06)
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -506,7 +506,7 @@ let rec fmt_extension c ctx key (ext, pld) =
                 ( Pstr_value _ | Pstr_type _ | Pstr_exception _
                 | Pstr_open {popen_override= Fresh; _}
                 | Pstr_include _ | Pstr_module _ | Pstr_recmodule _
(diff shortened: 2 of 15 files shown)


test/passing/extensions-indent.ml.ref
```diff
@@ -209,3 +209,87 @@ let () =
...
+
+module type%ext E = P
+
+class%ext x = y
+
+class%ext x = y
+
+class%ext x = y
+
+   and y = z
```
hhugo, line 279: Why is this indented ?

gpetiot: it was an issue with the existing boxes, it's fixed

lib/Fmt_ast.ml
```diff
@@ -501,8 +501,18 @@ let rec fmt_extension c ctx key (ext, pld) =
       str s
   | ( _
     , _
-    , PStr [({pstr_desc= Pstr_value _ | Pstr_type _; _} as si)]
-    , (Pld _ | Str _ | Top) ) ->
+    , PStr
+        [ ( { pstr_desc=
+                ( Pstr_value _ | Pstr_type _ | Pstr_exception _
+                | Pstr_open {popen_override= Fresh; _}
```
emillon, line 507: I was surprised that there was no `open!%ext M`. So I checked the parser, and actually it exists, but it needs to be spelled out `open! %ext M`. This can be fixed a separate PR if you prefer.

gpetiot: No I will add it into this one

CHANGES.md
```diff
@@ -44,6 +44,8 @@
 
   + Fix misalignment of cases in docked `function` match (#1498, @gpetiot)
 
+  + Preserve short-form extensions for structure item extensions (#1502, @gpetiot)
```
emillon, line 47: I think an example would be useful here, for example that `open%ext M` will not get rewritten to `[%%ext open M]`.

emillon (approved):
Looks good (with an example in the changelog)!

CHANGES.md
gpetiot: done

PR #1504, opened 2020-10-21
Implement the Fmt module on top of a small kernel

emillon:
Fmt is a lightweight abstraction on top of `Format` boxes (or rather `Format_` boxes). In its current implementation, we have `type t = Format.formatter -> unit`, and most of the module relies on that. It makes the current abstraction pretty leaky (even if it's just within the module), and hard to experiment with.

This PR defines a small kernel of functions that need to rely on the formatter abstraction, and implements the rest of the module on top of that, for example `list_pn`.

One pattern that has emerged is the need for a `lazy_` combinator, that defers side effects until corresponding bit is actually laid out. A test is added as an example. This missing combinator the part that was blocking previous attempts at modifying this module.

Inside of the module, this uses `Staged.t` in order to make staging explicit, which should make the move to another representation easier, but technically it's not required.

### DIFF
--- a/test/unit/test_fmt.ml
+++ b/test/unit/test_fmt.ml
@@ -8,42 +8,39 @@ let eval_fmt term =
   Format_.pp_print_flush ppf () ;
   Buffer.contents buffer
 
-let join l = List.fold_left l ~init:Fmt.noop ~f:Fmt.( $ )
-
-let map_string s ~f = String.to_list s |> List.map ~f |> join
-
 let tests_lazy =
-  let test name term ~expected =
-    ( "lazy_: " ^ name
+  [ ( "lazy_: not using lazy"
     , `Quick
     , fun () ->
+        let r = ref None in
+        let pp s =
+          r := Some s ;
+          Fmt.str s
+        in
+        let term = Fmt.fmt_if_k false (pp "hello") in
+        let expected = "" in
+        let expected_r = Some "hello" in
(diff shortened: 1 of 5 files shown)


test/unit/test_fmt.ml
```diff
@@ -0,0 +1,49 @@
...
+      (let i = ref 0 in
+       let pp_letter c =
+         let open Fmt in
+         lazy_ (fun () ->
+             Int.incr i ;
+             str (Int.to_string !i) $ char c)
+       in
+       let pp_if_vowel c = Fmt.fmt_if_k (is_vowel c) (pp_letter c) in
+       map_string "hello" ~f:pp_if_vowel)
+      ~expected:"1e2o" ]
```
gpetiot, line 47: These tests are a bit hard to follow, wouldn't it possible to have something simple with just some int or some bool `ref` that is modified before/after unstaging?

gpetiot (approved):
Minor comment on the tests, but otherwise it looks very good, thanks!

test/unit/test_fmt.ml
emillon: That's fair! I tried something closer to how ocamlformat uses that, but it's a bit complicated to read. I tried something simpler, let me know what you think.

gpetiot: It looks better indeed, thanks!

Issue #1505, opened 2020-10-22
ocamlformat does not build on 4.12


emillon:
Hi,

At the moment ocamlformat does not build with the 4.12 branch:

```
File "vendor/parse-wyc/lib/docstrings.ml", line 49, characters 25-47:
49 |   if Warnings.is_active (Warnings.Bad_docstring true) then begin
                              ^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound constructor Warnings.Bad_docstring
File "lib/Parse_with_comments.ml", line 47, characters 10-32:
47 |         | Warnings.Bad_docstring _ when conf.comment_check ->
               ^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound constructor Warnings.Bad_docstring
File "lib/Fmt_ast.ml", line 370, characters 47-50:
370 | let docstring_epi ?(standalone = false) ?next ?epi ~floating =
                                                     ^^^
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.
File "lib/Fmt_ast.ml", line 370, characters 41-45:
370 | let docstring_epi ?(standalone = false) ?next ?epi ~floating =
                                               ^^^^
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.
File "lib/Fmt_ast.ml", line 370, characters 20-38:
370 | let docstring_epi ?(standalone = false) ?next ?epi ~floating =
                          ^^^^^^^^^^^^^^^^^^
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.
```

Will send PRs to fix these.

emillon:
Now passing!

PR #1507, opened 2020-10-22
Add `is_unexpected_docstring` to compat

emillon:
In 4.12, this warning has been renamed to `Unexpected_docstring`, so we add this to the compat layer.

The vendored `Docstrings` in pwyc refers to that in an unused function, so we can just remove it.

Part of #1505


gpetiot (approved):
Looks good to me!

PR #1509, opened 2020-10-22
Do the partition of comments with CmtSet.partition

gpetiot:
This is the first step of #1193, distinguishing the structural improvements from the formatting improvements, no behavioral change. (each commit can be reviewed separately)

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -78,14 +78,17 @@ module CmtSet : sig
   val to_list : t -> Cmt.t list
   (** ordered by start location *)
 
-  val empty : t
-
   val is_empty : t -> bool
 
   val split : t -> Location.t -> t * t * t
   (** [split s {loc_start; loc_end}] splits [s] into the subset of comments
       that end before [loc_start], those that start after [loc_end], and
       those within the loc. *)
+
+  val partition :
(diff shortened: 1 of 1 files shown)


lib/Cmts.ml
```diff
@@ -87,69 +118,45 @@ end = struct
...
+        List.group cmtl ~break:(fun l1 l2 ->
+            not (is_adjacent src (Cmt.loc l1) (Cmt.loc l2)))
+      with
+      | [cmtl] when is_adjacent src (List.last_exn cmtl).loc next ->
+          let open Location in
+          let same_line_as_prev l =
+            prev.loc_end.pos_lnum = l.loc_start.pos_lnum
+          in
+          let decide loc =
+            match
```
emillon, line 135: I'm not too convinced by the pattern matching on differences. What do you think about matching against the boolean result (`match ... = ..., ... = ... with true, true -> ...`).

gpetiot: I made it is this because I have future plans of matching on a difference of 1, as for now there are issues with things like:
```ocaml
let foo = foooooooooo
(* comment not fitting on the previous line *)

let bar = bar
```

as the comment would be wrongfully attached to bar instead of foo and thus being moved one line lower (right in the middle)

emillon: ok, let's go with integers then!

PR #1510, opened 2020-10-23
Ppxlib: replace some map uses by fold and iter

emillon:
In most places, we are not interested by the return value and just using `map` for the side effects. These cases can be adapted to use `iter` instead.

In some places, this removes a good chunk of code that was used only to build tree fragments that were only ignored later.

Going further, most places only use side effects to build a list or a map. These can be adapted to use `fold`. This makes state management more explicit in parse-wyc.

### DIFF
--- a/lib/Loc_tree.ml
+++ b/lib/Loc_tree.ml
@@ -48,5 +48,5 @@ let fold src =
 
 (** Use Ast_mapper to collect all locs in ast, and create tree of them. *)
 let of_ast fragment ast src =
-  let locs = Mapper.fold_ast fragment (fold src) ast [] in
+  let locs = Traverse.fold fragment (fold src) ast [] in
   (of_list locs, locs)
--- a/lib/Loc_tree.ml
+++ b/lib/Loc_tree.ml
@@ -13,8 +13,6 @@ open Migrate_ast
 open Parsetree
 include Non_overlapping_interval_tree.Make (Location)
 
-let cons_option xo xs = match xo with Some x -> x :: xs | None -> xs
-
 let fold src =
   object
     inherit [Location.t list] Ppxlib.Ast_traverse.fold as super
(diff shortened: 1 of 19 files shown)


emillon:
cc @NathanReb

lib/Normalize.ml
```diff
@@ -43,13 +42,11 @@ let dedup_cmts fragment ast comments =
...
-              docs := Set.add !docs (Cmt.create ("*" ^ doc) pexp_loc) ;
-              atr
-          | _ -> super#attribute atr
+              Set.add docs (Cmt.create ("*" ^ doc) pexp_loc)
+          | _ -> super#attribute atr docs
       end
     in
-    Mapper.map_ast fragment mapper ast |> ignore ;
-    !docs
+    Mapper.fold_ast fragment iter ast (Set.empty (module Cmt))
```
emillon, line 49: @NathanReb suggested to rename `Mapper` to `Traverse`

lib/Loc_tree.ml
```diff
@@ -13,39 +13,40 @@ open Migrate_ast
...
+      if Ast.Attr.is_doc attr then locs else super#attribute attr locs
+
+    (** Ast_traverse recurses down to locations in stacks *)
+    method! location_stack _ l = l
+
+    method! expression e locs =
+      let extra_loc =
+        match e.pexp_desc with
+        | Pexp_constant _ -> Some (Source.loc_of_expr_constant src e)
+        | _ -> None
```
gpetiot, line 44: I would find it easier to read to return the new list of locations instead of returning a location option, so we wouldn't need cons_option something like:
```
super#expression
@@
match e.pexp_desc with
| Pexp_constant _ -> Source.loc_of_expr_constant src e :: locs
| _ -> locs
```

and something similar for pattern

gpetiot (approved):
Minor comment about the style, otherwise it looks good to me, thanks!

lib/Loc_tree.ml
emillon: I got rid of the option. There's still some work to do because one function returns an option but I used something in between!

gpetiot: Alright it was just about the style anyway, it looks good otherwise!

Issue #1508, opened 2020-10-22
Tests do not pass on 4.12


emillon:
Even with the fixes linked in #1505, `dune runtest` doesn't pass.
This is not an exhaustive list, but for example:

Warning syntax changed (can be polyfilled with a postprocessor):

```diff
-Warning 47: illegal payload for attribute 'ocamlformat'.
+Warning 47 [attribute-payload]: illegal payload for attribute 'ocamlformat'.
```

Some comments move:

```diff
-let {a (*a*) : a} = e
+let {a : a (*a*)} = e
```

Functor sugar is not recognized:

```diff
-  module Make : functor (TV : TYPEVIEW with type t = t) -> USERCODE(TV).F
+  module Make (TV : TYPEVIEW with type t = t) : USERCODE(TV).F
```

emillon:
N.B. when we close this issue, let's not forget to update the opam metadata.

gpetiot:
@emillon do you still have non-passing tests?

emillon:
It's all good now!

PR #1511, opened 2020-10-26
Add symbolic names when printing warnings

emillon:
This is the new behavior on 4.12. For 4.08 to 4.11 it is possible to map the string in the report value, but for 4.06 and 4.07 we need to parse the string that is being printed.

See #1508


emillon:
Warning messages change with this PR, should this go in the changelog?

gpetiot (approved):
Looks good to me, indeed we can add an entry in the changelog to say we're synchronizing the warning messages pre-4.12 to be up-to-date with 4.12

PR #1514, opened 2020-10-27
Preserve functor syntax for OCaml 4.12

### LINKED ISSUE #1512: Functor syntax is not recognized in 4.12
(Part of #1508)

In 4.12 (more precisely since ocaml/ocaml#8987), some locs have been changed and we can't determine functor concrete syntax.

Here's the diff on the following program:

```ocaml
module Z = functor (_ : sig end) (_ : sig end) (_ : sig end) -> struct end
```

```diff
   [
     structure_item (./f.ml[1,0+0]..[1,0+74])
       Pstr_module
       "Z" (./f.ml[1,0+7]..[1,0+8])
-        module_expr (./f.ml[1,0+11]..[1,0+74])
+        module_expr (./f.ml[1,0+19]..[1,0+74])
           Pmod_functor "_" (./f.ml[1,0+20]..[1,0+21])
           module_type (./f.ml[1,0+24]..[1,0+31])
             Pmty_signature
             []
-          module_expr (./f.ml[1,0+11]..[1,0+74])
+          module_expr (./f.ml[1,0+33]..[1,0+74])
             Pmod_functor "_" (./f.ml[1,0+34]..[1,0+35])
             module_type (./f.ml[1,0+38]..[1,0+45])
               Pmty_signature
               []
-            module_expr (./f.ml[1,0+11]..[1,0+74])
+            module_expr (./f.ml[1,0+47]..[1,0+74])
               Pmod_functor "_" (./f.ml[1,0+48]..[1,0+49])
               module_type (./f.ml[1,0+52]..[1,0+59])
                 Pmty_signature
```

Which corresponds to:

```
module Z = functor (_ : sig end) (_ : sig end) (_ : sig end) -> struct end
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ all module exprs, 4.11
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ module expr 1, 4.12
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ module expr 2, 4.12
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~ module expr 3, 4.12
```

To determine the concrete syntax, we look for a `functor` token within the range of the module expr, which now fails.

gpetiot:
Fix #1512

### DIFF
--- a/lib/Source.ml
+++ b/lib/Source.ml
@@ -114,6 +114,18 @@ let tokens_between t ~filter loc_start loc_end =
 let tokens_at t ~filter (l : Location.t) =
   tokens_between t ~filter l.loc_start l.loc_end
 
+let last_token_before t loc =
+  let loc_start = Lexing.{loc with pos_lnum= 0; pos_bol= 0; pos_cnum= 0} in
+  let s = string_at t loc_start loc in
+  let lexbuf = Lexing.from_string s in
+  lexbuf_set_pos lexbuf loc_start ;
+  let rec loop acc =
+    match Lexer.token lexbuf with
+    | Parser.EOF -> acc
+    | tok -> loop (Some (tok, Location.of_lexbuf lexbuf))
+  in
+  loop None
+
 let find_after t f (loc : Location.t) =
   let pos_start = loc.loc_end in
   let lexbuf =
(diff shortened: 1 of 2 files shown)


lib/Source.ml
```diff
@@ -114,6 +114,18 @@ let tokens_between t ~filter loc_start loc_end =
 let tokens_at t ~filter (l : Location.t) =
   tokens_between t ~filter l.loc_start l.loc_end
 
+let last_token_before t loc =
```
hhugo, line 117: Maybe it's time to tokenize the source only once instead of lexing the same tokens over and over

Julow: I agree, searching into a sorted array will be more efficient than running the lexer all the time.

gpetiot: I agree on the principle, I'm having a look at that but it can take some time before we come up with a better solution, can we merge this one in the meantime so that ocamlformat can run for 4.12 as soon as possible?

emillon: I agree but let's do it for the right reasons - at the moment it does not show up in profiles at all so it's to have a better API rather than for performance. I agree that this shouldn't block merging this one.

lib/Source.ml
```diff
@@ -209,16 +221,27 @@ let is_long_pexp_open source {Parsetree.pexp_desc; _} =
...
   | _ -> false
 
-let is_long_pmod_functor source Parsetree.{pmod_desc; pmod_loc= from; _} =
+let is_long_functor_syntax src ~from = function
+  | Parsetree.Unit -> false
+  | Parsetree.Named ({loc= upto; _}, _) -> (
+      (* before 4.12 the functor keyword is the first token of the functor
+         parameter *)
+      contains_token_between src ~from ~upto Parser.FUNCTOR
+      ||
```
emillon, line 230: Rather than `||`, what do you think of making a different test depending on `Sys.ocaml_version` (which we can expose as a parsed pair in `Parse.parser_version`)? That'll make the seam easier to remove once we require 4.12+.

lib/Source.ml
hhugo: I've created https://github.com/ocaml-ppx/ocamlformat/pull/1515 with a small PoC

lib/Source.ml
gpetiot: Done, it looks better indeed

emillon (approved):

PR #1515, opened 2020-10-28
Lex the entire source once and cache tokens

hhugo:
There is room for improving the complexity and not traverse the list of tokens every time.


hhugo:
It seems the diff in the test is both a regression and an improvement. I'm not sure what's going on.

hhugo:
The diff was due to some weird interaction with Location.compare

gpetiot (approved):
Great, thanks! No diff with test_branch, it looks good to me!

lib/Migrate_ast.ml
```diff
@@ -150,7 +150,13 @@ module Location = struct
 
   let sexp_of_t x = Sexp.Atom (to_string x)
 
-  let compare : t -> t -> int = Poly.compare
+  let compare {loc_start; loc_end; loc_ghost} b =
+    match Position.compare loc_start b.loc_start with
+    | 0 -> (
+      match Position.compare loc_end b.loc_end with
+      | 0 -> Poly.compare loc_ghost b.loc_ghost
```
emillon, line 157:
```suggestion
      | 0 -> Bool.compare loc_ghost b.loc_ghost
```

We might as well get rid of this one too. FWIW, this function is showing up a lot in the profiles so I'm glad it's not using poly compare anymore

emillon (approved):

hhugo:
The complexity increase is noticeable. It increases the time to format test/passing/source.ml by x3.
I'll open a PR with a fix shortly.

hhugo:
fixed in #1530

PR #1516, opened 2020-10-29
opam metadata: bump the bound to < 4.13

emillon:
The tests do not pass for now but ocamlformat can be built.

This will also make sure OCaml health check knows about 4.12
compatibility in master.


kit-ty-kate:
All green.

PR #1517, opened 2020-10-30
Comments of type constrained label in record pattern have to be relocated in 4.12

### LINKED ISSUE #1513: Some comments are attached differently in 4.12
Hi,

For this program:

```ocaml
let {a (*a*) : a} = e
```

The locs are different in 4.12 (since ocaml/ocaml#8987):

```diff
           pattern (./c.ml[1,0+4]..[1,0+17])
             Ppat_record Closed
             [
-              "a" (./c.ml[1,0+5]..[1,0+6])
-                pattern (./c.ml[1,0+5]..[1,0+16])
+              "a" (./c.ml[1,0+5]..[1,0+6]) ghost
+                pattern (./c.ml[1,0+5]..[1,0+16]) ghost
                   Ppat_constraint
-                  pattern (./c.ml[1,0+5]..[1,0+16])
+                  pattern (./c.ml[1,0+5]..[1,0+6])
                     Ppat_var "a" (./c.ml[1,0+5]..[1,0+6])
                   core_type (./c.ml[1,0+15]..[1,0+16])
                     Ptyp_constr "a" (./c.ml[1,0+15]..[1,0+16])
```

This causes a diff in the test suite (part of #1508):

```diff
-let {a (*a*) : a} = e
+let {a : a (*a*)} = e
```

The difference in ghostness is due to the fact that when punning is used, [the pattern is not present in the source, only the label is](https://github.com/ocaml/ocaml/pull/8987/files#r329310705). The change seems innocuous but I wonder if it would make sense to make sure that we don't attach comments to ghost locs.

On the other hand, the difference of loc in the `Ppat_constraint` node might cause the comment to be attached elsewhere:

```
let {a (*a*) : a} = e
     ~~~~~~~~~~~       4.11
     ~                 4.12
```

gpetiot:
Fix #1513 no change are required when the pattern is more complicated (other pattern matching cases of fmt_pattern) and same for expressions

The changes of `type_constraint_is_first` are not related to the bugfix but it makes the code easier to read.

### DIFF
--- a/lib/Migrate_ast.ml
+++ b/lib/Migrate_ast.ml
@@ -86,6 +86,8 @@ module Parse = struct
     | Traverse.Structure -> implementation lexbuf
     | Traverse.Signature -> interface lexbuf
     | Traverse.Use_file -> use_file lexbuf
+
+  let parser_version = Ocaml_version.of_string_exn Sys.ocaml_version
 end
 
 module Printast = struct
--- a/lib/Migrate_ast.ml
+++ b/lib/Migrate_ast.ml
@@ -87,7 +87,7 @@ module Parse = struct
     | Traverse.Signature -> interface lexbuf
     | Traverse.Use_file -> use_file lexbuf
 
-  let parser_version = Ocaml_version.of_string_exn Sys.ocaml_version
+  let parser_version = Ocaml_version.sys_version
 end
 
 module Printast = struct
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,8 @@
 
   + Do not change the spaces within the code spans in docstrings (#1499, @gpetiot)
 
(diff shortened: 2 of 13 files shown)


lib/Migrate_ast.ml
```diff
@@ -86,6 +86,8 @@ module Parse = struct
     | Traverse.Structure -> implementation lexbuf
     | Traverse.Signature -> interface lexbuf
     | Traverse.Use_file -> use_file lexbuf
+
+  let parser_version = Ocaml_version.of_string_exn Sys.ocaml_version
```
emillon, line 90:
```suggestion
  let parser_version = Ocaml_version.sys_version
```

emillon (approved):

PR #1523, opened 2020-11-02
Support injectivity type annotations (OCaml 4.12 feature)

### LINKED ISSUE #1444: Support injectivity annotations (4.12)
Injectivity annotations are a new feature in 4.12:

> * #9500, #9727: Injectivity annotations
  One can now mark type parameters as injective, which is useful for
  abstract types:
    module Vec : sig type !'a t end = struct type 'a t = 'a array end
  On non-abstract types, this can be used to check the injectivity of
  parameters. Since all parameters of record and sum types are by definition
  injective, this only makes sense for type abbreviations:
    type !'a t = 'a list
  (Jacques Garrigue, review by Jeremy Yallop and Leo White)

ocaml/ocaml#9500
ocaml/ocaml#9727

gpetiot:
Fix #1444
this requires ppxlib to merge https://github.com/ocaml-ppx/ppxlib/pull/193


lib/Fmt_ast.ml
```diff
@@ -3103,15 +3109,16 @@ and fmt_tydcl_params c ctx params =
     ( wrap_fits_breaks_if ~space:false c.conf
         (List.length params > 1)
         "(" ")"
-        (list params (Params.comma_sep c.conf) (fun (ty, vc) ->
-             fmt_variance vc $ fmt_core_type c (sub_typ ~ctx ty) ) )
+        (list params (Params.comma_sep c.conf) (fun (ty, (vc, inj)) ->
+             fmt_variance vc $ fmt_injectivity inj
```
emillon, line 3113: it's minor but since variance and injectivity are always formatted together, what do you think about passing the tuple to a `fmt_variance_and_injectivity` function?

lib/Fmt_ast.ml
```diff
@@ -2736,7 +2740,9 @@ and fmt_class_expr c ?eol ?(box = true) ({ast= exp; _} as xexp) =
   @@
   match pcl_desc with
   | Pcl_constr (name, params) ->
-      let params = List.map params ~f:(fun x -> (x, Invariant)) in
+      let params =
+        List.map params ~f:(fun x -> (x, (NoVariance, NoInjectivity)))
```
emillon, line 2744: if we pass tuples as suggested in another comment, this tuple could be named default_variance_and_injectivity (or Variance_and_injectivity.default next to a Variance_and_injectivity.fmt). This could use a better name though. Just Variance maybe?

emillon:
Can you add a pin-depends to ocamlformat.opam.template so that we have CI for these PRs?

lib/Fmt_ast.ml
gpetiot: Done

lib/Fmt_ast.ml
gpetiot: Done

gpetiot:
> Can you add a pin-depends to ocamlformat.opam.template so that we have CI for these PRs?

I couldn't find any syntax to setup a pin-depends in the dune-project file by the way.

emillon:
It's not supported in dune-project, but you can add the `pin-depends: ...` (in opam syntax) in a `ocamlformat.opam.template`, regenerate the opam file and it will work.

ocamlformat.opam.template
```diff
@@ -0,0 +1,3 @@
+pin-depends: [
+  "ppxlib" "git+https://github.com/ocaml-ppx/ppxlib.git#ast-4.12"
```
emillon, line 2:
```suggestion
  "ppxlib.dev" "git+https://github.com/ocaml-ppx/ppxlib.git#ast-4.12"
```

ocamlformat.opam.template
```diff
@@ -0,0 +1,3 @@
+pin-depends: [
+  "ppxlib.dev" "git+https://github.com/ocaml-ppx/ppxlib.git#ast-4.12"
```
emillon, line 2:
```suggestion
  "ppxlib.dev" "git+https://github.com/ocaml-ppx/ppxlib.git#8056a3386a0832dee1ab985760321a996e90a442"
```

ocamlformat.opam
```diff
@@ -47,3 +47,6 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocaml-ppx/ocamlformat.git"
+pin-depends: [
+  "ppxlib.dev" "git+https://github.com/ocaml-ppx/ppxlib.git#ast-4.12"
```
emillon, line 51:
```suggestion
  "ppxlib.dev" "git+https://github.com/ocaml-ppx/ppxlib.git#8056a3386a0832dee1ab985760321a996e90a442"
```

emillon:
I fixed the CI; the remaining part is a possible fix in ocaml-ci to support pins in lint-fmt but it doesn't seem necessary.

emillon:
I removed the opam constraint in order to have an ocamlci build and all the tests pass.

PR #1524, opened 2020-11-02
Fix position of comments around list constructor (::) 

### LINKED ISSUE #1521: Bug: comments moved to before (::) application
**Describe the bug**
Before styling:
```
let _ = if a then b :: c (* d *) else e
```
After styling:
```
let _ = if a then (* d *) b :: c else e
```
The comment should stay where it was.

**How to Reproduce**
Run ocamlformat built from 1bf68a70f1480df80a8ab0bd20a799b8e1df0081 with `--profile=janestreet`.

gpetiot:
Fix #1521, no diff with test_branch.sh

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,8 @@
 
   + Fix inconsistencies of the closing parentheses with indicate-multiline-delimiters (#1377, @gpetiot)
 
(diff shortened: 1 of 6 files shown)


emillon (approved):

Julow (approved):
Looks good.

PR #1525, opened 2020-11-03
Fix comments position in extensions

### LINKED ISSUE #1520: Bug: comments moved outside of extensions
**Describe the bug**
Before styling:
```
let _ = [%expr (* comment *) foo]
```
After styling:
```
let _ = (* comment *) [%expr foo]
```
I'd like the comment to stay where it was.

**How to Reproduce**
Styled with ocamlformat built from 1bf68a70f1480df80a8ab0bd20a799b8e1df0081, using `--profile=janestreet`.

gpetiot:
Fix #1520, no diff with test_branch.sh


lib/Cmts.ml
```diff
@@ -247,19 +247,43 @@ let relocate_pattern_matching_cmts (t : t) src tok ~whole_loc ~matched_loc =
...
+          match Source.loc_of_first_token_at src whole_loc PERCENT with
+          | Some loc -> loc
+          | None -> impossible "expect token starting extension" )
+      in
+      let f map =
+        Multimap.partition_multi map ~src:pstr_loc ~dst:whole_loc
+          ~f:(fun Cmt.{loc; _} -> Location.compare loc kwd_loc < 0)
+      in
+      update_cmts t `Before ~f ;
+      update_cmts t `Within ~f
```
emillon, line 284: This is the same pattern as in `relocate_pattern_matching_cmts`  (partition_multi_map based on loc comparison, and update before and within) - can you pull a function for it?

gpetiot: Good idea, done!

emillon (approved):

PR #1526, opened 2020-11-03
Using tokens instead of string manipulation when inspecting the original source

hhugo:
Do less string manipulation, rely on the list of token instead.


hhugo:
The diff in tests is an improvement / a fix

hhugo:
Note that tokens now include COMMENT and DOCSTRING

gpetiot (approved):

gpetiot (approved):
It looks good to me thanks! Maybe you can add a changelog entr and a nice name for the PR for the git log.
No diff with test_branch.sh

PR #1528, opened 2020-11-04
Preserve the syntax of infix set/get operators

### LINKED ISSUE #1500: String and array operator rewriting can change program semantics
Hi,

As of 0.15.0, OCamlformat decides whether to use `a.[i]` or `String.get a i` depending on the context (similarly for `String.set`, `Array.get` and `Array.set`). This works under the assumption that `a.[i]` is parsed as `String.get i`, so both constructs have the same semantics.

However, that's not always true. [If `Clflags.unsafe` is set, the operators are parsed as the unsafe variants](https://github.com/ocaml/ocaml/blob/5e15dd8eb5b3b5f0e66f9166d4de2bcee22aa57d/parsing/parser.mly#L243-L245), e.g. `String.unsafe_get a i`.

Consider the following program:

    let () = String.get "hello world" (-8) |> Char.code |> Printf.printf "%d\n"

Built with the following dune file:

```
(executable
 (name t)
 (flags :standard -unsafe))
```

When executed, it outputs an error:

    Fatal error: exception Invalid_argument("index out of bounds")

This is because `String.get` always does the check.

When reformatting the above program, it gets rewritten to:

    let () = "hello world".[-8] |> Char.code |> Printf.printf "%d\n"

When executed, it outputs `252`, the value of `String_tag` (this is just to show that the program is now unsafe; it could of course cause kind of memory error later in the program).

So, what should we do? Given that the two constructions do not always have the same semantics depending on how the file is compiled, I think that the only valid way to handle the situation is to preserve what can be found in the file.

gpetiot:
Fix #1500, no diff with test_branch.sh

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -158,7 +158,8 @@ module Indexing_op = struct
     | Some (lhs :: args) -> (
       match ident with
       | {pexp_desc= Pexp_ident {txt= ident; loc}; pexp_attributes= []; _}
-        -> (
+      (* We only use the sugared form if it was already used in the source. *)
+        when loc.loc_ghost -> (
         match get_sugar_ident ident args with
         | None -> None
         | Some (op, rhs) -> Some {lhs; op; rhs; loc} )
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
   + Fix parsing of invalid file wrt original source handling (#1542, @hhugo)
 
+  + Preserve the syntax of infix set/get operators (#1528, @gpetiot)
+    `String.get` and similar calls used to be automatically rewritten to their corresponding infix form `.()`, that was incorrect when using the `-unsafe` compilation flag. Now the concrete syntax of these calls is preserved.
+
 ### 0.16.0 (2020-11-16)
 
 #### Removed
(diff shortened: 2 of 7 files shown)


hhugo:
cc @ceastlund, this might also affect formatting of auto-generated files.

ceastlund:
Thanks, @hhugo, but I'm less concerned about this one. I'm fine with `String.get` showing up in our code from time to time.

Julow (approved):
Looks good.

This only works as long as `Clflags.unsafe` isn't set by default in the compiler OCamlformat is compiled with.
Thought, the only change is that operators are changed to explicit calls to the unsafe functions.

lib/Ast.ml
```diff
@@ -158,7 +158,8 @@ module Indexing_op = struct
     | Some (lhs :: args) -> (
       match ident with
       | {pexp_desc= Pexp_ident {txt= ident; loc}; pexp_attributes= []; _}
-        -> (
+      (* We only use the sugared form if it was already used in the source. *)
+        when loc.loc_ghost -> (
```
emillon, line 162: In what cases does `get_sugar_ident` returns None now?

Julow: In the case `ident` is not an indexing operator (`Array.get`, `.()`, etc..) or the number of argument is not right (if there is not enough or too many, it cannot be a "sugar" expression).

Julow: That didn't change. `get_sugar` changed however, it returns None in case of labelled arguments, attributes and now there is this new check.

PR #1529, opened 2020-11-04
Add location of invalid docstring in warning messages

### LINKED ISSUE #1448: Feature request: (optionally) exit with error on bad doc comments
When a bad doc comment is encountered, `ocamlformat` prints a warning, gives up on the comment, and exits with code 0 (success). It would be nice to have the option in `ocamlformat` to fail with an error code in that case, for instance to more easily keep a code base clean of errors in doc comments, especially as `ocamlc` won't complain about ill-formed doc comments either, although `ocamldoc` will.

Repro with ocamlformat version 0.14.1-13-g000da99:
```console
$ echo '(** [ *)' | ocamlformat --impl -
Warning: Invalid documentation comment:
File "<standard input>", line 1, characters 6-6:
End of text is not allowed in '[...]' (code).
(** [ *)
$ echo $?
0
```

gpetiot:
~Now the invalid docstrings are not reported as warnings anymore but make ocamlformat exit with code 1. I made it the default behavior but we can also choose to hide it behind a new option.~ (outdated)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,9 @@
   + Preserve the syntax of infix set/get operators (#1528, @gpetiot)
     `String.get` and similar calls used to be automatically rewritten to their corresponding infix form `.()`, that was incorrect when using the `-unsafe` compilation flag. Now the concrete syntax of these calls is preserved.
 
+  + Fail on invalid docstrings (#1529, @gpetiot)
+    Invalid docstrings are no longer reported as warnings but make ocamlformat exit with code 1.
+
 ### 0.16.0 (2020-11-16)
 
 #### Removed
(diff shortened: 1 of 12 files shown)


Julow (commented):
I wonder if this can break the workflow of some people ? For example, when enabling `parse-docstrings` on a project that contains a lot of comments that don't parse, every comments will now have to be fixed at the same time.
I suspect this breaks parse-wyc, we should keep the warning in this case.

gpetiot:
I reverted some of these changes, so ocamlformat still doesn't fail on invalid docstrings, all the warnings are printed during the parsing phase instead of in the middle of Fmt_ast, and missing locations are added.

Julow (commented):
The cost is to parse each docstrings twice. It's not a huge cost but what are the advantages ?
Showing the location is nice. (grepping for comments is not the best UI)

gpetiot:
True, it made more sense when we would fail on invalid docstrings at the beginning, well I find it slightly better to group the checks when parsing the file. We could memoize the parsed docstrings if the cost is too visible to users

gpetiot:
I moved the checks back where they were to reduce the diff.

Julow (approved):
Looks good.

PR #1530, opened 2020-11-05
Change the complexity of tokens lookup from linear to log(n)

hhugo:
following #1515 

|  | Time |
| --- | --- |
| before 1515 | 0m0.968s |
| after 1515 | 0m2.993s |
| new |  0m0.895s |

### DIFF
--- a/lib/Source.ml
+++ b/lib/Source.ml
@@ -103,17 +103,9 @@ let has_cmt_same_line_after (t : t) (loc : Location.t) =
     String.is_prefix str ~prefix:"(*"
 
 let find_first_token_after t pos =
-  let comp c = Position.compare pos (snd t.tokens.(c)).Location.loc_start in
-  let len = Array.length t.tokens in
-  let rec loop a b =
-    match b - a with
-    | 0 -> a
-    | 1 -> if comp a <= 0 then a else b
-    | _ -> (
-        let c = (a + b) / 2 in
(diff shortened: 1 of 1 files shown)


gpetiot (approved):
Looks good to me, thanks!

emillon:
Thanks for the optimization. This seems to work but I'm not completely comfortable with having a custom binary search here. Do you think this can be made to work with `Array.binary_search` or using a Set to do the search (this compares start and end pos so the latter might be tricky)?

hhugo:
Now using `Array.binary_search`

gpetiot (approved):

lib/Source.ml
```diff
@@ -12,13 +12,13 @@
...
 (** Concrete syntax. *)
-type t = {text: string; tokens: (Parser.token * Location.t) list}
+type t = {text: string; tokens: (Parser.token * Location.t) array}
 
 let create text =
   let lexbuf = Lexing.from_string text in
   let rec loop acc =
     match Lexer.token lexbuf with
-    | Parser.EOF -> List.rev acc
+    | Parser.EOF -> Array.of_list (List.rev acc)
```
emillon, line 21:
```suggestion
    | Parser.EOF -> Array.of_list_rev acc
```

emillon (commented):
Nice, thanks!

PR #1531, opened 2020-11-05
Remove extension from executable in error messages

emillon:
This ensures that on Windows the messages start with "ocamlformat: ..." and not "ocamlformat.exe: ..." (these appear in tests for example).


gpetiot (approved):
Looks good to me, I think we can still add a changelog entry, maybe in a Windows section?

PR #1532, opened 2020-11-05
Use last_token_before instead of position_before to help place comments

gpetiot:
Use the token based approach instead of testing the strings, no diff with test_branch.sh

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -55,9 +55,10 @@ let is_adjacent src (l1 : Location.t) (l2 : Location.t) =
           && Position.column l1.loc_start < Position.column l2.loc_start
       | _ -> false )
 
-(** Whether the symbol preceding location [loc] is an infix symbol or a
-    semicolon. If it is the case, comments attached to the following item
-    should be kept after the infix symbol. *)
+(** Whether the symbol preceding location [loc] is an infix symbol
+    (corresponding to [Ast.String_id.is_infix]) or a semicolon. If it is the
+    case, comments attached to the following item should be kept after the
+    infix symbol. *)
 let infix_symbol_before src (loc : Location.t) =
   match
     Source.find_token_before src ~filter:(function _ -> true) loc.loc_start
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -59,15 +59,16 @@ let is_adjacent src (l1 : Location.t) (l2 : Location.t) =
     semicolon. If it is the case, comments attached to the following item
     should be kept after the infix symbol. *)
 let infix_symbol_before src (loc : Location.t) =
-  let pos_cnum = loc.loc_end.pos_cnum - 1 in
-  let loc_end = {loc.loc_end with pos_cnum} in
-  match Source.position_before src loc_end with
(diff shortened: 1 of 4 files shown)


emillon (approved):

lib/Cmts.ml
```diff
@@ -137,7 +136,7 @@ end = struct
               , next.loc_start.pos_lnum - loc.loc_end.pos_lnum )
             with
             | 0, 0 -> impossible "already tested by previous if"
-            | 0, _ when infix_symbol_before src prev -> `Before_next
+            | 0, _ when infix_symbol_before src loc -> `Before_next
```
hhugo, line 139: What is this diff about ?

lib/Cmts.ml
```diff
@@ -59,15 +59,14 @@ let is_adjacent src (l1 : Location.t) (l2 : Location.t) =
...
-  match Source.position_before src loc_end with
-  | Some loc_start ->
-      if loc_start.pos_cnum < loc.loc_end.pos_cnum then
-        let str = Source.string_at src loc_start loc.loc_end in
-        String.equal str ";" || Ast.String_id.is_infix str
-      else false
-  | None -> false
+  match Source.last_token_before src loc.loc_start with
+  | Some
+      ( ( SEMI | INFIXOP0 _ | INFIXOP1 _ | INFIXOP2 _ | INFIXOP3 _
```
hhugo, line 64: Can you explain this list of token ?

hhugo:
This PR need to be rebased now that #1526 has been merged

hhugo:
we should add this PR to the changelog entry 
`Using tokens instead of string manipulation when inspecting the original source (#1526, #1533)`

lib/Cmts.ml
gpetiot: the former implementation was using the location of the `prev` element (before the comment) and we then looked for the position of the last token at this position; and now we look for the token preceding the location `loc` (which is the location of the comment)

lib/Cmts.ml
gpetiot: I added a comment saying it corresponds to `Ast.String_id.is_infix`

lib/Cmts.ml
```diff
@@ -55,19 +55,21 @@ let is_adjacent src (l1 : Location.t) (l2 : Location.t) =
...
-  let loc_end = {loc.loc_end with pos_cnum} in
-  match Source.position_before src loc_end with
-  | Some loc_start ->
-      if loc_start.pos_cnum < loc.loc_end.pos_cnum then
-        let str = Source.string_at src loc_start loc.loc_end in
-        String.equal str ";" || Ast.String_id.is_infix str
-      else false
-  | None -> false
+  match
+    Source.find_token_before src ~filter:(function _ -> true) loc.loc_start
```
hhugo, line 64: Should you filter out comments and docstrings here ? or not ?

gpetiot: It doesn't make a difference, for comments because in this function we are in the process of placing them so we are reasoning about comments location we don't ant to filter them and they are considered as location groups anyway; and docstrings are not allowed in the middle of expressions

PR #1533, opened 2020-11-05
Detect empty line looking by looking at tokens

hhugo:
using tokens instead of string manipulation.
Based on #1526

### DIFF
--- a/lib/Migrate_ast.mli
+++ b/lib/Migrate_ast.mli
@@ -45,8 +45,6 @@ module Position : sig
   val distance : t -> t -> int
 
   val compare : t -> t -> int
-
-  val to_string : t -> string
 end
 
 module Location : sig
--- a/lib/Migrate_ast.mli
+++ b/lib/Migrate_ast.mli
@@ -45,6 +45,8 @@ module Position : sig
   val distance : t -> t -> int
 
   val compare : t -> t -> int
+
+  val to_string : t -> string
 end
 
 module Location : sig
@@ -140,7 +142,7 @@ end
 module Parser = Token_latest
 
 module Lexer : sig
-  val token : Lexing.lexbuf -> Parser.token
+  val tokens : string -> (Parser.token * Location.t) array
 
   type error
 
--- a/lib/Source.ml
+++ b/lib/Source.ml
@@ -14,19 +14,7 @@ open Migrate_ast
 (** Concrete syntax. *)
 type t = {text: string; tokens: (Parser.token * Location.t) array}
 
-let create text =
-  let lexbuf = Lexing.from_string text in
-  let rec loop acc =
-    match Lexer.token lexbuf with
-    | Parser.EOF -> Array.of_list_rev acc
-    | Parser.EOL -> loop acc
-    (* The location in lexbuf are invalid for comments *)
-    | Parser.COMMENT (_, loc) as tok -> loop ((tok, loc) :: acc)
-    | Parser.DOCSTRING ds as tok ->
-        loop ((tok, Docstrings.docstring_loc ds) :: acc)
-    | tok -> loop ((tok, Location.of_lexbuf lexbuf) :: acc)
-  in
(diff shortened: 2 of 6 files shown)


lib/Source.ml
```diff
@@ -302,31 +309,6 @@ let char_literal t (l : Location.t) =
...
-  let rec ends_line_ cnum =
-    if cnum >= len then true
-    else
-      match t.text.[cnum] with
-      | '\n' | '\r' -> true
-      | c when Char.is_whitespace c -> ends_line_ (cnum + 1)
-      | _ -> false
-  in
-  if Location.compare Location.none l = 0 then false
-  else ends_line_ l.loc_end.pos_cnum
```
gpetiot, line 328: I understand the benefit to distinguish between string or token based functions, but this produces some noise in the diff, maybe they can be rewritten with a token analysis later anyway.

hhugo: you can review commit by commit

gpetiot: Yes but these commits will be squashed when merging

hhugo: done

lib/Source.ml
```diff
@@ -116,6 +83,17 @@ let tokens_between (t : t) ~filter loc_start loc_end =
...
       loop i []
 
+let empty_line_between (t : t) p1 p2 =
+  let l = tokens_between t ~filter:(function _ -> true) p1 p2 in
+  let rec loop prev_lnum l =
+    match l with
+    | [] -> p2.pos_lnum - prev_lnum > 1
+    | (_, x) :: xs ->
+        x.Location.loc_start.pos_lnum - prev_lnum > 1
+        || loop x.loc_end.pos_lnum xs
```
gpetiot, line 93: If there are tokens in the line between it can still return true with the recursive calls, we need to return false and stop if there are tokens on the line between.
There is a large diff with test_branch.sh with the conventional profile.

hhugo: Isn't it the semantic previously implemented by `&& List.exists (lines_between t p1 p2) ~f:is_line_empty`

gpetiot: Yes it is

hhugo: The locations of COMMENTs where wrong, fixed

gpetiot:
Here is the diff with test_branch (profile conventional):
```diff
diff --git a/configure.ml b/configure.ml
index c0e156b9..d4924b6a 100644
--- a/configure.ml
+++ b/configure.ml
@@ -18,15 +18,18 @@ let () =
   let dir_of_string s =
     if Filename.is_relative s then Filename.concat (Lazy.force cwd) s else s
   in
+
   let set_libdir s =
     let dir = dir_of_string s in
     library_path := Some [ dir ];
     library_destdir := Some dir
   in
+
   let set_mandir s =
     let dir = dir_of_string s in
     mandir := Some dir
   in
+
   let args =
     [
       ( "--libdir",
@@ -37,6 +40,7 @@ let () =
         "DIR where man pages are installed are for the default build context" );
     ]
   in
+
   let anon s = bad "Don't know what to do with %s" s in
   Arg.parse (Arg.align args) anon
     "Usage: ocaml configure.ml [OPTIONS]\nOptions are:";
```
so this time the issue is with the IN tokens

hhugo:
There is something fishy with shebang at the beginning of the files.. looking

hhugo:
Fixed. We were resetting Location after lexing shebang.

hhugo:
no diff with test_branch

lib/Migrate_ast.mli
```diff
@@ -45,6 +45,8 @@ module Position : sig
   val distance : t -> t -> int
 
   val compare : t -> t -> int
+
+  val to_string : t -> string
```
gpetiot, line 49: It looks like this isn't used

gpetiot (approved):
Minor comment, but otherwise looks good to me, thanks!

Issue #1534, opened 2020-11-05
Bug: formatting invalid files can fail because of incoherent locations.


hhugo:
**Describe the bug**

When recovering an invalid file, we generate a new valid source but feed the original source to `Source.create`.
Attempting to inspect the source will either return random information or fail.

Note that feeding the recovered source to `Source.create` should work ~but will result in invalid locations for warnings~. 

**How to Reproduce**
```sh
cat y.ml
```
```ocaml
let x = 1 +
;;

let x = "\064"
```
```sh
$ ocamlformat y.ml --enable-outside-detected-project --format-invalid-files=auto
Warning: /home/hheuzard/y.ml is invalid, recovering.
ocamlformat: Cannot process "/home/hheuzard/y.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: unhandled exception. Use [--debug] for details.
```

hhugo:
If we had control over tokens fed to the parser, we could replace the tokens part of invalid sections and keep other locations intact.

PR #1535, opened 2020-11-05
fix invalid source handling .. (quick fix, invalid locations in warnings)

hhugo:
This is a quick fix for #1534 

- The addition of the `;;` should only be done if present in the original source
- This gives bad location for warnings


hhugo:
superseded by https://github.com/ocaml-ppx/ocamlformat/pull/1542

PR #1536, opened 2020-11-05
Fix invalid source handling (with correct locations)

hhugo:
Alternative to #1535, as proposed in #1534 

It works by wrapping the lexer and emitting alternative tokens for invalid items
It rely on some ppxlib internals and compiler-libs, so it cannot be merged as is.

Consider this PR as a PoC. I'll let other developers  decide what to do with this.


hhugo:
superseded by https://github.com/ocaml-ppx/ocamlformat/pull/1542

Issue #1503, opened 2020-10-15
Many Cmts.fmt_* calls are not tested


emillon:
## Presentation

This issue tracks "comment coverage", a test metric defined as follows:
Consider all places in `Fmt_ast` where `Cmts.fmt*` functions are called. Such a call site has comment coverage if when the test suite runs, at least a comment is laid out by this call site (ie, the list of comments at the corresponding position is non-empty at least once).

A call site with no comment coverage can correspond to two issues:

- the comments could be in a different position from what we expect, or trigger a divergence error when some are actually present.
- the call site could be dead, in the sense that no comment can be at that location.

Because of these issues, it seems that 100% comment coverage is a valuable target (though probably low priority).

On the tested version (298dafa666835e96d799f38a4d3c4bd2199591f1), out of 158 call sites, only 91 had comment coverage (58%). The list is available below to track progress on this issue.

## Methodology

Here is how the results were gathered:

- the `Cmts.fmt_*` functions get a new `here:Lexing.position` argument
- the call sites in `Fmt_ast` are patched so that `~here:[%here]` is added
- the library gets preprocessed with `ppx_here`, which replaces this extension by the corresponding loc
- in the nonempty case of `fmt_cmts` (called by all comment formatting functions), a function logs a line with the right format:

```ocaml
let log_loc ~(here : Lexing.position) =
  let path = "/tmp/found.txt" in
  Stdio.Out_channel.with_file path ~append:true ~f:(fun oc ->
      let filename = here.pos_fname in
      let line = here.pos_lnum in
      let column = here.pos_cnum - here.pos_bol in
      Stdio.Out_channel.fprintf oc "%s:%d:%d\n" filename line column)
```

Running the test suite creates a `/tmp/found.txt` file with all nonempty calls. It is deduplicated using `sort -u` to get unique results, saved as `found-unique.txt`.

On the other hand, we can get a list of call sites using `git grep -n --column %here |cut -d: -f 1-3 |sort -u > callsites-raw.txt`. The column numbers need to be adjusted manually, 2 must be subtracted. The result is saved to `callsites.txt`.

Then we can `diff -Nru found-unique.txt callsites.txt`. The diff reads as follows:

- context lines, and lines not in the diff, are call sites with comment coverage.
- lines with a `+` are call sites that are not hit dynamically. In other words, call sites with no comment coverage. Some comments can be found here and they can be ignored.
[...]

PR #1538, opened 2020-11-06
Automatic testing of comments

Julow:
Related issue: https://github.com/ocaml-ppx/ocamlformat/issues/1503

Finding a test case for every call sites is not possible: some calls are unreachable and others are reachable in specific cases that are very hard to guess due to side effects and unspecified internals.

This PR tries to fix this in an automated way. For each tokens in a source file, a comment is inserted (then removed), formatting is called and the diff is printed.
I added a sample containing a big part of the language, more is still needed.

This finds:
- a lot of inserted newlines (sometimes 2)
- moved comments that can be considered bugs
- dropped comments

There is also a lot of false-positive:
- wrapping because of the margin on long lines (the sample file could be improved)
- moved comments that are definitely not bugs (eg. `M.(* comment *)N`)

The gained coverage is not huge (116/154 covered vs 91 on `master`) and we still cannot say which are unreachable.

The diff algorithm is from https://github.com/craigfe/diff. Other released libraries have incompatible dependencies.


gpetiot:
Is this still useful? I believe we have now enough tests on comments. What should happen to this PR?

Julow:
I don't think we have enough test on comments.
Before closing this, it would be nice to rebase and run this again to have an idea of the current state, which has probably improved.

PR #1540, opened 2020-11-10
Add missing space between parentheses of infix operators in pattern-matching

gpetiot:
Follow up of #1377
I'm not really satisfied with the way parentheses are handled right now, but it sounds like a bad idea to revamp all this just before a release. I also added similar tests in if-then-else, that should be fixed later (but we really need to remove some values of the if-then-else option), but I'm not sure this kind of code is encountered often anyway.
test_branch.sh shows that the diff of https://github.com/realworldocaml/mdx/pull/297/files/6d23cf8f384718ba3b533f668aecb0ff792a3ee8 is fixed on the conventional profile.


test/passing/ite-fit_or_vertical.ml.ref
```diff
@@ -149,3 +149,13 @@ let foo =
   else if Poly.(assoc_of_prec prec_ast = which_child && which_child <> Non)
   then
     foo
+
+let _ =
+  if fooo then (
```
emillon, line 154: these ones are not too great. what was the situation for this code in 0.15.0?

gpetiot: I observe the same results with 0.15.0

PR #1541, opened 2020-11-10
More tokens

hhugo:
based on #1533
- no longer expose Source.sub
- implement begins_line, ends_line by looking at tokens
- fix the detection of `begin` keywork


gpetiot (approved):
Looks great, thanks!

PR #1542, opened 2020-11-11
Fix parsing of invalid file wrt original source handling

### LINKED ISSUE #1534: Bug: formatting invalid files can fail because of incoherent locations.

hhugo:
fix #1534
There are multiple parts here:
- move all the parsing + lexing logic in the same place: `Parse_with_comments.ml`
- tie `ast` and `Source.t` together in `Parse_with_comments`, we want location in the ast and the source to be coherent.
- output `;;` after an invalid fragment if it was present in the source


test/passing/format_invalid_files_with_locations.ml.opts
```diff
@@ -0,0 +1 @@
+--format-invalid-files=auto -n 3
```
gpetiot, line 1: We usually use `--max-iter` instead of `-n` we should keep it that way to grep everything in one go

gpetiot (commented):
Looks good to me, thanks! I appreciate the simplifications in all the parse_with_comments code.
Maybe wait until 0.16.0 is released to merge this one though, since we are freezing master now for the release (cc @emillon )

hhugo:
will wait for 0.17

hhugo:
@emillon, Can you update `CHANGES.md` with the 0.16 release ? I'm rebase and merge this PR then.

PR #1544, opened 2020-11-13
Fix formatting of field override with constraint

### LINKED ISSUE #1537: Bug: Invalid AST on field override with constraint
```ocaml
let _ = {< x = (x : t) >}
```

This formats to `let _ = {<x : t>}`, which fails.
This case is explicitly handled (`Pexp_override`) but I think no syntax exist for this.

gpetiot:
Fix #1537, no diff with test_branch.sh


Julow (approved):
Looks good !

Julow (changes requested):
I approved too fast. Here is an input that still fails:
```ocaml
let _ = {< x = x [@a] >}
```
The attribute is removed too.

gpetiot:
Good catch! I fixed the simplified case and added some tests.

test/passing/override.ml.ref
```diff
@@ -0,0 +1,9 @@
+let _ = {<x = (x : t)>}
+
+let _ = {<x = ((x [@a]) : (t[@b])) [@c]>}
+
+let _ = {<x>}
+
+let _ = {<x = (x [@a])>}
```
Julow, line 7: These parentheses are optional but I think it's fine to keep them for readability. We need to keep track of this kind of decisions we take somewhere.

Julow (approved):
Let's merge !

PR #1546, opened 2020-11-18
Remove ocaml-migrate-parsetree dependency in dune file

gpetiot:


vendor/parse-wyc/lib/dune
```diff
@@ -1,6 +1,6 @@
 (library
  (name parse_wyc)
- (libraries menhirLib ocaml-migrate-parsetree ppxlib)
+ (libraries menhirLib ppxlib)
```
hhugo, line 3: we still directly depend on OMP in parse_wyc ..

gpetiot: oh indeed

Julow (approved):

PR #1547, opened 2020-11-18
Use tokens_between instead of string_between

gpetiot:
Some follow-up on recent work on using tokens in `Source.ml`

There remains 1 use of `Source.string_between` in `Cmts.add_cmts`, but it's to display the string between 2 positions so it makes sense to keep it (I tried to use `ppx_sexp_conv` to have some formatting functions for the tokens but it doesn't seem to be compatible with the current version of `ppxlib` anyway) (on the other hand it's something that is printed in debug mode only, and honestly I've never found that useful to me)

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -48,8 +48,7 @@ let find_at_position t loc pos =
 let is_adjacent src (l1 : Location.t) (l2 : Location.t) =
   match
     Source.tokens_between src l1.loc_end l2.loc_start ~filter:(function
-      | Token_latest.(COMMENT _ | DOCSTRING _) -> false
-      | _ -> true )
+        | _ -> true )
   with
   | [] -> true
   | [(Token_latest.BAR, _)] ->
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -46,14 +46,16 @@ let find_at_position t loc pos =
     [|] character and the first location begins a line and the start column
     of the first location is lower than that of the second location. *)
 let is_adjacent src (l1 : Location.t) (l2 : Location.t) =
-  Option.value_map (Source.string_between src l1.loc_end l2.loc_start)
-    ~default:false ~f:(fun btw ->
-      match String.strip btw with
-      | "" -> true
-      | "|" ->
-          Source.begins_line src l1
-          && Position.column l1.loc_start < Position.column l2.loc_start
-      | _ -> false )
+  match
+    Source.tokens_between src l1.loc_end l2.loc_start ~filter:(function
+      | Token_latest.(COMMENT _ | DOCSTRING _) -> false
+      | _ -> true )
+  with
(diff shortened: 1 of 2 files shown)


lib/Cmts.ml
```diff
@@ -46,14 +46,16 @@ let find_at_position t loc pos =
...
-    ~default:false ~f:(fun btw ->
-      match String.strip btw with
-      | "" -> true
-      | "|" ->
-          Source.begins_line src l1
-          && Position.column l1.loc_start < Position.column l2.loc_start
-      | _ -> false )
+  match
+    Source.tokens_between src l1.loc_end l2.loc_start ~filter:(function
+      | Token_latest.(COMMENT _ | DOCSTRING _) -> false
```
hhugo, line 51: do we need to filter out comments ?

gpetiot: Yes otherwise some comments are lost as they are not considered to be adjacent to some term because there is docstrings or other comments in the way (it fails while formatting on lib/Prec.ml)

hhugo: There is something weird here. This is not what the previous implementation was doing.

emillon:
I can't comment on this particular change but what are you trying to use `ppx_sexp_conv` for? It should be possible to write the converters by hand if needed.

gpetiot:
> I can't comment on this particular change but what are you trying to use `ppx_sexp_conv` for? It should be possible to write the converters by hand if needed.

The idea was to replace string_between with tokens_between for the last remaining usage, but we would need to print the tokens and I don't think there is a formatting function for that yet, that's why I tried ppx_sexp_conv but I admit it was my first try on that, I think it failed because it couldn't generate to sexp_of function for the fields that involve Location.t

emillon:
For that I think it's ok to just write `sexp_of_token` by hand in `Migrate_ast` (you can't really do it at the definition point without doing it 3 times in token_latest). For cases where you have a missing converter and you use a ppx (so, not the case here), one technique is to define an alias (`type location = Location.t`) and use that in the type definition. It will pick `sexp_of_location` in scope.

hhugo:
I think it's fine to keep using string_between for debug purposes. pretty printing tokens would make debug messages harder to read.

lib/Cmts.ml
hhugo: I don't see any failure when i don't filter out comments ...

gpetiot: It's weird indeed, I still had the errors until I tried with a fresh new switch, I didn't find where it was coming from, I removed the filter of comments/docstrings then.

gpetiot:
I feel like the CI failure on 4.06 and 4.07 is due to some erroneous locations somewhere in the ast that were then fixed since 4.08, I'm not sure but there are a lot of loc changes in the changelog.

gpetiot:
Waiting for #1549

PR #1549, opened 2020-11-24
Bump minimal ocaml version requirement to 4.08

gpetiot:
~(it looks like ocaml-ci is switched off)~ fixed!


gpetiot:
All tests are passing, can I merge this one?

emillon (commented):
Sounds good. What about this:

https://github.com/ocaml-ppx/ocamlformat/blob/2327d0e29c810e5e7d568e2684b75d215f62f492/lib/Fmt_ast.ml#L4354-L4356

Is this something that can be simplified given that all supported versions include that commit (it happened between 4.05 and 4.06 so it was the case before this PR too)?

gpetiot:
Maybe this can be simplified but it requires more changes on how the sugared polynewtypes are handled, this whole abstraction looks very messy to me, I wouldn't say no to a massive rewriting of this part in another PR.

hhugo:
What is the strategy for dropping support for old ocaml compilers ? Do we have a policy written down anywhere ?

gpetiot:
> What is the strategy for dropping support for old ocaml compilers ? Do we have a policy written down anywhere ?

ocaml 4.08 is a 1st class version for dune so we aim at supporting it for now, others are best effort, and since we need some 4.08 bugfixes about locations in ocamlformat I think it's the right time to drop 4.06 and 4.07 support.

PR #1550, opened 2020-11-24
Use dune instrumentation backend for bisect_ppx

tmattio:
This PR refactors the test coverage setup to use the newly introduced Dune's instrumentation stanza.

`bisect_ppx` can be used as a backend for Dune's instrumentation, but it is currently incompatible with ocamlformat, since it did not migrate to `ppxlib` yet. We can merge this once the migration of `bisect_ppx` to `ppxlib` is complete: https://github.com/aantron/bisect_ppx/pull/327

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@
 /ocamlformat.install
 /test-extra/code
 /_opam
-/coverage
+/_coverage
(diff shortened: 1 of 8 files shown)


gpetiot (approved):
Looks good to me, thanks!
cc @hhugo that used this feature in the past, in case of last minute comments before merging!

PR #1551, opened 2020-11-25
Fix the disabled structure items with attributes

### LINKED ISSUE #1545: Docstring parsing fails with [@@@ocamlformat "disable"]
**How to Reproduce**
Running
```sh
ocamlformat foo.ml --inplace
```
on this file
```ml
[@@@ocamlformat "disable"]

(** hello *)
let foo = 42
```
produces this error
```
ocamlformat: Cannot process "foo.ml".
  BUG: doc comments changed.
File "_none_", line 1:
Error: Formatting of (** hello *) is unstable (e.g. parses as a list or not depending on the margin), please tighten up this comment in the source or disable the formatting using the option --no-parse-docstrings.
```
when using ocamlformat `0.15.0`.

gpetiot:
fix #1545, no diff with test_branch.sh (alright just working for 4.11+ apparently)

`Source.extend_loc_to_include_attributes` had to also extend the location when the attributes precede the item, and we need to fetch the attributes of structure_item as we will rely on the item's location to write it verbatim when disabled.

I'm not really a fan of having some `[@@@ocamlformat "disable"]` attribute anywhere, personally I would rather only allow to disable the formatting of whole files with .ocamlformat-ignore and that's it, so we can remove this trick and the `maybe_disable` functions everywhere. It looks to me that most usage of this feature is to put the disable attribute at the beginning of file anyway. Any thoughts? (for 1.0 maybe)

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -447,6 +447,32 @@ module Tyd = struct
 end
 
 module Structure_item = struct
+  let attributes itm =
+    match itm.pstr_desc with
+    | Pstr_attribute attr -> [attr]
+    | Pstr_eval (_, attrs) -> attrs
+    | Pstr_value (_, vb) -> List.concat_map vb ~f:(fun x -> x.pvb_attributes)
+    | Pstr_primitive {pval_attributes= attrs; _} -> attrs
+    | Pstr_type (_, t) -> List.concat_map t ~f:(fun x -> x.ptype_attributes)
+    | Pstr_typext {ptyext_attributes= attrs; _} -> attrs
+    | Pstr_recmodule m ->
+        List.concat_map m ~f:(fun x -> x.pmb_expr.pmod_attributes)
+    | Pstr_modtype {pmtd_attributes= attrs; _} -> attrs
+    | Pstr_open {popen_attributes= attrs; _} -> attrs
+    | Pstr_extension (_, attrs) -> attrs
+    | Pstr_class_type ct -> List.concat_map ct ~f:(fun x -> x.pci_attributes)
+    | Pstr_class c -> List.concat_map c ~f:(fun x -> x.pci_attributes)
+    | Pstr_include
+        {pincl_mod= {pmod_attributes= attrs1; _}; pincl_attributes= attrs2; _}
+     |Pstr_exception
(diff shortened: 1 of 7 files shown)


lib/Fmt_ast.ml
```diff
@@ -4540,7 +4540,7 @@ let fmt_toplevel c ctx itms =
   let break_struct = c.conf.break_struct || is_top ctx in
   let fmt_item c ~last = function
     | `Item i ->
-        maybe_disabled c i.pstr_loc []
+        maybe_disabled c i.pstr_loc (Ast.Structure_item.attributes i)
```
Julow, line 4543: This allows:
```ocaml
let x=unformatted
[@@ocamlformat "disable"]
```

But not:
```ocaml
type t = { x : int }
[@@ocamlformat "field-space=loose"]
```

The first case seems very useful to me, it'll definitely be used. Someone will ask for the second for consistency.

Julow (commented):
Looks good.

lib/Ast.ml
```diff
@@ -455,6 +455,35 @@ module Tyd = struct
 end
 
 module Structure_item = struct
+  let attributes itm =
+    match itm.pstr_desc with
+    | Pstr_attribute attr -> [attr]
+    | Pstr_eval (_, attrs) -> attrs
+    | Pstr_value (_, vb) ->
+        List.map vb ~f:(fun x -> x.pvb_attributes) |> List.join
```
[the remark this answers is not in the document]
gpetiot: done

hhugo:
I've been disabling formatting locally on multiple occasions. I'd be sad to see that feature go

lib/Fmt_ast.ml
gpetiot: The second case already works for me.

Julow: You are right !

hhugo:
I suspect this is not the right fix. Let me have a look.

gpetiot:
> I suspect this is not the right fix. Let me have a look.

@hhugo do you have an idea on how to properly fix this?

gpetiot:
Wrong approach

PR #1553, opened 2020-11-26
Lower bound on menhir version

giltho:
Hi ! This is really a small thing but :
Minimum version for menhir is 20180528. It's when --infer-write-query was added.

For some reason, esy resolved menhir 2017xxx and ocamlformat would not compile..

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -53,7 +53,8 @@
   dune-build-info
   fix
   fpath
-  menhir
+  (menhir
+   (>= 20180528))
   (ocp-indent :with-test)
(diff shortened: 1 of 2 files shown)


gpetiot:
Hi, thanks for your contribution, the dependencies in the opam file are generated from the dune-project file, can you also edit the dune-project file then?

giltho:
Oh right sorry 😅 didn't notice you were using a generated opam file

gpetiot (approved):
Thanks, looks good to me!

PR #1554, opened 2020-11-27
Preserve spacing of toplevel comments

gpetiot:
Motivation:
```ocaml
let _ = assert (foo (bar + baz <= quux))
(* this comment should stay attached to the preceding item *)

let _ = foo
```
was formatted as:
```ocaml
let _ = assert (foo (bar + baz <= quux))

(* this comment should stay attached to the preceding item *)

let _ = foo
```
because the comment was wrongfully attached to the second item instead of the first one.

edit: no diff with test_branch.sh after the unnecessary commits were dropped.

### DIFF
--- a/lib/Migrate_ast.ml
+++ b/lib/Migrate_ast.ml
@@ -180,6 +180,8 @@ module Location = struct
 
   let compare_end_col x y = Position.compare_col x.loc_end y.loc_end
 
+  let line_difference fst snd = snd.loc_start.pos_lnum - fst.loc_end.pos_lnum
+
   let contains l1 l2 = compare_start l1 l2 <= 0 && compare_end l1 l2 >= 0
 
   let width x = Position.distance x.loc_start x.loc_end
--- a/test/passing/module_attributes.ml.ref
+++ b/test/passing/module_attributes.ml.ref
@@ -5,14 +5,11 @@ include struct
 end
 [@warning "item"] [@@warning "structure"]
 
-include M
-[@warning "item"] [@@warning "structure"]
+include M [@warning "item"] [@@warning "structure"]
 
-include (M : S)
-[@warning "item"] [@@warning "structure"]
+include (M : S) [@warning "item"] [@@warning "structure"]
 
-include M (N)
-[@warning "item"] [@@warning "structure"]
+include M (N) [@warning "item"] [@@warning "structure"]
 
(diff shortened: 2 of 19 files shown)


gpetiot:
Waiting for #1549

lib/Migrate_ast.ml
```diff
@@ -180,6 +180,8 @@ module Location = struct
 
   let compare_end_col x y = Position.compare_col x.loc_end y.loc_end
 
+  let line_difference fst snd = snd.loc_start.pos_lnum - fst.loc_end.pos_lnum
```
Julow, line 183: This function doesn't make sense if `fst` is after (both start and end positions) `snd`. I think we should at least test `assert (fst.loc_start.pos_lnum <= snd.loc_start.pos_lnum)`.
Also I'm not sure what this function means if `snd` is inside `fst`, we should probably assert that the result is `>= 0` too.
Could you add some documentation about that ?

test/passing/module_attributes.ml.ref
```diff
@@ -5,14 +5,11 @@ include struct
...
-[@warning "item"] [@@warning "structure"]
+include M [@warning "item"] [@@warning "structure"]
 
-include (M : S)
-[@warning "item"] [@@warning "structure"]
+include (M : S) [@warning "item"] [@@warning "structure"]
 
-include M (N)
-[@warning "item"] [@@warning "structure"]
+include M (N) [@warning "item"] [@@warning "structure"]
```
Julow, line 12: Is this expected ? The diff doesn't look bad but I wonder why it's related to the issue.

Julow (approved):
Looks good.

test/passing/module_attributes.ml.ref
gpetiot: Yes it is, before the structure_item would not always be put into a hvbox, now it is, and it is consistent with the following lines in this test!

Julow: Thanks

lib/Migrate_ast.ml
gpetiot: I agree on the principle, but I've just discovered that this invariant doesn't hold anymore when comments have been relocated, so for now I've just added a docstring for this function and I will investigate more deeply this relocation issue in another PR. It doesn't look blocking for this PR as this is what was already done in master (the difference computation) anyway.

PR #1556, opened 2020-12-02
Fix comments on the same line as prev and next elements

### LINKED ISSUE #1449: Bug: comment spanning lines does not stabilize
**Describe the bug**
```
let xxxxxx =
  let%map (* _____________________________
             __________ *)()            = yyyyyyyy in
  { zzzzzzzzzzzzz }
```
yields
```
ocamlformat: Cannot process "does-not-stabilize.ml".
```

**How to Reproduce**
Using ocamlformat built from 10aabb33581dd528c06493cb31849300186dba95 with `--profile=janestreet`.

gpetiot:
Fix #1449, no diff with test_branch on conventional and janestreet profiles.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -138,7 +138,7 @@ end = struct
               ( loc.loc_start.pos_lnum - prev.loc_end.pos_lnum
               , next.loc_start.pos_lnum - loc.loc_end.pos_lnum )
             with
-            | 0, 0 -> impossible "already tested by previous if"
+            | 0, 0 -> `Before_next
             | 0, _ when infix_symbol_before src loc -> `Before_next
             | 0, _ -> `After_prev
             | _ -> `Before_next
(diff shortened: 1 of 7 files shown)


lib/Cmts.ml
```diff
@@ -138,7 +138,7 @@ end = struct
               ( loc.loc_start.pos_lnum - prev.loc_end.pos_lnum
               , next.loc_start.pos_lnum - loc.loc_end.pos_lnum )
             with
-            | 0, 0 -> impossible "already tested by previous if"
+            | 0, 0 -> `Before_next
```
Julow, line 141: I couldn't track where exactly the exception was caught and how it results in a stabilization error. Do you have an idea ?

gpetiot: This was introduced in #1509 I just thought this case wouldn't be possible (that's just an assert), but the code of #1449 raises this exception, we just hadn't tested it since #1509 has been merged.

Julow: My question was where is this exception caught ? I couldn't find how this results in the exact error in https://github.com/ocaml-ppx/ocamlformat/issues/1449

gpetiot: This exception is not caught, we get "  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: unhandled exception. Use [--debug] for details." when trying to format the code of #1449 with master (the error message in te issue is not up-to-date)

Julow: Thanks

Julow (approved):
Looks good to me.

PR #1557, opened 2020-12-02
ocamlformat.el: add buffer filename in logs

dannywillems:
It is a bit more convenient to see on which file ocamlformat has been applied in logs.


gpetiot (approved):
It looks good to me, thanks! Can you add an entry in CHANGES.md?

gpetiot:
Thanks!

PR #1562, opened 2020-12-04
Fix stack overflow on large string constants

### LINKED ISSUE #1561: Bug: failure on large string constant
**Describe the bug**
Very large string constants cause ocamlformat to fail.
```
ocamlformat: Cannot process "very_large_string_constant.ml".
```

**How to Reproduce**
Run `ocamlformat --profile=janestreet --enable-outside-detected-project --disable-conf-files` on the attached file, renamed to .ml: [very_large_string_constant.txt](https://github.com/ocaml-ppx/ocamlformat/files/5639094/very_large_string_constant.txt)

I'm using ocamlformat built from 6d95522e1eca95c5050836d979deff4b0f5e2cf7 (https://github.com/janestreet/ocamlformat/tree/just-before-omp-2.0), which is a branch based on 1bf68a70f1480df80a8ab0bd20a799b8e1df0081 plus some cherry-picked PRs. That branch is just before the adoption of OMP 2.0. I'm trying to get our code base up to date with that, so I can subsequently switch us to OMP 2.0.

If there's a fix from this in the master branch, please point me to a PR or some commits I can add to my branch to work around this. Thanks!

gpetiot:
Fix #1561, no diff with test_branch.


test/passing/dune.inc
```diff
@@ -1534,6 +1534,16 @@
  (alias runtest)
  (action (diff label_option_default_args.ml.ref label_option_default_args.ml.output)))
 
+(rule
+ (deps .ocamlformat )
+ (action
+   (with-outputs-to large_string.ml.output
+     (run %{bin:ocamlformat} --profile=janestreet %{dep:large_string.ml}))))
```
emillon, line 1541: how long does this test take to run? also I'm wondering if it makes sense to generate the string instead of storing it?

lib/Fmt.ml
```diff
@@ -124,16 +124,17 @@ let list_pn x1N pp =
...
+          | [] -> List.rev acc
+          | [xI] -> aux (xI, (Some prev, xI, None) :: acc) []
+          | xI :: (xJ :: _ as xJN) ->
+              aux (xI, (Some prev, xI, Some xJ) :: acc) xJN
+        in
+        aux (x1, [(None, x1, Some x2)]) x2N
       in
-      list_pn_ x1 x2N
+      with_pp (fun fs ->
+          List.iter l ~f:(fun (prev, x, next) -> eval fs (pp ~prev x ~next)) )
```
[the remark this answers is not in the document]
gpetiot: I did something like this at first but it generated an error on some other tests (comments.ml I think, the comments were not printed in the correct order), I hope if something was wrong with this implementation it would show on the tests or with test_branch, the issue with the minimal interface is that it made it difficult to write list_pn in a tail recursive way

test/passing/dune.inc
gpetiot: It's almost instant for me, and I'm running on a slow vm

lib/Fmt.ml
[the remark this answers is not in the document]
emillon: The thing's that's happening is that side effects were not properly delayed. I'll write a unit test for list_pn instead of having formatting tests fail for that.

The following works:

```ocaml
List.fold_left l ~init:noop ~f:(fun acc (prev, x, next) ->
    acc $ lazy_ (fun () -> pp ~prev x ~next) )
```

test/passing/dune.inc
emillon: What do you think of generating this test, either as a dune rule or as a cram test?

lib/Fmt.ml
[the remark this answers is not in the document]
gpetiot: Indeed it's better, thanks! +1 for some unit tests

test/passing/dune.inc
gpetiot: It looks to be difficult to generate a string that is already formatted (-n 1), maybe that would be enough to have a unit test for list_pn instead of testing the absence of exception of the ml file?

emillon: ha yes I didn't realize the file was already format. let's scratch that idea and keep the huge file. the unit test is complementary for that, it won't check the resources, just the result. so it all seems good!

emillon (approved):

emillon:
I added a test in #1566.

emillon:
This is still segfaulting on ppc64 though, interesting

test/passing/dune.inc
gpetiot: I added some cram test that passes with this branch but not on master, but if we increase again the length of the string we get a segment fault this time, but this one is probably out of our control and machine dependent

emillon: can you repro the segfault locally? getting a stacktrace should give us some info about that

gpetiot: Yes I can reproduce it with `echo "let _ = \"$(printf '%*s' 300000 | sed 's/ /_ _/g')\"" | ocamlformat - --impl -o /dev/null --debug` but I don't get any trace

emillon: OK I'll try if this can be fixed with a different encoding.

emillon: I've got a fix for this, I'll open a PR

emillon: Done in #1568. Then you can write:

```ocaml
       let l =
         let rec aux (prev, acc) = function
          | [] -> acc
           | [xI] -> aux (xI, (Some prev, xI, None) :: acc) []
           | xI :: (xJ :: _ as xJN) ->
               aux (xI, (Some prev, xI, Some xJ) :: acc) xJN
         in
         aux (x1, [(None, x1, Some x2)]) x2N
       in
      List.rev_map l ~f:(fun (prev, x, next) ->
          lazy_ (fun () -> pp ~prev x ~next) )
      |> sequence
```

gpetiot: Thanks, I rebased this one, the CI is now ok!

PR #1563, opened 2020-12-04
Remove the deprecated 'let-open' option

gpetiot:


lib/Fmt_ast.ml
```diff
@@ -2203,10 +2199,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       , e0 ) ->
       let override = is_override flag in
       let let_open =
-        match c.conf.let_open with
-        | `Preserve ->
-            if Source.is_long_pexp_open c.source exp then `Long else `Short
-        | (`Long | `Short | `Auto) as x -> x
+        if Source.is_long_pexp_open c.source exp then `Long else `Short
```
Julow, line 2202: There is still some handling of `` `Auto `` below. Maybe it can be simplified with an `if then else`.

Julow (approved):
Looks good. 65 to go.

lib/Fmt_ast.ml

PR #1566, opened 2020-12-08
Add a unit test for `Fmt.list_pn`

emillon:
This function needs particular care in handling its function argument: it should be applied only in cases when the corresponding elements are laid out.

(Note that other higher order combinators like `opt` do not have this lazy behavior.)

This adds some tests:
- `pp` is applied left to right (this would catch the case where `fold_left` is replaced by `fold_right` for example)
- when the corresponding node is not printed, `pp` is not evaluated


gpetiot (approved):
Looks good, thanks!

PR #1567, opened 2020-12-08
Fix comment position around list cons operator

### LINKED ISSUE #1564: Bug: comment moved to before ( :: ) in named argument
**Describe the bug**
`let _ = f ~x:(a :: b) (* comment *) ~y`
gets rewritten to
`let _ = f ~x:((* comment *) a :: b) ~y`

**How to Reproduce**
Use ocamlformat built from 6d95522e1eca95c5050836d979deff4b0f5e2cf7 (a branch of 1bf68a70f1480df80a8ab0bd20a799b8e1df0081 plus cherry-picked PRs) with `--profile=janestreet`.

gpetiot:
Fix #1564

this removes the sugaring that was made for lists like `x :: y :: z` so there can be some significant diff on expressions like these, but it makes it much more easier to maintain this part of the code (in addition to fixing the comments' position that was messy because of the relocations). Any thoughts on the diff? (extract)

Unnecessary parens are removed:
```diff
diff --git a/examples/minesweeper/minesweeper.ml b/examples/minesweeper/minesweeper.ml
index e55d3b80d..7140990a5 100644
--- a/examples/minesweeper/minesweeper.ml
+++ b/examples/minesweeper/minesweeper.ml
@@ -102,7 +102,7 @@ let cells_to_see bd cf (i, j) =
         if bd.(x).(y).nbm <> 0 then c :: cells_to_see_rec l
         else
           let l1, l2 = relevant (neighbours cf c) in
-          (c :: l1) @ cells_to_see_rec (l2 @ l)
+          c :: l1 @ cells_to_see_rec (l2 @ l)
   in
```

the `::` operator is now indented with the expression instead of the opening parenthesis, and expressions are put below `::` if they break instead of floating on the right (reduces the amount of artificial indentation):
```diff
diff --git a/src/cache/messages.ml b/src/cache/messages.ml
index 4e8f41c9e..d6675a9db 100644
--- a/src/cache/messages.ml
+++ b/src/cache/messages.ml
@@ -37,13 +37,14 @@ let sexp_of_message : type a. version -> a message -> Sexp.t =
   | Lang versions ->
       cmd "lang"
         (Sexp.Atom "dune-cache-protocol"
-        :: (List.map ~f:(fun { major; minor } ->
-                Sexp.List
-                  [
-                    Sexp.Atom (string_of_int major);
-                    Sexp.Atom (string_of_int minor);
-                  ]))
-             versions)
+         ::
+         (List.map ~f:(fun { major; minor } ->
+              Sexp.List
+                [
+                  Sexp.Atom (string_of_int major);
+                  Sexp.Atom (string_of_int minor);
+                ]))
+           versions)
```

of course I tried to just fix the position of comments without having such huge diff, but didn't manage to do so with the way  fmt_infix_op_args is handled, and that's an abstraction I wanted to remove for some time as it artificially complicates the code and the cost/benefit is not clear on this one.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@
 
   + Fix stack overflow on large string constants (#1562, @gpetiot)
 
+  + Fix comment position around list cons operator (#1567, @gpetiot)
+
 #### Changes
 
   + Add buffer filename in the logs when applying ocamlformat (#1557, @dannywillems)
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2006,7 +2006,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         @@ hvbox indent_wrap
              ( fmt_expression c (sub_exp ~ctx x)
              $ fmt "@ "
-             $ hvbox 0
+             $ hovbox 0
                  ( Cmts.fmt c ~pro:noop loc (fmt "::@ ")
                  $ fmt_expression c (sub_exp ~ctx y) )
              $ fmt_atrs ) )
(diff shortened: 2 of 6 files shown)


gpetiot:
Would this kind of diff be alright for you @hhugo @ceastlund @jberdine @samoht @emillon @Julow ?

ceastlund:
I think that diff is acceptable, if it fixes this comment bug.

PR #1568, opened 2020-12-09
Add Fmt.sequence as a generalization for ($)

emillon:
It is specially optimized so that large lists can be formatted without chaining ($), which consumes linear stack space.

It is implemented as a balanced tree of ($), which will take logarithmic stack space.

(this also implement list_k in terms of sequence)


gpetiot (approved):
Thanks! it looks good to me :+1:

PR #1572, opened 2020-12-17
Remove part of comment that is causing merlin to report an error, masking the real errors

gpetiot:
Having `{%ext delim|...|delim}` in the comment is causing an issue with merlin, it reports an error on the pattern-matching even though the matching is fine, and it's preventing from jumping to the real compilation errors, so I just remove this part, and move the part about the comments and attributes to a more suitable place.

(I use merlin.3.4.2)


emillon:
Has this been reported to merlin?

PR #1574, opened 2020-12-19
Precise that ocamlformat is disabled by default

bobot:
It could help setting up a project to use ocamlformat.

Edit the wording as much as you like in this PR. I though to open an issue but it was easier to propose the wording this way.

Additionally the message of the editor commmand `ocamlformat applied` could be replaced by `ocamlformat disabled in this project`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ This can be a hard price to pay, since this means losing the corresponding git h
 If you use a custom configuration, options you rely on might also get removed in
 a later release.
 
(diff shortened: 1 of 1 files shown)


gpetiot (approved):
Thank you for your contribution! I already took the liberty of rephrasing some bits, it looks good to me now.
Regarding the status message of the emacs integration I'm a bit reluctant to add a message saying that we won't do anything it sounds like extra noise to me.

PR #1581, opened 2021-01-04
Consistent positioning of lambda return type annotations when no-break-infix-before-func and pre/post extensions

### LINKED ISSUE #1580: Feature request: consistent positioning of lambda return type annotation
OCamlformat sometimes chooses to position lambda return type annotations after the final argument, as in:

```ocaml
let encoder f =
  let field_encode = unstage (t f.ftype) in
  fun x k : unit -> field_encode (f.fget x) k
```

However, this is broken when the lambda is inside an infix operator application, e.g. with `@@`:

```diff
 |let encoder f =
 |  let field_encode = unstage (t f.ftype) in
-|  staged @@ fun x k : unit -> field_encode (f.fget x) k
+|  staged @@ fun x k -> (field_encode (f.fget x) k : unit)
```

My feeling is that this choice should not be a contextual one (and that the latter diff should not be proposed by OCamlformat), but other opinions may differ.

gpetiot:
Fix #1580, no diff with test_branch.sh

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,8 @@
 
   + Add buffer filename in the logs when applying ocamlformat (#1557, @dannywillems)
 
+  + Consistent positioning of lambda return type annotations when no-break-infix-before-func and pre/post extensions (#1581, @gpetiot)
+
 ### 0.16.0 (2020-11-16)
 
 #### Removed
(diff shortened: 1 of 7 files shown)


Julow (approved):
Looks good !

PR #1586, opened 2021-01-12
Expose a RPC interface through a new binary ocamlformat-rpc and a new library ocamlformat-rpc-lib

gpetiot:
cc @emillon @voodoos @rgrinberg

The current state of this PR is:

We define a new binary `ocamlformat-rpc` that starts a listening loop, accepting the following commands from stdin:
- halt
- format-type T, the formatted type would be printed on stdout
- invalid requests do nothing
So far the version is not part of the protocol, but we can imagine that the client specifies it when initiating the connection with the server, and the server could return ACK/NACK or propose another version.


rgrinberg:
> So far the version is not part of the protocol, but we can imagine that the client specifies it when initiating the connection with the server, and the server could return ACK/NACK or propose another version.

That seems like a good idea. Is the `rpc` library going to be public?

gpetiot:
> That seems like a good idea. Is the `rpc` library going to be public?

Yes I guess we can release this lib as a new `ocamlformat-rpc-lib` package.

bin/ocamlformat-rpc/main.ml
```diff
@@ -0,0 +1,127 @@
...
+              | Ok formatted ->
+                  ignore (Format.flush_str_formatter ()) ;
+                  Ocamlformat_rpc_lib.V1.output stdout (`Format formatted) ;
+                  Stop ()
+              | Error e ->
+                  Translation_unit.Error.print ~fmt:Format.str_formatter
+                    ~exe:"ocamlformat-rpc" ~debug:false ~quiet:false
+                    ~input_name:x e ;
+                  Continue () )
+            [ (fun () -> format Signature x)
```
hhugo, line 82: Note that some input could format successfully with multiple "kind" and that we have no guarantee that they would format the same way.

gpetiot: Yes, I tried to order the formatting functions in such a way that the ones expecting a keyword first (like signatures) are placed before the more general ones (like toplevel phrases), but I try to find ambiguous cases and add them in the tests.

gpetiot:
@emillon there are some issues related to the dune tests, I've listed `ocamlformat-rpc` as a dependency of the tests but I have to invoke it with `../../bin/ocamlformat-rpc/main.exe` in the code, I don't know if there is a way to say these 2 are the same binary, as the dependency is not properly taken into account in the CI.

bin/ocamlformat-rpc/main.ml
hhugo: Aren't you missing Structure here ?

gpetiot: It won't be used when called by ocaml-lsp, that should use types, module types and signatures, and mdx should use only toplevel phrases, and maybe signatures.
And parsing a file as `--impl` in ocamlformat processes it as a use_file (toplevel phrases) anyway.

hhugo: Adding a small comment about this will certainly help future contributers

gpetiot: done!

gpetiot:
This PR fulfills the needs of ocaml-lsp (@voodoos) and mdx (@NathanReb), @emillon @julow do you have some comments on it before I release it for 0.17.1 ?

bin/ocamlformat-rpc/main.ml
```diff
@@ -0,0 +1,172 @@
...
+          Ocamlformat_rpc_lib.Init.output stdout (`Version vstr) ;
+          Out_channel.flush stdout ;
+          rpc_main (Version_defined (v, Conf.default_profile))
+      | None -> (
+        match V.propose_another vstr with
+        | Some v ->
+            let vstr = V.to_string v in
+            Ocamlformat_rpc_lib.Init.output stdout (`Version vstr) ;
+            rpc_main Waiting_for_version
+        | None -> Ok () ) ) )
```
Julow, line 58: What does this `Ok` mean ? I suggest merging `is_handled` and `propose_another`: 
```ocaml
val handshake : string -> [ `Handled of t | `Propose_another of t ]
```

bin/ocamlformat-rpc/main.ml
```diff
@@ -0,0 +1,172 @@
...
+let format fg conf source =
+  let input_name = "<rpc input>" in
+  let opts =
+    Conf.{debug= false; margin_check= false; format_invalid_files= false}
+  in
+  Translation_unit.parse_and_format fg ~input_name ~source conf opts
+
+let rec rpc_main = function
+  | Waiting_for_version -> (
+    match Ocamlformat_rpc_lib.Init.read_input stdin with
```
Julow, line 43: `open Ocamlformat_rpc_lib` at the top ?

lib/Migrate_ast.mli
```diff
@@ -98,6 +98,8 @@ module Traverse : sig
     | Structure : Parsetree.structure fragment
     | Signature : Parsetree.signature fragment
     | Use_file : Parsetree.toplevel_phrase list fragment
+    | Core_type : Parsetree.core_type fragment
+    | Module_type : Parsetree.module_type fragment
```
Julow, line 102: Is this enough ?
For example, formatting expressions would be handy for Mdx: https://github.com/realworldocaml/mdx/pull/316/files#diff-7436767382485ab85681e67eda27b5af703dddc879b9605ec4390bd7042fb47cR113

Julow (approved):

bin/ocamlformat-rpc/main.ml
gpetiot: This means that we have no other version to propose (just ending the main loop), but this should not happen since we should always suggest at least v1 indeed. that's a better handshake function indeed!

lib/Migrate_ast.mli
gpetiot: Done! it should simplify the mdx integration indeed

gpetiot:
I tried several APIs for ocamlformat-rpc-lib but I'm not satisfied with what we've got (I tried moving the client in the exposed api so that the users don't have to define it but types escape their scope everywhere, not sure I'm on the right track), if you have some ideas for improvements I'm open to it @emillon @rgrinberg

lib-rpc/ocamlformat_rpc_lib.ml
```diff
@@ -0,0 +1,197 @@
...
+
+  type cmd
+
+  val pid : t -> int
+
+  val mk : pid:int -> in_channel -> out_channel -> t
+
+  val query : cmd -> t -> cmd
+
+  val halt : t -> (unit, [`Msg of string]) result
```
voodoos, line 35:
```suggestion
  val halt : t -> (unit, [> `Msg of string]) result
```
I would suggest using open sets of variant in the Client's signature.
It allows for more flexible handling of `Result` mapping.

gpetiot: Good idea, done!

PR #1587, opened 2021-01-12
Support more sugared extension points

### LINKED ISSUE #410: Recognise all sugars for extension points.
Also, consider always preserving the syntax of extension points, rewriting between syntaxes can be done outside ocamlformat.

### LINKED ISSUE #1584: Feature request: Option to prevent transform `assert%lwt false` to `[%lwt assert false]`
I've tried both `extension-sugar={preserve|always}`. But ocamlformat always transform my `assert%lwt false` to `[%lwt assert false]`

gpetiot:
Fix #1584 and fix #410, no diff with test_branch.

~I will probably remove the option `extension-sugar` that is deprecated since 0.14.0 first, this will reduce the diff of this PR => #1588~

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -1751,7 +1751,8 @@ end = struct
                               ( Pexp_new _ | Pexp_object _ | Pexp_while _
                               | Pexp_for _ | Pexp_function _ | Pexp_fun _
                               | Pexp_try _ | Pexp_match _ | Pexp_let _
-                              | Pexp_assert _ | Pexp_lazy _ )
+                              | Pexp_assert _ | Pexp_lazy _
+                              | Pexp_ifthenelse _ )
                           ; _ } as e )
                       , _ )
                 ; _ } ] )
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -1750,7 +1750,8 @@ end = struct
                       ( ( { pexp_desc=
                               ( Pexp_new _ | Pexp_object _ | Pexp_while _
                               | Pexp_for _ | Pexp_function _ | Pexp_fun _
-                              | Pexp_try _ | Pexp_match _ | Pexp_let _ )
+                              | Pexp_try _ | Pexp_match _ | Pexp_let _
+                              | Pexp_assert _ )
                           ; _ } as e )
                       , _ )
                 ; _ } ] )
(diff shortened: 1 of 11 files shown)


hhugo:
Does this add support all sugars ?

gpetiot:
> Does this add support all sugars ?

I noticed the support was missing for signature items, I'm working on it and it should be alright when I'll undraft this PR.

hhugo:
`begin%ext ... end` is most probably missing but it is another kind of beast and should not be addressed here.

gpetiot:
I think every case is handled now judging from the source.ml test file, I will have a look at the `begin%ext ... end` in another PR. No diff with test branch on the conventional and janestreet profile.

hhugo:
> I will have a look at the `begin%ext ... end` in another PR

Can you add a test for it now ?

emillon:
There's `;%ext` too, not sure if it's already supported or not.

gpetiot:
> There's `;%ext` too, not sure if it's already supported or not.

It is!

hhugo:
What about the expression `(module%ext S : S)` ?

gpetiot:
> What about the expression `(module%ext S : S)` ?

It is handled now

hhugo:
Have you check in the ocaml parser that all sugar cases are handled by ocamlformat ? (except begin%ext)

lib/Ast.ml
```diff
@@ -1742,28 +1742,9 @@ end = struct
...
-                ; _ } ] )
-        when Source.extension_using_sugar ~name:ext ~payload:e.pexp_loc ->
-          Some Apply
-      | Pexp_extension
-          ( ext
-          , PStr
-              [ { pstr_desc=
-                    Pstr_eval (({pexp_desc= Pexp_sequence _; _} as e), _)
-                ; _ } ] )
+      | Pexp_extension (ext, PStr [{pstr_desc= Pstr_eval (e, _); _}])
```
hhugo, line 1745: can you explain this change ?

lib/Ast.ml
```diff
@@ -1742,28 +1742,9 @@ end = struct
...
         when Source.extension_using_sugar ~name:ext ~payload:e.pexp_loc ->
-          Some Semi
+          prec_ast (Exp e)
```
hhugo, line 1747: can you explain this change ?

lib/Ast.ml
gpetiot: Because in the case where we use the sugared syntax this function should return the same for `keyword%ext` and `keyword`

gpetiot:
> Have you check in the ocaml parser that all sugar cases are handled by ocamlformat ? (except begin%ext)

I had a look at the ocaml parser and I think all cases are handled now

hhugo (approved):

PR #1590, opened 2021-01-14
Rewrite types of AST to prepare the design of new passes

gpetiot:
The idea behind this PR is that at some point I would like to split the current processing in several passes and we should be able to handle transformations over ASTs with functions typed like `Ast0 -> Ast1` or something similar.
And the current representation with the type `_ fragment` is dragging me down for this work (or maybe I'm missing something).

for the record the current fragment type is
```ocaml
  type 'a fragment =
    | Structure : Parsetree.structure fragment
    | Signature : Parsetree.signature fragment
    | Use_file : Parsetree.toplevel_phrase list fragment
```
and we would need to rewrite it as
```ocaml
  type fragment =
    | Pfrag_str of structure
    | Pfrag_sig of signature
    | Pfrag_usf of use_file

  type t = (* unused for now, should eventually replace fragment *)
    | Str of structure list
    | Sig of signature list
    | Usf of use_file list
```

as a consequence, all functions of the form `'a Traverse.fragment -> 'a -> ...` have been rewritten as `Parsetree.fragment -> ...`

This PR is only some preliminary work, it can be improved before being merged, in particular I'm not satisfied with handling the type `Parsetree.fragment` instead of `Parsetree.t` (that should in the future be AstN.t) in particular for parsing function of Parse_with_comments and the formatting one in Fmt_ast, there was a trick to parse `'a list fragment` (so list of structures, signatures etc) but the rewritting is not immediate to retrieve `Parsetree.t` instead since the parsing functions only return structures, signatures, etc. I'm still trying to understand this trick cc @emillon 
What do you think?

### DIFF
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -19,19 +19,11 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter)
 ;;
 Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 
-let format ?output_file ~kind:(Conf.Kind k) ~input_name ~source conf opts =
+let format ?output_file ~kind ~input_name ~source conf opts =
   if conf.Conf.disable then Ok source
   else
-    match k with
-    | Structure ->
-        Translation_unit.parse_and_format Structure k ?output_file
-          ~input_name ~source conf opts
-    | Signature ->
-        Translation_unit.parse_and_format Signature k ?output_file
-          ~input_name ~source conf opts
-    | Use_file ->
-        Translation_unit.parse_and_format Use_file k ?output_file ~input_name
-          ~source conf opts
+    Translation_unit.parse_and_format ~kind ?output_file ~input_name ~source
+      conf opts
 
 let to_output_file output_file data =
   match output_file with
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -22,7 +22,7 @@ Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 let format ?output_file ~kind ~input_name ~source conf opts =
   if conf.Conf.disable then Ok source
   else
-    Translation_unit.parse_and_format ~kind ?output_file ~input_name ~source
+    Translation_unit.parse_and_format kind ?output_file ~input_name ~source
       conf opts
 
 let to_output_file output_file data =
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -19,11 +19,11 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter)
 ;;
 Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 
-let format ?output_file ~kind:(Conf.Kind k) ~input_name ~source conf opts =
+let format ?output_file ~kind ~input_name ~source conf opts =
   if conf.Conf.disable then Ok source
   else
-    Translation_unit.parse_and_format k ?output_file ~input_name ~source conf
-      opts
+    Translation_unit.parse_and_format ~kind ?output_file ~input_name ~source
+      conf opts
 
 let to_output_file output_file data =
   match output_file with
--- a/bin/ocamlformat.ml
+++ b/bin/ocamlformat.ml
@@ -19,11 +19,19 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter)
 ;;
 Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 
-let format ?output_file ~kind ~input_name ~source conf opts =
+let format ?output_file ~kind:(Conf.Kind k) ~input_name ~source conf opts =
   if conf.Conf.disable then Ok source
   else
-    Translation_unit.parse_and_format ~kind ?output_file ~input_name ~source
-      conf opts
+    match k with
+    | Structure ->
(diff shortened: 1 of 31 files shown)


emillon:
The `'a fragment` type might be misnamed. It does not hold any data; it's only a witness encoding what the functions are supposed to do with the `'a` argument. You can encode that with a plain ADT (like the `Pfrag_str` constructor you mention - I don't understand what the `Str of structure list` is used for in your example - is that a list of lists?) but you'll have to handle cases that do not make sense in some cases.

For example in the GADT case, equality has just 3 cases:

```ocaml
  let equal (type a) (x : a fragment) : a -> a -> bool =
    match x with
    | Structure -> Parsetree.equal_structure
    | Signature -> Parsetree.equal_signature
    | Use_file -> List.equal Parsetree.equal_toplevel_phrase
```

On the other handm in the plain "dynamic" ADT type you have 3 x 3 = 9 cases to handle:

```ocaml
let equal x y = match x, y with
  | Frag_str sx, Frag_str sy -> Parsetree.equal_structure sx sy
  | Frag_sig sx, Frag_sig sy -> Parsetree.equal_signature sx sy
  | Frag_use ux, Frag_use uy -> List.equal Parsetree.equal_toplevel_phrase ux uy
  | Frag_str _, Frag_sig _  | ... -> false (* these 6 cases are not actually useful *)
```

(there's also a performance aspect to this since an additional layer of boxing is used in the ADT case, but here it doesn't matter).

In the ADT case you'll need a kind of nullary type to determine how to parse a string into the fragment type:

```ocaml
type how_to_parse = Parse_structure | Parse_signature | Parse_use_file

type parse_result = Structure of structure | Signature of signature | Use_file of use_file

let parse s how_to = match how_to with
| Parse_structure -> ........ in Structure str
...
```

In that case it's your responsibility to know that when you pass `Parse_structure`, you get a `Structure _`. You might have to insert some `assert false` branches. Using a GADT solves that (pass a `'a fragment` and a `string`, get a `'a`).

> there was a trick to parse 'a list fragment

I think you're refering to a function that was generalizing to `'a list fragment` instead of the more general `'a fragment`. The reason is that there's a place that returned `[]`. That can be solved by defining a function like:

```ocaml
let empty (type a) (f:a fragment) :a =
match f with
| Structure -> []
| Signature -> []
| Use_file -> []
```

It looks silly but it's a way to show to ocaml that all known fragments are some sorts of lists (of structure items, signature items or toplevel phrases).

Note that here you're replacing the GADT by a "dynamic" ADT, but this is not what the GADT was replacing: there used to be 3 "variants" of the function at every place. You can also go back to that situation.

If the number of entrypoints is going to increase (e.g. adding a way to format `core_type`s or `pattern`s), the number of nonsense combinations is going to increase too, quadratically. This is why I think a GADT makes sense.

If you want to have one AST type per pass, you'll have to somehow map the types. With a dynamic ADT, that means knowing that when you pass a `Ast0.Structure _` in, you get a `Ast1.Structure _` out. With a GADT, that means encoding a correspondence between `Ast0.structure` and `Ast1.structure`. It's genuinely hard to know what kind of structure will work better since it will depend on the number of passes and entrypoints and we don't know that.

I hope that clarifies the current use of `_ fragment`. I believe it's a useful tool that deduplicated a lot of code in the current implementation. But you're maintaining that code base, so I don't want to push that solution on you; there are solutions that are less type-heavy (but a bit more unsafe) that you can use instead if you're more comfortable with them.

That being said, what you are trying to do (the well-typed `AstN` passes) is very ambitious, so I believe that with all the solutions you'd see different symptoms: here the `_ fragment` type is getting in the way, but with the duplication you'd have to introduce duplication in more places, and with a dynamic ADT you'd have even more dead branches and "unwrap" functions.

gpetiot:
Thank you, that makes it clearer! I appreciate the `_ fragment` changes that removed a lot of duplications, I thought of making this change as I was struggling to get a clean solution in another PR where I had to juggle while writing the transformation functions.

If we keep the current design we would write something like: (for each AST)
```ocaml
module Ast0 = struct
  type 'a fragment =
    | Structure : structure fragment
    | Signature : signature fragment
    | Use_file : toplevel_phrase list fragment
end

module Ast1 = struct
  type 'a fragment =
    | Structure : structure fragment
    | Signature : signature fragment
    | Use_file : toplevel_phrase list fragment
end
```

and I encountered a lot of problems to make the transition from 1 fragment to another as a lot of functions are parameterized by the fragment witness. That's why having a new type `Syntax.t` independent from any type definition of the parsetree looked like a good solution to me.

emillon:
There was a version of this problem when porting parse-wyc to ppxlib. It was solved by having 2 type parameters on the fragment type:

```ocaml
type ('ast0, 'ast1) fragment =
  | Structure : (Ast0.structure, Ast1.structure) fragment
  | Signature : (Ast0.signature, Ast1.signature) fragment
  | Use_file : (Ast0.phrase list, Ast1.phrase list) fragment

val first_pass : ('ast0, 'ast1) fragment -> 'ast0 -> 'ast1
```

For more than 2 steps, adding more parameters work but I think it's possible to use one "common" fragment type and one "mapping" type per AST type. This probably makes it either to structure the program because the global type is independent from AST modules.

gpetiot:
I reverted the changes on the fragment gadt (as it will be useful in #1586), and the distinction between the initial ast and the final ast must be more obvious in the code now, let me know what you think!

PR #1593, opened 2021-01-19
Update Readme with instructions for use-package in emacs 

mattiasdrp:
I plan to update all readme of emacs related OCaml packages with their use-package configuration since it's extremely useful but not everyone know how to use it. This is my first attempt to upgrade one as it proposes an interesting challenge with multiple switches.


gpetiot (approved):
Looks good to me, thanks!

gpetiot:
Already merged, but during a github incident!

Issue #1211, opened 2020-01-18
Feature request: support BER MetaOCaml


ale64bit:
**Is your feature request related to a problem? Please describe.**
When using ocamlformat with [BER MetaOCaml](http://okmij.org/ftp/ML/MetaOCaml.html) sources, the bracket syntax is expanded to the corresponding attributes. For example:
```ocaml
let _ = .<1>.
```
expands to
```ocaml
let _ = (1 [@metaocaml.bracket])
```

**Describe the solution you'd like**
I guess we could argue that BER MetaOCaml is a "dialect" of OCaml and the fix doesn't belong in ocamlformat. However, it would be great if it was supported. Concretely, the bracket syntax (both bracket `.<expr>.` and escape `.~(expr)`) should be kept as is.

**Additional context**
none

gpetiot:
The problem is that we would need a way to tell the OCaml parser (that we rely on) that `.<expr>.` is a valid expression, formatting the corresponding attribute in the BER syntax is the easy part.
Once #1026 is merged we would be able to format invalid files, so the parts where you use BER syntax would be ignored and the rest would be formatted.

Julow:
@gpetiot  BER MetaOCaml is a fork of OCaml. Building OCamlformat with it automatically allows to parse BER syntax, since there is no new AST nodes.
This issue is about OCamlformat's output. BER parser uses `[@metaocaml. ..]` attributes internally but OCamlformat prints them.

I would be in favor of this feature :)
I wonder how compatibility with different version would work as we rely on `ocaml-migrate-parsetree` ? (OCamlformat only supports 4.07 to 4.09, though)

emillon:
This will at least make the test suite more complicated since we'll emit code that's invalid for the vanilla compiler, as @gpetiot said. So it would have to be gated behind a flag at least.

In any case, extensions are low priority at the moment. It would be easier to support than extensions like cppo or camlp4, but the focus for ocamlformat is vanilla ocaml code for now.

CraigFe:
I have a small (~50 line) patch that adds MetaOCaml dialect support [here](https://github.com/ocaml-ppx/ocamlformat/compare/master...CraigFe:metaocaml), which may be useful to anyone who just wants a quick solution. I have no comment on whether it's worth trying to maintain this upstream.

hhugo:
x-link https://github.com/ocaml/ocaml/pull/11871

gpetiot:
[...]

PR #1602, opened 2021-02-13
Support BER MetaOCaml N111 (in Unix)

moatom:
This PR is for supporting BER MetaOCaml N111, which is OCaml 4.11.1 extended to  support meta-programming (see http://okmij.org/ftp/ML/MetaOCaml.html). Because MetaOCaml adds `>.`, `.~`, and `.<` to original OCaml syntax, it adds `GREATERDOT`, `DOTTILDE`, and `DOTLESS` to token.

I tested it but `tools/test_branch.sh ber-n111` did't work, of course because HEAD was not buildable in MetaOCaml.
The test by `make test` worked fine :-)

### DIFF
--- a/vendor/token-latest/411ber/token_latest.ml
+++ b/vendor/token-latest/411ber/token_latest.ml
@@ -1,0 +1,131 @@
+type token = Parser.token =
+  | WITH
+  | WHILE
+  | WHEN
+  | VIRTUAL
+  | VAL
+  | UNDERSCORE
+  | UIDENT of string
+  | TYPE
+  | TRY
+  | TRUE
+  | TO
+  | TILDE
+  | THEN
+  | STRUCT
+  | STRING of (string * Location.t * string option)
+  | STAR
+  | SIG
+  | SEMISEMI
+  | SEMI
+  | RPAREN
+  | REC
+  | RBRACKET
+  | RBRACE
+  | QUOTED_STRING_ITEM of
+      (string * Location.t * string * Location.t * string option)
+  | QUOTED_STRING_EXPR of
+      (string * Location.t * string * Location.t * string option)
+  | QUOTE
+  | QUESTION
+  | PRIVATE
+  | PREFIXOP of string
+  | PLUSEQ
+  | PLUSDOT
+  | PLUS
+  | PERCENT
+  | OR
+  | OPTLABEL of string
+  | OPEN
+  | OF
+  | OBJECT
+  | NONREC
+  | NEW
+  | MUTABLE
+  | MODULE
+  | MINUSGREATER
+  | MINUSDOT
+  | MINUS
+  | METHOD
+  | MATCH
+  | LPAREN
+  | LIDENT of string
+  | LETOP of string
+  | LET
+  | LESSMINUS
+  | LESS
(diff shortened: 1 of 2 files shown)


hhugo:
I'm surprised this is enough. Don't you want to print theses   new tokens back ?

moatom:
Hi!
>Don't you want to print theses new tokens back ?

Are you mentioning about OCamlFormat's changing quoting syntax sugars to attributes representations (e.g. `.<1>.` -> `1[@metaocaml.bracket]`) ? Yes, I do. But I don't know how to do it.

moatom:
Although I am not familiar with the entire structure of OCamlFormat, I think Printast.expression's difference is related to this problem.
- https://github.com/ocaml-ppx/ocamlformat/blob/master/lib/Migrate_ast.mli#L124
- https://github.com/metaocaml/ber-metaocaml/blob/ber-n111/parsing/pprintast.ml#L639
- https://github.com/ocaml-ppx/ppxlib/blob/master/ast/pprintast.ml#L597

hhugo:
A special case will need to be added early in https://github.com/ocaml-ppx/ocamlformat/blob/master/lib/Fmt_ast.ml#L1553.
similar to what is done in metaocaml.
Note that Metaocaml uses attributes to encode theses new constructs. It is not ideal as it makes attributes order dependent, see discussion in https://github.com/ocaml/ocaml/pull/10131

CraigFe:
You may be interested in the [corresponding issue](https://github.com/ocaml-ppx/ocamlformat/issues/1211) for BER MetaOCaml support in OCamlFormat, and [this patch](https://github.com/ocaml-ppx/ocamlformat/compare/master...CraigFe:metaocaml) that provides a minimal implementation with the current use of attributes.

hhugo:
> You may be interested in the [corresponding issue](https://github.com/ocaml-ppx/ocamlformat/issues/1211) for BER MetaOCaml support in OCamlFormat, and [this patch](https://github.com/ocaml-ppx/ocamlformat/compare/master...CraigFe:metaocaml) that provides a minimal implementation with the current use of attributes.



> You may be interested in the [corresponding issue](https://github.com/ocaml-ppx/ocamlformat/issues/1211) for BER MetaOCaml support in OCamlFormat, and [this patch](https://github.com/ocaml-ppx/ocamlformat/compare/master...CraigFe:metaocaml) that provides a minimal implementation with the current use of attributes.

Nice
Note that your patch doesn't not care about the order of attributes. I think it will make both `.< .~x >.` and `.~(.<x>.)` format the same

CraigFe:
Indeed, thanks.

For what it's worth, I agree with your comments on the corresponding OCaml PR that extension points would be a better representation here & think it's worth waiting a little bit on the OCamlformat side until that question is resolved upstream.

moatom:
> You may be interested in the corresponding issue for BER MetaOCaml support in OCamlFormat, and this patch that provides a minimal implementation with the current use of attributes.

Thanks! It was very easy to understand.

>  think it's worth waiting a little bit on the OCamlformat side until that question is resolved upstream.

I roughly understand the current situation. I think so, personally.

PR #1607, opened 2021-02-16
Update the release procedure wrt. dune-release

gpetiot:


emillon:
Can you also add that until the corresponding bug is resolved, the master branch of `dune-release` should be used?

HACKING.org
```diff
@@ -23,7 +23,8 @@ Git is used to manage the script's state:
 - If a supported OCaml version does not have CI support, run test suite locally.
 
 - In =CHANGES.md=, change =(unreleased)= with the current version (=a.b.c=)
-  and date. Commit the result.
+  and date. In =README.md=, update the recommended version to use in
+  =.ocamlformat= files. Commit and push the changes.
```
Julow, line 27: I think the "push" step should be before the "dune-release" step, after tagging and checking of the version number, to give a chance to amend, just in case. It may also be worth doing `dune-release lint` just before pushing.

gpetiot: good idea, done!

HACKING.org
```diff
@@ -23,16 +23,26 @@ Git is used to manage the script's state:
...
 
 - Check that release binary has a release version number:
   =dune build @install && dune install --prefix _install && ./_install/bin/ocamlformat --version && rm -rf _install= should output just =a.b.c=
 
+- Linting
+  =dune-release lint=
+
+- Amend the release commit if necessary.
+
+- Push the changes.
```
Julow, line 40: The command needs to be `git push --tags`, otherwise the tag won't be pushed.
It is not necessary to push the tag now because `dune-release` will do it if necessary, however it can fail because it can only guesses the URIs and might be wrong.

HACKING.org
```diff
@@ -23,16 +23,26 @@ Git is used to manage the script's state:
...
 - Tag
   =dune-release tag=
 
 - Check that release binary has a release version number:
   =dune build @install && dune install --prefix _install && ./_install/bin/ocamlformat --version && rm -rf _install= should output just =a.b.c=
 
+- Linting
+  =dune-release lint=
+
+- Amend the release commit if necessary.
```
Julow, line 38: And update the tag, that's very important otherwise the tag will point to the previous commit. `dune-release tag -f`

HACKING.org
gpetiot: done

HACKING.org
gpetiot: done

Julow (approved):
Looks good to me.

PR #1608, opened 2021-02-17
Release process: document Windows binaries

emillon:


Julow:
What about building from the release tarball already uploaded to remove the dune-release part ? That can probably be done from a Github Action.

emillon:
This file documents what we're doing today so that others can do it.

Once ocaml-ci can do that we can use it but in the meantime it's not worth automating something that takes 10 minutes every 2 months :)

gpetiot (approved):
I agree about documenting the current process only, and if we use github Actions in the future we will update this document, looks good to me!

Issue #1583, opened 2021-01-05
Feature request: ocamlformat should not depend on ppxlib


dinosaure:
Since `ocamlformat.0.15.1`, the tool depends on `ppxlib`. For my point of view, `ocamlformat` is used by several projects which don't strictly depend on any ppx stuffs. Even if `ocamformat` is mostly not a part of the distribution of them - and it exists just as a linter - the introduction of `ppxlib` comes with a large dependency graph.

In my case, an upgrade from `ocamlformat.0.15.0` to `ocamlformat.0.16.0` requires a recompilation of 68 packages with OPAM and a deletion of some packages which did not upgrade to the last version of `ppxlib` (`0.20.0`). It's a large cost just to lint my codebase...

I did not really go deeper into such choice but, for my perspective, I should not upgrade all my stack just to be able to run my linter.

gpetiot:
We also had feedback from janestreet about the difficulty of upgrading to the last ppxlib just to upgrade ocamlformat

@emillon that's something we can try if it slows down adoption/upgrade

NathanReb:
To be honest that feels once more like a problem of executables installation through opam rather than a problem with ppxlib.

I think the energy would be better spent trying to make ocamlformat dependency free using vendoring or some new opam installation mode rather than re-implementing everything all the way down to the standard library inside ocamlformat.

There might some very good reasons for ocamlformat to use its own parser (and therefore to eventually cease depending on ppxlib) but I don't think this is one of them. I think it might be worth taking the time to consider what other options we have and how we can spend our time and effort towards fixing this problem in the long run and for other tools as well.

That being said, I totally understand the frustration for users and I'm not taking it lightly. I want to emphasize that I believe this should be fixed, just that it should be fixed differently so that tools such as ocamlformat are free to use any ocaml library they want without imposing them to their users.

hhugo:
What are the things we need ppxlib for ? Parser and AST mapper / iterator ? Anything else ? We should probably aim to use a single parser (namely parse_wyc)

gpetiot:
For now the recovery parser (parse_wyc) is not the default parser, as it is gated behind the `--format-invalid-files` option (soon to be replaced by the `--numeric` option that should only be used in editors), a lot of parsing rules are commented compared to the default parser, and I'm not sure this behavior would suit the default mode of ocamlformat. I've already experimented on that in the past and we would see things like incomplete structures being completed, so the user would see a `end` keyword magically appear for example.

[...]

PR #1610, opened 2021-02-25
Use opam monorepo to vendor ocamlformat dependencies

gpetiot:
Messing around with opam monorepo (see #1583)


gpetiot:
Looks like monorepo is not usable enough for now.

PR #1624, opened 2021-03-29
Disable failing tests on windows

emillon:
This adds `(enabled_if)` to tests that do not pass on Windows today.
The resulting situation is that `dune runtest` works on Windows, and corresponding tests can be fixed on a longer schedule.


test/passing/dune.inc
```diff
@@ -2015,12 +2015,14 @@
 
 (rule
  (deps .ocamlformat dir1/dir2/.ocamlformat dir1/dir2/print_config.ml)
+ (enabled_if (<> %{os_type} Win32))
  (action
    (with-outputs-to print_config.ml.output
      (run %{bin:ocamlformat} dir1/dir2/print_config.ml --print-config --config=max-iters=2 %{dep:print_config.ml}))))
```
Julow, line 2021: Would this one work if it was passed like this ? `%{dep:dir1/dir2/print_config.ml}`

test/passing/dune.inc
```diff
@@ -2489,22 +2491,26 @@
...
  (deps .ocamlformat )
+ (enabled_if (<> %{os_type} Win32))
  (action
    (with-outputs-to verbose1.ml.output
      (run %{bin:ocamlformat} --print-config --doc-comments=before --config=doc-comments=before %{dep:verbose1.ml}))))
 
 (rule
  (alias runtest)
+ (enabled_if (<> %{os_type} Win32))
  (action (diff verbose1.ml.ref verbose1.ml.output)))
```
Julow, line 2502: What's wrong with verbose* tests ?

test/projects/dune
```diff
@@ -25,6 +25,7 @@
 
 (rule
  (alias runtest)
+ (enabled_if (<> %{os_type} Win32))
  (action
   (diff outside_detected_project.expected outside_detected_project.output)))
```
Julow, line 30: Missing `enabled_if` on the rule for that file. Also, why is this one not working on Windows?

test/passing/gen/gen.ml
```diff
@@ -77,9 +81,12 @@ let emit_test test_name setup =
   in
   let extra_deps = String.concat " " setup.extra_deps in
   let enabled_if_line =
-    match setup.minimum_ocaml_version with
-    | None -> ""
-    | Some v -> Printf.sprintf "\n (enabled_if (>= %%{ocaml_version} %s))" v
+    match setup.minimum_ocaml_version, setup.enabled_if with
```
Julow, line 84: There's 6 uses of `minimum-ocaml-version`. Replacing it with `enabled-if` may be a smaller patch and would remove code.

test/passing/gen/gen.ml
```diff
@@ -7,7 +7,8 @@ type setup =
   ; mutable base_file: string option
   ; mutable extra_deps: string list
   ; mutable should_fail: bool
-  ; mutable minimum_ocaml_version: string option }
+  ; mutable minimum_ocaml_version: string option
+  ; mutable enabled_if: string option }
```
emillon, line 11: @gpetiot I had to add a more general mechanism of disabling tests. We can either:

- keep both `.minimum-ocaml-version` and `.enabled-if` files, where the former is a shortcut for a version check and there's an error if both files are provided
- just keep `.enabled-if` files and migrate the existing `.minimum-ocaml-version` files to this new format

What do you prefer?

gpetiot: I would prefer to have only one test, so only the `.enabled-if`

emillon: done!

test/projects/dune
gpetiot: I added the clause on the related rule

test/passing/dune.inc
gpetiot: I modified this line, let's wait to see if it works on windows indeed

emillon:
@gpetiot I just rebased, fixed the new failing tests under `failing/`, this is now passing on Windows.

gpetiot:
> @gpetiot I just rebased, fixed the new failing tests under `failing/`, this is now passing on Windows.

Do we know why `print_config` and `verbose1/2` tests are failing on windows?

emillon:
I think it's because they print paths, but the point of that PR is that we can investigate the reason of the failures later while keeping everything green.

gpetiot (approved):
Looks good to me, thanks! Feel free to merge when ready!

PR #1634, opened 2021-04-09
Fix formatting of odoc tags: the argument should be on the same line, indent description that wraps

### LINKED ISSUE #1631: Bug: Long URLs in ocamldoc `@see` tags lead to broken wrapping
**Describe the bug**
The following comment is impossible to both *format* and then *generate docs from*:

```ocaml
(** Registers all commands in a directory. The files must export a Command class
    constructor or instance. This is a short form for {!registerCommandsIn}.

    @param path The path to the directory

    @see <https://slash-create.js.org/#/docs/main/latest/class/SlashCreator?scrollTo=registerCommandsIn> *)
```

1. When parsed as-is, `ocamlformat` yells:

       ocamlformat: Cannot process "src/SlashCreate.mli".
         BUG: doc comments changed.
       File "_none_", line 1, characters -1--1::
       Error: Formatting of (** Registers all commands in a directory. The files m ... st/class/SlashCreator?scrollTo=register
       CommandsIn> *) is unstable (e.g. parses as a list or not depending on the margin), please tighten up this comment in th
       e source or disable the formatting using the option --no-parse-docstrings.

2. If I add a newline before the URL,

   ```ocaml
   (** Registers all commands in a directory. The files must export a Command class
       constructor or instance. This is a short form for {!registerCommandsIn}.

       @param path The path to the directory

       @see
       <https://slash-create.js.org/#/docs/main/latest/class/SlashCreator?scrollTo=registerCommandsIn> *)
   ```

   ... then ocamlformat will stabilize and format it, but I can no longer generate my docs:

   ```
   Warning: Invalid documentation comment:
   File "src/SlashCreate.mli", line 117, characters 6-10:
   '@see' should be followed by <url>, 'file', or "document title".
   ```
<details>
<summary><code>--print-config</code></summary>

```
profile=default
quiet=false
max-iters=10
comment-check=true
wrap-fun-args=true
wrap-comments=true (file .ocamlformat:11)
type-decl-indent=3 (file .ocamlformat:8)
type-decl=compact
stritem-extension-indent=0
space-around-variants=true (file .ocamlformat:18)
space-around-records=true (file .ocamlformat:15)
space-around-lists=true (file .ocamlformat:17)
space-around-arrays=true (file .ocamlformat:16)
single-case=compact
sequence-style=separator (file .ocamlformat:23)
parse-docstrings=true (file .ocamlformat:10)
parens-tuple=always
parens-ite=false
ocp-indent-compat=false
nested-match=wrap
module-item-spacing=sparse
max-indent=68
match-indent-nested=never
match-indent=0
margin=90 (file .ocamlformat:3)
let-open=preserve
let-module=compact
let-binding-spacing=sparse (file .ocamlformat:14)
let-binding-indent=3 (file .ocamlformat:7)
let-and=compact
infix-precedence=indent
indicate-nested-or-patterns=space (file .ocamlformat:20)
indent-after-in=0
if-then-else=compact
function-indent-nested=never
function-indent=3 (file .ocamlformat:6)
field-space=loose (file .ocamlformat:13)
extension-sugar=preserve
extension-indent=3 (file .ocamlformat:5)
exp-grouping=parens
dock-collection-brackets=true
doc-comments-tag-only=default
doc-comments-padding=2
doc-comments=after-when-possible
disable=false
cases-exp-indent=4
break-struct=force
break-string-literals=auto
break-sequences=true (file .ocamlformat:22)
break-separators=after
break-infix-before-func=false
break-infix=wrap
break-fun-sig=wrap
break-fun-decl=wrap
break-cases=fit-or-vertical (file .ocamlformat:21)
assignment-operator=end-line
align-variants-decl=false
align-constructors-decl=false
align-cases=false
```

</details>

gpetiot:
Fix #1631, no diff with test_branch.sh

### DIFF
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -220,7 +220,7 @@ let fmt_tag c = function
   | `See (`Url, sr, txt) -> fmt_tag_see c (wrap "<" ">") sr txt
   | `See (`File, sr, txt) -> fmt_tag_see c (wrap "'" "'") sr txt
   | `See (`Document, sr, txt) -> fmt_tag_see c (wrap "\"" "\"") sr txt
-  | `Since s -> at $ fmt "since@ " $ str s
+  | `Since s -> at $ str "since " $ str s
   | `Before (s, txt) ->
       at $ str "before " $ str s
       $ fmt_nestable_block_elements c ~prefix:space txt
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -227,7 +227,7 @@ let fmt_tag c = function
   | `Deprecated txt ->
       at $ fmt "deprecated" $ fmt_nestable_block_elements c ~prefix:space txt
   | `Param (s, txt) ->
-      at $ fmt "param@ " $ str s
+      at $ str "param " $ str s
       $ fmt_nestable_block_elements c ~prefix:space txt
   | `Raise (s, txt) ->
       at $ fmt "raise@ " $ str s
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -237,7 +237,7 @@ let fmt_tag c = function
   | `Inline -> at $ str "inline"
   | `Open -> at $ str "open"
   | `Closed -> at $ str "closed"
-  | `Canonical ref -> at $ fmt "canonical@ " $ fmt_reference ref
+  | `Canonical ref -> at $ str "canonical " $ fmt_reference ref
 
 let fmt_block_element c = function
   | `Tag tag -> hovbox 0 (fmt_tag c tag)
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -215,7 +215,7 @@ let fmt_tag_see c wrap sr txt =
   $ fmt_nestable_block_elements c ~prefix:space txt
 
 let fmt_tag c = function
-  | `Author s -> at $ fmt "author@ " $ str s
+  | `Author s -> at $ str "author " $ str s
   | `Version s -> at $ fmt "version@ " $ str s
   | `See (`Url, sr, txt) -> fmt_tag_see c (wrap "<" ">") sr txt
   | `See (`File, sr, txt) -> fmt_tag_see c (wrap "'" "'") sr txt
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -209,35 +209,30 @@ let at = char '@'
 
 let space = fmt "@ "
 
-let fmt_tag_see c wrap sr txt =
-  at $ str "see "
-  $ wrap (str sr)
-  $ fmt_nestable_block_elements c ~prefix:space txt
+let fmt_tag_args ?arg ?txt c tag =
+  at $ str tag
+  $ opt arg (fun x -> char ' ' $ x)
+  $ opt txt (fun x -> fmt_nestable_block_elements c ~prefix:space x)
+
+let wrap_see = function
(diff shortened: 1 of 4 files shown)


test/passing/tests/doc_comments.mli.ref
```diff
@@ -78,6 +78,8 @@ val k : k
     @author foo
     @version foo
     @see <foo> foo
+    @see <https://slash-create.js.org/#/docs/main/latest/class/SlashCreator?scrollTo=registerCommandsIn>
+    this url is very long
```
Julow, line 82: This text is part of the `@see` tag and it's not obvious from the indentation.
Should we indent tag arguments that wrap ?

Julow (commented):
That's the right fix. However, other tags are broken for the same reason, here a test case:

```ocaml
(** Tags taking an argument that shouldn't break then a description:

    @see <https://foooooooooooooooooooooooooooooo.com/baaaaaaaaaaaaaaaaaaaaaaaaaaaaaar> Foo bar
    @before Foooooooooooooooooooooooooooooooooooo.Baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar Foo bar
    @param Foooooooooooooooooooooooooooooooooooo_baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar Foo bar
    @raise Foooooooooooooooooooooooooooooooooooo_baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar Foo bar

    @deprecated Foooooooooooooooooooooooooooooooooooo Baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar *)

(** Tags taking one argument that shouldn't break:

    @author Foooooooooooooooooooooooooooooooooooo Baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
    @version Foooooooooooooooooooooooooooooooooooo Baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
    @since Foooooooooooooooooooooooooooooooooooo.Baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar
    @canonical Foooooooooooooooooooooooooooooooooooo.Baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar *)
```

test/passing/tests/doc_comments.mli.ref
gpetiot: I think we should indeed, I've added an indent of 2 cols

gpetiot:
I've added more tests and fixed all the cases you listed! (and a small refactoring to clear things up)

Julow (approved):
Let's merge !

lib/Fmt_odoc.ml
```diff
@@ -209,44 +209,39 @@ let at = char '@'
...
 let space = fmt "@ "
 
-let fmt_tag_see c wrap sr txt =
-  at $ fmt "see@ "
-  $ wrap (str sr)
-  $ fmt_nestable_block_elements c ~prefix:space txt
+let fmt_tag_args ?arg ?txt c tag =
+  at $ str tag
+  $ opt arg (fun x -> char ' ' $ x)
+  $ opt txt (fun x -> fmt_nestable_block_elements c ~prefix:space x)
```
Julow, line 215: I wonder if we should wrap the `txt` into a box to break between the `arg` and `txt` in priority. For example, with a margin of 20:
```
(** @param Foo Bar
      Baz *)
```
could be:

```
(** @param Foo
      Bar Baz *)
```

The word `Foo` is the `arg` and `Bar Baz` is the `txt`.

gpetiot: That would be a nice improvement indeed, we can do that in a follow-up PR

PR #1635, opened 2021-04-14
Improve formatting of odoc tags: break between the arg and the desc in priority

gpetiot:
Follow-up on #1634

### DIFF
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -201,11 +201,8 @@ and fmt_list_light c kind items =
   in
   vbox 0 (list items "@," fmt_item)
 
-and fmt_nestable_block_elements c ?(prefix = noop) = function
-  | [] -> noop
-  | elems ->
-      prefix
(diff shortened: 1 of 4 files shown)


lib/Fmt_odoc.ml
```diff
@@ -203,7 +203,9 @@ and fmt_list_light c kind items =
 
 and fmt_nestable_block_elements c ?(prefix = noop) = function
   | [] -> noop
-  | elems -> prefix $ list_block_elem elems (fmt_nestable_block_element c)
+  | elems ->
+      prefix
+      $ hovbox 0 (list_block_elem elems (fmt_nestable_block_element c))
```
Julow, line 208: That will double-box list elements. Can this be an issue ?
Otherwise, we can move the box and the prefix logic into `fmt_tag_args`.

Julow (approved):

lib/Fmt_odoc.ml
gpetiot: Good catch, should be fixed now!

PR #1639, opened 2021-04-22
Line/region indentation feature for emacs

gpetiot:
Simpler version of #1609: the *slicing* optimization has been removed and can be reviewed independently in #1609 once this PR is merged.

Idea: We don't need to parse invalid files outside of this feature, so the `--format-invalid-files` option has been removed. Instead, this PR defines a new `--numeric=X-Y` option that should have the same semantic as `--numeric --lines=X-Y` in `ocp-indent`.

Algorithm:
  -  try to parse either the source, or the recovered source (if the first one failed, using `menhir` recovery parsing)
  -  if the parsing is successful, we format the AST, then match the location tree of the formatted AST against the location tree of the unformatted AST to get the indentation produced by `ocamlformat` *[1]*
  -  if the parsing failed, we use the `ocp-indent` API to produce an indentation

*[1]* this does not return optimal results when multiple lines are made into one when formatted, for example "let x =\n a\nin\n ...." will be formatted as a single line "let x = a in\n ..." so this implementation cannot detect this and will return the same indentation for the 3 lines, which will not look good), `ocp-indent` is called on this range of lines to improve the indentation.

cc @charlesetc @olydis @emillon


bcc32:
Sorry to bother you @gpetiot, but emacs packages are often updated much sooner than ocaml ones (e.g., I already updated `ocamlformat.el` through MELPA but the ocamlformat package hasn't been released since this PR), so now I get this error every time I press `RET` in an ocaml buffer:

```
ocamlformat errors:
ocamlformat: unknown option `--numeric', did you mean `-n' ?
Usage: ocamlformat [OPTION]... [SRC]...
Try `ocamlformat --help' for more information.
```

Is it possible to make the emacs package work correctly with older versions of ocamlformat as well?  It certainly seems plausible that people would also have older versions of ocamlformat installed in local opam switches, in which case this error could also arise.

gpetiot:
> Is it possible to make the emacs package work correctly with older versions of ocamlformat as well? It certainly seems plausible that people would also have older versions of ocamlformat installed in local opam switches, in which case this error could also arise.

I'm having a look at how I could disable the new behavior for ocamlformat < 0.19 (that's the kind of moment where I wish this file would be maintained outside of ocamlformat by people more fluent than me in emacs lisp)

PR #1641, opened 2021-04-28
Extract portable parts of ocamlformat_lib to ocamlformat_kernel

pmwhite:
- Rename lib/ to lib-kernel/
- Move Translation_unit.Error.print into lib/

The purpose of this PR is to make ocamlformat usable from js_of_ocaml.


gpetiot:
Hi, thank you for your contribution! (let us know if you have some trouble signing the CLA by the way)

What do you think of keeping the original name `lib` (the non-IO `Translation_unit` can be put there), and instead creating a `lib-io` or `lib-unix` with the current version of `Translation_unit`?

lib-kernel/Conf.mli
```diff
@@ -122,3 +122,5 @@ val update_value :
      Result.t
 
 val print_config : t -> unit
+
+val janestreet_profile : t
```
gpetiot, line 126: I would put it just after the declaration of `default_profile`

pmwhite:
> Hi, thank you for your contribution! (let us know if you have some trouble signing the CLA by the way)

I signed it (twice, actually), but the check didn't pass. I did sign it after submitting the PR, so maybe it just hasn't re-run since I signed it. I signed it yesterday, though, so if it runs every hour, it should have run by now.

> What do you think of keeping the original name `lib` (the non-IO `Translation_unit` can be put there), and instead creating a `lib-io` or `lib-unix` with the current version of `Translation_unit`?

That sounds fine to me. I would have done that initially, except I was trying to avoid a breaking change. Are you suggesting I also change the dune/opam library `ocamlformat_lib` to `ocamlformat_unix`?

gpetiot:
> That sounds fine to me. I would have done that initially, except I was trying to avoid a breaking change. Are you suggesting I also change the dune/opam library `ocamlformat_lib` to `ocamlformat_unix`?

For now ocamlformat_lib is not public, there is no opam file. If this is closer to the convention used by most ocaml projects (separating the unix/non-unix parts) I think we should do it before we make this library public.

gpetiot:
Closing this one as requested by @pmwhite that will open another one

PR #1644, opened 2021-04-29
Remove unix dependency

pmwhite:
- Move Translation_unit.Error.print into lib-unix/
- Remove dependency of lib/ on the unix library

The purpose of this PR is to make ocamlformat usable from js_of_ocaml.


hhugo:
Unix is only used for debug purposes. I believe you could achieve the same result by just introducing a small hook or just replacing unix refs with their `Sys` equivalent.

pmwhite:
> Unix is only used for debug purposes. I believe you could achieve the same result by just introducing a small hook or just replacing unix refs with their `Sys` equivalent.

Wow, I didn't notice this; thanks for the tip. The overall diff is much smaller, although the commit history now has a bunch of unnecessary changes. Let me know if I should redo the PR with a single commit.

gpetiot:
> The overall diff is much smaller, although the commit history now has a bunch of unnecessary changes. Let me know if I should redo the PR with a single commit.

No don't worry I will squash the commits upon merging anyway!

gpetiot (approved):
Looks good to me, thanks!

PR #1649, opened 2021-05-03
Fix alignment of closing paren on separate line for anonymous functions

### LINKED ISSUE #1579: Bug: indicate-multiline-delimiters=closing-on-separate-line don't align parens for anonymous functions
**Describe the bug**

`indicate-multiline-delimiters=closing-on-separate-line` doesn't vertically align parens for anonymous functions.  The opening and closing delimiters should align in order to be consistent with other effects from this option.

**How to Reproduce**

```
$ ocamlformat --version
0.16.0
```

In each case below, the closing `)` is not vertically aligned with the opening `(` in `(fun`.  I don't have a global ocamlformat config and I made sure there wasn't a local `.ocamlformat` when setting up this example.  `break-cases=toplevel` is there to show how illustrate how this can be confusing in practice - it the first example below it looks like the `)` should pair with the `match` when it really pairs with `(fun`.  The third/last case shows a potentially confusing alignment when the anonymous function is labeled.  I'm not sure what the "right" result would be in that case.

```
$ ocamlformat test.ml --enable-outside-detected-project --indicate-multiline-delimiters=closing-on-separate-line --break-cases=toplevel
let raise fmt =
  Fmt.kstr
    (fun error_message (result : _ result) ->
      match result with
      | Ok v -> v
      | Error `Oh_no -> invalid_arg error_message
      )
    fmt

let raise fmt =
  Fmt.kstr
    (fun error_message aaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbb ->
      invalid_arg error_message
      )
    fmt

let raise fmt =
  Fmt.kstr
    (fun error_message (result : _ result) ->
      match result with
      | Ok v -> v
      | Error `Oh_no -> invalid_arg error_message
      )
    fmt

let raise fmt =
  Fmt.kstr
    (fun error_message aaaaaaaaaaaaaaaaaaaaaaaaaaaa bbbbbbbbbbbbbbbbbbbbb ->
      invalid_arg error_message
      )
    fmt

let contrived =
  List.map
    ~f:(fun aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ->
      f aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
      )
    l
```

gpetiot:
Fix #1579, the diff of test_branch.sh shows improvements of the paren alignment just like the tests I've added!

### DIFF
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -41,7 +41,7 @@ let parens_if parens (c : Conf.t) ?(disambiguate = false) k =
 let parens c ?disambiguate k = parens_if true c ?disambiguate k
 
 let wrap_exp (c : Conf.t) ?(disambiguate = false) ?(fits_breaks = true)
-    ~parens ~loc source k =
+    ?(offset_closing_paren = 0) ~parens ~loc source k =
   match parens_or_begin_end c source ~loc with
   | `Parens when disambiguate && c.Conf.disambiguate_non_breaking_match ->
       wrap_if_fits_or parens "(" ")" k
@@ -52,7 +52,8 @@ let wrap_exp (c : Conf.t) ?(disambiguate = false) ?(fits_breaks = true)
     | `Space ->
         Fmt.fits_breaks "(" "(" $ k $ Fmt.fits_breaks ")" ~hint:(1, 0) ")"
     | `Closing_on_separate_line ->
-        Fmt.fits_breaks "(" "(" $ k $ Fmt.fits_breaks ")" ~hint:(1000, 0) ")"
+        Fmt.fits_breaks "(" "(" $ k
+        $ Fmt.fits_breaks ")" ~hint:(1000, offset_closing_paren) ")"
     | _ -> wrap "(" ")" k )
   | `Begin_end ->
       vbox 2 (wrap "begin" "end" (wrap_k (break 1 0) (break 1000 ~-2) k))
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 
   + Preserve begin-end keywords delimiting match cases (#1651, @gpetiot)
 
+  + Fix alignment of closing paren on separate line for anonymous functions (#1649, @gpetiot)
+
 #### Changes
 
 #### New features
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2102,7 +2102,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       in
       hvbox_if (box || body_is_function) indent
         (Params.wrap_exp c.conf c.source ~loc:pexp_loc ~parens
-           ~disambiguate:true ~fits_breaks:false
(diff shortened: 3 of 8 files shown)


lib/Params.ml
```diff
@@ -41,7 +41,7 @@ let parens_if parens (c : Conf.t) ?(disambiguate = false) k =
 let parens c ?disambiguate k = parens_if true c ?disambiguate k
 
 let wrap_exp (c : Conf.t) ?(disambiguate = false) ?(fits_breaks = true)
-    ~parens ~loc source k =
+    ?(offset_closing_paren = 0) ~parens ~loc source k =
```
Julow, line 44: Could you add a bit of documentation, especially the fact that it's used in the `Closing_on_separate_line` case?

Julow (approved):
Looks good.

lib/Params.ml
gpetiot: done!

psafont:
This change introduces a lot of churn when formatting xapi-project/xen-api. This can make backports to releases formatted by ocamlformat 0.18.0 cumbersome for little benefit. Is there any way as a user to keep the previous formatting in newer versions?

gpetiot:
No there is no way to keep the previous formatting in newer versions, unless you switch to a profile that doesn't use `indicate-multiline-delimiters = closing-on-separate-line`. The `default`, `ocamlformat` and `janestreet` profiles don't use this option, they will become our main focus for the future versions (and will be more stable!), we thus deprecated a lot of options that are not popular enough to make it into these profiles.
Starting from `0.20.0` the option `indicate-multiline-delimiters = closing-on-separate-line` is thus deprecated. Please try one of the mentioned profiles, I'm sure one must fit your style, we try to encourage the ocaml community to adopt the default profile.

psafont:
That makes maintaining several releases with different versions of ocamlformat increasingly difficult.

In the xapi project we chose the options we did because they were the ones that matches more closely the existing style. This makes me question how is the team tracking styles used if their deprecation causes problems to the community.

Is there any way to see which are the 'non-popular options' so we can try applying all of them at once and avoid this kind of situations in the future?

gpetiot:
We made a survey on the opam packages: https://gist.github.com/Julow/110dc94308d6078225e0665e3eccd433

To explain further: on ocamlformat's side, removing some options, or at least some values for some options has become inevitable as it's making the maintainability of ocamlformat more difficult (and it's easy to miss bugs or inconsistencies, as the space of possible configurations is to big to test everything). So we decided it would be better to focus our efforts to improve a few widespread styles instead of allowing any combination of every tweak of style.

psafont:
> We made a survey on the opam packages: gist.github.com/Julow/110dc94308d6078225e0665e3eccd433

Most of the public repositories we use were missed because it makes little sense to publish daemons to opam :( https://github.com/search?q=org%3Axapi-project+org%3Axenserver+in%3Apath+.ocamlformat&type=code

PR #1650, opened 2021-05-04
Fix comment attachment so that comments starting a line are attached to the previous node

### LINKED ISSUE #1409: Bug: comment between module and attribute triggers failure
**Describe the bug**
ocamlformat fails on this code:
```
module M = struct end (* comment long enough to force line splitting *) [@ocaml.warning ""]
```

**How to Reproduce**
Ran with ocamlformat built from 10aabb33581dd528c06493cb31849300186dba95:
```
$ ocamlformat --profile=janestreet --enable-outside-detected-project --disable-conf-files /tmp/file.mlt 
ocamlformat: Cannot process "/tmp/file.mlt".
```

gpetiot:
Fix #1409, no diff with test_branch.sh

This fix is a bit ugly, as this is a temporary solution while waiting to have the comments in the AST some day. The `.ocp` test file only exhibits an issue after applying ocp-indent, but janestreet should eventually drop this behavior.


test/passing/tests/js_source.ml.ref
```diff
@@ -9865,3 +9865,6 @@ let xxxxxx =
              __________ *) () = yyyyyyyy in
   { zzzzzzzzzzzzz }
 ;;
+
+module M = struct end
+(* comment long enough to force line splitting *) [@ocaml.warning ""]
```
Julow, line 9870: We expect a newline between the comment and the attribute. I feel this should be solved with boxes and breaks but I couldn't come up with a fix.

Julow (approved):

PR #1651, opened 2021-05-04
Preserve begin-end keywords delimiting match cases

### LINKED ISSUE #1599: Bug: `exp-grouping=preserve` is not respected when formatting match cases
Formatting code turns `begin`...`end` blocks within match expressions into parenthesized blocks, even when `exp-grouping` is set to `preserve`.

**How to Reproduce**

Version: 0.16.0

In `test.ml`:

```ocaml
type t = Alpha | Beta

let _ = 
  match Alpha with
  | Alpha -> begin
    match Beta with
    | Alpha -> print_endline "Alpha, Alpha"
    | Beta -> print_endline "Alpha, Beta"
  end
  | Beta -> begin
    match Beta with
    | Alpha -> print_endline "Beta, Alpha"
    | Beta -> print_endline "Beta, Beta"
  end
```

Run:

`ocamlformat --enable-outside-detected-project --exp-grouping=preserve test.ml`

Produces:

```ocaml
type t = Alpha | Beta

let _ =
  match Alpha with
  | Alpha -> (
      match Beta with
      | Alpha -> print_endline "Alpha, Alpha"
      | Beta -> print_endline "Alpha, Beta")
  | Beta -> (
      match Beta with
      | Alpha -> print_endline "Beta, Alpha"
      | Beta -> print_endline "Beta, Beta")
```

I've also run into this issue in a project with an `.ocamlformat` file defined. For example, this file:

```
profile = conventional
indicate-multiline-delimiters = closing-on-separate-line
exp-grouping = preserve
```

Will respect the `indicate-multiline-delimiters` argument, but will still convert begin/end to parentheses.

gpetiot:
Fix #1599, no diff with test_branch.sh

We can come up with a cleaner solution when we rewrite the formatting of match cases (these ones depend on so many options!)

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -3022,7 +3022,7 @@ and fmt_case c ctx ~first ~last ~padding case =
     | (Pexp_match _ | Pexp_try _), `Align -> last
     | _ -> false
   in
-  let parens_here, parens_for_exp =
+  let parens_branch, parens_for_exp =
     if align_nested_match then (false, Some false)
     else if c.conf.leading_nested_match_parens then (false, None)
     else if is_displaced_infix_op xrhs then (false, None)
@@ -3042,9 +3042,9 @@ and fmt_case c ctx ~first ~last ~padding case =
       (fmt "@;<1000 0>")
   in
   let indent = if align_nested_match then 0 else indent in
-  let p = Params.get_cases c.conf ~first ~indent ~parens_here in
-  let grouping =
-    Params.parens_or_begin_end c.conf c.source ~loc:pc_rhs.pexp_loc
+  let p =
+    Params.get_cases c.conf ~first ~indent ~parens_branch c.source
+      ~loc:pc_rhs.pexp_loc
   in
   p.leading_space $ leading_cmt
   $ p.box_all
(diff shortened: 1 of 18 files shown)


test/passing/tests/exp_grouping.ml.ref
```diff
@@ -272,3 +272,16 @@ let _ =
     foo.fooooo <- Fooo.foo fooo foo.fooooo ;
     Fooo fooo
   end
+
+let _ =
+  match x with
+  | A -> begin
+    match B with A -> fooooooooooooo | B -> fooooooooooooo
+  end
```
Julow, line 280: Would it be rendered as a single line if it fits ? I think we need this test case.

lib/Fmt_ast.ml
```diff
@@ -3050,15 +3053,21 @@ and fmt_case c ctx ~first ~last ~padding case =
                     fmt "@;<1 2>when " $ fmt_expression c (sub_exp ~ctx g) )
               )
           $ p.break_before_arrow $ str "->" $ p.break_after_arrow
-          $ fmt_if parens_here " (" )
+          $ fmt_if parens_here
+              ( match grouping with
+              | `Parens -> " ("
+              | `Begin_end -> "@;<1 0>begin" ) )
```
Julow, line 3059: Can't this be moved into `Params.get_cases` ?
Also, that's yet an other piece of code rendering begin/end, I'd prefer if we can find a way to handle that in a consistent way (from `Params` maybe ?).

test/passing/tests/exp_grouping.ml.ref
gpetiot: It wasn't in my first commit, now it does and I added more tests to see what it looks like for combinations of `break-cases` and `exp-grouping`

lib/Fmt_ast.ml
gpetiot: done!

Julow (approved):
Looks good.

shrik450:
Thank you very much!

PR #1652, opened 2021-05-04
Preserve begin-end keywords around infix operators

### LINKED ISSUE #1594: Bug: exp-grouping is not respected around binary operators
The `begin` and `end` keywords are not preserved when `--exp-grouping=preserve` is set on the command line.

**How to Reproduce**

1. Create a file `test.ml` with the following contents:
  ```ocaml
  let () =
      begin
          add_test @@
          let test_name = "Test 1" in
          test_name >:: fun _ ->
              assert_equal "a" "a"
      end;
      begin
          add_test @@
          let test_name = "Test 2" in
          test_name >:: fun _ ->
              assert_equal "b" "b"
      end;
  ```

2. Run `ocamlformat --enable-outside-detected-project --exp-grouping=preserve test.ml`

The output of the above command is, in my 0.15.0 install, as follows:

```ocaml
let () =
  ( add_test
  @@
  let test_name = "Test 1" in
  test_name >:: fun _ -> assert_equal "a" "a" );
  add_test
  @@
  let test_name = "Test 2" in
  test_name >:: fun _ -> assert_equal "b" "b"
```

Note that `begin` and `end` have been replaced by `(` and `)`.

Interestingly, this does *not* happen unless the `@@` operator appears in the block.  The code

```ocaml
  let () =
      begin
          let test_name = "Test 1" in
          test_name >:: fun _ ->
              assert_equal "a" "a"
      end;
      begin
          let test_name = "Test 2" in
          test_name >:: fun _ ->
              assert_equal "b" "b"
      end;
  ```

transforms to

```ocaml
let () =
  begin
    let test_name = "Test 1" in
    test_name >:: fun _ -> assert_equal "a" "a"
  end;
  let test_name = "Test 2" in
  test_name >:: fun _ -> assert_equal "b" "b"
```

(Note: in the original artifact that inspired this bug report, the context in which the two `begin`-`end` blocks appear is within a `List.iteri`: hence why they are grouped together in this fashion.  I rather wish I could find a way to convince `ocamlformat` not to remove the last pair of "unnecessary" `begin`-`end` tokens as they help readability.)

Thank you!

gpetiot:
Fix #1594, no diff with test_branch.sh

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1462,7 +1462,7 @@ and fmt_sequence c ?ext ~has_attr parens width xexp pexp_loc fmt_atrs =
     opt prev f $ list_pn x fmt_seq
   in
   hvbox 0
-    (Params.wrap_exp c.conf c.source ~loc:pexp_loc ~parens
+    (Params.Exp.wrap c.conf c.source ~loc:pexp_loc ~parens
        ( Params.parens_if has_attr c.conf
            (hvbox_if (parens || has_attr) 0 @@ list_pn grps fmt_seq_list)
        $ fmt_atrs ) )
@@ -1520,15 +1520,14 @@ and fmt_infix_op_args c ~parens xexp op_args =
            $ fmt_if_k (not last) (break 1 0) ) )
     $ fmt_if_k (not last_grp) (break 1 0)
   in
-  Params.wrap_exp c.conf c.source ~loc:xexp.ast.pexp_loc ~parens
+  Params.Exp.Infix_op_arg.wrap c.conf c.source ~loc:xexp.ast.pexp_loc ~parens
     ~parens_nested:(Ast.parenze_nested_exp xexp)
-    ~fits_breaks:false ~infix_op_arg:true
     (list_fl groups fmt_op_arg_group)
 
 and fmt_match c ~parens ?ext ctx xexp cs e0 keyword =
   let indent = Params.match_indent c.conf ~ctx:xexp.ctx in
   hvbox indent
-    (Params.wrap_exp c.conf c.source ~loc:xexp.ast.pexp_loc ~parens
+    (Params.Exp.wrap c.conf c.source ~loc:xexp.ast.pexp_loc ~parens
        ~disambiguate:true
        ( hvbox 0
            ( str keyword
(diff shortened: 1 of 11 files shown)


test/passing/tests/extensions-indent.ml.ref
```diff
@@ -361,17 +361,45 @@ let _ = f [%ext new x]
...
+let _ =
+  [%ext
+       begin
+         y >>= z
+       end]
 
-let _ = [%ext y >>= z]
+let _ =
+  [%ext
+       begin
```
Julow, line 372: This indentation looks like a bug. Should be 2 or 4 but not the column it broke.

lib/Fmt_ast.ml
```diff
@@ -1528,11 +1528,18 @@ and fmt_infix_op_args c ~parens xexp op_args =
...
+  with
+  | `Parens ->
+      wrap_if_k
+        (parens || Ast.parenze_nested_exp xexp)
+        (fits_breaks "(" opn)
+        (fits_breaks ")" ?hint cls)
+        (list_fl groups fmt_op_arg_group)
+  | `Begin_end ->
+      wrap "begin@;<1 2>@[<0>" "@]@;<1000 0>end"
+        (list_fl groups fmt_op_arg_group)
```
Julow, line 1542: Can't this be calling `wrap_exp` below ?

lib/Fmt_ast.ml
```diff
@@ -1520,18 +1520,14 @@ and fmt_infix_op_args c ~parens xexp op_args =
            $ fmt_if_k (not last) (break 1 0) ) )
     $ fmt_if_k (not last_grp) (break 1 0)
   in
-  let opn, hint, cls =
-    if parens || Poly.(c.conf.infix_precedence = `Parens) then
```
gpetiot, line 1524: The `infix-precedence` option is not taken into consideration anymore, but the diff produced by test_branch.sh still looks good, only a few parens have been added.
We could remove this option since it is not used anymore anywhere else in ocamlformat.

lib/Fmt_ast.ml
gpetiot: I've done that now, a few changes in parens, but the code is much easier to maintain, and some cases of breaking the margin have been solved in the testbase.

test/passing/tests/extensions-indent.ml.ref
gpetiot: This indentation is due to the option `--extension-indent=5` for this file

Julow: right!

lib/Fmt_ast.ml
Julow: This option was used only here, it looks like you just un-implemented that option.

lib/Fmt_ast.ml
gpetiot: Right, so I think it's better to just keep the first commit, wrap_exp does not fit this usecase

Julow: What's the reason ?

gpetiot: The code was too different from `wrap_exp`, I tried to merge both into one, but there was arguments/options only used for this "infix_op_arg" usecase and others only used for the general case, so in my last commit I chose to rename `Params.wrap_exp` as `Params.Exp.wrap` and define a new function `Params.Exp.Infix_op_arg.wrap` to use here (in future PRs I would like to put more expression-related functions into the Exp module)

Julow (commented):
Sorry for the slow reply. I think this is good.

Issue #1388, opened 2020-06-05
Bug: ocamlformat tells me to report this bug


dinosaure:
I did not go further but it seems that we got a bug about documentation which can be easily reproducible when the CI got it too:

https://ci.ocamllabs.io/github/mirage/ocaml-conduit/commit/c8bbe6f6b7138da2c1dd2bbb015d09d872e8301a/variant/(lint-fmt)

The commit is: https://github.com/mirage/ocaml-conduit/pull/311/commits/c8bbe6f6b7138da2c1dd2bbb015d09d872e8301a

So if you clone at this commit and do a simple `dune build @fmt`, you should reproduce the problem. On our side, we added `parse-doctstrings=false` into the `.ocamlformat` file.

jberdine:
This is due to the docstring not parsing because of a `]` instead of a `}` [here](https://github.com/mirage/ocaml-conduit/blob/c8bbe6f6b7138da2c1dd2bbb015d09d872e8301a/lwt/conduit_lwt.mli#L29). If you change that typo it seems to work for me.

I wonder if there is something better we can do regarding the error message in the case that a docstring does not parse. @gpetiot, any ideas?

gpetiot:
As Josh said this isn't a bug but badly written error messages, I've improved the situation in #1654, that should be enough to detect what should be changed in the doc-comment. I'm closing this issue but feel free to re-open if necessary!

PR #1654, opened 2021-05-05
Improve the diff of unstable docstrings displayed in error messages

gpetiot:
Improves the situation of #1388 (enough to close it?):
- in case of unstable docstrings between 2 iterations of the formatting, ocamlformat used to display both docstrings, and then the diff of their sexp representation (which was huge and looked more like a bug than a feature). 
I suppose it's better to just display the diff of the 2 docstrings, as it was already done when the formatting of ml code is also unstable.
- I removed the use of ellipsis, there is no point in truncating docstrings if we want to display their diff.
- I also cleaned up the code to have Added/Removed constructors instead of putting everything in Unstable.


test/failing/tests/unstable_docstrings.ml.broken-ref
```diff
@@ -0,0 +1,15 @@
+ocamlformat: Cannot process "tests/unstable_docstrings.ml".
+  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
+  BUG: doc comments changed.
+File "_none_", line 0, characters -1--1::
+Error: Formatting of doc-comment is unstable (e.g. parses as a list or not depending on the margin):
```
Julow, line 5: This test case will disappear when this bug is fixed.

Julow (approved):
Looks good.

test/failing/tests/unstable_docstrings.ml.broken-ref
gpetiot: This is not really a bug as this is expected to have this errormsg if the docstring is invalid, that's why I think we can also close #1388 after this PR is merged

PR #1665, opened 2021-05-26
Display comments in the loc_tree for debugging

gpetiot:
This only affects the `--debug` mode, as shown in the `debug.t` test.

The goal is to display everything in the loc_tree printed when debugging, the implementation is a bit ugly, passing `cmts_before`/`cmts_within`/`cmts_after` functions to the interval tree, I couldn't think of a better design for now, let me know if you see how to improve this!
I find it more convenient to have less lines printed in debug mode, and to have everything in the same place.

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -158,34 +158,10 @@ end = struct
     | _ -> (empty, cmts)
 end
 
-let position_to_string = function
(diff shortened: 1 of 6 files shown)


Julow (approved):
Looks good !

PR #1669, opened 2021-06-02
Vendor ocaml 4.12 parser, drop ppxlib and ocaml-version dependency

### LINKED ISSUE #1583: Feature request: ocamlformat should not depend on ppxlib

gpetiot:
Closes #1583

cc @pitag-ha @emillon this is almost ready, there are just a few regressions to fix, all of the form:
```diff
-ocamlformat: ignoring "tests/unit_types.ml" (syntax error)
-File "tests/unit_types.ml", line 32, characters 7-8:
-Error: Syntax error
+ocamlformat: ignoring "tests/unit_types.ml"
+(Ocaml_412__Parser.MenhirBasics.Error)
```

The `MenhirBasics.Error` exception type should be caught in `Translation_unit.Error.print` but I can't find it defined anywhere, not even in MenhirLib, I will continue to investigate but help is welcome!


vendor/ocaml-4.12/dune
```diff
@@ -0,0 +1,12 @@
+(library
+ (name ocaml_412)
+ (modules_without_implementation let_binding)
+ (flags
+  (:standard -w -9))
+ (libraries compiler-libs.common menhirLib))
+
+(ocamllex lexer)
+
+(menhir
```
Julow, line 10: The `MenhirBasics.Error` thing is not supposed to happen. Here are the flags that are used in the compiler (I removed `--require-aliases`):

```
(menhir
 (flags
  --lalr
  --explain
  --dump
  --strict
  -lg
  1
  -la
  1
  --unused-token
  COMMENT
  --unused-token
  DOCSTRING
  --unused-token
  EOL
  --unused-token
  GREATERRBRACKET
  --fixed-exception
  --table
  --strategy
  simplified)
 (modules parser))
```

Julow (commented):
- Many tests under `test/failing/tests` are unstable (different result depending on OCaml's version) because the error messages changed. I think it's fine to stop running those tests for `ocaml < 4.10`.
- `List.find_map` needs to be implemented for the `Location` module (used for the improved error messages).

vendor/ocaml-4.12/dune
gpetiot: Great, thanks! the CI is alright now, the outputs of the tests are okay since the code printing the error messages is vendored anyway

vendor/ocaml-4.12/ocaml_412.ml
```diff
@@ -0,0 +1,15 @@
...
+module Docstrings = Docstrings
+module Let_binding = Let_binding
+module Lexer = Lexer
+module Location = Location
+module Longident = Longident
+module Parse = Parse
+module Parser = Parser
+module Parsetree = Parsetree
+module Pprintast = Pprintast
+module Printast = Printast
```
Julow, line 13: Do we need these modules ? (same for `Warnings`, `Profile`, `Terminfo`)

vendor/ocaml-4.12/location.ml
```diff
@@ -0,0 +1,956 @@
...
+
+  (* Since OCaml 4.10 *)
+  let rec find_map f = function
+    | [] -> None
+    | x :: l ->
+        begin match f x with
+        | Some _ as result -> result
+        | None -> find_map f l
+        end
+end
```
Julow, line 27: I would prefer if this was implemented outside of this module to keep the vendoring a bit reproducible.
For example, into a module `Parser_shim` that is passed with `-open`. This trick can help to replace the modules we do not care about (`Warning`, `Terminfo`, etc..) with dummy implementations that do nothing.

Ideally, vendored modules would be unchanged and every compatibility tweaks would be in a separate module. It will help a lot the future upgrades that these files can be overwritten easily.

gpetiot: Good point, done!

vendor/ocaml-4.12/ocaml_412.ml
gpetiot: We can remove `Profile` and `Terminfo` indeed. `Warnings` has to be vendored as `Warnings.Unexpected_docstring` has been added in 4.12.

`Printast` and `Pprintast` have to stay in sync with `Parsetree` since their implementation depend on the AST, so they have to stay vendored for now. But they are solely used for debugging I think (I will check) so we can choose to drop these before we merge this PR

vendor/ocaml-4.12/clflags.mli
```diff
@@ -0,0 +1,272 @@
+(**************************************************************************)
```
Julow, line 1: We don't need the whole `Clflags`, we just need some of the variables and I think we want to decide their value instead of using the defaults.

vendor/ocaml-4.12/ocaml_412.ml
Julow: I think it's fine to keep `Printast` and `Pprintast` as long as we know why. We can also change the debug output to use only one of the two.

Julow (commented):
Build time increased by 20% and we now need `menhir` as a build dependency. It's fine for me. The alternative is to commit the generated code (46K LoC), `menhir` would then be a test-dependency.

vendor/ocaml-4.12/clflags.mli
gpetiot: done

vendor/ocaml-4.12/ocaml_412.ml
gpetiot: done in #1671

gpetiot:
> Build time increased by 20% and we now need `menhir` as a build dependency. It's fine for me. The alternative is to commit the generated code (46K LoC), `menhir` would then be a test-dependency.

For most people they will only build ocamlformat once, we maintainers are the only ones impacted by a longer build time, but if we choose to version the generated code we can do it in a separate PR

vendor/ocaml-4.12/pprintast.ml
```diff
@@ -0,0 +1,1677 @@
+(**************************************************************************)
```
Julow, line 1: This file is different than the upstream version:

```diff
@@ -38,16 +38,6 @@ val string_of_structure: Parsetree.structure -> string
 val toplevel_phrase : Format.formatter -> Parsetree.toplevel_phrase -> unit
 val top_phrase: Format.formatter -> Parsetree.toplevel_phrase -> unit
 
-val class_field: Format.formatter -> Parsetree.class_field -> unit
-val class_type_field: Format.formatter -> Parsetree.class_type_field -> unit
-val class_expr: Format.formatter -> Parsetree.class_expr -> unit
-val class_type: Format.formatter -> Parsetree.class_type -> unit
-val module_expr: Format.formatter -> Parsetree.module_expr -> unit
-val module_type: Format.formatter -> Parsetree.module_type -> unit
-val structure_item: Format.formatter -> Parsetree.structure_item -> unit
-val signature_item: Format.formatter -> Parsetree.signature_item -> unit
-val binding: Format.formatter -> Parsetree.value_binding -> unit
-val payload: Format.formatter -> Parsetree.payload -> unit
 
 val tyvar: Format.formatter -> string -> unit
   (** Print a type variable name, taking care of the special treatment
```

vendor/ocaml-4.12/clflags.mli
Julow: Perhaps this module should be moved out of this library to make it clear that it's a stubs and not the origin source. For example in the `parser-shims` new library.

Julow (commented):
The `Let_binding` module is not present in ocaml 4.12, where the `let_binding` type is defined in `Parser`.
This is the only remaining diff with the upstream version. I think the vendored code should never be modified to make upgrading it easier.

Here is a script that reproduces the vendoring:

```bash
# Run with bash

set -e

V=4.12.0
MODULES=(
  ast_helper.ml{,i} ast_mapper.ml{,i} asttypes.mli docstrings.ml{,i}
  lexer.mll location.ml{,i} longident.ml{,i} parse.ml{,i}
  parser.mly parsetree.mli pprintast.ml{,i} syntaxerr.ml{,i}
)

rm vendor/ocaml-4.12/*.*

git fetch "https://github.com/ocaml/ocaml" "$V"

for m in "${MODULES[@]}"; do
  git show "FETCH_HEAD:parsing/$m" > "vendor/ocaml-4.12/$m"
done
```

vendor/ocaml-4.12/asttypes.ml
```diff
@@ -0,0 +1,67 @@
+(**************************************************************************)
```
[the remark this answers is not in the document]
gpetiot: Yes I did that in prupose otherwise menhir would give me this error:
`Error: Module Parser in directory _build/default/vendor/ocaml-4.12 depends on
Ocaml_412.`

I tried to keep the vendored code as close as possible to the 4.12 release and only did big changes like this to avoid getting some error.

gpetiot:
> The `Let_binding` module is not present in ocaml 4.12, where the `let_binding` type is defined in `Parser`.
> This is the only remaining diff with the upstream version. I think the vendored code should never be modified to make upgrading it easier.

Here as well I made that change out of necessity, otherwise I would get this error:
```
File "vendor/parse-wyc/lib/parser.mli", line 446, characters 32-44:
446 |     | N_let_bindings_no_ext_ : (let_bindings) nonterminal
```
and changing the `menhir` flags didn't help

vendor/ocaml-4.12/clflags.ml
```diff
@@ -0,0 +1,35 @@
...
+let principal = ref false               (* -principal *)
+let recursive_types = ref false         (* -rectypes *)
+let applicative_functors = ref true     (* -no-app-funct *)
+let for_package = ref (None: string option) (* -for-pack *)
+let transparent_modules = ref false     (* -trans-mod *)
+let locations = ref true                (* -d(no-)locations *)
+
+let unsafe_string =
+  if Config.safe_string then ref false
+  else ref (not Config.default_safe_string)
```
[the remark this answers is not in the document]
gpetiot: done

vendor/ocaml-4.12/clflags.mli
gpetiot: done

Julow:
I've managed to remove the `Let_binding` module and to keep `Parsetree` and `Asttypes` as mli-only, I made a PR to your PR: https://github.com/gpetiot/ocamlformat/pull/5

vendor/ocaml-4.12/dune
```diff
@@ -0,0 +1,44 @@
+(library
+ (name ocaml_412)
+ (flags
+  (:standard -w -9 -open Parser_shims))
+ (libraries compiler-libs.common menhirLib parser_shims))
```
Julow, line 5:
```suggestion
 (libraries menhirLib parser_shims))
```
Covered by `parser-shims`

vendor/parser-shims/dune
```diff
@@ -0,0 +1,3 @@
+(library
+ (name parser_shims)
+ (libraries compiler-libs.common))
```
Julow, line 3: Compiler-libs is only needed for the `Misc` module. We already shim that module for compatibility reasons, let's vendor `Misc`.

Julow: Actually, vendoring `Misc` is not enough. `Misc` depends on other modules and the code for printing errors too, we'd need to vendor most of the `utils/` directory from the compiler.

PR #1671, opened 2021-06-03
Use Pprintast instead of Printast for debugging

gpetiot:
Allows #1669 to not vendor `printast.ml`.
This only impacts debug mode. The `*.ast` files generated in Translation_unit will print the pretty-printed ast instead of the sexp representation, but it's a good thing as the changes in locations would add too much noise and make these files not exploitable.


Julow (approved):
Looks good.

PR #1672, opened 2021-06-03
Simplify normalization of AST

gpetiot:
Sanitize and simplify the code normalizing the AST, no behavioral change.

There were some duplications with the code checking whether an attribute is a doc-attribute and unnecessary overloading of methods:
- either just redefining the default behavior, calling each function on each member;
- or double checking things like ignore_doc_comments (checked by the `attributes` method anyway), maybe these were considered to be optimizations but I didn't observe any slowdown;
- or sugaring some ast nodes (see the patterns and the value_bindings) but that's not necessary for this feature, that's just made the code harder to maintain.

The hidden goal behind this PR is to reduce the diff on this file with #1669 (replacing ppxlib with ast_mapper)

### DIFF
--- a/lib/Normalize.ml
+++ b/lib/Normalize.ml
@@ -28,8 +28,7 @@ let dedup_cmts fragment ast comments =
 
         method! attribute atr docs =
           match atr with
-          | { attr_name= {txt= "ocaml.doc" | "ocaml.text"; _}
-            ; attr_payload=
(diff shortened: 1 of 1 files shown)


hhugo:
This PR introduced a regression regarding `--no-comment-check` that I'm trying to fix in #2456.
Some logic was removed wrongly assuming it was just an optimization.

PR #1673, opened 2021-06-04
Use UTF8 length of strings, not only in wrapped comments

jberdine:
This improves string length computation, and hence box breaking, in
cases where string literals contain unicode code points, as well as in
docstrings.


gpetiot:
Thanks! do you have a testcase that can be associated to this change?

jberdine:
How about the comment added in the recent commit?

gpetiot (approved):
The test looks good, thanks!

PR #1680, opened 2021-06-07
Implement OCaml 4.13 features

gpetiot:
- [x] no regression from 4.12
- [x] ocaml#10013: Let-punning
- [x] ocaml#9584: named existentials in pattern matching
- [x] ocaml#10133: module type substitutions
- [x] no diff with test_branch.sh


gpetiot:
No idea what is causing some existing tests to break, it looks like Location is losing some data somewhere, but no code related to this has changed between 4.12 and 4.13. I'm open to any idea!

test/failing/tests/unit_modtypes.ml.broken-ref
```diff
@@ -2,8 +2,4 @@ ocamlformat: ignoring "tests/unit_modtypes.ml" (syntax error)
 File "tests/unit_modtypes.ml", line 66, characters 13-14:
 66 |   module M (X) (Y):
                   ^
-Error: Syntax error: 'end' expected
-File "tests/unit_modtypes.ml", line 34, characters 0-3:
-34 | sig
-     ^^^
-  This 'sig' might be unmatched
+Error: Syntax error
```
hhugo, line 5: Why are errors printed differently ?

gpetiot: This is due to https://github.com/ocaml/ocaml/pull/10095

PR #1681, opened 2021-06-08
Remove double parens around a functor in a module application

### LINKED ISSUE #1677: Feature request: avoid redundant parentheses in higher-order functor application

```ocaml
module _ = F ((functor (X : T) -> X))

module _ =
  F
    ((functor
       (X____________________________ : T)
       ->
       X____________________________))
```

In both cases, the argument to the functor `F` is wrapped in two sets of parens when one would do (and would probably be neater).

I leave it to a maintainer to decide whether this is a bug or a feature request :slightly_smiling_face:

gpetiot:
Fix #1677, no diff with test_branch.sh


Julow (approved):
Looks good !

PR #1683, opened 2021-06-08
Depend on odoc-parser instead of odoc

kit-ty-kate:
This PR has been done on top of #1669 because the support for ppxlib master is for the moment required to get ocamlformat on OCaml 4.13. For the same reason, ocamlformat also requires odoc (master) but the API changed a little so here is a PR to fix this.

With this we get it working on OCaml 4.13 for free. Feel free to use https://github.com/kit-ty-kate/opam-alpha-repository/ for testing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,9 @@
 
   + Improve position of `;;` tokens (#1688, @gpetiot)
 
+  + Depend on `odoc-parser` instead of `odoc` (#1683, @kit-ty-kate, @jonludlam)
+    The parser from odoc has been split from the main odoc package and put into its own package, `odoc-parser`.
+
 #### New features
 
   + Implement OCaml 4.13 features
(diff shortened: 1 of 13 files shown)


Julow (commented):
Thanks!

The parser will change again before Odoc 2.0. It will be packaged as a separate opam package, the API may change a bit again and the library name will not be the same.
The opam file needs to be updated with a `pin-depends` to pass CI.

This PR should be marked draft in the meantime I think.

gpetiot:
Any idea why the CI is not passing since the `odoc-parser` package is found on opam?

kit-ty-kate:
That looks like a bug in ocaml-ci. I'll have a look

kit-ty-kate:
@gpetiot https://github.com/ocurrent/ocaml-ci/pull/353 should fix this. Hopefully it'll be merged soon.

gpetiot:
Thanks for having a look!

talex5:
I've just merged it.

gpetiot:
Thank you :)

gpetiot (approved):

PR #1685, opened 2021-06-14
Disable the indentation feature in emacs for ocamlformat < 0.19.0

gpetiot:
Follow-up of #1639, tested with the current master branch, works well when simulating 0.16.* and 0.19.* versions as well.
cc @bcc32 let me know if that fixes the issue.

### DIFF
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -398,7 +398,8 @@ With ARG, perform this action that many times."
   "Get the version of the installed ocamlformat."
   (car
    (split-string
-    (shell-command-to-string (format "%s --version" ocamlformat-command))
+    (shell-command-to-string
+     (format "%s --version" (shell-quote-argument ocamlformat-command)))
     "-")))
 
 (defun enable-indent ()
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -408,12 +408,11 @@ With ARG, perform this action that many times."
   "Whether the indentation feature is enabled."
   (version<= "0.19.0" (ocamlformat-version)))
 
-(if (enable-indent)
-  (progn
-    (add-hook 'tuareg-mode-hook 'ocamlformat-setup-indent t)
-    (add-hook 'tuareg-mode-hook 'set-newline-and-indent)
-    (add-hook 'caml-mode-hook 'ocamlformat-caml-mode-setup  t)
-    (add-hook 'caml-mode-hook 'set-newline-and-indent)))
+(when (enable-indent)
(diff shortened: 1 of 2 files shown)


emacs/ocamlformat.el
```diff
@@ -394,11 +394,23 @@ With ARG, perform this action that many times."
...
+  (car
+   (split-string
+    (shell-command-to-string (format "%s --version" ocamlformat-command))
+    "-")))
+
+(defun enable-indent ()
+  "Whether the indentation feature is enabled."
+  (version<= "0.19.0" (ocamlformat-version)))
+
+(if (enable-indent)
```
bcc32, line 408: FYI, you can replace `if...progn` with `when`.

emacs/ocamlformat.el
```diff
@@ -394,11 +394,23 @@ With ARG, perform this action that many times."
...
   (local-set-key (kbd "RET") 'custom_newline-and-indent))
 
-(add-hook 'tuareg-mode-hook 'ocamlformat-setup-indent t)
-(add-hook 'tuareg-mode-hook 'set-newline-and-indent)
-
-(add-hook 'caml-mode-hook 'ocamlformat-caml-mode-setup  t)
-(add-hook 'caml-mode-hook 'set-newline-and-indent)
+(defun ocamlformat-version ()
+  "Get the version of the installed ocamlformat."
+  (car
```
bcc32, line 399: I think `version<=` should be able to handle the likely snapshot version syntax (I don't know if you had a particular syntax in mind that it doesn't handle, see `version-to-list`).

bcc32:
This does solve the issue for me, thanks @gpetiot.  I made some small elisp suggestions as well, hope that's okay.

bcc32:
Actually, I think you also need to trim trailing whitespace from `shell-command-to-string` in order to get the correct version string, and you may want to `shell-quote-argument` the `ocamlformat-command` as well.

emacs/ocamlformat.el
gpetiot: The unreleased versions are of the form `0.18.0-37-g587b36a0-dirty` and they are not handled by `version<=`, but that's okay to only keep what's on the left hand side of the first `-` for the comparison

emacs/ocamlformat.el
gpetiot: done

gpetiot:
> Actually, I think you also need to trim trailing whitespace from `shell-command-to-string` in order to get the correct version string, and you may want to `shell-quote-argument` the `ocamlformat-command` as well.

Good suggestion, I fixed that!

bcc32:
Thanks for fixing this!

PR #1687, opened 2021-06-17
Enable indentation buffer by buffer

bcc32:
When enabling indentation, we should check each buffer individually,
as $PATH may have changed (local variables, direnv-mode, etc.), or the
version of ocamlformat in $PATH may have changed (local opam
switches).

This also prevents ocamlformat.el from failing to load if ocamlformat
cannot be found at emacs startup.


bcc32:
This PR also contains some minor fixups in separate commits.  Hopefully, they are pretty unobjectionable.

gpetiot (approved):
I tried it and it looks good to me, thanks!

PR #1667, opened 2021-05-28
Preserve position of comments attached to the last node of a subtree

gpetiot:
Fix #1662 
Fix #1963

Diff for the `janestreet` profile:

<details>

```diff
diff --git a/infer/src/topl/ToplAst.ml b/infer/src/topl/ToplAst.ml
index 3b606a25c..c964bc840 100644
--- a/infer/src/topl/ToplAst.ml
+++ b/infer/src/topl/ToplAst.ml
@@ -31,7 +31,6 @@ type predicate =
   | Value of (* bool *) value
 
 type condition = predicate list (* conjunction *)
-
 type assignment = register_name * variable_name
 
 (** a regular expression *)
diff --git a/src/dune_engine/rules.mli b/src/dune_engine/rules.mli
index dac6da247..95e18a85d 100644
--- a/src/dune_engine/rules.mli
+++ b/src/dune_engine/rules.mli
@@ -26,7 +26,7 @@ module Dir_rules : sig
 
            When passing [--force] to Dune, these are exactly the actions that
            will be re-executed. *)
-          Action of
+        Action of
           Rule.Anonymous_action.t Action_builder.t
 
     type t = { expansions : (Loc.t * item) Appendable_list.t } [@@unboxed]
diff --git a/src/dune_engine/scheduler.ml b/src/dune_engine/scheduler.ml
index 428a6ba56..e09e45d3f 100644
--- a/src/dune_engine/scheduler.ml
+++ b/src/dune_engine/scheduler.ml
@@ -680,23 +680,23 @@ type status =
   | (* Ready to start the next build. Waiting for a signal from the user, the
        test harness, or the polling loop. The payload is the collection of
        filesystem events. *)
-      Standing_by of
+    Standing_by of
       Memo.Invalidation.t
   | (* Waiting for file changes to start a new a build *)
-      Waiting_for_file_changes of
+    Waiting_for_file_changes of
       Memo.Invalidation.t Fiber.Ivar.t
   | (* Waiting for the propagation of inotify events to finish before starting a
        build. *)
-      Waiting_for_inotify_sync of
+    Waiting_for_inotify_sync of
       Memo.Invalidation.t * unit Fiber.Ivar.t
   | (* Running a build *)
-      Building
+    Building
   | (* Cancellation requested. Build jobs are immediately rejected in this
        state *)
-      Restarting_build of
+    Restarting_build of
       Memo.Invalidation.t
   | (* Shut down requested. No new new builds will start *)
-      Shutting_down
[...]

PR #1694, opened 2021-06-28
Preserve bracketed lists in the Parsetree

gpetiot:
This simplifies the formatting of lists, removes the need of desugaring the `(::)` constructs, and will help improve the position of comments in lists when I tackle the issue (the lists were a big regression in #1667).

Some adjustments are made in Ast.ml to minimize the diff, we thus keep unnecessary parentheses around a list element in some cases (ppat_alias, ppat_or).

The diff of test_branch.sh looks good, most of it is caused by comments wrapping around list elements, and that's a fix:
<details>
```diff
diff --git a/infer/src/absint/ConcurrencyModels.ml b/infer/src/absint/ConcurrencyModels.ml
index dd999c965..a7b14ede0 100644
--- a/infer/src/absint/ConcurrencyModels.ml
+++ b/infer/src/absint/ConcurrencyModels.ml
@@ -156,24 +156,15 @@ end = struct
   let guards =
     (* TODO std::scoped_lock *)
     [
-      (* no lock/unlock *)
-      "apache::thrift::concurrency::Guard";
-      (* no lock/unlock *)
-      "apache::thrift::concurrency::RWGuard";
-      (* only unlock *)
-      "folly::SharedMutex::ReadHolder";
-      (* only unlock *)
-      "folly::SharedMutex::WriteHolder";
-      (* read/write locks under operator() etc *)
-      "folly::LockedPtr";
-      (* no lock/unlock *)
-      "folly::SpinLockGuard";
-      (* no lock/unlock *)
-      "std::lock_guard";
-      (* everything *)
-      "std::shared_lock";
-      (* everything *)
-      "std::unique_lock";
+      (* no lock/unlock *) "apache::thrift::concurrency::Guard";
+      (* no lock/unlock *) "apache::thrift::concurrency::RWGuard";
+      (* only unlock *) "folly::SharedMutex::ReadHolder";
+      (* only unlock *) "folly::SharedMutex::WriteHolder";
+      (* read/write locks under operator() etc *) "folly::LockedPtr";
+      (* no lock/unlock *) "folly::SpinLockGuard";
+      (* no lock/unlock *) "std::lock_guard";
+      (* everything *) "std::shared_lock";
+      (* everything *) "std::unique_lock";
     ]
 
   let ( get_guard_constructor,
diff --git a/infer/src/absint/annotations.ml b/infer/src/absint/annotations.ml
index 5fdb8e507..936d72831 100644
--- a/infer/src/absint/annotations.ml
+++ b/infer/src/absint/annotations.ml
@@ -190,13 +190,13 @@ let ia_is_nullable ia =
       (* From org.checkerframework.checker.nullness.compatqual package. Extensively used in Guava library.
          Identical to {@code @Nullable}, but can only be written at declaration locations. *);
       propagates_nullable (* @PropagatesNullable is implicitly nullable *);
-      recently_nullable;
+      recently_nullable
       (* @RecentlyNullable is a special annotation that was added to solve backward compatibility issues
          for Android SDK migration.
          See https://android-developers.googleblog.com/2018/08/android-pie-sdk-is-now-more-kotlin.html for details.
          From nullsafe point of view, such annotations should be treated exactly as normal @Nullable annotation.
          (Actually, it might even be shown as @Nullable in IDE/source code)
-      *)
+      *);
     ]
 
 let ia_is_nonnull ia =
@@ -205,13 +205,13 @@ let ia_is_nonnull ia =
       nonnull;
       notnull;
       camel_nonnull;
-      recently_nonnull;
+      recently_nonnull
       (* @RecentlyNonNull is a special annotation that was added to solve backward compatibility issues
          for Android SDK migration.
          See https://android-developers.googleblog.com/2018/08/android-pie-sdk-is-now-more-kotlin.html for details.
          From nullsafe point of view, such annotations should be treated exactly as normal @NonNull annotation.
          (Actually, it might even be shown as @NonNull in IDE/source code)
-      *)
+      *);
     ]
 
 let ia_is_nullsafe_strict ia = ia_ends_with ia nullsafe_strict
diff --git a/infer/src/base/CommandDoc.ml b/infer/src/base/CommandDoc.ml
index 108aca35d..bd5767fe0 100644
--- a/infer/src/base/CommandDoc.ml
+++ b/infer/src/base/CommandDoc.ml
@@ -53,7 +53,8 @@ let mk_command_doc ~see_also:see_also_commands ?environment:environment_opt
         ]
   in
   CLOpt.mk_command_doc ~section ~version:Version.versionString
-    ~date:Version.man_pages_last_modify_date ~synopsis:[ `Pre synopsis ]
+    ~date:Version.man_pages_last_modify_date
+    ~synopsis:[ `Pre synopsis ]
     ~environment ~files ?see_also
 
 let analyze =
diff --git a/infer/src/base/Config.ml b/infer/src/base/Config.ml
index 5b7a2b91a..df1e509ec 100644
--- a/infer/src/base/Config.ml
+++ b/infer/src/base/Config.ml
@@ -680,7 +680,8 @@ and biabduction_join_cond =
 
 and biabduction_memleak_buckets =
   CLOpt.mk_symbol_seq ~deprecated:[ "-ml-buckets" ]
-    ~long:"biabduction-memleak-buckets" ~default:[ `MLeak_cf ]
+    ~long:"biabduction-memleak-buckets"
+    ~default:[ `MLeak_cf ]
     ~symbols:ml_bucket_symbols ~eq:PolyVariantEqual.( = )
     "Specify the memory leak buckets to be checked in C++."
 
diff --git a/infer/src/bufferoverrun/bufferOverrunModels.ml b/infer/src/bufferoverrun/bufferOverrunModels.ml
index 58ddce6ea..c2c632d34 100644
--- a/infer/src/bufferoverrun/bufferOverrunModels.ml
+++ b/infer/src/bufferoverrun/bufferOverrunModels.ml
@@ -2024,8 +2024,7 @@ module Call = struct
         $ capt_arg_of_typ (-"std" &:: "vector")
         $--> StdVector.constructor_empty;
         -"google" &:: "StrLen" <>$ capt_exp $--> strlen;
-        (* Java models *)
-        -"java.lang.Object" &:: "clone" <>$ capt_exp $--> id;
+        (* Java models *) -"java.lang.Object" &:: "clone" <>$ capt_exp $--> id;
         +PatternMatch.Java.implements_arrays
         &:: "asList" <>$ capt_exp $!--> create_copy_array;
         +PatternMatch.Java.implements_arrays
diff --git a/infer/src/checkers/impurity.ml b/infer/src/checkers/impurity.ml
index 624f971ce..564509d19 100644
--- a/infer/src/checkers/impurity.ml
+++ b/infer/src/checkers/impurity.ml
@@ -109,7 +109,8 @@ let add_to_modified pname ~pvar ~access ~addr pre_heap post modified_vars =
                        attrs_post access_list modified_vars)
                     ~addr_to_explore ~visited))
   in
-  aux ([], modified_vars) ~addr_to_explore:[ (access, addr) ]
+  aux ([], modified_vars)
+    ~addr_to_explore:[ (access, addr) ]
     ~visited:AbstractValue.Set.empty
 
 let get_modified_params pname post_stack pre_heap post formals =
diff --git a/infer/src/nullsafe/modelTables.ml b/infer/src/nullsafe/modelTables.ml
index 3e8e67ac4..c38694dea 100644
--- a/infer/src/nullsafe/modelTables.ml
+++ b/infer/src/nullsafe/modelTables.ml
@@ -770,8 +770,7 @@ let annotated_list_nullability_other =
     ( (n, [ o; n; n ]),
       "org.w3c.dom.Node.setUserData(java.lang.String,java.lang.Object,org.w3c.dom.UserDataHandler):java.lang.Object"
     );
-    (* References *)
-    (ng, "java.lang.ref.Reference.get():java.lang.Object");
+    (* References *) (ng, "java.lang.ref.Reference.get():java.lang.Object");
     (ng, "java.lang.ref.PhantomReference.get():java.lang.Object");
     (ng, "java.lang.ref.SoftReference.get():java.lang.Object");
     (ng, "java.lang.ref.WeakReference.get():java.lang.Object");
diff --git a/infer/src/unit/abstractInterpreterTests.ml b/infer/src/unit/abstractInterpreterTests.ml
index 6da02797a..cfd6ec417 100644
--- a/infer/src/unit/abstractInterpreterTests.ml
+++ b/infer/src/unit/abstractInterpreterTests.ml
@@ -150,8 +150,7 @@ let tests =
                 (* note: each instruction in try block is treated as potentially-excepting... *)
                 (* point 1 *)
                 invariant "1";
-                (* point 2 *)
-                invariant "1";
+                (* point 2 *) invariant "1";
               ],
               (* point 4 *)
               [
diff --git a/sledge/cli/frontend.ml b/sledge/cli/frontend.ml
index 6719f49f3..6c3a4b9c0 100644
--- a/sledge/cli/frontend.ml
+++ b/sledge/cli/frontend.ml
@@ -1010,7 +1010,7 @@ let xlate_intrinsic_inst emit_inst x name_segs instr num_actuals loc =
   | [ "_Znam" (* operator new[](unsigned long) *) ]
   | [ "_Znwm" (* operator new(size_t num) *) ]
   | [
-      "_ZnwmSt11align_val_t"; (* operator new(unsigned long, std::align_val_t) *)
+      "_ZnwmSt11align_val_t" (* operator new(unsigned long, std::align_val_t) *);
     ]
     when num_actuals > 0 ->
       let reg = xlate_name x instr in
@@ -1019,11 +1019,11 @@ let xlate_intrinsic_inst emit_inst x name_segs instr num_actuals loc =
       emit_inst ~prefix (Inst.alloc ~reg ~num ~len ~loc)
   | [ "_ZdlPv" (* operator delete(void* ptr) *) ]
   | [
-      "_ZdlPvSt11align_val_t"; (* operator delete(void* ptr, std::align_val_t) *)
+      "_ZdlPvSt11align_val_t" (* operator delete(void* ptr, std::align_val_t) *);
     ]
   | [
-      "_ZdlPvmSt11align_val_t";
-      (* operator delete(void* ptr, unsigned long, std::align_val_t) *)
+      "_ZdlPvmSt11align_val_t"
+      (* operator delete(void* ptr, unsigned long, std::align_val_t) *);
     ]
   | [ "free" (* void free(void* ptr) *) ] ->
       let prefix, ptr = xlate_value x (Llvm.operand instr 0) in
diff --git a/benchmarks/sources/ml/nucleic.ml b/benchmarks/sources/ml/nucleic.ml
index ec7b8ba628..4fb1cd94e5 100644
--- a/benchmarks/sources/ml/nucleic.ml
+++ b/benchmarks/sources/ml/nucleic.ml
@@ -7107,11 +7107,8 @@ let anticodon_domains =
     stacked3' rA 36 37;
     stacked3' rA 35 36;
     stacked3' rG 34 35;
-    (* <-. Distance      *)
-    p_o3' rCs 32 31;
-    (*   | Constraint    *)
-    p_o3' rUs 33 32;
-    (* <-' 3.0 Angstroms *)
+    (* <-. Distance      *) p_o3' rCs 32 31;
+    (*   | Constraint    *) p_o3' rUs 33 32 (* <-' 3.0 Angstroms *);
   ]
 
 (* Anticodon constraint *)
@@ -7139,41 +7136,29 @@ let pseudoknot_domains =
     helix3' rC 20 21;
     wc_dumas rG 11 20;
     helix3' rU' 19 20;
-    (* <-.               *)
-    wc_dumas rA 12 19;
+    (* <-.               *) wc_dumas rA 12 19;
     (*   | Distance      *)
     (*                             | Constraint    *)
     (*  Helix 1                    | 4.0 Angstroms *)
     helix3' rC 3 19;
-    (*   |               *)
-    wc_dumas rG 13 3;
-    (*   |               *)
-    helix3' rC 2 3;
-    (*   |               *)
-    wc_dumas rG 14 2;
-    (*   |               *)
-    helix3' rC 1 2;
-    (*   |               *)
-    wc_dumas rG' 15 1;
+    (*   |               *) wc_dumas rG 13 3;
+    (*   |               *) helix3' rC 2 3;
+    (*   |               *) wc_dumas rG 14 2;
+    (*   |               *) helix3' rC 1 2;
+    (*   |               *) wc_dumas rG' 15 1;
     (*   |               *)
     (*                             |               *)
     (*  L2 LOOP                    |               *)
     p_o3' rUs 16 15;
-    (*   |               *)
-    p_o3' rCs 17 16;
-    (*   |               *)
-    p_o3' rAs 18 17;
+    (*   |               *) p_o3' rCs 17 16;
+    (*   |               *) p_o3' rAs 18 17;
     (* <-'               *)
     (*                                             *)
     (*  L1 LOOP                                    *)
     helix3' rU 7 8;
-    (* <-.               *)
-    p_o3' rCs 4 3;
-    (*   | Constraint    *)
-    stacked5' rU 5 4;
-    (*   | 4.5 Angstroms *)
-    stacked5' rC 6 5;
-    (* <-'               *)
+    (* <-.               *) p_o3' rCs 4 3;
+    (*   | Constraint    *) stacked5' rU 5 4;
+    (*   | 4.5 Angstroms *) stacked5' rC 6 5 (* <-'               *);
   ]
 
 (* Pseudoknot constraint *)
diff --git a/compiler/lib/reserved.ml b/compiler/lib/reserved.ml
index 8c2881d088..cc4ccb2017 100644
--- a/compiler/lib/reserved.ml
+++ b/compiler/lib/reserved.ml
@@ -24,8 +24,7 @@ let keyword =
     ~f:(fun acc x -> StringSet.add x acc)
     ~init:StringSet.empty
     [
-      (* keyword *)
-      "break";
+      (* keyword *) "break";
       "case";
       "catch";
       "continue";
@@ -51,8 +50,7 @@ let keyword =
       "void";
       "while";
       "with";
-      (* reserved in ECMAScript 5 *)
-      "class";
+      (* reserved in ECMAScript 5 *) "class";
       "enum";
       "export";
       "extends";
@@ -67,18 +65,15 @@ let keyword =
       "public";
       "static";
       "yield";
-      (* other *)
-      "null";
+      (* other *) "null";
       "true";
       "false";
       "NaN";
       "undefined";
       "this";
-      (* Unexpected eval or arguments in strict mode *)
-      "eval";
+      (* Unexpected eval or arguments in strict mode *) "eval";
       "arguments";
-      (* also reserved in ECMAScript 3 *)
-      "abstract";
+      (* also reserved in ECMAScript 3 *) "abstract";
       "boolean";
       "byte";
       "char";
@@ -95,8 +90,7 @@ let keyword =
       "throws";
       "transient";
       "volatile";
-      (* also reserved in ECMAScript 6 *)
-      "await";
+      (* also reserved in ECMAScript 6 *) "await";
     ]
 
 let provided =
diff --git a/compiler/tests-ocaml/prim-revapply/apply.ml b/compiler/tests-ocaml/prim-revapply/apply.ml
index 4f58586251..aaa938a0c5 100644
--- a/compiler/tests-ocaml/prim-revapply/apply.ml
+++ b/compiler/tests-ocaml/prim-revapply/apply.ml
@@ -18,15 +18,10 @@ let _ =
       print_newline ())
     [
       f @@ 3;
-      (* 6 *)
-      g @@ f @@ 3;
-      (* 36 *)
-      f @@ g @@ 3;
-      (* 18 *)
-      h @@ g @@ f @@ 3;
-      (* 37 *)
-      add 4 @@ g @@ f @@ add 3 @@ add 2 @@ 3;
-      (* 260 *)
+      (* 6 *) g @@ f @@ 3;
+      (* 36 *) f @@ g @@ 3;
+      (* 18 *) h @@ g @@ f @@ 3;
+      (* 37 *) add 4 @@ g @@ f @@ add 3 @@ add 2 @@ 3 (* 260 *);
     ]
 
 external ( @@ ) : ('a -> 'b) -> 'a -> 'b = "%apply"
@@ -46,13 +41,8 @@ let _ =
       print_newline ())
     [
       f @@ 3;
-      (* 6 *)
-      g @@ f @@ 3;
-      (* 36 *)
-      f @@ g @@ 3;
-      (* 18 *)
-      h @@ g @@ f @@ 3;
-      (* 37 *)
-      add 4 @@ g @@ f @@ add 3 @@ add 2 @@ 3;
-      (* 260 *)
+      (* 6 *) g @@ f @@ 3;
+      (* 36 *) f @@ g @@ 3;
+      (* 18 *) h @@ g @@ f @@ 3;
+      (* 37 *) add 4 @@ g @@ f @@ add 3 @@ add 2 @@ 3 (* 260 *);
     ]
diff --git a/compiler/tests-ocaml/prim-revapply/revapply.ml b/compiler/tests-ocaml/prim-revapply/revapply.ml
index bf320ab6b9..fafed316ba 100644
--- a/compiler/tests-ocaml/prim-revapply/revapply.ml
+++ b/compiler/tests-ocaml/prim-revapply/revapply.ml
@@ -18,13 +18,8 @@ let _ =
       print_newline ())
     [
       3 |> f;
-      (* 6 *)
-      3 |> f |> g;
-      (* 36 *)
-      3 |> g |> f;
-      (* 18 *)
-      3 |> f |> g |> h;
-      (* 37 *)
-      3 |> add 2 |> add 3 |> f |> g |> add 4;
-      (* 260 *)
+      (* 6 *) 3 |> f |> g;
+      (* 36 *) 3 |> g |> f;
+      (* 18 *) 3 |> f |> g |> h;
+      (* 37 *) 3 |> add 2 |> add 3 |> f |> g |> add 4 (* 260 *);
     ]
diff --git a/src/dune_engine/process.ml b/src/dune_engine/process.ml
index 506501baf..c352d49db 100644
--- a/src/dune_engine/process.ml
+++ b/src/dune_engine/process.ml
@@ -495,7 +495,8 @@ end = struct
                     [
                       (has_unexpected_stdout, "stdout");
                       (has_unexpected_stderr, "stderr");
-                    ] ~f:(fun (b, name) -> Option.some_if b name)
+                    ]
+                    ~f:(fun (b, name) -> Option.some_if b name)
                 in
                 match (n, unexpected_outputs) with
                 | 0, _ :: _ ->
diff --git a/src/dune_rules/findlib/meta.ml b/src/dune_rules/findlib/meta.ml
index df9d6f78a..385ba13ef 100644
--- a/src/dune_rules/findlib/meta.ml
+++ b/src/dune_rules/findlib/meta.ml
@@ -141,7 +141,8 @@ let archives ?(kind = [ Mode.Byte; Mode.Native ]) name =
       (Mode.Native, archive, Mode.compiled_lib_ext);
       (Mode.Byte, plugin, Mode.compiled_lib_ext);
       (Mode.Native, plugin, Mode.plugin_ext);
-    ] ~f:(fun (k, f, ext) ->
+    ]
+    ~f:(fun (k, f, ext) ->
       if List.mem kind k ~equal:Mode.equal then
         Some (f (Mode.to_string k) (name ^ ext k))
       else None)
diff --git a/lambda/matching.ml b/lambda/matching.ml
index 1a40adb53..a2da2d86b 100644
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -702,7 +702,7 @@ end = struct
                the row is not empty. *)
           match (filter_one p1 ps, filter_one p2 ps) with
           | [], row | row, [] -> row @ filter_rec rem
-          | [ (arg1 :: _) ], [ (arg2 :: _) ] ->
+          | [ arg1 :: _ ], [ arg2 :: _ ] ->
               (* By the row shape property,
                  the wildcard patterns can only be ps. *)
               (* The output below is a single row,
@@ -3234,7 +3234,8 @@ let for_trywith ~scopes loc param pat_act_list =
 
 let simple_for_let ~scopes loc param pat body =
   compile_matching ~scopes loc ~failer:Raise_match_failure None param
-    [ (pat, body) ] Partial
+    [ (pat, body) ]
+    Partial
 
 (* Optimize binding of immediate tuples
 
@@ -3487,7 +3488,8 @@ let do_for_multiple_match ~scopes loc paraml pat_act_list partial =
     let partial = check_partial pat_act_list partial in
     let rows = map_on_rows (fun p -> (p, [])) pat_act_list in
     toplevel_handler ~scopes loc ~failer:Raise_match_failure partial
-      [ (arg, Strict) ] rows
+      [ (arg, Strict) ]
+      rows
   in
   handler (fun partial pm1 ->
       let pm1_half =
diff --git a/lambda/translclass.ml b/lambda/translclass.ml
index f2f5931e5..ef35bfa13 100644
--- a/lambda/translclass.ml
+++ b/lambda/translclass.ml
@@ -212,7 +212,8 @@ let rec build_object_init ~scopes cl_table obj params inh_init obj_init cl =
diff --git a/lambda/matching.ml b/lambda/matching.ml
index 1a40adb53..a2da2d86b 100644
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -702,7 +702,7 @@ end = struct
                the row is not empty. *)
           match (filter_one p1 ps, filter_one p2 ps) with
           | [], row | row, [] -> row @ filter_rec rem
-          | [ (arg1 :: _) ], [ (arg2 :: _) ] ->
+          | [ arg1 :: _ ], [ arg2 :: _ ] ->
               (* By the row shape property,
                  the wildcard patterns can only be ps. *)
               (* The output below is a single row,
@@ -3234,7 +3234,8 @@ let for_trywith ~scopes loc param pat_act_list =
 
 let simple_for_let ~scopes loc param pat body =
   compile_matching ~scopes loc ~failer:Raise_match_failure None param
-    [ (pat, body) ] Partial
+    [ (pat, body) ]
+    Partial
 
 (* Optimize binding of immediate tuples
 
@@ -3487,7 +3488,8 @@ let do_for_multiple_match ~scopes loc paraml pat_act_list partial =
     let partial = check_partial pat_act_list partial in
     let rows = map_on_rows (fun p -> (p, [])) pat_act_list in
     toplevel_handler ~scopes loc ~failer:Raise_match_failure partial
-      [ (arg, Strict) ] rows
+      [ (arg, Strict) ]
+      rows
   in
   handler (fun partial pm1 ->
       let pm1_half =
diff --git a/lambda/translclass.ml b/lambda/translclass.ml
index f2f5931e5..ef35bfa13 100644
--- a/lambda/translclass.ml
+++ b/lambda/translclass.ml
@@ -212,7 +212,8 @@ let rec build_object_init ~scopes cl_table obj params inh_init obj_init cl =
               loc = of_location ~scopes pat.pat_loc;
               body =
                 Matching.for_function ~scopes pat.pat_loc None (Lvar param)
-                  [ (pat, rem) ] partial;
+                  [ (pat, rem) ]
+                  partial;
             }
         in
         match obj_init with
@@ -505,7 +506,8 @@ let rec transl_class_rebind ~scopes obj_init cl vf =
             loc = of_location ~scopes pat.pat_loc;
             body =
               Matching.for_function ~scopes pat.pat_loc None (Lvar param)
-                [ (pat, rem) ] partial;
+                [ (pat, rem) ]
+                partial;
           }
       in
       ( path,
@@ -590,13 +592,15 @@ let transl_class_rebind ~scopes cl vf =
                 ( Pmakeblock (0, Immutable, None),
                   [
                     mkappl (Lvar new_init, [ lfield cla 0 ]);
-                    lfunction [ (table, Pgenval) ]
+                    lfunction
+                      [ (table, Pgenval) ]
                       (Llet
                          ( Strict,
                            Pgenval,
                            env_init,
                            mkappl (lfield cla 1, [ Lvar table ]),
-                           lfunction [ (envs, Pgenval) ]
+                           lfunction
+                             [ (envs, Pgenval) ]
                              (mkappl
                                 ( Lvar new_init,
                                   [ mkappl (Lvar env_init, [ Lvar envs ]) ] ))
diff --git a/lambda/translcore.ml b/lambda/translcore.ml
index d8cc8ac2d..bb6b21336 100644
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -915,7 +915,8 @@ and transl_curried_function ~scopes loc return repr partial (param : Ident.t)
               cases'
           in
           ( (Curried, (param, kind) :: params, return),
-            Matching.for_function ~scopes loc None (Lvar param) [ (pat, body) ]
+            Matching.for_function ~scopes loc None (Lvar param)
+              [ (pat, body) ]
               partial )
         else (
           (match partial with
@@ -1251,7 +1252,9 @@ and transl_match ~scopes e arg pat_expr_list partial =
     | arg, _ :: _ ->
         let val_id = Typecore.name_pattern "val" (List.map fst val_cases) in
         let k = Typeopt.value_kind arg.exp_env arg.exp_type in
-        static_catch [ transl_exp ~scopes arg ] [ (val_id, k) ]
+        static_catch
+          [ transl_exp ~scopes arg ]
+          [ (val_id, k) ]
           (Matching.for_function ~scopes e.exp_loc None (Lvar val_id) val_cases
              partial)
```
</details>


Julow:
```diff

**Review by Julow — APPROVED** [review]
Good idea ! A lot of "preserving" code is gone.

**Comment by hhugo** [comment]
Should something similar be done for String, Array, Bigarray getter and setter ?

**Comment by gpetiot** [comment]
> Should something similar be done for String, Array, Bigarray getter and setter ?

We should look into it indeed

**Comment by gpetiot** [comment]
I removed unnecessary parens, the parens are only necessary in cases such as `x :: ([y; z] [@attr])` but we always did it without checking for the context, that's a big diff but it looks better.

**Comment by jberdine** [comment]
Maybe I'm paranoid, but having a vendored parser, different parsetree, and so on makes me nervous about correctness. Has it been considered to perform the check that the round-trip through parsing-formatting-parsing gives "equal" parsetrees using the unmodified compiler-libs parser? The "formatting" step would then involve parsing with the modified parser. This would keep the strong property that reformatting does not change how the compiler interprets the code, while still allowing the modified vendored parser to be used to simplify the implementation.

**Comment by Julow** [comment]
> Has it been considered to perform the check that the round-trip through parsing-formatting-parsing gives "equal" parsetrees using the unmodified compiler-libs parser?

That's a great idea. However this would require maintaining a second normalization function and making OCamlformat slower.
The problem with the normalization function is that it needs to elide some data that OCamlformat is allowed to change and these details will need to be kept in sync between the two parsetrees. (the normalization function is needed to do that but also to print a readable diff)

Alternatively, we could use OCaml's `Pprintast` to normalize the upstream AST. We do not need to match on the parsetree, so maintainance is a lot easier.
However, we'll always find diffs so we'll need to run this on the tests only and to promote the diff to the source tree. That way, we'll always know what are the diffs OCamlformat is allowed to make and it will be a test failure only if these diffs change.
This test can only run on a specific version of OCaml. It should be possible to run it on external repositories (like we do with `test_branch`) and keep the diffs here.

That's less useful than what you suggested but it's easier to maintain and has no overhead for users. I think that's fine for now, as long as changes to the parser are small. (which are only going to increase)

File: lib/Ast.ml [hunk 1]
```diff
@@ -2286,6 +2286,11 @@ end = struct
     | Exp {pexp_desc= Pexp_field (e, _); _}, {pexp_desc= Pexp_construct _; _}
       when e == exp ->
         true
+    | ( Exp {pexp_desc= Pexp_apply (op, [_; (Nolabel, e)]); _}
+      , {pexp_desc= Pexp_list _; pexp_attributes= _ :: _; _} )
+      when Exp.is_infix op && e == exp ->
+        true
+    | _, {pexp_desc= Pexp_list _; _} -> false
```
Julow, line 2293: This case seems too easy and shortcuts the next case, which checks a lot of things. Here is an example where the AST changes:
```ocaml
let _ = f 1 ([e; f] [@a])
```
I think this case should be moved to after the next one, which already does what we expect (except in the local open case). I'm still not 100% confident this will be ok in every cases.

Julow (changes requested):

jberdine:
> That's a great idea. However this would require maintaining a second normalization function and making OCamlformat slower.
> The problem with the normalization function is that it needs to elide some data that OCamlformat is allowed to change and these details will need to be kept in sync between the two parsetrees. (the normalization function is needed to do that but also to print a readable diff)

I was thinking that the normalization function could be changed to only operate on the compiler-libs parsetree. But maybe there are now uses of normalization other than to perform this correctness check? I did a quick scan and it seems that it would be possible to split the current normalization into the part dealing with comments and docstrings, which would continue to use the vendored parsetree, and the part that deals with non-comments, which could use the upstream parsetree.

As for being slower, yes, but only slightly, and by less than one round of reformatting that is done for the single-line sensitive formatting that is done in a number of cases.

> Alternatively, we could use OCaml's `Pprintast` to normalize the upstream AST. We do not need to match on the parsetree, so maintainance is a lot easier.
> However, we'll always find diffs so we'll need to run this on the tests only and to promote the diff to the source tree. That way, we'll always know what are the diffs OCamlformat is allowed to make and it will be a test failure only if these diffs change.
> This test can only run on a specific version of OCaml. It should be possible to run it on external repositories (like we do with `test_branch`) and keep the diffs here.
> 
> That's less useful than what you suggested but it's easier to maintain and has no overhead for users. I think that's fine for now, as long as changes to the parser are small. (which are only going to increase)

In our use case, I really want to perform this check on absolutely every execution of ocamlformat. So far, since the very first version, there has not been a production bug due to a formatting change, and I want to keep that streak going. :-)

In terms of end user usage, I think it makes sense to check every run as well. The time when someone is thinking most carefully about some code is right before they save and run the formatter. If the formatting introduces a bug at that point, there is less and less close attention paid in the steps going forward. So it is easiest and most valuable to do the check every time.

hhugo:
What about we track and review the diffs between the upstream parser and the one in ocamlformat ? Ideally, the diff should remain small and would give us enough confidence about correctness. Some changes, such as the one from this very PR should make there way upstream at some point.

gpetiot:
I proposed this change upstream https://github.com/ocaml/ocaml/pull/10459 and it doesn't seem like it will make it in upstream ocaml

jberdine:
Yes, my impression of that PR is that it is not likely to be welcomed upstream. I can understand that there are pros and cons and that for the compiler itself there are mostly cons. This is the classic tension between *abstract* syntax trees and *concrete* syntax trees: the compiler is mostly just complicated by concrete syntax, while ocamlformat (or tools such as refactoring tools, etc.) benefit from moving a bit toward a concrete syntax tree.

Tracking the diff between the parsers would be helpful, but not I think a replacement for a translation validation check. There is a lot to keep in mind and many opportunities to leave something our or be surprised when mentally computing the difference between parse vs nonstandard-parse; normalize-format-print; parse. The differences that can lead to translation validation failures can be very minor-seeming. For example remember how we got tripped by changing e.g. `(f a b) c d` to `f a b c d`?

lib/Ast.ml
gpetiot: I fixed this case and made sure we don't break the parenze_exp function

Julow (approved):
The code looks good to me and I think the benefits of tweaking the parser outweigh the maintenance cost of backporting new syntaxes.
I agree with jberdine on checking correctness on every runs. We need to use OCaml's parser during the validation step.

gpetiot:
> I agree with jberdine on checking correctness on every runs. We need to use OCaml's parser during the validation step.

I will try to work on that before merging this one

gpetiot:
I will split this PR before merging, the changes that involve using std_ast/extended_ast in the rest of the code can be made in a separate PR

PR #1636, opened 2021-04-15
Consistently format let bindings and monadic let bindings, do not drop comments before monadic bindings

gpetiot:
Fix #1623, no diff with test_branch.sh

- factorize formatting of `Pexp_let` and `Pexp_letop` with a new function `fmt_let_bindings`
- define a new sugared type `let_binding` in `Sugar`, with conversion functions; use it for `Pstr_value`, `Pcl_let`, `Pexp_let`, `Pexp_letop`
- side-effect: `indicate-multiline-delimiters` is now taken into consideration for letop bindings

PR #1695, opened 2021-06-28
Fix normalization of odoc paragraphs

### LINKED ISSUE #1686: Bug: "comments changed" | Cannot process a doc comment with a new line at the beginning
**Describe the bug**

ocamlformat 0.17.0 cannot process a doc comment with a new line at the beginning

**How to Reproduce**

`ocamlformat --impl -` and put this code:

```ocaml
(**

works?
*)
let a = 1
```

results in 

```
  BUG: comments changed.
```

`.ocamlformat`:

```
version=0.17.0
break-separators=before
dock-collection-brackets=false
break-sequences=true
doc-comments=before
field-space=loose
let-and=sparse
sequence-style=terminator
type-decl=sparse
wrap-comments=true
if-then-else=k-r
let-and=sparse
space-around-records
space-around-lists
space-around-arrays
cases-exp-indent=2
break-cases=all
parse-docstrings=true
```

```
profile=default
quiet=false
max-iters=10
comment-check=true
wrap-fun-args=true
wrap-comments=true (file .ocamlformat:10)
type-decl-indent=2
type-decl=sparse (file .ocamlformat:9)
stritem-extension-indent=0
space-around-variants=true
space-around-records=true (file .ocamlformat:13)
space-around-lists=true (file .ocamlformat:14)
space-around-arrays=true (file .ocamlformat:15)
single-case=compact
sequence-style=terminator (file .ocamlformat:8)
parse-docstrings=true (file .ocamlformat:19)
parens-tuple=always
parens-ite=false
ocp-indent-compat=false
nested-match=wrap
module-item-spacing=sparse
max-indent=68
match-indent-nested=never
match-indent=0
margin=80
let-module=compact
let-binding-spacing=compact
let-binding-indent=2
let-and=sparse (file .ocamlformat:12)
infix-precedence=indent
indicate-nested-or-patterns=unsafe-no (file .ocamlformat:18)
indent-after-in=0
if-then-else=k-r (file .ocamlformat:11)
function-indent-nested=never
function-indent=2
field-space=loose (file .ocamlformat:6)
extension-indent=2
exp-grouping=parens
dock-collection-brackets=false (file .ocamlformat:3)
doc-comments-tag-only=default
doc-comments-padding=2
doc-comments=before (file .ocamlformat:5)
disable=false
cases-exp-indent=2 (file .ocamlformat:16)
break-struct=force
break-string-literals=auto
break-sequences=true (file .ocamlformat:4)
break-separators=before (file .ocamlformat:2)
break-infix-before-func=false
break-infix=wrap
break-fun-sig=wrap
break-fun-decl=wrap
break-cases=all (file .ocamlformat:17)
assignment-operator=end-line
align-variants-decl=false
align-constructors-decl=false
align-cases=false
```

gpetiot:
I added a comment to keep trace of it, paragraphs must not be considered for the normalized form. They are only structural and the shape of a docstring organized in paragraphs only depend on linebreaks, which may be changed by the formatting.

I also changed the normalized output of other nodes, using `Node(X,Y)` instead of `Node,X,Y` as it is more convenient for debugging. No diff with test_branch.sh

Fix #1686

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,9 @@
 
   + Improve breaking of comments to avoid violating the margin (#1676, @jberdine)
 
+  + Fix normalization of odoc paragraphs (#1695, @gpetiot)
+    Paragraphs must not be considered for the normalized form. They are only structural and the shape of a docstring organized in paragraphs only depend on linebreaks, which may be changed by the formatting.
+
 #### Changes
 
   + Improve the diff of unstable docstrings displayed in error messages (#1654, @gpetiot)
--- a/lib/Normalize.ml
+++ b/lib/Normalize.ml
@@ -79,23 +79,28 @@ let rec odoc_inline_element fmt = function
       let txt =
         String.filter txt ~f:(function '\\' -> false | _ -> true)
       in
-      fpf fmt "Word,%a" str txt
-  | `Code_span txt -> fpf fmt "Code_span,%a" str txt
-  | `Raw_markup (Some lang, txt) -> fpf fmt "Raw_html:%s,%a" lang str txt
-  | `Raw_markup (None, txt) -> fpf fmt "Raw_html,%a" str txt
+      fpf fmt "Word(%a)" str txt
+  | `Code_span txt -> fpf fmt "Code_span(%a)" str txt
+  | `Raw_markup (Some lang, txt) -> fpf fmt "Raw_markup(%s,%a)" lang str txt
+  | `Raw_markup (None, txt) -> fpf fmt "Raw_markup(%a)" str txt
   | `Styled (style, elems) ->
-      fpf fmt "Styled,%a,%a" odoc_style style odoc_inline_elements elems
+      fpf fmt "Styled(%a,%a)" odoc_style style odoc_inline_elements elems
   | `Reference (_kind, ref, content) ->
-      fpf fmt "Reference,%a,%a" odoc_reference ref odoc_inline_elements
+      fpf fmt "Reference(%a,%a)" odoc_reference ref odoc_inline_elements
         content
   | `Link (txt, content) ->
-      fpf fmt "Link,%a,%a" str txt odoc_inline_elements content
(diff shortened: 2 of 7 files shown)


hhugo:
I don't understand this change. I don't understand what structural mean in this context. Surely, we don't want ocamlformat to alter paragraph inside doc comment. The only test added here is about a blank leading blank line. What about other cases such that trailing newline or blank lines in the middle ?

hhugo:
I don't understand this change. Surely, we don't want ocamlformat to alter paragraph inside doc comment. The only test added here is about a blank leading blank line. What about other cases such that trailing newline or blank lines in the middle of a comment ?

gpetiot:
My assumption was that we didn't want to preserve these, but if that's an issue I will try to improve that before the release.

hhugo:
We probably don't care about preserving leading blank lines. But preserving paragraph seems essential. Take a look at the document of any library from @dbuenzli.

hhugo:
We probably don't care about preserving leading blank lines. But preserving paragraph seems essential. Take a look at the document of any library from @dbuenzli.

gpetiot:
I'm adding more tests in #1706 and it looks like paragraphs are preserved, only the leading and trailing blank lines are removed.

hhugo:
It's good that they are preserved, but are we still checking that ? What I understand from your change is that it's removing the check for paragraphs ..

hhugo:
Any update regarding my last comment ?

gpetiot:
Yes this change removed the check, I didn't have time to work on a different change that would keep this check but also fix #1636 since I'm trying to wrap up the 0.19.0 release. What I can do is revert this PR's change and work on a fix for 0.20.0.

hhugo:
I would imagine that normalizing parsed doc comment to remove empty paragraph would be enough

gpetiot:
I tried that but no that's not enough unfortunately.

PR #1697, opened 2021-06-30
Type identifiers can have a disambiguation suffix

gpetiot:
This change is made for https://github.com/ocaml/ocaml-lsp/pull/386

No diff with test_branch.sh

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,8 @@
     - Break the line and reindent the cursor when pressing <ENTER>
   (#1639, #1685, @gpetiot) (#1687, @bcc32)
 
+  + Type identifiers can have a disambiguation suffix (e.g. `t/2`), as produced by the compiler (#1697, @trefis)
+
 ### 0.18.0 (2021-03-30)
 
 #### Bug fixes
--- a/test/passing/tests/sig_value.mli
+++ b/test/passing/tests/sig_value.mli
@@ -7,3 +7,5 @@ val f : f:((string * string)[@att]) (** doc *) -> unit
 val f : f:((string * string)[@att]) (** doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc doc *) -> unit
 val f : f:((string * string)[@att]) -> unit
 val f : f:((string * string)) (** doc *) -> unit
+
+val f : t/1 -> f:(unit -> unit t/2) -> unit M/2.t/2
--- a/test/passing/tests/sig_value.mli.ref
+++ b/test/passing/tests/sig_value.mli.ref
@@ -21,3 +21,5 @@ val f :
 val f : f:(string * string[@att]) -> unit
 
 val f : f:string * string (** doc *) -> unit
+
+val f : t/1 -> f:(unit -> unit t/2) -> unit M/2.t/2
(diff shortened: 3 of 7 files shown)


hhugo:
How is that not breaking parsing ?
Can you add tests for things like `a/2` or even `a/2n`

trefis:
To expand a bit on @hhugo's message: the lexer changes are going to have an impact in other contexts, not just for identifiers in type expressions.
So it's likely to break the parsing of arithmetic expressions.

As I see it, you want to handle `t/n` only when formatting type expressions coming from merlin. It would make sense to have a separate entry point, and a dedicated lexer (or entry point in the lexer), for those.
The parser changes seem fine as they are though.

gpetiot:
I'm having a look at that, but the change is very involved.

hhugo:
Not ideal but what about changing the parser to accept `LIDENT SLASH INT` in the right context instead ?

voodoos:
> Not ideal but what about changing the parser to accept `LIDENT SLASH INT` in the right context instead ?

Yes, to only allow type idents to have slashes and not value idents. This would prevent the confusion with the arithmetic `/ `.

trefis:
There is no `SLASH` token, you'd have to allow `LIDENT INFIXOP3 INT` in that context which is a bit more permissive than intended, but that doesn't really matter too much.
Not sure what the conflicts would look like though.

trefis:
Looks like there are no conflicts:
```
--- a/parsing/parser.mly
+++ b/parsing/parser.mly
@@ -3600,13 +3600,25 @@ label_longident:
     mk_longident(mod_longident, LIDENT) { $1 }
 ;
 type_longident:
-    mk_longident(mod_ext_longident, LIDENT)  { $1 }
+    mk_longident(mod_ext_longident, disambiguated_lident)  { $1 }
+;
+disambiguated_lident:
+    LIDENT { $1 }
+  | LIDENT op=INFIXOP3 INT {
+    assert (op = "/");
+    $1 ^ op ^ fst $3 }
+;
+disambiguated_uident:
+    UIDENT { $1 }
+  | UIDENT op=INFIXOP3 INT {
+    assert (op = "/");
+    $1 ^ op ^ fst $3 }
 ;
 mod_longident:
     mk_longident(mod_longident, UIDENT)  { $1 }
 ;
 mod_ext_longident:
-    mk_longident(mod_ext_longident, UIDENT) { $1 }
+    mk_longident(mod_ext_longident, disambiguated_uident) { $1 }
   | mod_ext_longident LPAREN mod_ext_longident RPAREN
       { lapply ~loc:$sloc $1 $3 }
   | mod_ext_longident LPAREN error
```
```
# let x : t / 3 -> int = fun _ -> ();;
Error: Unbound type constructor t/3
# let x : M/1 . t / 3 -> int = fun _ -> ();;
Error: Unbound module M/1
# let x : t lor 3 -> int = fun _ -> ();;
Fatal error: exception File "parsing/parser.mly", line 3610, characters 4-10: Assertion failed
```
(I guess you'll want to raise `Syntaxerr` instead of asserting)

gpetiot:
I tried this approach as well but there is an issue with `M/2.t/2` (this can also be produced by the disambigation right?) I think the dot is considered part of the INT and so it doesn't find the dot to be part of the module identifier.

nojb:
Hello, I **think** this change breaks parsing of expressions of the form `(???)/2`

image

gpetiot:
This change has been reverted to give us more time to work out a better solution. Sorry for the inconvenience!

PR #1698, opened 2021-06-30
Handle merlin typed holes

gpetiot:
I experience something weird with test_branch.sh where it's failing on master, gonna clear up everything and start again later


ulugbekna:
Wow, that was quick, thanks! It works like a charm for me so far :-)

gpetiot:
Any objection to this being merged @emillon @Julow @hhugo @jberdine  ?

Julow (commented):
Sorry if this has already been considered. Merlin already has a parser that supports this feature and is rebased regularly. What about vendoring merlin's parser ? (without taking advantage of the recovery for now)

gpetiot:
> Sorry if this has already been considered. Merlin already has a parser that supports this feature and is rebased regularly. What about vendoring merlin's parser ? (without taking advantage of the recovery for now)

We will make more changes and diverge further from the upstream ocaml parser ([ ] lists soon, I also want to try a few things for the comments, same for literal strings, add a location to constants, etc.) anyway

Julow (commented):
Could it be useful to have a "hole" case in longident too ?
For example:
- `module_type` (`module type T = _`, or `module M : _` in mli)
- `Pexp_field` (`x._`), `Pexp_new` (`new _`), module pack (`(module _)`)

It would also supersede `Pmod_hole` and `Pexp_hole` and add the hole syntax everywhere it's possible.

This suggestion doesn't make sense for the merlin use case, which only need the current patch. It would be useful for quickly fixing the syntax in order to format and makes the patch easier to rebase (as you said, we are diverging from merlin's parser too).

gpetiot:
I tried doing that, the issue is that the underscore is already a valid rule for patterns, types and modules names, so it creates conflicts in the parser, the modification to solve all of this would be more involved

Julow:
The fix is to remove the "any" pattern and type too. My claim that it'll make the patch simpler to rebase might be wrong.

Julow (approved):
LGTM

PR #1699, opened 2021-07-02
Use a separate lexer for merlin-specific inputs (type disambiguations)

gpetiot:
#1697 follow-up, fixes the regression on arithmetic use of `/`. Having a separate lexer is the only solution that worked, as I didn't find any way of disabling some rules in the lexer (and having conditional branching made some characters or lexems disappear).

cc @hhugo  @trefis @voodoos

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,7 +45,7 @@
     - Break the line and reindent the cursor when pressing <ENTER>
   (#1639, #1685, @gpetiot) (#1687, @bcc32)
 
-  + Type identifiers can have a disambiguation suffix (e.g. `t/2`), as produced by the compiler (#1697, @trefis)
+  + Type identifiers can have a disambiguation suffix (e.g. `t/2`), as produced by the compiler (#1697, #1699, @trefis, @gpetiot)
 
 ### 0.18.0 (2021-03-30)
 
--- a/lib/Ast_passes.ml
+++ b/lib/Ast_passes.ml
@@ -57,11 +57,11 @@ module Ast0 = struct
   module Parse = struct
     let implementation = Parse.implementation
 
-    let interface = Parse.interface
+    let interface = Parse_merlin.interface
 
     let use_file = Parse.use_file
 
-    let core_type = Parse.core_type
+    let core_type = Parse_merlin.core_type
 
     let module_type (lx : Lexing.lexbuf) =
       let pre = "module X : " in
--- a/lib/Parse_with_comments.ml
+++ b/lib/Parse_with_comments.ml
@@ -29,9 +29,21 @@ end
 
 exception Warning50 of (Location.t * Warnings.t) list
 
-let tokens lexbuf =
+let token_with_comments (type a) : a Ast_passes.Ast0.t -> _ -> _ = function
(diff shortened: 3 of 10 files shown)


voodoos:
Small comment: The `Lexer_merlin` does not reflect the specificity of the lexer. Maybe `Lexer_types` or `Lexer_nonambigous_types` whould be better names ?

In the future OCamlformat might format other outputs from Merlin, and not only types (I am thinking of Destruct and Construct).

hhugo:
> and having conditional branching made some characters or lexems disappear

If I understand what you mean by conditional branching , you'd need to adjust locations and fix the production to not use plain`Lexing.lexeme` as it only captures chars matching the rule on the LHS.

I'd be better to not have to maintain two nearly identical lexers.

Also note that structure item can in theory appear inside signatures using `modue type of struct let x = t/2 end`

Yet another approach would be to use the incremental API of menhir to lex+feed non ambiguous type ident when the parser chock on a slash after an ident. I've used the approach in js_of_ocaml (parsing JavaScript) to disambiguate between division `/` and regexp `/regexp/`

gpetiot:
What I mean is I tried something like:
``` diff
-rule token = parse
+rule token_common = parse
...
+  | "val"
+      { merlin_mode := true; VAL }
...

+and token = parse
+  | _
+    { (if !merlin_mode then token_merlin_mode else token_common)
+      (extend lexbuf) }
+
+and token_merlin_mode = parse
+  | lowercase identchar * type_disambig as name
+      { try Hashtbl.find keyword_table name
+        with Not_found -> TLIDENT name }
+  | lowercase_latin1 identchar_latin1 * type_disambig as name
+      { warn_latin1 lexbuf; TLIDENT name }
+  | uppercase identchar * type_disambig as name
+      { TUIDENT name } (* No capitalized keywords *)
+  | uppercase_latin1 identchar_latin1 * type_disambig as name
+      { warn_latin1 lexbuf; TUIDENT name }
+  | _
+      { token_common (extend lexbuf) }
+
```

here I defined `extend` as a function that extends the buffer for one char, to compensate for the one being read but it doesn't work as expected. Previously I tried modifying the fields `lex_curr_pos` and so on but always ended up in a bad state, I wish the lexer provided a simple backtracking function.

I'm having a look at what is done in js_of_ocaml but the structure of parse.ml is a bit different to what we have (copy of 4.12), I will see if I can make something out of it.

gpetiot:
> Yet another approach would be to use the incremental API of menhir to lex+feed non ambiguous type ident when the parser chock on a slash after an ident. I've used the approach in js_of_ocaml (parsing JavaScript) to disambiguate between division `/` and regexp `/regexp/`

@hhugo I am trying this approach in #1701 but I didn't get the locations right so far, I will git it another shot later.

hhugo:
@gpetiot, I don't think #1701 implement what I had in mind.
What I meant was:
- feed tokens to the parser
- if the parser goes into error when fed with `INFIXOP3 "/` and the previous token is an ident  (and tokens are adjacent ?)
  - backtrack the parser AND the lexer to theirs state before the ident and continue by lexing and feeding a disambiguated ident

Note: I'm note saying it is the way to go for sure, I'm just proposing alternatives

PR #1703, opened 2021-07-03
Add 'line-endings' option to specify line endings used in the formatted output

### LINKED ISSUE #49: Line endings changed to \r\n on Windows.

### LINKED ISSUE #1666: Bug: ocamlformat changes the end-of-lines style on Windows
**Describe the bug**
On Windows, if I use files using Unix end-of-lines (a single LF character) and run ocamlformat on them, ocamlformat will change the end-of-line to Windows style (CR LF characters), introducing noise in the diff. I'd like ocamlformat to respect the original EOL style, or provide an option to configure the end of line characters.

**How to Reproduce**

This assumes an unix shell in a Cygwin environment on Windows.

```shell
$ dune init executable shaihulud && git init .
$ dos2unix dune shaihulud.ml # convert to Unix style
$ chmod -x dune shaihulud.ml
$ sed -i 's/^/    /g' shaihulud.ml # weird indent so that ocamlformat may process the file
$ dune build @fmt --auto-promote
diff --git a/shaihulud.ml b/.formatted/shaihulud.ml
index 513361f..9cc7de4 100755
--- a/shaihulud.ml
+++ b/.formatted/shaihulud.ml
@@ -1 +1 @@
-             let () = print_endline "Hello, World!"
m+let () = print_endline "Hello, World!"
Promoting _build/default/.formatted/shaihulud.ml to shaihulud.ml.
$ hexdump -C shaihulud.ml
00000000  6c 65 74 20 28 29 20 3d  20 70 72 69 6e 74 5f 65  |let () = print_e|
00000010  6e 64 6c 69 6e 65 20 22  48 65 6c 6c 6f 2c 20 57  |ndline "Hello, W|
00000020  6f 72 6c 64 21 22 0d 0a                           |orld!"..|
00000028
```

You can see that ocamlformat has converted the file to CR LF end-of-lines.

nojb:
This PR is a simpler version of #274, which still solves the common use-cases. It adds a new configuration option `line-endings` with two possible values:

- `native`: use the host OS line endings (= channels are opened in text mode). The formatted output will use LF on Unix and CRLF on Windows.
- `unix`: the formatted output will use LF regardless of host OS (= channels are opened in binary mode).

Unix-centric projects (such as this one), should use `line-endings = unix` so that if a developer checks out the project on Windows and does `dune build @fmt` the line-endings are not changed in the process.

Fixes #49 
Fixes #1666


emillon:
Hi!

Thanks for working on this. This problem happens in several platform tools (mdx, dune, really anything that uses promotion) and I've been thinking about the right way to solve this in a uniform way.

I think that being able to emitting both kinds of newlines is part of the right solution, but I don't think that a new configuration option is the right way to approach this.

I strongly believe that the right approach is to always preserve what was found in the original source file. That way, reformatting never changes the kind of newlines used by a file. The source of newline types is what is in the original file, not the OS type or a new configuration option.

The main benefits are that it does not add a new configuration option, and that it does not create a situation (`native`) where the last one to edit rewrites the whole file.

Note that this means that a Unix ocamlformat needs to be able to write windows-style line endings. So it's not possible to use the "text mode" translation layer - it has to happen in-library.

The flow would look like:
- detect input file line ending type (either using a hook in the lexer, or by running a lightweight line ending lexer)
- store that kind with the parsed AST (maybe in `Parse_with_comments.t`?)
- continue internally operating on `\n` only.
- when writing the file (at the end of `with_buffer_formatter`), if CRLF is required, do the translation here.

I can help you with the implementation if you'd like. This is something that will greatly help ocamlformat. Thanks!

hhugo:
I'd vote for an option. A project should be able to enforce a specific style. Detecting line ending type can't be done in a few edge cases (Single lines or mixed encoding)

emillon:
If there's no choice but to add an option, let it be at least `auto | lf | crlf` with the default being `auto` or `lf` (this is what prettier does) - I don't think that `native` is ever the right choice.

But I hope we can make it work with no option! (single lines with no newline can be preserved too, and mixed encodings can never be properly supported, option or not)

nojb:
> this is what prettier does

Sorry, what is "prettier"? I random google search didn't turn anything up...

emillon:
[It's a popular code formatter that supports various languages](https://prettier.io/). Given that it's very succesful I think it's always a good idea to check what approach it takes to for this kind of language-agnostic questions. The bits about end of line is [here](https://prettier.io/docs/en/options.html#end-of-line). The general [page about option philosophy](https://prettier.io/docs/en/option-philosophy.html) is very interesting I think.

hhugo:
Of course we don't want to preserve mixed encoding!
But mixed encoding is what you get when editing files on different platform. You don't want the behavior to change depending on weather you add a new line at the very top of the file for example

nojb:
I pushed a commit implementing `line-endings = lf | crlf`. If there are no objections, I will add the auto-detection option as well.

emillon:
Yes, it's a good point. I assumed it's rare to create mixed encoding files but we don't want to converge to CRLF every time a file is edited on a broken editor on windows.

To sum up the discussion, I believe that we all agree on:

- `lf` and `crlf` modes are valuable
- `native` (or text mode in general) is not useful here

And an auto mode could be better than a default `lf` but we're not sure it can be implemented in the right way ([prettier switched from auto by default to lf by default because of this](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-endofline-to-lf-7435httpsgithubcomprettierprettierpull7435-by-kachkaevhttpsgithubcomkachkaev))

nojb:
> And an auto mode could be better than a default `lf` but we're not sure it can be implemented in the right way ([prettier switched from auto by default to lf by default because of this](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-endofline-to-lf-7435httpsgithubcomprettierprettierpull7435-by-kachkaevhttpsgithubcomkachkaev))

Thinking more about this, is `auto` really useful? What would be its use-case? The only one I can imagine is having files with different line endings in the same project which you want to preserve, but I have never seen this in practice...

hhugo:
I'm not sure changing out_ newline is enough here. Can you add tests with newlines inside `{||}` string literals or verbatim blocks in comments ?

nojb:
> I'm not sure changing out_ newline is enough here. Can you add tests with newlines inside `{||}` string literals or verbatim blocks in comments ?

I'll do this, but what is the specification? Newlines should be converted in both cases to the `line-endings` setting?

nojb:
> I'm not sure changing out_ newline is enough here. Can you add tests with newlines inside `{||}` string literals or verbatim blocks in comments ?

I switched to a simpler implementation (by postprocessing the final output). I added some tests that show that both in quoted strings and verbatim odoc blocks the newlines are normalized to the `line-endings` setting.

nojb:
Any opinions on actually including `auto` as an option? My own opinion is that it is not very useful. Otherwise, what do you think of the current implementation?

hhugo:
> > I'm not sure changing out_ newline is enough here. Can you add tests with newlines inside `{||}` string literals or verbatim blocks in comments ?
> 
> I switched to a simpler implementation (by postprocessing the final output). I added some tests that show that both in quoted strings and verbatim odoc blocks the newlines are normalized to the `line-endings` setting.

The feature can probably alter the exact content of string literals.
I think this is fine, in witch case we should probably document the behavior and make sure ocamlformat doesnt spot that as an AST change.

nojb:
> > > I'm not sure changing out_ newline is enough here. Can you add tests with newlines inside `{||}` string literals or verbatim blocks in comments ?
> > 
> > 
> > I switched to a simpler implementation (by postprocessing the final output). I added some tests that show that both in quoted strings and verbatim odoc blocks the newlines are normalized to the `line-endings` setting.
> 
> The feature can probably alter the exact content of string literals.
> I think this is fine, in witch case we should probably document the behavior and make sure ocamlformat doesnt spot that as an AST change.

Yes, indeed, we alter the exact content of literal strings. But like you, I think this is generally what is expected. Because the change happens after the final AST is produced, this change is not "seen" by the AST checker.

gpetiot (approved):
I rebased and checked that nothing broke with test_branch, looks good to me, thanks!

hhugo:
@jberdine, since you spent some time thinking about this topic, do you want to review this PR

nojb:
Friendly ping.

gpetiot:
Since it's been reviewed enough I'm fine with adding it to the release, just rebasing and we're good to go!

jberdine:
Sorry I couldn't get to this earlier. I haven't tested, but the only point I would be particularly concerned about is the treatment of literal string constants as denoted by `{|` `|}`. For those, it would seem to be incorrect to change any embedded newlines to some other format. This might result in files that appear to have mixed encoding, but I don't think that changing the literal strings is semantically ok.

nojb:
> Sorry I couldn't get to this earlier. I haven't tested, but the only point I would be particularly concerned about is the treatment of literal string constants as denoted by `{|` `|}`. For those, it would seem to be incorrect to change any embedded newlines to some other format. This might result in files that appear to have mixed encoding, but I don't think that changing the literal strings is semantically ok.

We discussed this briefly above https://github.com/ocaml-ppx/ocamlformat/pull/1703#issuecomment-875675086, and there was a weak agreement that the change was OK. My reasoning was as follows.

Typically quoted strings `{|...|}` are used to avoid having to escape characters which are typically quoted: double quotes, backslashes, newlines, etc. With the PR as it is it means that the embedded newlines will be normalized to the `line-endings` setting.

- If the quoted string is destined to a text channel then the change will not be visible.
- Leaving the quoted string with its original newlines is also not always "correct" since the newlines may be "accidental" (eg you had an LF file and then edited your file in a Windows environment and changed all your newlines to CRLF and then `ocamlformat` reverted the newlines to LF, except the ones inside the quoted string)
- In general, `ocamlformat` doesn't have enough information to know whether preserving or changing the embedded newlines is the "right" option; and preserving them unconditionally is unlikely to be the right choice in some cases (eg the scenario above) *and* may leave files with a mixed encoding.

On balance, I went for simplicity, and if you find yourself in the "bad" case of absolutely needing to embed a CRLF in a LF file or vice-versa, then you can always work around it by using explicit escapes. I agree that this edge-case is not ideal, but I have never seen it in the wild so I assumed it was unusual (but would be interested to learn otherwise).

Happy to hear further opinions on this point!

jberdine:
I completely agree that the actual user intent can't be guessed, but that at least sometimes people want to also change the newlines in literal strings. My worry is more that it seems that the property that processing source code with ocamlformat and compiling should not be able to change the behavior of the resulting program. (This is not quite true due to programs having access to source code location strings, but code whose behavior crucially depends on those is already playing with fire.) Changing the content of literal strings is, I think, the only way that the extensional runtime semantics of code can be changed by ocamlformatting its source. That makes me rather uncomfortable. For literal strings in comments, no worries. But I have encountered code before that used literal strings to produce the content of generated files, where the code that wrote the files had to be careful to emit the expected newlines on ms vs unix systems. That code would have broken by changing the encoding of newlines of the literal strings. I agree that such code is very fragile, so it is perhaps more robust to specify the line endings in the ocamlformat config and make sure that the implementation of the code is in agreement.

How would people feel about an intermediate point where ocamlformat would issue a warning when newlines in a literal string are changed?

nojb:
> That code would have broken by changing the encoding of newlines of the literal strings. I agree that such code is very fragile, so it is perhaps more robust to specify the line endings in the ocamlformat config and make sure that the implementation of the code is in agreement.

I agree that this kind of code is overly fragile.

> How would people feel about an intermediate point where ocamlformat would issue a warning when newlines in a literal string are changed?

It sounds sensible. The warning can be implemented independently of the EOL normalization: if a literal string contains LF and the `line-endings` setting is CRLF (or vice-versa), then the warning should trigger.

nojb:
> Changing the content of literal strings is, I think, the only way that the extensional runtime semantics of code can be changed by ocamlformatting its source. That makes me rather uncomfortable.

I just noticed that the input channels are opened (since before this PR) in text mode which means that the distinction between CRLF and LF has never been preserved.

https://github.com/ocaml-ppx/ocamlformat/blob/1c4960bad6dd1cf3d931e41b6b07c0be24800c6f/bin/ocamlformat/main.ml#L32-L34

jberdine:
Oof, so broken from the beginning then. We have really not had much success with testing on windows.

I should probably be quiet and let those who interact with windows choose what they find the best path to be. But just for the record: I don't think ocamlformat should change the bytes of string literals without warning, and I think that the possibility of changing them should be documented.

hhugo:
@nojb , in #1748  @bcc32 reportted a use case broken by this PR. I think this is only an issue with quoted strings. We could recover positions of such token (using the lexer) and not apply normalization in their range. What do you think ?

nojb:
> @nojb , in #1748 @bcc32 reportted a use case broken by this PR. I think this is only an issue with quoted strings. We could recover positions of such token (using the lexer) and not apply normalization in their range. What do you think ?

Thanks for the ping! To be honest I *still* think that applying normalization to quoted strings is the right thing to do **in general**, given that we do not know the intent of the code in question. Also excluding quoted strings will complicate the implementation and make it harder to explain and reason about the semantics of the option.

@jberdine suggested outputting a warning when rewriting string literals and that seems a good idea to me. Do you think the warning would be enough? I will add a comment to the linked PR with the workaround of escaping the string literal in question.

PR #1706, opened 2021-07-07
Add odoc paragraph tests

gpetiot:
Adding some tests following the discussion of #1695


hhugo:
is the `parse-docstring` option set for theses tests ?

gpetiot:
Oh I just realized it's not indeed, neither for doc_comments.ml nor doc_comments.mli tests, it should have been. So this part of the code was not tested all along.
I will add some tests but I don't want to delay the next release too much.

gpetiot:
Replaced by #1712

PR #1709, opened 2021-07-08
Include LGPL 2.1 license of compiler vendored code

gpetiot:


ocamlformat-rpc-lib.opam
```diff
@@ -5,7 +5,8 @@ description:
   "OCamlFormat is a tool to automatically format OCaml code in a uniform style. This package defines a RPC interface to OCamlFormat"
 maintainer: ["OCamlFormat Team <ocamlformat-dev@lists.ocaml.org>"]
 authors: ["Josh Berdine <jjb@fb.com>"]
-license: "MIT"
```
emillon, line 8: I think the client library still only links to MIT code - you should be able to set `(license MIT)` in the `package` stanza to override the global one.

gpetiot: done

PR #1712, opened 2021-07-09
Do not consider leading star '*' when checking the diff of doc comments

hhugo:
Restore the check on docstring while still fixing #1686

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,12 +34,11 @@
 
   + Fix parentheses around successive unary operations (#1696, @gpetiot)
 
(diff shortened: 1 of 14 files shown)


hhugo:
@gpetiot, Do you want to take it from there ?

gpetiot (approved):
I've added some tests (replacing #1706) to further test the formatting of odoc paragraphs, especially when `parse-docstrings` is set, no diff with test_branch.sh as well.
It looks good to me, thanks!

hhugo:
The title and change entry are misleading. Formatting was already discarding the leading star. Only the check after formatting was wrong.

PR #1713, opened 2021-07-09
Format Odoc code block metadata fields

Julow:
Since `odoc-parser.0.9.0`, Odoc's code blocks have a metadata field:

```
(** {@metadata[
      ...
    ]} *)
```

This was added into Odoc after the first version of <https://github.com/ocaml-ppx/ocamlformat/pull/1683> and is not implemented (currently drops the metadata).
It is an arbitrary string and has no escaping syntax, so it is printed as-is.

An alternative formatting would be to add a space after it: `{@metadata [ ... ]}`.

This string might start with the name of the language used within the code block, OCamlformat could use this information to format only blocks containing ocaml code. This might be added to the AST in the future, I'll submit a PR when this is released.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,7 +48,7 @@
 
   + Improve position of `;;` tokens (#1688, @gpetiot)
 
-  + Depend on `odoc-parser` instead of `odoc` (#1683, @kit-ty-kate, @jonludlam)
+  + Depend on `odoc-parser` instead of `odoc` (#1683, #1713, @kit-ty-kate, @jonludlam, @julow)
     The parser from odoc has been split from the main odoc package and put into its own package, `odoc-parser`.
 
 #### New features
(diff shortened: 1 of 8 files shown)


gpetiot (approved):
It looks good to me, thanks!
I've added a few tests, maybe it would be better to add a space after the metadata field to break in case we reach the margin indeed, otherwise looks good.

Julow:
There's a typo in your test, the break is after the `[`:
```diff
-(** {@some_tag[ Code block with metadata field. This is a big block that
-    should hopefully break} ] *)
+(** {@some_tag[
+      Code block with metadata field. This is a big block that should hopefully break
+    ]} *)
```

If we add a space there, it shouldn't be breaking otherwise it would render like this:
```
(** {@some_tag
    [
      ...
    ]} *)
```

gpetiot:
Indeed, good catch!
Everything looks good to me then, feel free to merge if everything is ready!

Julow:
Good to merge then :)

PR #1714, opened 2021-07-09
Fix unstable broken-ref test

Julow:
In this test, OCamlformat calls git diff then tries to hide the temporary file names in the output. This didn't work on my system, where temporary files are created into `/run/user/...` instead of `/tmp`.


gpetiot (approved):
Looks good, thanks!

PR #1716, opened 2021-07-12
Allow disambiguated global identifiers (like t/2) so they can be formatted by tools like OCaml-LSP

let-def:
This is an alternative implementation that solves the same problem as #1715 

By the way, if I am right, I updated the specification to better match the output of the typed printer: `M/2.t/2` is not allowed, disambiguators are only allowed on the first identifier of a path.

The patch is quite close to #1715: try to parse one way, fall back to the other in case of failure.
But contrary to #1715, we approach the problem differently: first we try to parse using normal OCaml rules (lexer is untouched).
In case of failure, we check if the token could be mistaken for a disambiguation suffix, which we call a `TYPE_DISAMBIGUATOR`. (FIXME: find a better name, it disambiguates global identifiers, not just types).

The benefit of this approach are two folds:
- first, by construction, the original grammar is not affected: more sentences are recognized, but all valid OCaml phrases are accepted in exactly the same way (modulo the fact that we distinguish "/" from other INFIXOP3, it is not necessary but it is cleaner)
- second, by shifting `/` inside the identifier rules, we should get some input from Menhir if that introduced a grammar conflict (luckily, there is none with the current rules)

### DIFF
--- a/vendor/diff-parsers-ext-parsewyc.patch
+++ b/vendor/diff-parsers-ext-parsewyc.patch
@@ -1,6 +1,48 @@
 --- ocaml-4.13-extended/lexer.mll
 +++ parse-wyc/lib/lexer.mll
 @@@@
+ let is_keyword name = Hashtbl.mem keyword_table name
+ 
+ let check_label_name lexbuf name =
+   if is_keyword name then error lexbuf (Keyword_as_label name)
+ 
+-(* To "unlex" a few characters *)
+-let set_lexeme_length buf n = (
+-  let open Lexing in
+-  if n < 0 then
+-    invalid_arg "set_lexeme_length: offset should be positive";
+-  if n > buf.lex_curr_pos - buf.lex_start_pos then
+-    invalid_arg "set_lexeme_length: offset larger than lexeme";
+-  buf.lex_curr_pos <- buf.lex_start_pos + n;
+-  buf.lex_curr_p <- {buf.lex_start_p
+-                     with pos_cnum = buf.lex_abs_pos + buf.lex_curr_pos};
+-)
+-
+-let disambiguate lexbuf txt =
+-  let pos = ref 0 in
+-  let len = String.length txt in
+-  let is_digit c = c >= '0' && c <= '9' in
+-  while !pos < len && is_digit txt.[!pos] do incr pos done;
+-  let txt =
+-    if !pos < len then (
+-      set_lexeme_length lexbuf !pos;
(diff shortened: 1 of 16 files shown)


hhugo:
The downside of this approach is that we would accept `t / 2` instead of  just `t/2`. That said, I like this approach better.

let-def:
> The downside of this approach is that we would accept `t / 2` instead of just `t/2`. That said, I like this approach better.

If that is a problem, we can always check for lack of spaces between tokens.

gpetiot:
Sorry for the late answer, I also like this approach better. As Hugo said if it's feasible to not accept the spaces around the `/` that would be perfect.
Otherwise, the implementation looks good, thanks!

let-def:
No problem. Looking at the implementation again, I don't think it is worth making it more complex just for the spaces around `/`.
We already parse regular OCaml correctly; by luck this extension is not ambiguous, so it is not a problem to parse a little bit more than the printer can produce..

gpetiot:
Now that we distinguished the standard ocaml 4.13 parser and the extended one that ocamlformat uses, I've reported your changes on the extended one, I didn't find where the errors come from though

let-def:
The `rpc_test` failing, you mean? If yes, I can look at that.

hhugo:
This should be rebased after #1783 is merged.

vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -1,5 +1,47 @@
 --- ocaml-4.13-extended/lexer.mll
 +++ parse-wyc/lib/lexer.mll
+@@@@
+ let is_keyword name = Hashtbl.mem keyword_table name
+ 
+ let check_label_name lexbuf name =
+   if is_keyword name then error lexbuf (Keyword_as_label name)
+ 
+-(* To "unlex" a few characters *)
```
hhugo, line 9: Don't we need the same fix in parse-wyc ?

gpetiot: This is only necessary to print the disambiguated identifiers, that should be called by ocaml-lsp, which wouldn't set the `--numeric` option (only set to get the re-indent behavior in emacs) that is the only way to run parse-wyc for recovery parsing.

gpetiot (approved):
Looks good to me, waiting for more opinions before merging

Julow (approved):
Looks good to me :)

PR #1718, opened 2021-07-13
Convert test for option 'parse-docstrings=false'

gpetiot:
I just realized that the `parse-docstrings=true` option was already set in `test/passing/tests/.ocamlformat`:

```
profile = ocamlformat
break-cases = fit
margin = 77
parse-docstrings = true
wrap-comments = true
max-iters = 2
```

~and I don't think it would be worth it to reconvert this test to check `parse-docstrings=false` so we can just remove it!~
edit: wait finally maybe it's worth it, I'm gonna reconvert this test instead!

### DIFF
--- a/test/passing/dune.inc
+++ b/test/passing/dune.inc
@@ -759,13 +759,13 @@
  (deps tests/.ocamlformat )
  (package ocamlformat)
  (action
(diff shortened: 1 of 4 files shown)


gpetiot:
The parse-docstrings=true option was already set in test/passing/tests/.ocamlformat

PR #1720, opened 2021-07-13
Remove unused dependency to uutf

gpetiot:
`dune external-lib-deps --missing @fmt` does not return any error for me but somehow does in the CI, so probably an `ocaml-ci` error.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -66,9 +66,7 @@
   (stdio
    (< v0.15))
   (uuseg
-   (>= 10.0.0))
-  (uutf
-   (>= 1.0.1))))
+   (>= 10.0.0))))
 
 (package
  (name ocamlformat-rpc)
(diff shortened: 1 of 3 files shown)


Julow (changes requested):
We don't use `uutf` directly but we use `uuseg.string` which is only installed if `uutf` is available (it's a depopt): https://github.com/dbuenzli/uuseg/blob/master/pkg/pkg.ml#L20
We actually depend on "uuseg with uutf depopt".

gpetiot:
Alright I don't think dune can handle this one, closing this PR then

jberdine:
Does this point to a potential improvement in the packaging of uuseg? It seems that it is probably unintentional on the part of the uuseg packaging that we have to install uutf to use uuseg. @dbuenzli ?

PR #1722, opened 2021-07-14
README: use ocamlformat from current switch in Vim

silenc3r:
Different projects may use different versions of ocamlformat.


gpetiot (approved):
Looks good, thanks!

PR #1731, opened 2021-07-21
Fix normalization of sequences of expressions

### LINKED ISSUE #1728: BUG: ast changed
0.19.0 with default options
```ocaml
let read_traces filename =
  let ic = open_in_bin filename in
  begin
    read_hashtable ~t:[%t: contracts_trace] 0 40 ic tbl1;
    read_hashtable ~t:[%t: variables_trace] 40 70 ic tbl2;
    read_hashtable ~t:[%t: expressions_trace] 70 100 ic tbl3;
  end;
  close_in ic
```

### LINKED ISSUE #1730: BUG: ast changed
0.19.0 with default options
```ocaml
let compute_leverage_surface
  =
    fun () ->
        begin
          V.div case_2_s case_2_m ~dst:case_2_s;
          V.mul case_2_m case_2_dw ~dst:case_2_s;
        end;
        index_s2 := 0
```

gpetiot:
Fix #1728 
Fix #1730
No diff with test_branch.sh

The normalization is now consistent with how sequences are formatted, to preserve the parentheses/begin-end keywords we would need to find a way to have them in the AST, the current implementation is not able to distinguish between
```ocaml
begin
  a;
  b;
  c;
  d
end
```
and
```ocaml
begin
  a;
  b
end;
begin
  c;
  d
end
```
that's why the begin-end keywords are not preserved (I am not even sure we would want to preserve them)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 #### Bug fixes
 
+  + Fix normalization of sequences of expressions (#1731, @gpetiot)
+
 #### Changes
 
 #### New features
--- a/lib/Normalize.ml
+++ b/lib/Normalize.ml
@@ -219,10 +219,21 @@ let make_mapper conf ~ignore_doc_comments =
   in
   let expr (m : Ast_mapper.mapper) exp =
     let exp = {exp with pexp_loc_stack= []} in
-    match exp.pexp_desc with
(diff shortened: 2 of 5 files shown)


jberdine (approved):

nojb:
Thanks!

PR #1734, opened 2021-07-21
Add missing parenthesis around constrained pattern in class

### LINKED ISSUE #1729: BUG: ast changed
0.19.0 with default options
```ocaml
class virtual accumulate_flows =
  let (mc_exit : _) =
    match nb_trajectories_ratio with
    | _ when errs ->
        invalid_arg ""
  in
  object
  end
```

gpetiot:
Fix #1729, no diff with test_branch.sh

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
 
   + Fix normalization of sequences of expressions (#1731, @gpetiot)
 
-  + Add missing parenthesis around constrained pattern in class (#1734, @gpetiot)
+  + Type constrained patterns are now always parenthesized, parentheses were missing in a class context (#1734, @gpetiot)
 
 #### Changes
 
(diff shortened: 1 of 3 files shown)


Julow (approved):
Looks good.

Is there a `Ppat_constraint` that isn't parenthesed in other contexts ?

Julow:
> Is there a Ppat_constraint that isn't parenthesed in other contexts ?

Of course, there's the pattern of a `let` or a letop. It is already handled above, however, so this can be simplified into 
```ocaml
    | _, Ppat_constraint _ -> true
```
eg. add parentheses by default for `Ppat_constraint`. I think this pattern tends to have parentheses more often than not.

gpetiot:
I factorized these cases, there was no diff with test_branch.sh, and updated the changelog

Julow (approved):
LGTM

nojb:
Thanks!

Issue #1735, opened 2021-07-21
`[...]` vs `{[...]}`: missing odoc functionality?


jberdine:
I'm trying to handle the difference between `[...]` and `{[...]}` docstring elements more correctly, and it seems that perhaps we need the odoc parser to preserve more of the info that is in the source.

Ideally, the content of `[...]` should be formatted, while `{[...]}` is expected to be preformatted and so should be left untouched. I am not finding a distinction between these two cases in the parsed form, am I missing something?

Cc: @Julow @jonludlam

Julow:
`[ ... ]` is an "inline" element and is expected to contain a small amount of text (eg. an identifier or a function call), I'm not sure it's worth trying to format it.
`{[ ... ]}` is a "block" element and is expected to contain a multi-line piece of OCaml code and is currently formatted by OCamlformat. (falls back to the initial text if parsing fails)

Since recently, there's a way to attach metadata to code blocks:
```ocaml
(** {@ocaml[
    ...
    ]} *)
```

The first word is the language used within the block. We could disable formatting of code blocks with a language that is not "ocaml". (currently, not looking at the language tag)
Also, there's the `{v ... v}` block for preformatted text, that OCamlformat preserves.

Issue #1675, opened 2021-06-04
Feature request: Stronger notion of formatting stability


jberdine:
A desirable but not currently achieved form of stability is for `ocamlformat -m 80 -i file.ml` to be equivalent to `ocamlformat -m 999 -i file.ml; ocamlformat -m 80 -i file.ml` (or some number even larger than 999).

This is not currently achieved for a number of reasons, that seem to have varying degrees of seriousness / difficulty:

1. Single line detection gets tripped up and some phrases that are stable either as a single line or as multiple lines can change from the multiple-line form to the single line form when first formatting with a huge margin.
2. Comments, particularly those that end a line, can be attached differently, and cause formatting changes. Perhaps detecting if a comment ends a line and adjusting the formatting to ensure a break is added after it would resolve this.
3. Open lines around comments at the top of structures (and signatures?) are not always preserved. Perhaps this is another incarnation of the end-of-line comment behavior.
4. The indentation of some comments affects their attachment, and this is not preserved.
5. The detection of banner comments seems to be sensitive to the margin and they are not preserved by formatting with a huge margin.

Comment attachment is already complicated, but I wonder if there is something that can reasonably be done regarding the changes based on single-line detection. For example, with a margin of 76, the code
```ocaml
let marshal program file =
  Out_channel.with_file file ~f:(fun oc ->
      Marshal.to_channel oc program [] )
```
is stable and so is the single-line form:
```ocaml
let marshal program file =
  Out_channel.with_file file ~f:(fun oc -> Marshal.to_channel oc program [])
```

PR #1736, opened 2021-07-21
Add missing docstring normalization

jberdine:
Without this docstrings with UTF8 withing `[]` have miscalculated lengths.

This adds (some of?) the normalization needed for `[...]` docstring elements. Currently this change also incorrectly applies to preformatted code block elements `{[...]}`. I do not know the best way to distinguish these cases in odoc, see https://github.com/ocaml-ppx/ocamlformat/issues/1735.


lib/Fmt_odoc.ml
```diff
@@ -89,7 +89,8 @@ let fmt_code_block conf s1 s2 =
       let box = match lines with _ :: _ :: _ -> vbox 0 | _ -> hvbox 0 in
       box (wrap_code (vbox 0 (list_fl lines fmt_line)))
 
-let fmt_code_span s = hovbox 0 (wrap "[" "]" (str (escape_brackets s)))
+let fmt_code_span s =
+  hovbox 0 (wrap "[" "]" (str_normalized ~escape:escape_brackets s))
```
Julow, line 93: `str_normalized` inserts a break between each word. UTF-8 length calculation is also done by `str`.
Code span were lacking breaks on purpose because they are expected to contain a small amount of text but I'm fine with adding breaks.

Julow (commented):
> I do not know the best way to distinguish these cases in odoc

Multi-line code blocks are `` `Code_block `` in the AST and are handled in `fmt_nestable_block_element`, immediately calling `fmt_code_block`.
Inline code spans are `` `Code_span ``.

jberdine:
I don't know the best implementation strategy. The issues I am trying to resolve are, I think, similar in nature to the following example. Formatting:
```ocaml
  val trms : t -> trm iter
  (** [trms a] enumerates the maximal foreign or noninterpreted proper
      subterms of [a]. Considering an arithmetic term as a polynomial,
      [trms (c × (Σᵢ₌₁ⁿ cᵢ × Πⱼ₌₁ᵐᵢ Xᵢⱼ^pᵢⱼ))] is the sequence of monomials
      [Πⱼ₌₁ᵐᵢ Xᵢⱼ^pᵢⱼ] for each [i]. If the arithmetic term is a monomial,
      [trms (Πⱼ₌₁ᵐ Xⱼ^pⱼ)] is the sequence of factors [Xⱼ] for each [j]. *)
```
yields
```ocaml
  val trms : t -> trm iter
  (** [trms a] enumerates the indeterminate terms appearing in [a].
      Considering an arithmetic term as a polynomial,
      [trms (c × (Σᵢ₌₁ⁿ cᵢ × Πⱼ₌₁ᵐᵢ
      Xᵢⱼ^pᵢⱼ))] is the sequence of terms [Xᵢⱼ] for each [i] and
      [j]. *)
```
while with this PR this docstring is unchanged, as desired.

There seems to be something amiss where without this PR, the length of strings with UTF8 characters is not calculated correctly, where with this PR they are. @Julow, this behavior that I see doesn't match up with your explanation in the inline comment, so I'm confused. Do you see a different / better solution?

Julow:
Are you using 0.18.0 as a comparison ?

I get no diff between this PR and the current master with your example but I do see it compared with 0.18.0.
UTF8 length (grapheme clusters) is used on every strings (including keywords and identifiers) since https://github.com/ocaml-ppx/ocamlformat/pull/1673

jberdine:
Hmm, I am literally comparing this PR with master. I'll dig some more to see if I can understand what is happening.

jberdine:
Ugh, this is an instance of https://github.com/ocaml-ppx/ocamlformat/issues/1675. If you start with master and format
```
val trms : t -> trm iter
  (** [trms a] enumerates the indeterminate terms appearing in [a].
      Considering an arithmetic term as a polynomial,
      [trms (c × (Σᵢ₌₁ⁿ cᵢ × Πⱼ₌₁ᵐᵢ
      Xᵢⱼ^pᵢⱼ))] is the sequence of terms [Xᵢⱼ] for each [i] and
      [j]. *)
```
then it is unchanged. Switch to this PR and format and it produces:
```
  val trms : t -> trm iter
  (** [trms a] enumerates the maximal foreign or noninterpreted proper
      subterms of [a]. Considering an arithmetic term as a polynomial,
      [trms (c × (Σᵢ₌₁ⁿ cᵢ × Πⱼ₌₁ᵐᵢ Xᵢⱼ^pᵢⱼ))] is the sequence of monomials
      [Πⱼ₌₁ᵐᵢ Xᵢⱼ^pᵢⱼ] for each [i]. If the arithmetic term is a monomial,
      [trms (Πⱼ₌₁ᵐ Xⱼ^pⱼ)] is the sequence of factors [Xⱼ] for each [j]. *)
```

Julow:
The remaining question is: How should be change the formatting of code spans.

This PR allows every spaces in a code span to break. This makes it more stable (there's only one stable formatting) but it changes Odoc's AST.

```ocaml
(** [a
    b] *)
```

This is parsed as `"a\n    b"` by Odoc. (The HTML output is "a b" but not in other backends) 
Perhaps we could change Odoc to normalize whitespaces inside code spans ? For now we can't break code spans.

jberdine:
Hmm, this feels like another case of odoc just doing something different from what is in the ocaml manual, so I don't know. To my understanding, `{[...]}` are described as preformatted, and hence parsing their contents in a whitespace-preserving way makes sense, while `[...]` are not described as preformatted, and so their contents should be normalized. If the current parsing in odoc prevents this, then it seems like an odoc improvement is needed. Does that make sense to anyone other than me?

lib/Fmt_odoc.ml
jberdine: You're right about length calculation. The issue is missing breaks, which get added with this patch.

Julow:
I would support normalizing `[ ... ]` code blocks but this cannot be done yet. With the current Odoc, this means changing the AST and inserting unwanted newlines into the rendered output.
Also, we must be careful at how code blocks are used, this might break some documentations:
```diff
- (** [trim "  "] is [""] *)
+ (** [trim " "] is [""] *)
```

I believe that this PR doesn't fix the original issue (which was fixed in https://github.com/ocaml-ppx/ocamlformat/pull/1673).

Perhaps we should open an issue on Odoc ?

jberdine:
In the original report I was mistaken to bring up utf length computation, that turned out not to be the issue. The issue in the [comment](https://github.com/ocaml-ppx/ocamlformat/pull/1736#issuecomment-884488828) above remains though, and is fixed by this PR.

@Julow, are you saying that any change in ocamlformat that allows every space in a code span to break would cause problems with odoc? Are the odoc ast changes you mention something that are checked in ocamlformat analogous to how we check the main parsetree is preserved by formatting? I'm curious as we have been running with this PR in production for a while with no issues.

Julow:
> @Julow, are you saying that any change in ocamlformat that allows every space in a code span to break would cause problems with odoc? Are the odoc ast changes you mention something that are checked in ocamlformat analogous to how we check the main parsetree is preserved by formatting? I'm curious as we have been running with this PR in production for a while with no issues.

Any change in whitespace/newlines in code spans would change the rendered documentation. Every strings in doc comments are normalized: https://github.com/ocaml-ppx/ocamlformat/blob/main/lib/Normalize.ml#L79-L83 which is wrong at least for code spans with the current implementation of Odoc.

I believe it is reasonable to change Odoc to normalize line breaks in code spans (but not consecutive spaces).

gpetiot:
> I believe it is reasonable to change Odoc to normalize line breaks (and the following spaces) in code spans (but not consecutive spaces).

@Julow So is this something worth fixing in odoc? we will synchronize with odoc as necessary. Otherzise we should close this PR.

Julow:
I think it's worth clarifying Odoc's code span syntax independently from this PR.
This PR still makes sense in my opinion.

Julow:
Yes, I think this still makes sense. Let's discuss that on Odoc's side.

PR #1743, opened 2021-07-23
Improve spacing of module items

### LINKED ISSUE #1253: Feature request: Simpler spacing heuristics for modules

OCamlformat `conventional` spaces module definitions according to some heuristic of the 'simplicity' of the module being defined. In particular, it seems to decide that any functor application longer than a single line or with more than one argument deserves a space:

```ocaml
module A = Make_A
module B = Make_B (Foo)
module C = Make_C (Foo)

module D = Make_D (Foo) ()

module E = Make_E (Foo) ()

module F = Make_F

module G = Make_G (struct
  let x = 0000_0000_0000_0000_0000_0000_0000_0000_0000
end)

module H = Make_H
module I = Make_I
```

Personally, it seems better to only space module definitions that require more than a single line:

```ocaml
module A = Make_A
module B = Make_B (Foo)
module C = Make_C (Foo)
module D = Make_D (Foo) ()
module E = Make_E (Foo) ()
module F = Make_F

module G = Make_G (struct
  let x = 0000_0000_0000_0000_0000_0000_0000_0000_0000
end)

module H = Make_H
module I = Make_I
```

**Additional context**

This is similar to https://github.com/ocaml-ppx/ocamlformat/issues/1220 in that it advocates for more consistency & simplicity in spacing heuristics, so that the user is less often left wondering what the formatter is thinking. The situation for `open` statements (described in that issue) also suffers from strange line-spacing heuristics (but not the same ones):

```ocaml
module M : sig
  open Foo
  open Foo(A)
  open Foo(A)(B)

  open Foo.Aaaaaaaaa.Bbbbbbbbb.Ccccccccccc.Ddddddddddd.Eeeeeeeeee

  open Foo(A)
  open Foo
end = struct end
```

gpetiot:
Fix #1253


Julow (approved):
The implementation looks good to me :)

The diff might be big on some projects, can you try `test_branch` on more of them ?

jberdine:
A higher-level question that this PR raises is what is `module-item-spacing = sparse` meant to be? That is, `compact` and `preserve` are, I think, rather clear, but `sparse` is less so. This PR moves spares in the direction of compact. This seems fine to me, but I think it would help to state what objective there is for sparse that prevents just making it the same as compact. Prior to this PR, I think that the main point for sparse was that it did not rely on `Location.is_single_line`, but this PR changes that. So honest question: would changing `sparse` to be the same as `compact` be an option instead? Another question: would resolving the odd choices in the linked issue by just always adding an open line be an option? I would be fine with either of these. I ask these questions since the answers could help to identify what it is that `sparse` is aiming for.

gpetiot:
I was seriously considering only keeping one behavior, it looks like there are exceptions for both compact and sparse modes (in addition to some bugs/omissions).

To ease the maintenance of ocamlformat, only having `sparse` would be a blessing, and it would not have the weird diff of adding/removing blank lines when an item becomes shorter/longer/gets or loses comments. But janestreet uses `compact`, but I think most other big projects use `sparse`.

jberdine:
My suggestion was actually to only have `compact` by making `sparse` the same as compact. :-) To remove `compact` I think we would need `sparse` to become much more compact.

gpetiot:
Alright it works for me as well, let's start by deprecating this option first for the next release. The work done in this PR wlll be simpler once the option has been removed.

jberdine:
When deprecating it, would the plan be to also not set it in the profiles? That's going to lead to some big whitespace diffs. I'm not arguing for it, but the path of simplifying `sparse` basically to make `is_simple` always be `false` might be a smoother transition for existing projects.

gpetiot:
For now when we deprecate an option we don't change the behavior at all (only a warning is printed if the option is explicitly set), the change only happens in the version where the option is removed.

jberdine:
I think it makes more sense to change the defaults and profiles so that the behavior that is planned to be removed is only used if an option is explicitly set. That is, if `module-item-spacing` is deprecated now, and later the only behavior will be what now corresponds to `module-item-spacing = compact`, then the defaults and profiles should be set so that the `module-item-spacing = compact` behavior is used unless someone has `module-item-spacing = {sparse,preserve}` in their config file. Otherwise it is the same as removing the current (default even) behavior without notice.

PR #1745, opened 2021-07-26
Make the formatting mode of comments more explicit

gpetiot:
Some cleanup of the Cmt.ml file, that was a bit difficult to follow.
- now `fmt` is split in 2 parts: determine the formatting mode and calling the right function
- all other functions are moved out of `fmt`
- make the impossible cases more explicit
- group dependent functions in modules (Asterisk_prefixed and Unwrapped)

### DIFF
--- a/lib/Cmt.ml
+++ b/lib/Cmt.ml
@@ -29,6 +29,7 @@ end
 
 include T
 include Comparator.Make (T)
+open Fmt
 
 let create txt loc = {txt; loc}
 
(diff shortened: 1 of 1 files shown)


lib/Cmt.ml
```diff
@@ -29,148 +31,138 @@ end
...
-        let line = String.rstrip ~drop (String.drop_prefix txt pos) in
-        if String.is_empty line then [" "]
-        else if Char.equal line.[String.length line - 1] '\n' then
-          [String.drop_suffix line 1; ""]
-        else if Char.is_whitespace txt.[String.length txt - 1] then
-          [line ^ " "]
-        else [line]
-  in
-  split_asterisk_prefixed_ 0
+module Asterisk_prefixed = struct
```
jberdine, line 38: @gpetiot in case you don't like this name, these comments are also known as "banner" comments due to how they hang off a pole (of asterisks).

PR #1752, opened 2021-07-29
Fix the handling of .ocamlformat-ignore on Windows

### LINKED ISSUE #1751: Bug: .ocamlformat-ignore is not honoured on Windows
Create an empty directory with `a.ml`
```
let
_
=
1
```
and an `.ocamlformat-ignore` with
```
a.ml
```
Then `ocamlformat a.ml` should leave `a.ml` alone, but instead reformats it.

Tested with `ocamlformat` built from `master` with `mingw64` running under Cygwin.

nojb:
Fixes #1751

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 
   + Add missing parentheses around constructor used as indexing op (#1740, @gpetiot)
 
+  + Honour .ocamlformat-ignore on Windows (#1752, @nojb)
+
 #### Changes
 
 #### New features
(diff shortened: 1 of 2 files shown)


gpetiot (approved):
The code looks good to me, if you confirm that it fixes the bug on windows and that .ocamlformat-ignore/.ocamlformat-enable files are properly handled I'm fine to merge it

nojb:
> The code looks good to me, if you confirm that it fixes the bug on windows and that .ocamlformat-ignore/.ocamlformat-enable files are properly handled I'm fine to merge it

I confirm! Thanks.

Issue #1748, opened 2021-07-27
Bug: trailing carriage returns removed from string literals


bcc32:
**Describe the bug**
ocamlformat removes trailing carriage returns from delimited string literals.  I think this might be related to the recent change to CRLF-handling behavior, but it seems like ocamlformat should never touch string literals in a way that would change the actual contents of the string.

**How to Reproduce**

Below is an example file.  Please replace `^M` with actual carriage return characters.

```ocaml
let%expect_test _ =
  do_test (fun () ->
    let%bind () = system "cat foo.csv" in
    [%expect
      {|
    e,d,f^M
    "x
    x",x,xxx^M
    e,d,f
    "x
    x",x,xxx^M
    yy,y,"y
    yy"
   |}];
    return ())
;;
```

When ocamlformat is run on the above file, the carriage returns disappera.

Options:
```
--profile=janestreet
--enable-outside-detected-project
```

bcc32:
FYI, this bug was observed at a572aee651f859a8ca600c59e76c4d34db6954c8.

nojb:
> ocamlformat removes trailing carriage returns from delimited string literals. I think this might be related to the recent change to CRLF-handling behavior, but it seems like ocamlformat should never touch string literals in a way that would change the actual contents of the string.

Thanks for the bug report! Indeed we were aware that the changes to the handling of line endings could cause a problem like the one you encountered (see discussion in #1703). My view is that embedding control characters in string literals is inherently fragile, and that explicitly escaping the CRLF is an acceptable workaround when the difference between CRLF and LF is crucial.

> it seems like ocamlformat should never touch string literals in a way that would change the actual contents of the string.

The problem is that without knowing the intent of the code it is hard to say if this is the right thing to do or not. If you edit your file on a Windows editor you may be inserting CRLF instead of LF "by accident" and normalizing the newline in that case is the right thing to do. Leaving the content of literal strings untouched means that `ocamlformat` may be producing mixed-encoding files routinely even when not needed.

In the end, I went for the simplest possible semantics of the new option, which is to normalize newlines across the board. In #1703 it was suggested that `ocamformat` should output a warning when rewriting string literals with embedded newlines to warn the user of this problem. This looks like a good mitigation to me. Would you be happy with such a warning?

ceastlund:
[...]

PR #1753, opened 2021-07-29
Add `line-endings=lf-unnormalized` option

nojb:
Following discussion in #1748, this PR introduces an third option for the `line-endings` option which uses `lf` as the EOL for the pretty-printing algorithm but does not normalize embedded newlines in string literals (and comments, which are mostly treated the same way as strings). The `janestreet` profile is set to use this option by default.

On the implementation side this is achieved by not doing any newline normalization after pretty printing, making use of the fact that the output channel is already opened in binary mode.

The analogous option `crlf-unnormalized` is not implemented in this PR but could be implemented in a future PR if there is a demand for it (I haven't looked at the details).

The fact that embedded newlines in comments are also not normalized is somewhat unsatisfying, but doing something specific for comments would complicate the implementation. Unless there are bug reports, I prefer to keep it as simple as possible.

cc @ceastlund @hhugo @jberdine

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,10 @@
 
   + Handle merlin typed holes (#1698, @gpetiot)
 
+  + Add `line-endings=lf-unnormalized` option to use LF as line endings in the
+    formatted output, but without modifying newlines in string literals or
+    comments. (#1703, @nojb)
+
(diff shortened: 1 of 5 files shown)


nojb:
Also @bcc32

nojb:
Closing this one in favour of #1754

PR #1754, opened 2021-07-29
Do not normalize newlines inside quoted strings

### LINKED ISSUE #1748: Bug: trailing carriage returns removed from string literals

nojb:
This is an alternative instead of #1753 to solve the problem in #1748, suggested by @hhugo: we simply avoid touching the newlines inside quoted strings `{|...|}`.

The implementation is simple (also suggested by @hhugo): after the final AST has been produced we iterate once over it, recording the locations of such strings, sorted by character position. Later, when normalizing the newlines according to the setting of `line-endings`, we skip those newlines that fall inside one of the locations recorded previously.

Fixes #1748


hhugo:
I took the liberty to rewrite your patch a bit.
- It should now handle the string pattern case (`Ppat_constant ..`).
- The fact that strings are untouched is more explicit.
- `\r\r\r\r\n` now gets normalized as `\n' in `Lf` mode and `\r\n` in `Crlf` mode. Consecutive Cr were previously kept

nojb:
> I took the liberty to rewrite your patch a bit.
> 
> * It should now handle the string pattern case (`Ppat_constant ..`).
> * The fact that strings are untouched is more explicit.
> * `\r\r\r\r\n` now gets normalized as `\n' in `Lf`mode and`\r\n`in`Crlf` mode. Consecutive Cr were previously kept

Looks good, thanks! I'm closing #1753 in favour of this one.

hhugo:
@nojb, can you rebase the pr and fix the conflict.
@jberdine , how do you like this approach ?
@ceastlund, does this work for you ?

nojb:
> @nojb, can you rebase the pr and fix the conflict.

Done.

nojb:
Friendly ping

gpetiot:
The implementation looks good to me, thanks!

@bcc32 does this change fix the bug you repoprted?

bcc32:
@nojb @gpetiot Yes, this patch seems to work!  Thank you!

gpetiot (approved):
Alright looks good then!

nojb:
Thanks!

jberdine:
> @jberdine , how do you like this approach ?

This seems ok to me. But this has all made me wonder whether this is actually simpler than https://github.com/ocaml-ppx/ocamlformat/pull/274 ? To me it seems much cleaner to override the formatter's output function to do translation when needed.

I also think that all defaults, not just the jane profile, should default to not changing string literals. Potentially semantics-changing behavior should need to be explicitly asked for, I think.

nojb:
> I also think that all defaults, not just the jane profile, should default to not changing string literals. Potentially semantics-changing behavior should need to be explicitly asked for, I think.

Unlike the aborted #1753, this PR applies across the board: all users get the new (better) behaviour where newlines inside `{|...|}` are left untouched.

> But this has all made me wonder whether this is actually simpler than #274 ? To me it seems much cleaner to override the formatter's output function to do translation when needed.

Reviving #274 is the first thing I did #1703, but if I remember well there was a problem with newlines in verbatim blocks in docstrings not being correctly normalized, so I took the present approach (post-processing). But perhaps the original approach (overriding `out_newline`) can be made to work with a little more effort, not sure.

jberdine:
> > I also think that all defaults, not just the jane profile, should default to not changing string literals. Potentially semantics-changing behavior should need to be explicitly asked for, I think.
> 
> Unlike the aborted #1753, this PR applies across the board: all users get the new (better) behaviour where newlines inside `{|...|}` are left untouched.

Thanks, my confusion between the two then.

> > But this has all made me wonder whether this is actually simpler than #274 ? To me it seems much cleaner to override the formatter's output function to do translation when needed.
> 
> Reviving #274 is the first thing I did #1703, but if I remember well there was a problem with newlines in verbatim blocks in docstrings not being correctly normalized, so I took the present approach (post-processing). But perhaps the original approach (overriding `out_newline`) can be made to work with a little more effort, not sure.

Ok, thanks for pointing out that issue. Do you know if that was the only one? I'm not sure how the implementation complexity of the two approaches would end up comparing, but it could be preferable to override `out_newline` and then explicitly transform strings (such as the bodies of verbatim blocks) that contain embedded newlines that should be normalized, rather than postprocessing all newlines except those explicitly collected in the preprocessing scan. The override approach would be a lighter touch, and I expect it would be clearer since the code would identify where action is taken rather than acting on everything except specially identified points. It would perhaps be slightly more efficient, though I don't know if it is significant.

nojb:
We may want to revisit this issue following ocaml/ocaml#12514.

PR #1755, opened 2021-07-30
Update vendored parser with ocaml#10441

gpetiot:
Apply the changes of https://github.com/ocaml/ocaml/pull/10441, no behavioral change


hhugo:
Could we commit the diff between upstream and our parser ? This would be nice to make sure we don't diverge to much

gpetiot:
> Could we commit the diff between upstream and our parser ? This would be nice to make sure we don't diverge to much

Yes I'm opening PRs will the remaining changes, but some require to implement new things in Fmt_ast so I want to keep the PRs distinct

Issue #633, opened 2019-02-16
Support for old versions of the compiler.


hhugo:
Extracted from https://github.com/ocaml-ppx/ocamlformat/issues/630:

@hhugo:
> This makes me think that we might need a flag to specify the targeted version of the OCaml compiler.
For instance, it would be incorrect to use the following syntax in 4.07.
`module M(_:S) = struct end`

@jberdine:
>My position has always been to try to support the latest released version and the one preceding for "a while" to make it possible to upgrade non-atomically. That's not very generous, I know. But I'm not sure of the cost:value ratio for supporting old versions for dev tools, especially when they are so closely tied to compiler internals. Also, I don't know if it makes sense to commit/try to ensure that everyone can use the same compiler version to build ocamlformat as to build the code they are formatting. I'm definitely open to alternative, but just to say that I haven't already been convinced that something like dune's approach to legacy support is the right choice here.

> For examples like the particular case of `module M(_:S) = struct end`, would it make more sense wrt community-wide effort to make a ppx that could be used with eg 407 that would convert the 408 ast to the expanded form that 407 can handle?

hhugo:
While formatting some part of js_of_ocaml, I had to manually tune few places to stay compatible with ocaml 4.02.3. The churn was not very big, work around were trivial. Unsure if it could become painful overtime.

emillon:
This came up again a couple times:
- `(type a b)` as a shorthand for `(type a) (type b)` (only available on 4.03+)
- functor keyword vs module argument #1285 

Initially I was thinking that an option for target version would work, but maintaining concrete syntax seems a better choice in general.

jberdine:
Can I just note that there are noteworthy downsides to just punting on such questions and trying to preserve the concrete syntax.

- Preserving the concrete syntax undermines all the objectives of using an auto-formatter, and should really be left as a last-ditch escape hatch mechanism for when things are necessarily very ad hoc.
- The use of autoformatters to automatically update code to the current/modern language is blocked by this sort of thing. It would be better if the formatter could automatically rewrite old workaround to new solutions based on a target version.
- The implementation mechanisms for preserving the concrete syntax will never be as efficient.

Do we know how much of a maintenance burden it would be to add a target version for the generated concrete syntax? Most importantly, do we know if it would interact with other configuration options in a multiplicative or additive way?

gpetiot:
[...]

PR #1756, opened 2021-07-30
Handle punned labelled arguments with type constraint in function applications (ocaml#10434)

gpetiot:
Update the parser with https://github.com/ocaml/ocaml/pull/10434

Accept labelled argument punning with type constraint in pexp_apply

For example, function application of the form `foo ~(x:int)` instead of the explicit `foo ~x:(x:int)`.

No diff with test_branch.sh

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,9 @@
 
   + Handle merlin typed holes (#1698, @gpetiot)
 
+  + Handle punned labelled arguments with type constraint in function applications.
+    For example, function application of the form `foo ~(x:int)` instead of the explicit `foo ~x:(x:int)`. (ocaml#10434) (#1756, @gpetiot)
+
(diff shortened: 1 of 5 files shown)


nojb:
Hello, naive question: how does `ocamlformat` handle the dependency on the OCaml version in order to avoid using a syntax that is not understood by "old" OCaml versions (such as the one used by this PR)?

hhugo:
#633

nojb:
> #633

Thanks!

nojb:
> > #633
> 
> Thanks!

So do I understand correctly that you cannot use the syntax `~x:(x : ...)` together with the next release of `ocamlformat` unless you are running the version of OCaml that includes that syntax (at the moment, only `trunk` I think).

gpetiot:
I've opened #1759 to take care of this

PR #1759, opened 2021-08-03
Add an ocaml-version option, format punned labelled arguments when OCaml 4.14+ syntax is selected

gpetiot:
Follow-up on #1756

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -59,6 +59,7 @@ type t =
   ; max_iters: int
   ; module_item_spacing: [`Compact | `Preserve | `Sparse]
   ; nested_match: [`Wrap | `Align]
+  ; ocaml_version: Ocaml_version.t
   ; ocp_indent_compat: bool
   ; parens_ite: bool
   ; parens_tuple: [`Always | `Multi_line_only]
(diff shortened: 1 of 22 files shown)


nojb (commented):
I think the version should be 4.14+, not 4.13+.

gpetiot:
Ah indeed, I thought it would be included in the 4.13 release

nojb:
> Ah indeed, I thought it would be included in the 4.13 release

I don't know anything about it, but I don't think it is at the moment.

hhugo:
@gpetiot, what do you think about happening an issue for listing all syntaxes that need to be conditional on the ocaml version (e.g. (type a b))

gpetiot:
I will add a list to #633

lib/Conf.ml
```diff
@@ -935,6 +944,16 @@ module Formatting = struct
       (fun conf x -> {conf with nested_match= x})
       (fun conf -> conf.nested_match)
 
+  let ocaml_version =
+    let docv = "V" in
+    let doc = "Version of OCaml syntax of the output." in
+    let default = Ocaml_version.sys_version in
+    let default_doc = "the currently installed OCaml version" in
```
jberdine, line 951: @gpetiot I think that this is not as clear as it could be. Wouldn't "the version of OCaml used to build OCamlFormat" be better?

gpetiot: Indeed, done in #1765

Issue #1750, opened 2021-07-29
Bug: quadratic behaviour with nested lists in argument position


nojb:
The following example shows a quadratic behaviour in `ocamlformat`: if you add an extra `all [`, the running time roughly doubles.
```
let _ =
  all
    [
      f
       (all
        [
          all
            [
              all
                [
                  all
                    [
                      all
                        [
                          all
                            [
                             all
                               [
                                 all
                                   [
                                     all
                                       [
                                         all
                                           [
                                             all
                                               [
                                                 all
                                                   [
                                                     all
                                                       [
                                                         all
                                                          [
                                                            identify]]]]]]]]]]]]]])]
```

gpetiot:
@emillon After profiling it looks like the slowdown comes from the infix operators `Fmt.( $ )`, tried removing the Staged abstraction and further flatten the list formatting but the performance is still as bad

emillon:
I just had a look. Part of the issue in this particular case is that there's a code path that triggers layout on a subexpression (`Cmts.preserve` does this):

https://github.com/ocaml-ppx/ocamlformat/blob/68e3256b5551933d5b12a1215e61318b40a85d29/lib/Fmt_ast.ml#L1392-L1403

So the first layer checks if the expression is simple by triggering layout on the second layer, which... and after checking that, it still has layout again (for real now!) the second layer, etc.

By adding a cache on the function that computes if a given expression breaks, the runtime on this file becomes instantaneous (6s -> 40ms). I'm not sure that it's the best solution, but it works here!

nojb:
> By adding a cache on the function that computes if a given expression breaks, the runtime on this file becomes instantaneous (6s -> 40ms). I'm not sure that it's the best solution, but it works here!

Do you have a branch with your patch? I would like to try it on some other files where `ocamlformat` feels a bit slow.

emillon:
I just pushed that to https://github.com/emillon/ocamlformat/tree/breaks-cache.

I haven't tried to repro, but it looks like heavily nested patterns and nested arrays could suffer from the same issue since they're donig this sublayout too.

emillon:
(note that the cache might not be correct in the presence of comments)

nojb:
> I just pushed that to https://github.com/emillon/ocamlformat/tree/breaks-cache.
> 

Thanks. For info, on a file where `ocamlformat` feels slow (~ 0.59s to reformat a file which is already formatted), the patch brings it down to ~0.48s.

PR #1763, opened 2021-08-04
Cmts.preserve: improve API

emillon:
This simplifies how `Cmts.preserve` deals with mutation. Instead of setting a `remove` flag, it can do a shallow copy and restore the original value of mutable fields at the end.

### DIFF
--- a/lib/Cmts.mli
+++ b/lib/Cmts.mli
@@ -126,6 +126,6 @@ val diff :
   Conf.t -> Cmt.t list -> Cmt.t list -> (string, string) Either.t Sequence.t
 (** Difference between two lists of comments. *)
 
-val preserve : (t -> Fmt.t) -> t -> string
-(** [preserve fmt_x x] formats like [fmt_x x] but returns a string and does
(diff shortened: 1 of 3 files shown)


emillon:
That's in preparation to fix #1750.

lib/Cmts.mli
```diff
@@ -126,6 +126,6 @@ val diff :
   Conf.t -> Cmt.t list -> Cmt.t list -> (string, string) Either.t Sequence.t
 (** Difference between two lists of comments. *)
 
-val preserve : (t -> Fmt.t) -> t -> string
-(** [preserve fmt_x x] formats like [fmt_x x] but returns a string and does
-    not consume comments from the internal state. *)
+val preserve : (unit -> Fmt.t) -> t -> string
+(** [preserve t f] formats like [f ()] but returns a string and does not
```
gpetiot, line 130: I think it should be `preserve f t` instead

gpetiot (approved):
It looks good to me, thanks!

PR #1749, opened 2021-07-28
Attach location to constants

gpetiot:
No diff with test_branch.sh

This allows us to remove all the code that parses the tokens extracted from Source.ml to retrieve the location.
Upstream PR: https://github.com/ocaml/ocaml/pull/10544

PR #1764, opened 2021-08-04
Use OCaml's parser for validation step

gpetiot:
- Removed `Ast_passes`: when I added this module at first I imagined we could add passes to recreate the sugaring, add locations, etc. but it's become clear that these changes will be added in the same ast, this new ast will become `Ocamlformat_ast`
- Add `Ocamlformat_ast`, this AST will differ from the OCaml's AST: for now #1694 #1749 in the future adding locations to underscore, preserving begin/end, adding comments, etc.
- Add `Std_ast`, that is the vendored 4.13 parser
- For now these 2 are identical but they will diverge as soon as #1694 is merged

The only diff is that parsing errors are printed twice in case of error, which is weird since the second parser must not be called if the first one errored out, but that's not a huge deal I guess.


test/passing/tests/error4.ml.ref
```diff
@@ -6,6 +6,14 @@ File "tests/error4.ml", line 3, characters 8-16:
...
             ^^^^^^^^
 Warning 50 [unexpected-docstring]: unattached documentation comment (ignored)
+File "tests/error4.ml", line 2, characters 0-13:
+2 | (** a or b *)
+    ^^^^^^^^^^^^^
+Warning 50 [unexpected-docstring]: ambiguous documentation comment
+File "tests/error4.ml", line 3, characters 8-16:
+3 | let b = (** ? *) ()          
+            ^^^^^^^^
+Warning 50 [unexpected-docstring]: unattached documentation comment (ignored)
```
jberdine, line 16: This happens because when warning 50 is reported, a valid parsetree is still returned. Look at the call to `Docstrings.warn_bad_docstrings` in the `wrap` function in `parsing/parse.ml` of the compiler. It seems like it would be enough to disable the warning during one of the calls to parse, probably better for the custom one. I'm not sure if the compiler's `Warnings.without_warnings` is exposed, but it ought to work I think.

jberdine (commented):
Thanks for working on this, I think it is important.

This overall looks good to me. It makes me think that the `Ocamlformat_ast` is perhaps not the best name, as this change would be clearer if it was named something like `Extended_ast`.

It would be good to resolve the duplicated warnings issue. See inline comment for my interpretation of what is happening.

test/passing/tests/error4.ml.ref
gpetiot: I've added a more specific argument `?disable_w50` to `Parse_with_comments.parse instead`, I previously tried to disable all warnings with `Warnings.without_warnings` but some warnings should not be disabled as it got me this diff:
```diff
------ test/failing/tests/unit_lex.ml.broken-ref
++++++ test/failing/unit_lex.ml.output
File "test/failing/tests/unit_lex.ml.broken-ref", line 1, characters 0-1:
-|File "tests/unit_lex.ml", line 18, characters 4-10:
-|18 |     ������ (* this file must be iso-8859-1 *)
-|         ^^^^^^
-|Alert deprecated: ISO-Latin1 characters in identifiers
 |ocamlformat: ignoring "tests/unit_lex.ml" (syntax error)
 |File "tests/unit_lex.ml", line 55, characters 2-8:
 |55 |   '\999'; (* wrong, but yet... *)
 |       ^^^^^^
 |Error: Illegal backslash escape in string or character ('\999'): 999 is outside the range of legal characters (0-255).
     patdiff (internal) (exit 1)
(cd _build/default && /home/gpe/.opam/ocamlformat-4.11.0/bin/patdiff -keep-whitespace -location-style omake test/failing/tests/unit_values.ml.broken-ref test/failing/unit_values.ml.output)
------ test/failing/tests/unit_values.ml.broken-ref
++++++ test/failing/unit_values.ml.output
File "test/failing/tests/unit_values.ml.broken-ref", line 1, characters 0-1:
-|File "tests/unit_values.ml", line 6, characters 10-11:
-|6 | let i32 = −1073741824, 1073741823
-|              ^
-|Alert deprecated: ISO-Latin1 characters in identifiers
 |ocamlformat: ignoring "tests/unit_values.ml" (syntax error)
 |File "tests/unit_values.ml", line 6, characters 11-12:
 |6 | let i32 = −1073741824, 1073741823
 |               ^
 |Error: Illegal character (\136)
     patdiff (internal) (exit 1)
(cd _build/default && /home/gpe/.opam/ocamlformat-4.11.0/bin/patdiff -keep-whitespace -location-style omake test/passing/tests/cinaps.ml test/passing/cinaps.ml.output)
------ test/passing/tests/cinaps.ml
++++++ test/passing/cinaps.ml.output
File "test/passing/tests/cinaps.ml", line 49, characters 0-1:
 |  (* fooooooo *)
 |  let y = 2 in
 |  (* foooooooo *)
 |  z
 |$*)
 |(*$*)
 |
 |let foo = foo
 |
 |(*$QR foo Q.small_int (fun i-> foo i (+) [1;2;3] = List.fold_left (+) i
 |  [1;2;3] ) *)
 |let foo = foo
 |
 |(* Cinaps comment should not wrap if they don't parse. The first one would
 |   crash and the second become a mess *)
 |
-|(*$(**)"
-|  "*)
+|(*$ (**) "\n  " *)
 |
 |(*$
 |  print_newline () ;
 |  <SYNTAX ERROR>
 |  List.iter
 |    (fun s -> Printf.printf "let ( %s ) = Pervasives.( %s )\n" s s)
 |    ["+"; "-"; "*"; "/"]
 |*)
 |(*$*)
```

lib/Translation_unit.ml
```diff
@@ -325,30 +324,40 @@ let format (type a b) (fg0 : a Ast_passes.Ast0.t)
...
         |> List.filter_map ~f:(fun (s, f_opt) ->
                Option.map f_opt ~f:(fun f -> (s, String.sexp_of_t f)) )
       in
-      ( match parse Ast_passes.Ast0.Parse.ast fg0 conf ~source:fmted with
+      ( match
+          parse Extended_ast.Parse.ast ~disable_w50:true fg conf
+            ~source:fmted
+        with
       | exception Sys_error msg -> Error (Error.User_error msg)
       | exception Warning50 l -> internal_error (`Warning50 l) (exn_args ())
```
jberdine, line 340: This is dead now I guess.

jberdine (approved):
Thanks, this looks good.

lib/Translation_unit.ml
gpetiot: done

PR #1766, opened 2021-08-05
Cache results of Cmts.preserve

### LINKED ISSUE #1750: Bug: quadratic behaviour with nested lists in argument position

emillon:
The `preserve` function returns a string with the layout of a part of the AST. It is used in several places to determine if a subpart has a break for example.

This can be a problem when these constructions are nested, since laying out the external construction requires laying out the internal ones first, and this creates quadratic behavior.

This is solved by caching the results of these intermediate layouts. This cache is unbounded so there are bad corner cases. At worst, each cached point is associated to the result string which could mean quadratic memory consumption.

Some ways to fix this:
- use a LRU cache like the one from `core_kernel`
- store only a summary rather than the full string (length, number of lines, max indent)

In addition to #1750, a similar case can be triggered using nested cases. It seems that `expression_width` would be affected but I have not been able to reproduce that using that code path, so this part stays not cached.

Closes #1750

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1366,7 +1366,10 @@ and fmt_label_arg ?(box = true) ?epi ?parens ?eol c
   | _ -> fmt_label lbl ":@," $ fmt_expression c ~box ?epi ?parens xarg
 
 and expression_width c xe =
-  String.length (Cmts.preserve (fun () -> fmt_expression c xe) c.cmts)
+  String.length
+    (Cmts.preserve ~cache_key:(Expression xe.ast)
+       (fun () -> fmt_expression c xe)
+       c.cmts )
 
 and fmt_args_grouped ?epi:(global_epi = noop) c ctx args =
   let fmt_arg c ~first:_ ~last (lbl, arg) =
@@ -1395,6 +1398,7 @@ and fmt_args_grouped ?epi:(global_epi = noop) c ctx args =
     let xexp = sub_exp ~ctx x in
     let output =
       Cmts.preserve
+        ~cache_key:(Arg (lbl, x))
         (fun () ->
           let cmts = Cmts.drop_before c.cmts x.pexp_loc in
           fmt_arg ~first:false ~last:false {c with cmts} (lbl, x) )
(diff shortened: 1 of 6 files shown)


emillon:
MaxRSS does not change too much when formatting `source.ml` so even if this introduces a memory leak, this stays small enough not to matter.

lib/Fmt_ast.ml
```diff
@@ -1366,7 +1409,7 @@ and fmt_label_arg ?(box = true) ?epi ?parens ?eol c
   | _ -> fmt_label lbl ":@," $ fmt_expression c ~box ?epi ?parens xarg
 
 and expression_width c xe =
-  String.length (Cmts.preserve (fun () -> fmt_expression c xe) c.cmts)
+  String.length (Cmts.preserve c (fun () -> fmt_expression c xe))
```
gpetiot, line 1412: Out of curiosity, even if the issue was not reproduced on expressions, wouldn't it be better to cache the results as well for consistency, and just in case it breaks in the future?

gpetiot (approved):
The performance improvement is impressive, it looks good to me, thanks!

lib/Fmt_ast.ml
```diff
@@ -19,12 +19,48 @@ open Extended_ast
...
+end
+
 type c =
   { conf: Conf.t
   ; debug: bool
   ; source: Source.t
   ; cmts: Cmts.t
-  ; fmt_code: Conf.t -> string -> (Fmt.t, unit) Result.t }
+  ; fmt_code: Conf.t -> string -> (Fmt.t, unit) Result.t
+  ; layout_cache: Layout_cache.t }
```
gpetiot, line 63: I'm not sure it's worth to add this field to the type c, as we don't need to pass it to every function, it's only used by Cmts.preserve, why not add it to Cmts.t and directly modify the implementation of Cmts.preserve?

emillon: I did so because it's not really related to comment handling, but yes we can do that.

lib/Fmt_ast.ml
emillon: That takes more memory for no known reason. I thought about adding a warning when we detect > 2 nested calls of `expression_width` (which would indicate there's a way to nest arbitrarily - I haven't found a way to do that) but that's not super useful. I just pushed a commit that does that.

gpetiot (approved):
Thanks!

PR #1768, opened 2021-08-06
Fix misaligned higher-order functor parameters

### LINKED ISSUE #1678: Bug: misaligned higher-order functor parameters

```ocaml
module F1 (G : functor (_ : T) ->
  T) (A : sig
  val x : int
end) =
struct end

module F2 (G : functor (_ : T) ->
  T_________________________________________________________________________)
                                                                            (A : sig
  val x : int
end) =
struct end

module F3
    (G : functor
      (_ : T____________________________________________)
      (_ : T____________________________________________)
      ->
  T) (A : sig
  val x : int
end) =
struct end
```

It appears that using immediate `functor ..` types confuses OCamlformat's normal placement strategy for functor parameters, resulting in a variety of unusual indentation. I'm not sure entirely what I'd expect this to format to, but roughly speaking:
 - In the first case, `F1`, I'd expect `G` to not break, and `A` to either start on the subsequent line or dock after `G`.
 - For `F2`, I'd expect `A` to start much further to the left.
 - For `F3`, I'd expect the `T` result type of `G` to be indented by another couple of levels (and perhaps `A` to start on its own line, left-aligned with `G`).

I don't know whether these are known limitations of the placement strategy, in which case feel free to reclassify this issue as a feature request or close it altogether.

gpetiot:
I made some experiments to fix #1678, so far the diff on the conventional profile looks good.
cc @CraigFe since you opened the issue.

The idea is to not put the beginning of the functor on the same line as the module opening if the functor signature is going to be complex, I find it easier to read. Thoughts?

The example of #1678 looks like this:
```ocaml
module F1            
    (G : functor (_ : T) -> T) (A : sig
      val x : int
    end) =
struct end

module F2
    (G : functor (_ : T) ->
           T_________________________________________________________________________
    )
    (A : sig
      val x : int
    end) =
struct end

module F3
    (G : functor
           (_ : T____________________________________________)
           (_ : T____________________________________________)
           ->
           T)
    (A : sig
      val x : int
    end) =
struct end
```

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -3361,11 +3361,15 @@ and fmt_functor_arg c {loc; txt= arg} =
   match arg with
   | Sugar.Unit -> Cmts.fmt c loc (str "()")
   | Sugar.Named (name, mt) ->
+      let mty = fmt_module_type c mt in
       Cmts.fmt c loc
         (wrap "(" ")"
            (hovbox 0
-              ( hovbox 0 (fmt_str_loc_opt c name $ fmt "@ : ")
-              $ compose_module (fmt_module_type c mt) ~f:Fn.id ) ) )
+              ( hovbox 0 (fmt_str_loc_opt c name $ fmt "@ :")
+              $
+              match mty.pro with
+              | Some _ -> str " " $ compose_module mty ~f:Fn.id
+              | None -> break 1 2 $ compose_module mty ~f:(hvbox 0) ) ) )
 
 and fmt_module_type c ({ast= mty; _} as xmty) =
   let ctx = Mty mty in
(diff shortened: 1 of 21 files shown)


jberdine:
I agree that it would be more readable to be less eager about putting the beginning of complex functors on the same line. But I think that it would still be better to put `sig` at the end of the preceding line rather than on its own line and increasing the indentation of the whole signature body. The `sig` token itself does not add any readability.

gpetiot:
@CraigFe does that fix every case you encountered?

Julow (commented):
The code from the issue now formats to:
```ocaml
module F1
    (G : functor (_ : T) -> T) (A : sig
      val x : int
    end) =
struct end

module F2
    (G :
      functor (_ : T) ->
        T_________________________________________________________________________
    )
    (A : sig
      val x : int
    end) =
struct end

module F3
    (G :
      functor
        (_ : T____________________________________________)
        (_ : T____________________________________________)
        -> T)
    (A : sig
      val x : int
    end) =
struct end
```

`F1`: Shouldn't `A` start on its own line ? It does when `G` breaks, so I think it should do when `A` itself breaks.

`F2`: The break before the closing parenthese of `G` is due to the identifier exceeding the margin. I think that's fine, but would look slighly better if then parenthese was still on the right.

Maybe unrelated: There's a break between `= struct` but not between `: sig`:

```ocaml
module F1
    (G : functor (_ : T) -> T) (A : sig
      val x : int
    end) : sig end
```

lib/Fmt_ast.ml
```diff
@@ -3442,11 +3442,14 @@ and fmt_functor_arg c {loc; txt= arg} =
...
+      let mty = fmt_module_type c mt in
       Cmts.fmt c loc
         (wrap "(" ")"
            (hovbox 0
-              ( hovbox 0 (fmt_str_loc_opt c name $ fmt "@ : ")
-              $ compose_module (fmt_module_type c mt) ~f:Fn.id ) ) )
+              ( hovbox 0 (fmt_str_loc_opt c name $ fmt "@ :")
+              $ fmt_or (Option.is_some mty.pro) " " "@;<1 2>"
+              $ compose_module mty ~f:(fun k ->
+                    hvbox_if (Option.is_none mty.pro) 0 k ) ) ) )
```
Julow, line 3452: I think this would be easier to read if `mty.pro` was matched only once:
```ocaml
    match mty.pro with
    | Some _ -> fmt " " $ compose_module mty ~f:(fun k -> k)
    | None -> fmt "@;<1 2>" $ compose_module mty ~f:(fun k -> hvbox 0 k)
```

Julow (approved):
Looks good :)

test/passing/tests/module.ml
```diff
@@ -71,3 +71,43 @@ let helper ?x =
...
+end
+
+module M = struct
+  module F
+      (X : sig
+        module type T
+      end)
+      (Res : X.T) =
+  Res
+  end
```
gpetiot, line 113: to fix

gpetiot:
Replaced by #2322

PR #1775, opened 2021-08-23
Remove the version handshake from the RPC protocol

gpetiot:
This simplifies the RPC protocol by removing the whole "version handshake" part, the old `Version is not a command anymore but if some software used the csexp representation of the version command it will just get ignored.

This is necessary for the usecase that mdx has for ocamlformat-rpc (https://github.com/realworldocaml/mdx/pull/335)


Julow:
> Is it reasonable to send from the server a message containing the version of the protocol as soon as the client connects

Actually, I think we should keep the `version` command as it is. The RPC is released at the same time as ocamlformat, clients that want to support several versions of ocamlformat will have to support several versions of the RPC too.

I agree that the server doesn't need to support several versions of the client. If it really needs to be simplified, what about:

- Move the command to the main protocol (remove the "wait for version" phase).
  Maybe some clients don't need to bother with that because they use Opam constraints or vendor ocamlformat ? Or they cached the result.
- Document that the command doesn't take an argument anymore: it has a constant response.

gpetiot:
I don't remember all the details but I think what happened for our `mdx` usecase https://github.com/realworldocaml/mdx/pull/335 is that (for some reason I forgot) we need an intermediate layer between the ocamlformat-rpc server and the client, and we can't have the handshake made in these both layers, and it doesn't work either to have it in only one, and the modules used would not be the same.

PR #1783, opened 2021-08-27
Track differences between the various ocaml parsers

### LINKED ISSUE #1781: Track differences between the various ocaml parsers.
We have two (or three ?) OCaml parsers vendor-ed into the repo.
I think we need to make sure they don't diverge too much from each other and from the upstream parser.

I propose that we do code review on the differences and that we commit at least 2 diffs in the repo :
- between upstream and the vendor-ed ocaml-4.13
- between ocaml-4.13 and ocaml-4.13-extented

We should probably do something similar with the parser in parse-wyc

gpetiot:
Closes #1781

I still have to figure out why the dune rules are not working, but opening the PR in case someone has some time to look at it in the meantime.


hhugo:
I don't think the runtest rules are necessary. We just need rules to generate diffs with automatic promotion

hhugo:
I've updated the PR:
- diff all files from the compiler parser, instead of just parser.mly
- add a small tool to compute diffs between two parsers and import the upstream one
- commit the upstream parser, we don't want dune to call curl
- update diff files

hhugo:
the diff vs upstream is quite large, is there a description of how to import the parser ?

gpetiot:
I got a dune error `Error: _build/tools/diff-parsers/diff_parsers.exe refers to unknown build
context: tools`, maybe I'm missing something. cc @emillon in case there is a simpler solution we didn't see.


> the diff vs upstream is quite large, is there a description of how to import the parser ?

I'm not sure what you mean, but no there is no description/

hhugo:
how did you get the error ? by running the import script manually ? `vendor/import-ocaml-upstream.sh` should be run from inside vendor

hhugo:
I've updated the source and diff with the right version. The diff is now much smaller :)

gpetiot:
I've added the diff of parse-wyc with respect to ocaml-4.13-extended.
I was trying to remove the modification time from the diff with `sed` or something else but I'm not sure it's possible. The PR looks good to me.

vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -0,0 +1,978 @@
...
+ let comment_list = ref []
+ 
+ let add_comment com =
+@@ -272,63 +272,10 @@
+   in
+     add_comment com
+ 
+ let comments () = List.rev !comment_list
+ 
+-(* Error report *)
```
hhugo, line 27: This diff could be made smaller by commenting out the code instead of deleting it

vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -0,0 +1,978 @@
...
+   | LABEL simple_expr %prec below_HASH
+@@ -2495,12 +2490,12 @@
+ ;
+ let_binding_body:
+   | let_binding_body_no_punning
+       { let p,e = $1 in (p,e,false) }
+ /* BEGIN AVOID */
+-  | val_ident %prec below_HASH
+-      { (mkpatvar ~loc:$loc $1, mkexpvar ~loc:$loc $1, true) }
++  (*| val_ident %prec below_HASH
```
hhugo, line 746: why is this commented

vendor/diff-parsers-upstream-std.patch
```diff
@@ -0,0 +1,821 @@
...
+   Pat.mk ~loc:lbl.loc  (Ppat_var (loc_last lbl))
+ 
+ let mk_newtypes ~loc newtypes exp =
+@@ -822,13 +822,13 @@
+ %nonassoc HASH                         /* simple_expr/toplevel_directive */
+ %left     HASHOP
+ %nonassoc below_DOT
+ %nonassoc DOT DOTOP
+ /* Finally, the first tokens of simple_expr are above everything else. */
+-%nonassoc BACKQUOTE BANG BEGIN CHAR FALSE FLOAT INT
```
hhugo, line 229: Why do we add OBJECT here ?

vendor/diff-parsers-upstream-std.patch
```diff
@@ -0,0 +1,821 @@
...
+   | LET MODULE ext_attributes mkrhs(module_name) module_binding_body IN seq_expr
+       { Pexp_letmodule($4, $5, $7), $3 }
+@@ -2312,14 +2317,10 @@
+       { Pexp_for($3, $5, $7, $6, $9), $2 }
+   | ASSERT ext_attributes simple_expr %prec below_HASH
+       { Pexp_assert $3, $2 }
+   | LAZY ext_attributes simple_expr %prec below_HASH
+       { Pexp_lazy $3, $2 }
+-  | OBJECT ext_attributes class_structure END
+-      { Pexp_object $3, $2 }
```
hhugo, line 275: What is this diff about ?

vendor/diff-parsers-upstream-std.patch
```diff
@@ -0,0 +1,821 @@
...
+   | mod_longident DOT LPAREN seq_expr error
+       { unclosed "(" $loc($3) ")" $loc($5) }
+   | LBRACE record_expr_content RBRACE
+@@ -2465,10 +2472,13 @@
+   | LABEL simple_expr %prec below_HASH
+       { (Labelled $1, $2) }
+   | TILDE label = LIDENT
+       { let loc = $loc(label) in
+         (Labelled label, mkexpvar ~loc label) }
++  | TILDE LPAREN label = LIDENT ty = type_constraint RPAREN
```
hhugo, line 317: Why this diff ?

vendor/diff-parsers-upstream-std.patch
```diff
@@ -0,0 +1,821 @@
...
+         (Optional label, mkexpvar ~loc label) }
+   | OPTLABEL simple_expr %prec below_HASH
+       { (Optional $1, $2) }
+@@ -2494,19 +2504,15 @@
+         let loc = Location.(t.ptyp_loc.loc_start, t.ptyp_loc.loc_end) in
+         let typ = ghtyp ~loc (Ptyp_poly([],t)) in
+         let patloc = ($startpos($1), $endpos($2)) in
+         (ghpat ~loc:patloc (Ppat_constraint(v, typ)),
+          mkexp_constraint ~loc:$sloc $4 $2) }
+-  | let_ident COLON typevar_list DOT core_type EQUAL seq_expr
```
hhugo, line 331: Why this diff ?

vendor/diff-parsers-upstream-std.patch
```diff
@@ -0,0 +1,821 @@
...
+      pcd_loc: Location.t;
+      pcd_attributes: attributes; (* C of ... [@id1] [@id2] *)
+     }
+@@ -524,15 +529,16 @@
+     ptyexn_loc: Location.t;
+     ptyexn_attributes: attributes; (* ... [@@id1] [@@id2] *)
+   }
+ 
+ and extension_constructor_kind =
+-    Pext_decl of constructor_arguments * core_type option
```
hhugo, line 591: What's this diff ?

hhugo (commented):
I've reviewed both the PR and the content of the parsers diffs.

vendor/diff-parsers-upstream-std.patch
gpetiot: This is due to https://github.com/ocaml-ppx/ocamlformat/pull/1755, maybe we need to compare with the `trunk` branch instead of `4.13`

vendor/diff-parsers-upstream-std.patch
gpetiot: Same https://github.com/ocaml-ppx/ocamlformat/pull/1755

vendor/diff-parsers-upstream-std.patch
hhugo: Well trunk is a moving target. What about a specific commit that we would bump from time to time ?
Also if we're already supporting stuff from after 4.13, we should maybe not name directories with 4.13 in it.

vendor/diff-parsers-upstream-std.patch
gpetiot: Due to https://github.com/ocaml-ppx/ocamlformat/pull/1756

vendor/diff-parsers-upstream-std.patch
gpetiot: Due to https://github.com/ocaml-ppx/ocamlformat/pull/1757

vendor/diff-parsers-upstream-std.patch
gpetiot: Also due to https://github.com/ocaml-ppx/ocamlformat/pull/1757

vendor/diff-parsers-ext-parsewyc.patch
gpetiot: All the rules between `/* BEGIN AVOID */ ... /* END AVOID */` (I didn't comment the full declaration between comments as it does not matter as long as I comment the calls) should be commented for `parse-wyc` otherwise the error rules of the parser are triggered instead of the recovery mode. If I uncomment this one I get regressions with parse-wyc.

vendor/diff-parsers-ext-parsewyc.patch
gpetiot: done

tools/diff-parsers/diff_parsers.ml
```diff
@@ -0,0 +1,59 @@
+let diff d1 d2 f =
```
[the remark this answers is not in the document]
gpetiot: I don't think so

vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -0,0 +1,978 @@
...
+-          Location.prerr_warning (Location.curr lexbuf) Warnings.Eol_in_string;
++      { (*if not (in_comment ()) then
++          Location.prerr_warning (Location.curr lexbuf) Warnings.Eol_in_string;*)
+         update_loc lexbuf None 1 false 0;
+         store_lexeme lexbuf;
+         string lexbuf
+       }
+   | eof
+--- ocaml-4.13-extended/parser.mly
++++ parse-wyc/lib/parser.mly
```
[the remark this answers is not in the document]
gpetiot: done

vendor/diff-parsers-upstream-std.patch
gpetiot: I updated the version used for the vendored upstream, I agree that "4.13" is not exact, that's some kind of 4.13+, but not yet 4.14 and it might still change before 4.14. Personally it does not confuse me (4.13 is more informative than ocaml_current or ocaml_trunk) we can change it in another PR if we find a better name, since there are more impact in the code.

vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -0,0 +1,953 @@
...
+-  | LPAREN error              { expecting $loc($2) "operator" }
+-  | LPAREN MODULE error       { expecting $loc($3) "module-expr" }
++  (*| LPAREN operator error     { unclosed "(" $loc($1) ")" $loc($3) }*)
++  (*| LPAREN error              { expecting $loc($2) "operator" }*)
++  (*| LPAREN MODULE error       { expecting $loc($3) "module-expr" }*)
+ ;
+ val_ident:
+     LIDENT                    { $1 }
+-  | val_extra_ident           { $1 }
++  (*| val_extra_ident           { $1 }*)
```
hhugo, line 907: Why this diff ?

vendor/diff-parsers-upstream-std.patch
```diff
@@ -0,0 +1,301 @@
...
++let payload = payload reset_ctxt
+--- ocaml-4.13-upstream/pprintast.mli
++++ ocaml-4.13/pprintast.mli
+@@@@
+ val module_expr: Format.formatter -> Parsetree.module_expr -> unit
+ 
+ val toplevel_phrase : Format.formatter -> Parsetree.toplevel_phrase -> unit
+ val top_phrase: Format.formatter -> Parsetree.toplevel_phrase -> unit
+ 
++val class_field: Format.formatter -> Parsetree.class_field -> unit
```
hhugo, line 289: Have you considered upstreaming theses changes ?

vendor/diff-parsers-ext-parsewyc.patch
gpetiot: It was a mistake, I've uncommented the line

vendor/diff-parsers-upstream-std.patch
gpetiot: I will open a PR

vendor/diff-parsers-std-ext.patch
```diff
@@ -0,0 +1,308 @@
...
+     | Ppat_lazy p -> lazy_ ~loc ~attrs (sub.pat sub p)
+--- ocaml-4.13/asttypes.mli
++++ ocaml-4.13-extended/asttypes.mli
+@@@@
+ 
+ type override_flag = Override | Fresh
+ 
+ type closed_flag = Closed | Open
+ 
++type closed_flag_loc = Closed | Open of Location.t
```
hhugo, line 98: Why define a new type instead of patching the one above ?

vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -0,0 +1,948 @@
...
++
+ %{
+ 
+ open Asttypes
+ open Longident
+ open Parsetree
+ open Ast_helper
+ open Docstrings
+ open Docstrings.WithMenhir
++open Let_binding
```
hhugo, line 120: Is this diff (aka adding let_binding.ml) necessary ?

vendor/diff-parsers-std-ext.patch
gpetiot: Because this type is used in other places where it was not necessary for ocamlformat to have a location (and I wanted to minimize the diff). e.g. Ptyp_variant: the open case is `[> `A|`B ]` where there can't be a comment between `[` and `>` anyway.

vendor/diff-parsers-upstream-std.patch
gpetiot: I have opened https://github.com/ocaml/ocaml/pull/10618

vendor/diff-parsers-ext-parsewyc.patch
gpetiot: Yes it's something I tried to fix in the past, but I get an error in the generated `parser.mli` file: `Unbound type constructor let_bindings`, even with the same flags in the dune file.

hhugo (approved):
looks good to me

PR #1784, opened 2021-08-30
Attach location to the underscore of an open record pattern

gpetiot:
Part of the effort to remove the "guessing the loc" code from ocamlformat and leverage more data from the parser.
No behavioral change.

### DIFF
--- a/vendor/ocaml-4.13-extended/parser.mly
+++ b/vendor/ocaml-4.13-extended/parser.mly
@@ -1094,8 +1094,8 @@ reversed_bar_llist(X):
 listx(delimiter, X, Y):
 | x = X ioption(delimiter)
     { [x], None }
-| x = X delimiter Y delimiter?
-    { [x], Some (make_loc $loc($3)) }
+| x = X delimiter y = mkloc(Y) delimiter?
+    { [x], Some y }
 | x = X
   delimiter
   tail = listx(delimiter, X, Y)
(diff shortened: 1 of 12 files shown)


vendor/ocaml-4.13-extended/parser.mly
```diff
@@ -1094,8 +1094,8 @@ reversed_bar_llist(X):
 listx(delimiter, X, Y):
 | x = X ioption(delimiter)
     { [x], None }
-| x = X delimiter y = Y delimiter?
-    { [x], Some y }
+| x = X delimiter Y delimiter?
+    { [x], Some (make_loc $loc($3)) }
```
Julow, line 1098: We can keep `y` and reduce the diff:
```suggestion
| x = X delimiter y = Y delimiter?
    { ignore y; [x], Some (make_loc $loc(y)) }
```

Julow (approved):
Looks good !

vendor/ocaml-4.13-extended/parser.mly
gpetiot: I chose a slightly different fix to reduce the diff:
```diff
- | x = X delimiter y = Y delimiter?
+ | x = X delimiter y = mkloc(Y) delimiter?
    { [x], Some y }
```

Julow: Nice :)

vendor/parse-wyc/lib/parser.mly
rushduhhp, line 1: Third party

rushduhhp (changes requested):

PR #1793, opened 2021-09-06
Deprecate --align-* options

gpetiot:
Deprecate options `--align-cases`, `--align-constructors-decl` and `--align-variants-decl` for ocamlformat.0.20.0

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -185,24 +185,29 @@ let ocaml_version_conv =
 module Formatting = struct
   let section = `Formatting
 
+  let no_align = "Vertical alignment is no longer a promoted behavior."
+
   let align_cases =
     let doc = "Align match/try cases vertically." in
     let names = ["align-cases"] in
-    C.flag ~default:false ~names ~doc ~section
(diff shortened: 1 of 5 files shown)


lib/Conf.ml
```diff
@@ -185,24 +185,29 @@ let ocaml_version_conv =
 module Formatting = struct
   let section = `Formatting
 
+  let no_align = "Vertical alignment is no longer a promoted behavior."
```
Julow, line 188: I think this would sound better:
```suggestion
  let no_align = "Vertical alignment is discouraged."
```

Julow (approved):

lib/Conf.ml
gpetiot: I'm not a big fan of discouraged, what do you think of inadvisable?

jberdine: Is it necessary to add any such sentence? It seems enough to state that it is deprecated and will be removed by version 1.0.

gpetiot: indeed, fixed!

PR #1797, opened 2021-09-09
Fix alignment of comments inside a tuple pattern and remove incorrect linebreak, and fix formatting of labelled arguments containing comments

### LINKED ISSUE #1656: Bug: blank line with trailing whitespace (before multiline comment, inside a tuple pattern)
The following code is obtained after reformatting using the default options of ocamlformat 0.15.0:

```
$ ocamlformat --enable-outside-detected-project repro.ml
let () =
  match x with
  | ( _,
      
    (* line 1
       line 2 *)
    Some _ ) ->
      ()
```

It contains a blank line containing trailing spaces, which is an issue for other linters that try to eliminate trailing whitespace:

image

This is the simplest example I could come up with that produces the blank line. In particular, the problem disappears if `Some _` is replaced by something else, including `None` and `(Some _, Some _)`.

Workaround
--

Move the comment as follows:
```
let a =
  match x with
  | ( _,
      Some _
    (* line 1
       line 2 *) ) ->
      0
```

gpetiot:
Fix #1656, no diff with test_branch.sh


gpetiot:
Any opinion on this one?

test/passing/tests/label_option_default_args.ml.ref
```diff
@@ -68,10 +68,10 @@ let (* 0 *) f
...
 
 let (* 0 *) f
-    ?l:(
-      (* 1 *)
-      (* 2 *)
-      (* 3 *) C (* 4 *) x (* 5 *) ) =
+    ?l:((* 1 *)
+        (* 2 *)
+        (* 3 *)
+      C (* 4 *) x (* 5 *) ) =
```
Julow, line 74: Missing an indentation here ?

gpetiot: This is not directly due to this PR (it just made it visible), I commited a fix in this PR but it can also be extracted in another PR

Julow (approved):
Let's merge :)

test/passing/tests/label_option_default_args.ml.ref
Julow: The fix is small and related so we can keep them together.

mjambon:
thank you!

PR #1798, opened 2021-09-09
Take the 'in' location into consideration to decide whether there is a blank line after a let-binding

### LINKED ISSUE #1796: Bug: empty line introduced while formatting
Format the following file with default options (tested with `main` on Unix). The empty line before `in` is removed but a new empty line is inserted after `in`.
```ocaml
let _ =
  let collateralized_contracts_mtas =
    Cva_common.
      [
        (swap_1_c, { minimum_transfert_amount = 0. });
        (swap_2_c, { minimum_transfert_amount = 0. });
        (swap_3_c, { minimum_transfert_amount = 0. });
      ]
    
  in
  ()
```

gpetiot:
Fix #1796, no diff with test_branch.sh

The issue lies with the option `sequence-blank-line=preserve-one` that didn't take the location of `in` into consideration, so the linebreak before `in` (that should not be preserved) was confused with a linebreak after `in` (that should be preserved when this option is set)

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -4274,9 +4274,19 @@ and fmt_let c ctx ~ext ~rec_flag ~bindings ~parens ~fmt_atrs ~fmt_expr ~loc
   in
   let blank_line_after_in =
     let last_bind = List.last_exn bindings in
-    (* The location of the first binding (just after `let`) is wrong, it
-       contains the whole letop expression *)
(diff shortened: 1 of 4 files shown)


Julow (commented):
To me, this was intended. It's using the location of the end of the body to detect an empty line before the `in` too.

I'd expect code like this to exist:

```ocaml
  let x =
    y

  in z
```

and I think we should preserve that empty line too.

PR #1801, opened 2021-09-10
Add parentheses around type alias used as type constraint

### LINKED ISSUE #1800: Bug: generating invalid ocaml syntax
**Describe the bug**
```
echo "fun (x : int as 'a) -> (x : int as 'a)" | ocamlformat - --name a.ml
ocamlformat: Cannot process "a.ml".
```
This bug was found while testing ocamlformat on the ocaml codebase

gpetiot:
Fix #1800, no diff with test_branch.sh

Parentheses are also added in case of let-binding constraint, where it's not necessary, but it's more consistent with the `fun` case. To distinguish between the two would require to pass the context through several functions and is not as easy to maintain.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,7 +28,7 @@
 
   + Fix excessive break and wrong indentation after a short-open when `indicate-multiline-delimiters=closing-on-separate-line` (#1786, @gpetiot)
 
-  + Add parentheses around typa alias used as type constraint (#1801, @gpetiot)
+  + Add parentheses around type alias used as type constraint (#1801, @gpetiot)
 
 #### Changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,8 @@
 
   + Fix excessive break and wrong indentation after a short-open when `indicate-multiline-delimiters=closing-on-separate-line` (#1786, @gpetiot)
 
+  + Add parentheses around typa alias used as type constraint (#1801, @gpetiot)
+
 #### Changes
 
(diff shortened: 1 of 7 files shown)


jberdine:
Please give me a bit of time to evaluate the change in the `let` case. It's not obviously what we want.

Julow (commented):
Perhaps we should preserve the original syntax in this case ?
```ocaml
 fun (x : long_type) ->
  (x : long_type)
```
formats to
```ocaml
 fun (x : long_type) :
     long_type ->
  x
```
which is less readable to me.

gpetiot:
I fixed the regression on let-bindings, instead of adding a case in `Ast.pazenze_typ` I'm re-using some existing variable (`in_constraint` that is renamed according to what it really does, it was true if and only if the context was a `fun` anyway).
I'm wondering if I should merge this variable with `in_type_declaration` which basically does the same thing (provide a context of usage, instead of the context provided by the AST).

I've also considered improving the AST so that the type constraint information is carried on the right node (so we can get a meaningful context), but the change would be too involved (and I can already feel the rejection of the PR upstream).

jberdine:
When I looked into this last, it was important to the type-checker that the constraints were where they were in the parsetree. So without changes to type inference, upstreaming such a change would cause regressions in the compiler. Probably only worth the churn and effort if done as part of some larger endeavor.

Julow (approved):
Looks good :)

I wouldn't be in favor of huge change to the parsetree because that would make upgrades much harder.

PR #1803, opened 2021-09-13
Deprecate 'compact' and 'sparse' profiles

gpetiot:


Julow (approved):
Nice !

jberdine:
A higher level point than about this PR in particular, but I'm not sure that it is enough to deprecate things just by changing the docs. If we want people to notice and provide feedback in the time between deprecation and removal, I think that issuing a warning when running with a deprecated configuration is necessary.

tmattio:
Is there a reason for deprecating these? I'm using the `sparse` profile in lots of projects and find it nicer than the default one. I'm happy to use a custom configuration instead, but I'm curious what motivated the change?

Also, a suggestion: maybe some documentation with the configuration that mimics the deprecated profiles would help people who use these profiles to migrate.

tmattio:
Thanks for the detailed context @gpetiot 🙂

PR #1804, opened 2021-09-13
Deprecate option 'wrap-comments'

gpetiot:
This option prevents people from organizing their comments as they want, so most people don't use it.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -2,5 +2,4 @@ profile = ocamlformat
 break-cases = fit
 margin = 77
 parse-docstrings = true
-wrap-comments = true
 line-endings = lf
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -1147,7 +1147,8 @@ module Formatting = struct
        wrapped. Consecutive comments with both left and right margin \
        aligned are not wrapped either."
     in
-    C.flag ~default:false ~names:["wrap-comments"] ~doc ~section
+    let deprecated = C.deprecated ~since_version:"0.20.0" removed_by_v1_0 in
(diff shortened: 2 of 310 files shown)


jberdine:
Internally we use this (I think) exclusively. I do not like deprecating it, at least not before something like an ability to format comments as docstrings is supported well.

Julow (commented):
Are we sure nobody use it ?
I do use it pretty often from my editor (when formatting a single paragraph instead of the whole file).

jberdine:
Also e.g. https://github.com/ocaml-flambda/flambda-backend/blob/e5b3bdd1ee51b33dc90b9bce5d5bbe2774ec7de4/middle_end/flambda2/.ocamlformat#L13

Julow:
> Would it be fine to parse all comments as docstrings, and whether it is a valid docstring or not, wrap each paragraph of a comment, but not the whole text?

That sounds better than the current behavior to me. Docstrings are formatted that way already but we'd still need to fallback to what you suggested in case the parsing failed.

jberdine:
I'm not sure, but it seems worth trying. But, I think we probably still need to preserve banner-comments, at least changing that should be a separate step. Also, I think it would probably be better to add a comment form such as `(*= ... *)` to indicate a comment that should be parsed as a docstring, and warn/error if parsing fails. Once that is in place, the need for `wrap-comments` would probably go away and it could be removed. On the other hand, the current situation is somewhat suboptimal in that the docstring language is undocumented and seems to be not quite the best fit for our use and hard to change. From that perspective the current comment wrapping is nice.

PR #1805, opened 2021-09-13
Deprecate option 'disambiguate-non-breaking-match'

gpetiot:
This option is not set by any major project, and from Jules' script only transept.0.1.0 and preface.0.1.0 use it.

Maybe we can even remove it without deprecating it first?


Julow (approved):
Removing an option is always nice.

jberdine:
Just for tracking, this was added based on a request https://github.com/ocaml-ppx/ocamlformat/issues/423, but perhaps now most projects are breaking on all sequences and so the visual ambiguity does not arise.

PR #1807, opened 2021-09-15
Improve documentation of deprecated options and removed options

gpetiot:
- Avoid having a flag for `removed` and a flag for `deprecated`, introduce a `status`
- Introduce a `kind` of option (formatting/operational) instead of directly handling the doc section (since the section also depends on the status)
- Create a section for deprecated options in the command line manual
- Move the `--profile` documentation from the operational to the formatting section (makes more sense to me, but that one is arguable)

No behavioral change.
The next step is to do something similar for option values (that can be removed, but not deprecated for now).

### DIFF
--- a/lib/Config_option.ml
+++ b/lib/Config_option.ml
@@ -22,6 +22,8 @@ module Make (C : CONFIG) = struct
 
   type config = C.config
 
+  type kind = Formatting | Operational
+
   type parsed_from = [`File of Fpath.t * int | `Attribute]
 
   type updated_from = [`Env | `Commandline | `Parsed of parsed_from]
@@ -29,7 +31,11 @@ module Make (C : CONFIG) = struct
   type from =
     [`Default | `Profile of string * updated_from | `Updated of updated_from]
 
-  type deprecated = {msg: string; since_version: string}
+  type deprecated = {dmsg: string; dversion: string}
+
+  type removed = {rmsg: string; rversion: string}
+
+  type status = [`Valid | `Deprecated of deprecated | `Removed of removed]
 
   type 'a t =
     { names: string list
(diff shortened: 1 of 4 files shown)


lib/Config_option.ml
```diff
@@ -22,14 +22,20 @@ module Make (C : CONFIG) = struct
...
 
   type from =
     [`Default | `Profile of string * updated_from | `Updated of updated_from]
 
-  type deprecated = {msg: string; since_version: string}
+  type deprecated = {dmsg: string; dversion: string}
+
+  type removed = {rmsg: string; rversion: string}
+
+  type status = [`Valid | `Deprecated of deprecated | `Removed of removed]
```
Julow, line 38: I'm not sure about this compared to two values. Removed options were deprecated at some point and perhaps we can keep that information around ? Also, the code uses this as "is deprecated" and "is removed" and not both values at the same time.

Julow (approved):
The new doc looks good :)

lib/Config_option.ml
gpetiot: The way we use it is always: `valid --> deprecated --> removed` anyway, all deprecated options will be removed at some point, and all removed options have been deprecated first (unless removing an option is litterally a fix, but I'm not sure we had this for now). I don't see the value of being deprecated and removed at the same time, and it was difficult to maintain the code having 2 flags that are mutually exclusive.

gpetiot: But let me know if you see an easy improvement before I merge this one!

Julow: The "deprecated" flag contains since when it is deprecated and a message that we might want to keep when the option is removed. 

However, I agree having the two flags at the same time might make some part of the code harder. What do you think of `` `Removed of removed * deprecated option `` ?

gpetiot: `Config_option.removed` takes a string as parameter, just like `Config_option.deprecated`, so it's possible to just keep the same message and change the function called, or just tailor a new message. On the last batch of deprecated options the message is `It will be removed by version 1.0.` so it does not make sense to automatically preserve it (the `deprecated option` will be `None` far too often in my opinion)

PR #1811, opened 2021-09-17
[Indent] use ocp-indent when ocamlformat is deactivated

### LINKED ISSUE #1792: Bug: In emacs, indentation in project where ocamlformat is not activated
**Describe the bug**
Normally if ocamlformat is not activated in a project, it is transparent. For exemple a formatting save-hook in emacs will not break in a project where no `.ocamlformat` is present: no formatting is done. However since perhaps #1639 and #1687 indentation is not done if no `.ocamlformat` is present. Since it seems to take priority over `ocp-indent` hooks, the effect is to disable indentation.

**How to Reproduce**
- install only ocp-indent, user-setup
- indentation works in OCaml files in a project without any `.ocamlformat`
- install ocamlformat 0.19 and require it in emacs configuration

bobot:
Fixes #1792 . If ocamlformat is disable directly use ocp-indent for indentation instead of still trying to parse it.

But no tests have been added.

### DIFF
--- a/lib/Translation_unit.ml
+++ b/lib/Translation_unit.ml
@@ -534,12 +534,16 @@ let numeric (type a b) (fg : a list Extended_ast.t)
     | Ok parsed -> Ok parsed
     | Error _ -> parse_result recover fg conf ~source:src ~input_name
   in
-  match parse_or_recover ~src:source with
-  | Ok parsed -> (
-    match parse_result Std_ast.Parse.ast std_fg conf ~source ~input_name with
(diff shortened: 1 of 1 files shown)


gpetiot:
I think the `numeric` feature should never be called in emacs if ocamlformat is disabled.
Maybe ocamlformat will be the entry point for ocp-indent in the future but I'm not sure that's where we want to go, so I'm not fond of calling ocp-indent when ocamlformat is disabled.

I'm closing this one for now :)

PR #1812, opened 2021-09-17
Allow configuration options to have deprecated values 

gpetiot:
Here is what is produced when deprecating the value of some option:

```diff
diff --git a/ocamlformat-help.txt b/ocamlformat-help.actual
index 29694e95..abcc9a03 100644
--- a/ocamlformat-help.txt
+++ b/ocamlformat-help.actual
@@ -105,8 +105,9 @@ OPTIONS (CODE FORMATTING STYLE)
        --break-separators={after|before}
            Break before or after separators such as `;` in list or record
            expressions. after breaks the expressions after the separator.
-           before breaks the expressions before the separator. The default
-           value is after.
+           before breaks the expressions before the separator. Warning: Value
+           `before` is deprecated since version 0.20.0. Deprecation message.
+           The default value is after.

diff --git a/test/passing/tests/break_separators-before_docked_wrap.ml.err b/test/passing/break_separators-before_docked_wrap.ml.stderr
index 8cec0959..f2a39eff 100644
--- a/test/passing/tests/break_separators-before_docked_wrap.ml.err
+++ b/test/passing/break_separators-before_docked_wrap.ml.stderr
@@ -1 +1,2 @@
+Warning: Value `before` of option `after` is deprecated since version 0.20.0. Deprecation message.
 Warning: tests/break_separators.ml:212 exceeds the margin
```

A warning is printed in the ocamlformat documentation, and when using a deprecated value explicitly.

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -215,12 +215,12 @@ module Formatting = struct
     let doc = "Position of the assignment operator." in
     let names = ["assignment-operator"] in
     let all =
-      [ C.Value.valid ~name:"end-line" `End_line
+      [ C.Value.make ~name:"end-line" `End_line
           ~doc:
             "$(b,end-line) positions assignment operators (`:=` and `<-`) \
              at the end of the line and breaks after it if the whole \
              assignment expression does not fit on a single line."
-      ; C.Value.valid ~name:"begin-line" `Begin_line
+      ; C.Value.make ~name:"begin-line" `Begin_line
           ~doc:
             "$(b,begin-line) positions assignment operators (`:=` and `<-`) \
              at the beginning of the line and breaks before it if the whole \
@@ -237,12 +237,12 @@ module Formatting = struct
     in
     let names = ["break-before-in"] in
     let all =
-      [ C.Value.valid ~name:"fit-or-vertical" `Fit_or_vertical
+      [ C.Value.make ~name:"fit-or-vertical" `Fit_or_vertical
           ~doc:
             "$(b,fit-or-vertical) will always break the line before the \
              $(i,in) keyword if the whole $(i,let) binding does not fit on \
              a single line."
-      ; C.Value.valid ~name:"auto" `Auto
+      ; C.Value.make ~name:"auto" `Auto
           ~doc:
             "$(b,auto) will only break the line if the $(i,in) keyword does \
              not fit on the previous line." ]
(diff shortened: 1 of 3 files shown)


lib/Conf.ml
```diff
@@ -215,16 +215,16 @@ module Formatting = struct
...
-        , "$(b,begin-line) positions assignment operators (`:=` and `<-`) \
-           at the beginning of the line and breaks before it if the whole \
-           assignment expression does not fit on a single line." ) ]
+      [ C.Value.Ok.valid ~name:"end-line" `End_line
+          ~doc:
+            "$(b,end-line) positions assignment operators (`:=` and `<-`) \
+             at the end of the line and breaks after it if the whole \
+             assignment expression does not fit on a single line."
+      ; C.Value.Ok.valid ~name:"begin-line" `Begin_line
+          ~doc:
```
Julow, line 224: I think the labels are not necessary. Removing them might reduce the diff.

lib/Config_option.ml
```diff
@@ -188,46 +188,91 @@ module Make (C : CONFIG) = struct
...
-            (fun s -> Error (`Msg s))
-            "value `%s` has been removed in version %s. %s" name version msg
-      | None -> Arg.conv_parser conv s
-    in
-    Arg.conv (parse, Arg.conv_printer conv)
+  module Value = struct
+    module Ok = struct
+      type 'a t = string * 'a * string * [`Valid | `Deprecated of deprecated]
+
+      let valid ~name ~doc value = (name, value, doc, `Valid)
```
Julow, line 195: What about `make` rather than `valid` ? `deprecated` can be turned into `make_deprecated` too.

lib/Config_option.ml
```diff
@@ -188,46 +188,91 @@ module Make (C : CONFIG) = struct
...
-      match List.find values ~f:(fun {name; _} -> String.equal name s) with
-      | Some {name; version; msg} ->
-          Format.kasprintf
-            (fun s -> Error (`Msg s))
-            "value `%s` has been removed in version %s. %s" name version msg
-      | None -> Arg.conv_parser conv s
-    in
-    Arg.conv (parse, Arg.conv_printer conv)
+  module Value = struct
+    module Ok = struct
```
Julow, line 192: Instead of `Value.Ok` and `Value.Removed`, what about `Value` and `Value_removed` ? The extra level doesn't seem necessary and the name `Ok` feels weird.

lib/Config_option.ml
```diff
@@ -188,46 +188,91 @@ module Make (C : CONFIG) = struct
...
+    end
+  end
 
   let choice ~all ?(removed_values = []) ~names ~doc ~kind
-      ?(allow_inline = Poly.(kind = Formatting)) ?status =
-    let _, default, _ = List.hd_exn all in
-    let opt_names = List.map all ~f:(fun (x, y, _) -> (x, y)) in
-    let conv = add_removed_values removed_values (Arg.enum opt_names) in
+      ?(allow_inline = Poly.(kind = Formatting)) ?status update =
+    let name, default, _, _ = List.hd_exn all in
```
Julow, line 247:
```suggestion
    let _, default, _, _ = List.hd_exn all in
    let name = List.hd_exn names in
```
This is the name of the default value but we want the name of the option.

lib/Config_option.ml
```diff
@@ -188,46 +188,91 @@ module Make (C : CONFIG) = struct
...
+
+      let add_parse_errors values conv =
+        let parse s =
+          match
+            List.find values ~f:(fun {name; _} -> String.equal name s)
+          with
+          | Some {name; version; msg} ->
+              Format.kasprintf
+                (fun s -> Error (`Msg s))
+                "value `%s` has been removed in version %s. %s" name version
```
Julow, line 237: This doesn't say "of option '%s'" like the deprecated message.

lib/Config_option.ml
gpetiot: Ah yes, good catch!

lib/Config_option.ml
gpetiot: This is not necessary as the error message issued when parsing the value is preceded by the option name:
```
$ ocamlformat a.ml --break-string-literals=wrap
ocamlformat: option `--break-string-literals': value `wrap` has been removed
             in version 0.12. It has been replaced by the new default `auto`
             value, which breaks lines at newlines and wraps string literals
             at the margin.
```

lib/Conf.ml
gpetiot: done

lib/Config_option.ml
gpetiot: done

lib/Config_option.ml
gpetiot: done

lib/Config_option.ml
Julow: Indeed. It would be nice to use the same template for deprecation messages.

gpetiot: done:
```
Warning: option `break-string-literals`: value `never` is deprecated since version V. Deprecation message.
```

PR #1814, opened 2021-09-20
Emacs: only hook ocamlformat mode on tuareg/caml modes when ocamlformat is not disabled

### LINKED ISSUE #1792: Bug: In emacs, indentation in project where ocamlformat is not activated

gpetiot:
Trying to fix #1792, this should be improved to also take into account the value of `--root`. For now it just looks up the `.ocamlformat` file in the current directory, but it allows some basic testing to at least confirm if this is going in the right direction.
(This will be so much better once all this plumbing will be taken care of by dune)

cc @bobot let us know if it improves the situation

### DIFF
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -411,10 +411,23 @@ With ARG, perform this action that many times."
     t
     split-string-default-separators)))
 
-(add-hook 'tuareg-mode-hook 'ocamlformat-setup-indent t)
-(add-hook 'tuareg-mode-hook 'ocamlformat-set-newline-and-indent)
-(add-hook 'caml-mode-hook 'ocamlformat-caml-mode-setup t)
-(add-hook 'caml-mode-hook 'ocamlformat-set-newline-and-indent)
+(defun ocamlformat--add-hooks ()
+  "Link ocamlformat with tuareg-mode and caml-mode."
+  (progn
+    (add-hook 'tuareg-mode-hook 'ocamlformat-setup-indent t)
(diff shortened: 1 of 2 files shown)


emacs/ocamlformat.el
```diff
@@ -411,10 +411,23 @@ With ARG, perform this action that many times."
...
+
+(pcase ocamlformat-enable
+  ;; never hook
+  ('disable '())
+  ;; always hook
+  ('enable-outside-detected-project (ocamlformat--add-hooks))
+  ;; only hook if there is an .ocamlformat file at the root of the project
+  ('enable
+   (if (file-exists-p (concat default-directory ".ocamlformat"))
+    (ocamlformat--add-hooks))))
```
bobot, line 430: Is it not too restrictive? The `.ocamlformat` can be in a parent directory, no?

gpetiot: It should be fixed now thanks to vc-find-root, I made a few tests and it works for me.

PR #1815, opened 2021-09-20
Only sugarize RHS of sequences

### LINKED ISSUE #1787: AST of sequences is allowed to change
```ocaml
let _ = (A; B); C
```

The code above formats to the code below since https://github.com/ocaml-ppx/ocamlformat/pull/1731:

```ocaml
let _ = A ; B ; C
```

In the first case, the AST looks like:
```
  sequence (sequence (A, B), C)
```

and in the second case:
```
  sequence (A, sequence (B, C))
```

This change might be benign but it seems not to be made on purpose so I think it's worth talking about it.

gpetiot:
Fix #1787, better fix of #1728 and #1730.
No diff with test_branch.sh.
To not break the AST we should only sugarize the RHS of sequences, LHS should be parenthesized if it's also a sequence.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@
 
 #### Bug fixes
 
-  + Fix normalization of sequences of expressions (#1731, @gpetiot)
+  + Fix normalization of sequences of expressions (#1731, #1815, @gpetiot)
 
   + Type constrained patterns are now always parenthesized, parentheses were missing in a class context (#1734, @gpetiot)
 
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -2205,6 +2205,8 @@ end = struct
     in
     let exp_in_sequence lhs rhs exp =
       match (lhs.pexp_desc, exp.pexp_attributes) with
+      (* already parenthesized if attributes are attached *)
+      | Pexp_sequence _, [] when lhs == exp -> true
       | (Pexp_match _ | Pexp_try _), _ :: _ when lhs == exp -> true
       | _, _ :: _ -> false
       | ( Pexp_extension
--- a/lib/Normalize.ml
+++ b/lib/Normalize.ml
@@ -251,21 +251,11 @@ let make_mapper conf ~ignore_doc_comments =
   in
   let expr (m : Ast_mapper.mapper) exp =
     let exp = {exp with pexp_loc_stack= []} in
-    let {pexp_desc; pexp_loc= loc1; pexp_attributes= attrs1; _} = exp in
+    let {pexp_desc; _} = exp in
     match pexp_desc with
     | Pexp_poly ({pexp_desc= Pexp_constraint (e, t); _}, None) ->
         m.expr m {exp with pexp_desc= Pexp_poly (e, Some t)}
     | Pexp_constraint (e, {ptyp_desc= Ptyp_poly ([], _t); _}) -> m.expr m e
-    | Pexp_sequence
-        ( exp1
-        , { pexp_desc= Pexp_sequence (exp2, exp3)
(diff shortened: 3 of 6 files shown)


hhugo:
I believe is this one more instance where preserving too much from the original source go against the original design of ocamlformat. Should we have this kind of normalization under a flag ? and have the normalization run as an ast mapper, like proposed for dealing with begin-end ?
cc @jberdine

gpetiot:
My take is that if a user wrote `(A; B); C` instead of `A; B; C` then their intent is not the same, if a user wants a fully normalized sequence they would not have put parentheses in the first place (and I guess generated code would not put parentheses here as well, but this one is a wild guess).

Preserving this also reduces the amount of normalizing and sugaring code, and the code looks easier to maintain.

hhugo:
> My take is that if a user wrote `(A; B); C` instead of `A; B; C` then their intent is not the same, if a user wants a fully normalized sequence they would not have put parentheses in the first place (and I guess generated code would not put parentheses here as well, but this one is a wild guess).

One doesn't always put parents explicitly. Sometime, it is the result of moving code around.

> Preserving this also reduces the amount of normalizing and sugaring code, and the code looks easier to maintain.

I agree on this point. That's why I propose to move this out of the formatting logic.

Julow (approved):
> I believe is this one more instance where preserving too much from the original source go against the original design of ocamlformat

This PR doesn't preserve anything from the original source, right ? It's interpreting the AST in a more correct way.
The previous "desugaring of sequences" code incorrectly interpreted:

```ocaml
  Sequence (Sequence (a, b), Sequence (c, d))
```

as `[ a; b; c; d ]`, which gets formatted as if the input was:

```ocaml
  Sequence (a, Sequence (b, Sequence (c, d)))
```

Now it's reading `[ Sequence (a, b); c; d ]` and formats to the same AST.

gpetiot:
Exactly! I will wait @jberdine's opinion but I think this is an improvement

hhugo:
> This PR doesn't preserve anything from the original source, right ? It's interpreting the AST in a more correct way.
The previous "desugaring of sequences" code incorrectly interpreted:

This PR preserves some grouping of sequences that is currently normalized on purpose. The original goal of ocamlformat was to normalize code while keeping the same runtime semantic.
As far as I know, `(a;b);c` has the same semantic as `a;b;c`.

I don't like that we treat `(a;b);c` and `a;(b;c)` differently. If you want to preserve visual grouping, do it all the way, and preserve parens in the second case as well.

With the current proposal, you get some unfriendly (to me at least) interactive experience.
In `let () = a;(b;c);d`, if I decide de remove the expression d, parens will disappear around `(b;c)`

gpetiot:
> With the current proposal, you get some unfriendly (to me at least) interactive experience.
> In `let () = a;(b;c);d`, if I decide de remove the expression d, parens will disappear around `(b;c)`

That's a good point against this change indeed, I won't push too hard for this PR. I'm closing it, the current situation about sequences is pretty good right now, there is no urgent need for a change.

jberdine:
I'm late but agree with @hhugo. I'll overstate just for emphasis, but consider that the ideal for the auto-formatter is not to change only the whitespace in a file, but to resolve as many questions/disagreements about (static and dynamic) semantics-preserving changes between people working on a codebase in the config file for the formatter rather than in e.g. code review. The ideal would be for any nit-picky or cosmetic comment that could come up in code review like "you know, those parens are not necessary" or "you know, normally we use begin/end instead of parens in such cases instead" to be made in one single way by the formatter. Then those people who want can argue with each other over the config file, and leave code review to be as productive and effective as possible.

PR #1816, opened 2021-09-20
Fix documentation of option 'use-file', it should not be deprecated

gpetiot:
`--use-file` and `--impl` are not treated the same, contrary to what the documentation says, calling different functions from `Parse`, and also different formatting functions.
Maybe it was the case before we had rewritten the code to use fragment types, but not anymore.


Julow (commented):
`--impl` and `--use-file` are treated the same. They both means `Syntax.Use_file`. (confusingly, `--impl` doesn't mean `Syntax.Structure`)
This should be reverted.

gpetiot:
Oh indeed, I missed the most important part in Conf.ml, I will revert it then.

PR #1820, opened 2021-09-21
Use current-bench to benchmark 'format' and 'numeric' features

gpetiot:
Trying to use https://github.com/ocurrent/current-bench

A few things to fix but I think the architecture is right.


bench/dune
```diff
@@ -0,0 +1,20 @@
...
+(rule
+ (alias runbench)
+ (action
+  (run ./bench.exe)))
+
+(rule
+ (alias runtest)
+ (package ocamlformat-bench)
+ (deps bench.exe)
+ (action (progn)))
```
Julow, line 20: This rule is testing that the benchmark code compiles ?
Is it useful if the benchmarks are run in CI anyway ?

bench/dune
```diff
@@ -0,0 +1,20 @@
+(executable
+ (name bench)
+ (public_name bench)
+ (package ocamlformat-bench)
+ (libraries bechamel bechamel-js ocamlformat_lib stdio))
+
+(alias
+ (name bench)
+ (deps ./bench.exe))
```
Julow, line 9: What is this rule for ?

Makefile
```diff
@@ -43,3 +43,6 @@ coverage:
 headers:
 	tools/update_headers.sh
 	dune build @fmt --auto-promote
+
+bench:
+	@dune exec -- ./bench/bench.exe
```
Julow, line 48: Should be calling the corresponding dune rule:
```suggestion
	@dune build @runbench
```

Julow (commented):
Very interesting ! Will the results appear in the "CI" box on Github ?

bench/dune
Julow: I think that's actually a very good idea :)

gs0510:
Just need to add `ocaml-ppx` to list of users, will re-deploy once I do that :)

bench/dune
gpetiot: I admit I just picked it from `mirage/index` that is also using current-bench

bench/dune
gpetiot: Same, it came from mirage/index, but I removed it.

bench/dune
```diff
@@ -0,0 +1,16 @@
...
+(rule
+ (alias runbench)
+ (action
+  (run ./bench.exe)))
+
+(rule
+ (alias runtest)
+ (package ocamlformat-bench)
+ (deps bench.exe)
+ (action (progn)))
```
emillon, line 16: I think you can remove the action. That will just record `bench.exe` as a dependency to `@runtest`.

An alternative that I'm using sometimes is to check for `argv` in the executable, do nothing if there are no arguments, and run the bench otherwise. That way you can use a `(test)` stanza (running `dune runtest` will build the test but do nothing) and a `runbench` action that calls it with an argument.

bench/bench.ml
```diff
@@ -0,0 +1,83 @@
...
+type input =
+  { name: string
+  ; input_name: string
+  ; kind: Syntax.t
+  ; source: string
+  ; conf: Conf.t
+  ; action: [`Format | `Numeric of range] }
+
+let source_from_file f =
+  Stdio.In_channel.with_file f ~f:Stdio.In_channel.input_all
```
emillon, line 16: that's `In_channel.read_file`

bench/bench.ml
```diff
@@ -0,0 +1,83 @@
...
+
+let inputs =
+  let source_ml = source_from_file "test/passing/tests/source.ml" in
```
emillon, line 19: this means that your benchmark will move everytime you add things to that file. that can make it difficult to track what's going on across versions.

emillon: also is you read that file at runtime make sure to register it as a dependency otherwise the benchmark won't get reexecuted and things like `dune clean; dune build @runbench` won't work

gpetiot: done

gs0510:
When I run `make bench` locally and in the CI, both times this is the output:

```make: 'bench' is up to date.```

**Comment by emillon** [comment]
I think this can be fixed by adding `.PHONY: bench` before the rule. (otherwise the target name is interpreted as the directory name)

File: bench/dune [hunk 7]
```diff
@@ -0,0 +1,18 @@
+(executable
+ (name bench)
+ (public_name bench)
+ (package ocamlformat-bench)
+ (libraries bechamel bechamel-js ocamlformat_lib stdio))
+
+(rule
+ (alias runbench)
+ (package ocamlformat-bench)
+ (deps source_bench.ml)
```
jberdine, line 10: I don't suppose there is any way to write `../test/passing/tests/source.ml` here instead? Or a symlink? Or it would also help me appease the bots if test code like `bench_source.ml` was in a separate directory like `bench/test` or similar.

bench/bench.ml
```diff
@@ -0,0 +1,85 @@
...
+    List.map (fun instance -> Analyze.all ols instance raw_results) instances
+  in
+  let results = Analyze.merge ols instances results in
+  (results, raw_results)
+
+let nothing _ = Ok ()
+
+let () =
+  match Sys.argv with
+  | [| _; "run"; _; |] -> (
```
jberdine, line 76: It would help to have a comment somewhere explaining what is going on with testing for `run` and otherwise doing nothing. I don't understand why you do this.

jberdine (commented):
Also, please run `update_headers.sh`

bench/bench.ml
gpetiot: The idea was to do something similar to what @emillon suggested in one of his comments above, but this PR is still messy, I'm still trying to get something that is at least working and then I will clean it up.

bench/dune
gpetiot: done

gpetiot:
`make bench` now works properly and returns some json, I wish we could expand the error message we got on this page when the CI fails

gs0510:
@gpetiot `make bench` fails with this (Both on docker and my machine locally)

```
2021-10-01 17:58.04: Exec: "docker" "run" "--security-opt" "seccomp=./aslr_seccomp.json" 
                           "--mount" "type=volume,src=current-bench-data,dst=/home/opam/bench-dir/current-bench-data" 
                           "--tmpfs" "/dev/shm:rw,noexec,nosuid,size=4g" 
                           "--cpuset-cpus" "3" "--rm" "-i" "sha256:f1d48a7218abfaeb98d501a6feb373152c97cc7ff4275875c532c2b1520deeff" 
                           "/usr/bin/setarch" "x86_64" "--addr-no-randomize" 
                           "make" "bench"
       bench alias bench/runbench (exit 2)
(cd _build/default/bench && ../../install/default/bin/bench)
Fatal error: exception (Sys_error "test/source_bench.ml: No such file or directory")
Raised by primitive operation at Stdlib.open_in_gen in file "stdlib.ml", line 405, characters 28-54
Called from Stdio__In_channel.with_file in file "src/in_channel.ml", line 22, characters 45-66
Called from Stdio__In_channel.read_all in file "src/in_channel.ml" (inlined), line 91, characters 21-49
Called from Dune__exe__Bench.inputs in file "bench/bench.ml", line 16, characters 18-66
make: *** [Makefile:49: bench] Error 1
2021-10-01 17:58.47: Error: Command "docker" "run" "--security-opt" "seccomp=./aslr_seccomp.json" 
"--mount" "type=volume,src=current-bench-data,dst=/home/opam/bench-dir/current-bench-data" 
"--tmpfs" "/dev/shm:rw,noexec,nosuid,size=4g" "--cpuset-cpus" "3" "--rm" 
"-i" "sha256:f1d48a7218abfaeb98d501a6feb373152c97cc7ff4275875c532c2b1520deeff" 
"/usr/bin/setarch" "x86_64" "--addr-no-randomize" "make" "bench" exited with status 2

```

bench/bench.ml
```diff
@@ -0,0 +1,74 @@
...
+type input =
+  { name: string
+  ; input_name: string
+  ; kind: Syntax.t
+  ; source: string
+  ; conf: Conf.t
+  ; action: [`Format | `Numeric of range] }
+
+let inputs =
+  let source_ml = Stdio.In_channel.read_all "test/source_bench.ml" in
```
gs0510, line 16:
```suggestion
  let source_ml = Stdio.In_channel.read_all "test/source.ml" in
```

bench/test/dune
```diff
@@ -0,0 +1,7 @@
+(rule
+ (target source_bench.ml)
```
gs0510, line 2:
```suggestion
 (target bench.ml)
```

bench/bench.ml
gpetiot: The idea was to concatenate 5 times source.ml and make the result into source_bench.ml, dune should be able to generate source_bench.ml from the rules.

gs0510: Does it depend on a specific dune version? I can't seem to make it work even locally.

gs0510:
Because the ouptut is bechamel format and not current-bench format, the UI doesn't display the benchmarks. We already have a library in current-bench to convert bechamel format to current-bench format, working on a fix rn!

PR #1824, opened 2021-09-21
Set option 'enable-outside-detected-project' for test_branch.sh

gpetiot:
I got a lot of errors while running `test_branch.sh` because this option was not set, but what's funny is that it didn't appear after I updated my `test-extra` directory, just out of nowhere.
Maybe someone could double check before I merge.


Julow (approved):
I can't explain why you found these errors based on the state of test-extra I had locally.
Anyway, we want to keep the flag.

PR #1825, opened 2021-09-22
Add missing parentheses around application of prefix op when used as operator

### LINKED ISSUE #1823: Bug: AST changed with prefix operator
**Describe the bug**
```
let z = ((!) ~x:4)1 2 ~c:3
```
produces
```
  BUG: ast changed.
```

**How to Reproduce**
Style the above code with ocamlformat 0.19.0 and `--profile=janestreet`.

gpetiot:
Fix #1823, no diff with test_branch.sh

Couldn't find a way to better factorize with other cases in Ast.ml without hindering the readability. I wish we could find a better way of dealing with the parentheses one day.


lib/Ast.ml
```diff
@@ -2268,6 +2268,13 @@ end = struct
       , {pexp_desc= Pexp_construct _ | Pexp_variant _; _} )
       when e == exp ->
         true
+    | ( Exp {pexp_desc= Pexp_apply (e, _ :: _); _}
+      , {pexp_desc= Pexp_apply (op1, args); pexp_attributes; _} )
+      when e == exp && Exp.is_prefix op1 -> (
+      match (args, pexp_attributes) with
+      | _, _ :: _ -> true
+      | [(Nolabel, _)], _ -> false
```
jberdine, line 2276: Am I right that this case is not explicitly covered by the new test cases?

jberdine (commented):
This reminds me of the bug that was fixed by the revert commit https://github.com/ocaml-ppx/ocamlformat/commit/e976bc89ab9b37f34cfd5afcb06ae8b72aa7a572 . I wonder if in this current case, it would be easier to adjust `prec_ctx` instead of `parenze_exp`. Have you tried that approach? I'm not saying that the current change isn't ok, but it might be simpler to change `prec_ctx`. Also, the condition you have involving presence of attributes or labeled arguments seems like it would be also sufficient for the case of normal applications. That is, currently we leave `(f x) y` alone, but it would be better to normalize it to `f x y`. The commit https://github.com/ocaml-ppx/ocamlformat/commit/b3b201b97b9c17fefeb6b346592a7e7cfc880eb6 (that https://github.com/ocaml-ppx/ocamlformat/commit/e976bc89ab9b37f34cfd5afcb06ae8b72aa7a572 reverted) did that, but there were type-checking issues in the presence of labeled arguments. It would be a nice improvement if this bug and redoing https://github.com/ocaml-ppx/ocamlformat/commit/b3b201b97b9c17fefeb6b346592a7e7cfc880eb6 in a non-buggy way could be solved together, without special casing on whether a prefix op is involved or not. But perhaps I have not fully considered all the cases.

lib/Ast.ml
gpetiot: Not in the new tests but it's covered by apply.ml, source.ml and js_source.ml

gpetiot:
I tried this approach but I couldn't manage to make it work, the code of prec_ctx is also a nightmare to maintain, I wish we could move to a simpler model that would be more in favor or disambiguating even when not necessary by having unnecessary parentheses.

lib/Ast.ml
```diff
@@ -2268,6 +2268,13 @@ end = struct
...
+      | _ -> true )
```
Julow, line 2277: This is equivalent to

```suggestion
      | [(Nolabel, _)], [] -> false
      | _ -> true )
```

Julow (approved):

PR #1833, opened 2021-09-30
Bring vendored Format lib up-to-date with upstream stdlib and track differences

gpetiot:
- rename `diff_parsers` to `diff_ocaml`
- adapt the script to diff format.ml/format.mli with the files upstream
- commit the diff
- modify the vendored library to reduce the diff
- no behavioral change


hhugo:
What is the diff between the old vendored format and the new one ?

gpetiot:
Here is the diff:

<details>

```diff
diff --git a/vendor/ocamlformat_support/format_.ml b/vendor/ocamlformat_support/format.ml
index 1a6a4ed3..de6243f9 100644
--- a/vendor/ocamlformat_support/format_.ml
+++ b/vendor/ocamlformat_support/format.ml
@@ -42,6 +42,8 @@ end  = struct
   let is_known n = n >= 0
 end
 
+
+
 (* The pretty-printing boxes definition:
    a pretty-printing box is either
    - hbox: horizontal box (no line splitting)
@@ -685,6 +687,8 @@ let pp_print_as state isize s =
 let pp_print_string state s =
   pp_print_as state (String.length s) s
 
+let pp_print_bytes state s =
+  pp_print_as state (Bytes.length s) (Bytes.to_string s)
 
 (* To format an integer. *)
 let pp_print_int state i = pp_print_string state (Int.to_string i)
@@ -736,7 +740,7 @@ let pp_print_if_newline state () =
       { size = Size.zero; token = Pp_if_newline; length = 0 }
 
 
-(* Generalized break hint that allows to print strings before/after
+(* Generalized break hint that allows printing strings before/after
    same-line offset (width) or new-line offset *)
 let pp_print_custom_break state ~fits ~breaks =
   let before, width, after = fits in
@@ -893,29 +897,48 @@ let pp_set_margin state n =
 (** Geometry functions and types *)
 type geometry = { max_indent:int; margin: int}
 
+let validate_geometry {margin; max_indent} =
+  if max_indent < 2 then
+    Error "max_indent < 2"
+  else if margin <= max_indent then
+    Error "margin <= max_indent"
+  else Ok ()
+
 let check_geometry geometry =
-  geometry.max_indent > 1
-  &&  geometry.margin > geometry.max_indent
+  match validate_geometry geometry with
+  | Ok () -> true
+  | Error _ -> false
 
 let pp_get_margin state () = state.pp_margin
 
+let pp_set_full_geometry state {margin; max_indent} =
+  pp_set_margin state margin;
+  pp_set_max_indent state max_indent;
+  ()
+
 let pp_set_geometry state ~max_indent ~margin =
-  if max_indent < 2 then
-    raise (Invalid_argument "Format.pp_set_geometry: max_indent < 2")
-  else if margin <= max_indent then
-      raise (Invalid_argument "Format.pp_set_geometry: margin <= max_indent")
-  else
-    pp_set_margin state margin; pp_set_max_indent state max_indent
+  let geometry = { max_indent; margin } in
+  match validate_geometry geometry with
+  | Error msg ->
+    raise (Invalid_argument ("Format.pp_set_geometry: " ^ msg))
+  | Ok () ->
+    pp_set_full_geometry state geometry
 
 let pp_safe_set_geometry state ~max_indent ~margin =
-  if check_geometry {max_indent;margin} then
-    pp_set_geometry state ~max_indent ~margin
-  else
-    ()
+  let geometry = { max_indent; margin } in
+  match validate_geometry geometry with
+  | Error _msg ->
+     ()
+  | Ok () ->
+    pp_set_full_geometry state geometry
 
 let pp_get_geometry state () =
   { margin = pp_get_margin state (); max_indent = pp_get_max_indent state () }
 
+let pp_update_geometry state update =
+  let geometry = pp_get_geometry state () in
+  pp_set_full_geometry state (update geometry)
+
 (* Setting a formatter basic output functions. *)
 let pp_set_formatter_out_functions state {
       out_string = f;
@@ -1173,6 +1196,7 @@ and open_stag = pp_open_stag std_formatter
 and close_stag = pp_close_stag std_formatter
 and print_as = pp_print_as std_formatter
 and print_string = pp_print_string std_formatter
+and print_bytes = pp_print_bytes std_formatter
 and print_int = pp_print_int std_formatter
 and print_float = pp_print_float std_formatter
 and print_char = pp_print_char std_formatter
@@ -1201,6 +1225,7 @@ and get_max_indent = pp_get_max_indent std_formatter
 and set_geometry = pp_set_geometry std_formatter
 and safe_set_geometry = pp_safe_set_geometry std_formatter
 and get_geometry = pp_get_geometry std_formatter
+and update_geometry = pp_update_geometry std_formatter
 
 and set_max_boxes = pp_set_max_boxes std_formatter
 and get_max_boxes = pp_get_max_boxes std_formatter
@@ -1249,6 +1274,22 @@ let rec pp_print_list ?(pp_sep = pp_print_cut) pp_v ppf = function
     pp_sep ppf ();
     pp_print_list ~pp_sep pp_v ppf vs
 
+(* To format a sequence *)
+let rec pp_print_seq_in ~pp_sep pp_v ppf seq =
+  match seq () with
+  | Seq.Nil -> ()
+  | Seq.Cons (v, seq) ->
+    pp_sep ppf ();
+    pp_v ppf v;
+    pp_print_seq_in ~pp_sep pp_v ppf seq
+
+let pp_print_seq ?(pp_sep = pp_print_cut) pp_v ppf seq =
+  match seq () with
+  | Seq.Nil -> ()
+  | Seq.Cons (v, seq) ->
+    pp_v ppf v;
+    pp_print_seq_in ~pp_sep pp_v ppf seq
+
 (* To format free-flowing text *)
 let pp_print_text ppf s =
   let len = String.length s in
@@ -1279,6 +1320,10 @@ let pp_print_result ~ok ~error ppf = function
 | Ok v -> ok ppf v
 | Error e -> error ppf e
 
+let pp_print_either ~left ~right ppf = function
+| Either.Left l -> left ppf l
+| Either.Right r -> right ppf r
+
  (**************************************************************)
 
 let compute_tag output tag_acc =
@@ -1394,10 +1439,20 @@ let kfprintf k ppf (Format (fmt, _)) =
 and ikfprintf k ppf (Format (fmt, _)) =
   make_iprintf k ppf fmt
 
+let ifprintf _ppf (Format (fmt, _)) =
+  make_iprintf ignore () fmt
+
 let fprintf ppf = kfprintf ignore ppf
 let printf fmt = fprintf std_formatter fmt
 let eprintf fmt = fprintf err_formatter fmt
 
+let kdprintf k (Format (fmt, _)) =
+  make_printf
+    (fun acc -> k (fun ppf -> output_acc ppf acc))
+    End_of_acc fmt
+
+let dprintf fmt = kdprintf (fun i -> i) fmt
+
 let ksprintf k (Format (fmt, _)) =
   let b = pp_make_buffer () in
   let ppf = formatter_of_buffer b in
```

</details>

tools/diff-ocaml/diff_ocaml.ml
```diff
@@ -42,28 +41,38 @@ let files =
...
+    "usage:\n\n\
+     %s diff parser DIR1 DIR2\n\n\
+     %s diff stdlib DIR1 DIR2\n\n\
+     %s import VERSION DIR\n"
+    exe exe exe ;
   exit 1
 
+let diff files d1 d2 =
+  let codes = List.map (diff d1 d2) files in
+  let non_zero = List.filter (function 0 | 1 -> false | _ -> true) codes in
```
hhugo, line 58: We should rename non_zero.

hhugo (commented):
Why is it ok to rename format_.ml into format.ml ?

gpetiot:
> Why is it ok to rename format_.ml into format.ml ?

I just renamed because we need to match the names of the stdlib files.

hhugo:
> > Why is it ok to rename format_.ml into format.ml ?
> 
> I just renamed because we need to match the names of the stdlib files.

The codebase is referring to `Format_`. I don't understand how you change still pass the ci ..

jberdine:
Are you done changing the vendored format.ml to reduce the diff? It looks to me like it could be reduced more.

gpetiot:
> Are you done changing the vendored format.ml to reduce the diff? It looks to me like it could be reduced more.

The diff should be minimal now.

> The codebase is referring to `Format_`. I don't understand how you change still pass the ci ..

This is still the name of the library, and the entry point of the library (format_.ml) just includes format.ml.

gpetiot:
Once this one is merged I will have another look at the suggestions of the reviewers of https://github.com/ocaml/ocaml/pull/1229

jberdine:
I will have a look at this soon. It would be great to get back to the upstream PR. I fell behind on that too. One thing I was thinking was to try to get a concrete and focused understanding of Vladimir's "why can't it be simpler like this" sort of suggestions by implementing ocamlformat using them and seeing what happens, and reducing down to minimal examples. That would help either us to have a simpler implementation or a better explanation of the complication for upstream.

vendor/diff-stdlib-format.patch
```diff
@@ -0,0 +1,222 @@
...
+ and stag = ..
+ 
+@@@@
+   format_string state before;
+   pp_output_newline state;
+   state.pp_is_new_line <- true;
+   let indent = state.pp_margin - width + offset in
+   (* Don't indent more than pp_max_indent. *)
+-  let real_indent = Int.min state.pp_max_indent indent in
++  let real_indent = min state.pp_max_indent indent in
```
jberdine, line 31: This change does not seem to be needed.

vendor/diff-stdlib-format.patch
```diff
@@ -0,0 +1,222 @@
...
+    a cut is a break hint that prints nothing if the break does not split the
+    line. *)
+@@@@
+       if state.pp_max_indent <= state.pp_margin
+       then state.pp_max_indent else
+       (* If possible maintain pp_min_space_left to its actual value,
+          if this leads to a too small max_indent, take half of the
+          new margin, if it is greater than 1. *)
+-       Int.max (Int.max (state.pp_margin - state.pp_min_space_left)
++       max (max (state.pp_margin - state.pp_min_space_left)
```
jberdine, line 188: This change does not seem to be needed.

vendor/diff-stdlib-format.patch
```diff
@@ -0,0 +1,222 @@
...
+       | Pp_hbox -> break_same_line state fits
+       end
+     end
+ 
++  | Pp_or_newline (n, off, fits, breaks) ->
++    if state.pp_is_new_line
++    then format_string state breaks
++    else
++      let fits = ("", n, fits) and breaks = ("", off, breaks) in
++      format_pp_token state size (Pp_break {fits; breaks})
```
jberdine, line 76: Maybe it is better as a diff-reduction approach, but I would add a comment that when upstreaming this, it would still be better to make the `Pp_break` case into a function and also call it here instead of making `format_pp_token` recursive.

jberdine (commented):
Looks good, only a few minor comments.

I wonder if this PR is a good opportunity to rename `Format_` to e.g. `Format_ext` or to otherwise manage having both the stdlib `Format` and our extended `Format_` modules around at the same time. Did the work on vendoring the parser reveal a better way to manage this sort of thing?

vendor/diff-stdlib-format.patch
gpetiot: That's because min and max are only available since 4.13: https://github.com/ocaml/ocaml/blob/trunk/stdlib/int.mli

vendor/diff-stdlib-format.patch
gpetiot: I will revert this change to keep the function instead, that will be clearer and avoid missing it when upstreaming

vendor/diff-stdlib-format.patch
gpetiot: done, we had a small lib `parser_shims` redefining these for the parser files already.

gpetiot:
> I wonder if this PR is a good opportunity to rename `Format_` to e.g. `Format_ext` or to otherwise manage having both the stdlib `Format` and our extended `Format_` modules around at the same time. Did the work on vendoring the parser reveal a better way to manage this sort of thing?

I don't feel strongly either way about a name change.

We currently already have both Format modules around at the same time since they have a different names. If we manage to upstream our changes it will make sense to move these files to the `ocaml-common` vendored lib (that just shadows files from stdlib), but the current architecture looks good for now.

jberdine:
Just shadowing files probably won't work though due to using other libraries that use the `Stdlib.Format.formatter` type, like `Fpath.pp`.

PR #1834, opened 2021-09-30
Handle let operator punning uniformly with other punning forms

jberdine:
Other sorts of punning, such as record fields, normalize to the punned
form. For example `{x = x}` normalizes to `{x}`. This patch brings let
operator punning in line by normalizing e.g. `let* x = x` to `let* x`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,9 @@
 
   + Allow disambiguated global identifiers (like t/2) so they can be formatted by tools like OCaml-LSP (#1716, @let-def)
 
+  + Handle let operator punning uniformly with other punning forms.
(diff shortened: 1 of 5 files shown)


hhugo:
I'll bring back the topic of having ocamlformat work with older compilers. This will make ocamlformat unusable for many codebases.

jberdine:
In the diff I check `Ocaml_version.(compare c.conf.ocaml_version Releases.v4_13 >= 0)`. Is that not enough? In my manual testing passing `--ocaml-version 4.12.0` did prevent the punned form from being generated.

hhugo:
Oh ok, I didn't know / did not remember we implemented that :)

gpetiot (approved):
Looks good to me, thanks!
Can you add a changelog entry before merging?

gpetiot (approved):

PR #1839, opened 2021-10-02
Update README with git blame ignoreRevs example

rbjorklin:
Provide a basic example of how to make OCamlFormat code changes less impactful on `git blame` behavior.

Rendered here: https://github.com/rbjorklin/ocamlformat/tree/docs-ignore-revs

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Moreover if you adopt OCamlFormat in one project it will not break your workflow
 2ceaf76b9f84cb632327c1479d0f30acfa3eeba2
 ```
 * Run `git config --local blame.ignoreRevsFile .git-blame-ignore-revs`
-* All future use of `git blame` will now provide the relevant change message
+* All future use of `git blame` will now provide blame information omitting the reformatting commits.
 
 ### What configuration should I use?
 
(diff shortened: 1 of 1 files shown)


rbjorklin:
Signed the CLA.

Screenshot 2021-10-02 at 11-38-49 Contributing to Facebook Projects Facebook Code

README.md
```diff
@@ -49,6 +49,17 @@ a later release.
...
+### Using OCamlFormat without breaking `git blame`
+
+* Create `.git-blame-ignore-revs` in your project
+* Add revision(s) where the code was re-formatted
+```
+# Apply new formatting with OCamlFormat
+2ceaf76b9f84cb632327c1479d0f30acfa3eeba2
+```
+* Run `git config --local blame.ignoreRevsFile .git-blame-ignore-revs`
+* All future use of `git blame` will now provide the relevant change message
```
jberdine, line 61: For this line in particular, I wonder if it is as explicitly informative as possible. It seems that maybe something more explicit could be used instead of "provide" and/or "relevant". What happens is that git blame will omit the reformatting commits, and automatically blame further back into the history, correct? Is there a standard oneish-line explanation of such behavior?

jberdine (commented):
Thanks! I have a question inline.

README.md
gpetiot: I slightly rephrased: `provide blame information omitting the reformatting commits`, if that's alright for you we can merge, that's a great addition to the documentation!

Julow: The man for `git-blame` says:

> Ignore changes made by the revision when assigning blame, as if the change never happened. Lines that were changed or added by an ignored commit will be blamed on the previous commit that changed that line or nearby lines.

We could repeat the second sentence ?

PR #1845, opened 2021-10-05
Emit an error on line directives

### LINKED ISSUE #563: Line directives are silently dropped

gpetiot:
Fix #563, we could also emit a warning instead of an error although it would be easier to miss it.


hhugo:
How does that error interact with the `quiet` option ?

vendor/diff-parsers-upstream-std.patch
```diff
@@ -132,6 +132,36 @@
...
+   | ([' ' '\t']* (['0'-'9']+ as num) [' ' '\t']*
+-        ("\"" ([^ '\010' '\013' '\"' ] * as name) "\"") as directive)
++        ("\"" ([^ '\010' '\013' '\"' ] * as _name) "\"") as directive)
+         [^ '\010' '\013'] *
+       {
+         match int_of_string num with
+         | exception _ ->
+             (* PR#7165 *)
+             let explanation = "line number out of range" in
+             error lexbuf (Invalid_directive ("#" ^ directive, Some explanation))
```
Julow, line 149: Why try to decode the line number at all ? The whole match can be removed.

Julow (approved):
Looks good.

Handling these directives might be as hard as handling cppo directives.

vendor/diff-parsers-upstream-std.patch
gpetiot: done

PR #1837, opened 2021-10-01
Fix location of the let-binding part of a let-op

gpetiot:
First draft at fixing #1835, there are some regressions, comments moving when running test_branch.sh, maybe this location was not the only one wrong and they somehow cancelled each others.

PR #1850, opened 2021-10-07
Revert "Fix locations of the letop-bindings"

gpetiot:
Reverts ocaml-ppx/ocamlformat#1837

For context, apparently these locations were like that on purpose so our implementation in Fmt_ast.ml was correct, I'm reverting #1837 and removing the comment stating the locations were wrong.


gpetiot:
The comment attachment is a good point, maybe this change rightfully belongs to the extended parser and parse-wyc.

jberdine:
Another option to consider is, if all the needed info for the locs is in the standard parsetree, but in the "wrong" place, then it might be easier to maintain for ocamlformat if the locs are adjusted in a pass over the parsetree after parsing but before comment attachment, rather than modifying the parser. I'm not sure if the needed info is there but hidden though.

PR #1855, opened 2021-10-08
Apply option 'field-space' to values in signatures for more consistency

### LINKED ISSUE #1802: `field-space=tight-decl` should apply to signature items as well
The option `field-space=tight-decl` should also apply to signature items, which are very similar to record fields, eg:

```
val foo: ty
```

instead of

```
val foo : ty
```

gpetiot:
Fix #1802

It makes sense to apply this option to values in signatures since it's similar to types of record fields. There is no diff for `conventional` and `janestreet` profiles, but the diff for the `ocamlformat` profile can be seen on the files of ocamlformat itself (@jberdine what do you think?).
This PR adds consistency.

About the future of this option, I was already thinking of making `field-space=tight-decl` the new default, deprecating the 2 other values (loose and tight) in the meantime, that would be done in another PR if we choose to go with it. If it sounds like a good idea I can open another issue for that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 #### Changes
 
+  + Apply option 'field-space' to values in signatures for more consistency (#1855, @gpetiot)
+
 #### New features
 
 ### 0.20.1 (2021-12-13)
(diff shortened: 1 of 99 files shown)


gpetiot:
Closing this PR, I think having a separate option is also the best solution afterall. But not a priority for now.

PR #1865, opened 2021-10-11
Define 'break-infix=wrap-or-vertical' that only wraps high precedence infix ops

### LINKED ISSUE #1080: Feature request: Don't wrap pipelines
The option `break-infix=fit-or-vertical` allows to break before every infix operators if the whole expression does not fit on a single line:

```diff
-           get_succs parent |> Sequence.of_list
+           get_succs parent
+           |> Sequence.of_list
            |> Sequence.filter ~f:(fun n -> not (equal node n))
            |> Sequence.Generator.of_sequence)
```

This works well for pipelines, boolean operators and sometimes list cons/concat but not with most other expressions:

```diff
 let git_clone ~branch ~remote ~output_dir =
   run_and_log
     Cmd.(
-      v "git" % "clone" % "--depth" % "1" % "--branch" % branch % remote
+      v "git"
+      % "clone"
+      % "--depth"
+      % "1"
+      % "--branch"
+      % branch
+      % remote
       % p output_dir)
```

This is clearly not what we want when the operands are short or when the expression is dense (eg. arithmetic, cmdliner's `&` and `$`).

I'd like `fit-or-vertical` in the `default` profile but only for pipelines.
Should we implement that ? How to decide what is a pipeline and what is not ?

There have been discussions before, in https://github.com/ocaml-ppx/ocamlformat/issues/69 then `break-infix` has been added in https://github.com/ocaml-ppx/ocamlformat/pull/150.

gpetiot:
Fix #1080

@CraigFe @samoht please don't hesitate if you have any feedback on this new option

### DIFF
--- a/test/passing/tests/break_infix.ml.ref
+++ b/test/passing/tests/break_infix.ml.ref
@@ -1,0 +1,87 @@
+let _ =
+  get_succs parent
+  |> Sequence.of_list
+  |> Sequence.filter ~f:(fun n -> not (equal node n))
+  |> Sequence.Generator.of_sequence
+
+let git_clone ~branch ~remote ~output_dir =
+  run_and_log
+    Cmd.(
+      v "git"
+      % "clone"
+      % "--depth"
+      % "1"
+      % "--branch"
+      % branch
+      % remote
+      % p output_dir )
+
+let all_6_2char_hex a b c d e f =
+  is_2char_hex a
+  && is_2char_hex b
+  && is_2char_hex c
+  && is_2char_hex d
+  && is_2char_hex e
+  && is_2char_hex f
+
+let pf0 =
+  let open Utils.Parallel_folders in
+  open_ construct
+  |+ misc_stats_folder header
+  |+ elapsed_wall_over_blocks_folder header block_count
+  |+ elapsed_cpu_over_blocks_folder header block_count
+  |+ Span_folder.create header.initial_stats.timestamp_wall block_count
+  |+ cpu_usage_folder header block_count
+  |+ ( pack_folder
+     && foooooooooooooooooooo
+     && foooooooooooooooooooo
+     && foooooooooooooooooooo )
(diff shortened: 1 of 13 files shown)


gpetiot:
I refreshed the PR. I think the new behavior is close to what you described but doesn't break if the expression doesn't reach the margin. Breaking systematically sounds a bit extreme to me at first but given the direction we are giving ocamlformat with the future diff-friendly profile it kinda makes sense.

test/passing/tests/break_infix.ml.ref
```diff
@@ -0,0 +1,87 @@
...
+  run_and_log
+    Cmd.(
+      v "git"
+      % "clone"
+      % "--depth"
+      % "1"
+      % "--branch"
+      % branch
+      % remote
+      % p output_dir )
```
Julow, line 17: This is vertical because of the applications of `v` and `p`, it's correct but annoying. 1 char long function names are intended to be discreet, like an prefix operator.
Would it make sense to add an exception this case ?

Julow (approved):
Nice ! The advantages of `fit-or-vertical` but more concise.

test/passing/tests/break_infix.ml.ref
gpetiot: It makes sense but it should be done separately, considering this kind of function application as simple would be useful in other contextes.

Julow: I don't think it should be done separately. There's a risk of not implementing it due to timing and finding that generalizing it is too breaking.
This option needs a different "simple" predicate that can be tweaked.

gpetiot: This part of the algorithm doesn't rely on whether an expression is simple but on the precedence, we can tweak it to treat this kind of application differently but I'm afraid it would look weird (what about `f (fun x -> complex expr)`). I think it's reasonable to keep the current behavior until users complain too much, it's gated behind a new option anyway.

PR #1866, opened 2021-10-12
RFC: Add comments in a hashtbl during parsing

gpetiot:
Asking for opinions before sinking too much time into this.

The idea is to produce a `Location.t -> (string list * string list)` (comments before, comments after) hash table during the parsing step (I know that the lexer currently ignores the COMMENT tokens when feeding tokens to the parser, I didn't touch that yet). Having such a hashtbl would make Non_overlapping_interval_tree.ml, Loc_tree.ml and a big part of Cmts.ml obsolete.
Having this hashtbl means we don't need to add fields in the parsetree for each node that can have comments.

So I plan to edit the rules to accept comments in specific places. I was wondering if it could work, that would be a big diff on the parser (both ocaml-4.13-extended and parse-wyc), wrapping the whole set of rules doesn't work, e.g.

```
module_expr:
  cmts_around(
     ...
| ...
)
{ $1 }
```

so `cmts_around` has to be called on each call of node that could have comments, and apparently for some nodes it produces reduce conflicts. Unless I'm missing something.

Thoughts @emillon @Julow @jberdine @hhugo @let-def @trefis ?

### DIFF
--- a/vendor/ocaml-4.13-extended/parser.mly
+++ b/vendor/ocaml-4.13-extended/parser.mly
@@ -31,6 +31,18 @@ open Ast_helper
 open Docstrings
 open Docstrings.WithMenhir
 
+let cmt_tbl = Hashtbl.create 32
+
+let add_cmts_before loc cmts =
+  match Hashtbl.find_opt cmt_tbl loc with
+  | Some (before, after) -> Hashtbl.replace cmt_tbl loc (before @ cmts, after)
+  | None -> Hashtbl.add cmt_tbl loc (cmts, [])
+
+let add_cmts_after loc cmts =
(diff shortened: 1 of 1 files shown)


hhugo:
- you need to decide whether to attach a comment after an expression or before the next one.

```ocaml
let f () =
   call
     arg1 (*arg1*)
     arg2
```
vs 
```ocaml
let f () =
   call
     arg1
     (* arg2 *)
     arg2
```

- you need to decide how deep to attach a comment

```ocaml
let f () = match x with
  | None -> ()
  | Some s ->
     print s (* about s *)
```
vs

```ocaml
let f () = match x with
  | None -> ()
  | Some s ->
     print s
(* about f *)
```

PR #1868, opened 2021-10-13
Set 'module-item-spacing=compact' in the ocamlformat profile

gpetiot:
Follow-up on #1848 as changing this option for the `ocamlformat` was the probable next step, this PR gives a preview of that.
A lot of tests require `max-iter=3` because `module-item-spacing=compact` uses `is_single_line` to decide whether two items are simple enough (must fit on one line) to be packed together.

Mainly needing @jberdine's input before going through with this one.


gpetiot:
@jberdine what do you think of this?

PR #1879, opened 2021-10-21
Do not check if a comment fits on a single line when splitting comment groups

gpetiot:
No diff with test_branch.sh


test/passing/tests/source.ml.ref
```diff
@@ -1470,15 +1470,22 @@ let ty_list : type a e. (a, e) ty -> (a vlist, e) ty =
...
+  | Pnil : ('a,'e) ty_pvar
+  | Pconst : 't * ('b,'e) ty_pvar -> ('t -> 'b, 'e) ty_pvar
+  | Parg : 't * ('a,'e) ty * ('b,'e) ty_pvar -> ('t * 'a -> 'b, 'e) ty_pvar
+*)
+(*
+   An attempt at encoding omega examples from the 2nd Central European
+   Functional Programming School:
+     Generic Programming in Omega, by Tim Sheard and Nathan Linger
+          http://web.cecs.pdx.edu/~sheard/
+*)
```
Julow, line 1488: Why are these comments not wrapped anymore ?

Julow (commented):
> No diff with test_branch.sh

`test_branch` can't be trusted when we talk about comments because the two things it compares have been formatted. The second pass of formatting is made on top of the first one.

PR #1885, opened 2021-10-27
Break arg group when there is a comment after

gpetiot:
Fix a regression spotted in https://github.com/facebook/infer/pull/1546

ping @jberdine since it's originally found on `infer`, btw the comments are located before the arguments on `infer`, it would be better to put them after


gpetiot:
@jberdine do you have some feedback about this one?

PR #1908, opened 2021-11-03
Deprecate option 'parens-tuple-patterns'

gpetiot:


vch9:
Hi, I'm really late for the party but we just upgraded our ocamlformat and I was wondering why this option was deprecated? Is there a workaround?

gpetiot:
Hi, given the feedback on these deprecations I plan to de-deprecate them for 0.22.0, so don't mind the warning message.

vch9:
> Hi, given the feedback on these deprecations I plan to de-deprecate them for 0.22.0, so don't mind the warning message.

Thanks a lot! :)

PR #1920, opened 2021-11-08
Mention that operational options cannot be set in attributes or .ocamlformat files

### LINKED ISSUE #1917: Bug: Unknown option "margin-check"
**Describe the bug**

In v0.19, `ocamlformat --help` mentions `--margin-check` however I was surprised that it is not recognized in `.ocamlformat`. Is this the desired behavior? If so, maybe the man page should note that it's only for the command line.

**How to Reproduce**

Create an `.ocamlformat` file which includes the line:

```text
margin-check
```

…and run `ocamlformat` on any file.

gpetiot:
Fix #1917

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -169,7 +169,11 @@ let info =
          $(b,.ocamlformat-enable) file specifies a filename relative to the \
          directory containing the $(b,.ocamlformat-enable) file. \
          Shell-style regular expressions are supported. Lines starting with \
-         $(b,#) are ignored and can be used as comments." ]
+         $(b,#) are ignored and can be used as comments."
+    ; `S (C.section_name C.Operational `Valid)
+    ; `P
+        "Unless mentioned otherwise non-formatting options cannot be set \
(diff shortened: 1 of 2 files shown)


lib/Conf.ml
```diff
@@ -169,7 +169,11 @@ let info =
          $(b,.ocamlformat-enable) file specifies a filename relative to the \
          directory containing the $(b,.ocamlformat-enable) file. \
          Shell-style regular expressions are supported. Lines starting with \
-         $(b,#) are ignored and can be used as comments." ]
+         $(b,#) are ignored and can be used as comments."
+    ; `S (C.section_name C.Operational `Valid)
+    ; `P
+        "Unless mentioned otherwise non-formatting options cannot be set \
+         neither in attributes nor in $(b,ocamlformat) files." ]
```
jberdine, line 176:
```suggestion
         in attributes or $(b,ocamlformat) files." ]
```

jberdine (approved):
One minor rewording, otherwise LGTM.

PR #1921, opened 2021-11-09
Use Printast output to debug comment positioning

gpetiot:


Julow (commented):
Pretty nice !

My first reaction was to say "What about Pprintast ?!" but that wouldn't have helped much more and would have been harder to implement (would need to add call sites everywhere).

I've found some locations that are not printed: `pconst_loc`, `prf_loc`, `pdira_loc`, `pof_loc`, `Pexp_send` label's loc, `Pexp_newtype` first arg's loc, `popen_loc`, `extension` first arg's loc, `attribute` name loc, `attr_loc`, `ptyext_loc`. I've stopped at half the file so it's not very reliable yet.

gpetiot:
> I've found some locations that are not printed: `pconst_loc`, `prf_loc`, `pdira_loc`, `pof_loc`, `Pexp_send` label's loc, `Pexp_newtype` first arg's loc, `popen_loc`, `extension` first arg's loc, `attribute` name loc, `attr_loc`, `ptyext_loc`. I've stopped at half the file so it's not very reliable yet.

I've checked the whole file manually and I should have caught all locations now

vendor/diff-parsers-std-ext.patch
```diff
@@ -535,6 +535,659 @@
...
++      module_type_declaration i ppf x
+   | Psig_open od ->
+-      line i ppf "Psig_open %a %a\n" fmt_override_flag od.popen_override
+-        fmt_longident_loc od.popen_expr;
+-      attributes i ppf od.popen_attributes
++      line i ppf "Psig_open\n";
++      open_description i ppf od
+   | Psig_include incl ->
+       line i ppf "Psig_include\n";
+       module_type i ppf incl.pincl_mod;
```
Julow, line 1018: Missing `pincl_loc` (same above in signatures)

Julow (approved):
There's a missing loc in `typevars` and `toplevel_phrase`. After that I think it's good to go.

This patch is worth submitting to the compiler !

vendor/diff-parsers-std-ext.patch
gpetiot: done

gpetiot:
I will try to get it accepted upstream before merging this PR

Julow:
We can still merge this PR first and propose it upstream asynchronously. This kind of merge conflict is trivial to resolve (if it doesn't resolve automatically)

PR #1926, opened 2021-11-16
Improve comment attachment when followed but not preceded by a linebreak

### LINKED ISSUE #1925: Bug: Inconsistent spacing between comments and docstrings
**Describe the bug**
When combining documentation comments and regular comments in a signature, `ocamlformat` inserts blank lines inconsistently (and quite intrusively) between the documentations and the regular comments.
Only the last signature item seem to get formatted correctly (or at least as I would expect it to be)

**How to Reproduce**
With version 0.19.0 and no option, the following snippet
```ocaml
val f : unit
(** docstring *)
(* comment *)

val g : unit
(** docstring *)
(* comment *)
```
gets formatted as 
```ocaml
val f : unit
(** docstring *)

(* comment *)

val g : unit
(** docstring *)
(* comment *)
```

I would expect the input snippet to be already well formatted.

gpetiot:
Fix #1925, this is a cheap fix that only makes Cmts.ml dirtier, but it allows to fix this for 0.20.0, as a bigger rework looks out of scope.
No diff with test_branch.sh

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -106,27 +106,15 @@ let find_at_position t loc pos =
   in
   Map.find map loc
 
+let no_filter (_ : Parser.token) = true
+
 (** Heuristic to determine if two locations should be considered "adjacent".
     Holds if there is only whitespace between the locations, or if there is a
     [|] character and the first location begins a line and the start column
     of the first location is lower than that of the second location. *)
-let is_adjacent_prev src (l1 : Location.t) (l2 : Location.t) =
-  match
-    Source.tokens_between src l1.loc_end l2.loc_start ~filter:(function
-      | DOCSTRING _ -> false
-      | _ -> true )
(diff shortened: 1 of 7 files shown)


lib/Cmts.ml
```diff
@@ -110,6 +110,18 @@ let find_at_position t loc pos =
     Holds if there is only whitespace between the locations, or if there is a
     [|] character and the first location begins a line and the start column
     of the first location is lower than that of the second location. *)
```
Julow, line 112: This comment was attached to `is_adjacent`, I think it'll be better to leave it there and to add a comment for the new function.
Also, I think `is_adjacent_prev` could be implemented as an optional argument to `is_adjacent` (the filter function)

Julow (approved):
Minor comment on the code, otherwise looks good :)

PR #1929, opened 2021-11-18
Make the formatting of attributes and docstrings more consistent

gpetiot:
This PR removes a lot of local tweaks, remove some boilerplate code and improves consistency. I would like to go even further and remove every call of the function `doc_atrs` , and instead have something like `wrap_attributes_and_docstrings` that would correctly place attributes and docstrings (duh!), and automatically takes care of the docstrings that only go before and the ones that can go both depending of the `doc-comments` option. But this PR looks good enough to be merged in my opinion.

The tests look good, the small diff shows the improved consistency (you can see in each file that it becomes more consistent with the rest).

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -594,7 +594,39 @@ and fmt_attributes = fmt_attributes_aux ~key:Attr.Key.Regular
 
 and fmt_item_attributes = fmt_attributes_aux ~key:Attr.Key.Item
 
-and fmt_floating_attributes = fmt_attributes_aux ~key:Attr.Key.Floating
+and fmt_attributes_and_docstrings_aux c ~key attrs =
+  let standalone, pro, space =
+    match key with
+    | Attr.Key.Regular | Attr.Key.Item ->
+        (false, break c.conf.doc_comments_padding 0, break 1 0)
+    | Attr.Key.Floating -> (true, noop, noop)
+  in
+  let aux = function
+    | { attr_name=
+          {txt= "ocaml.doc" | "ocaml.text"; loc= {loc_ghost= true; _}}
+      ; attr_payload=
+          PStr
+            [ { pstr_desc=
+                  Pstr_eval
+                    ( { pexp_desc=
+                          Pexp_constant
+                            {pconst_desc= Pconst_string (txt, _, None); _}
+                      ; pexp_loc= loc
+                      ; pexp_attributes= []
+                      ; _ }
+                    , [] )
+              ; _ } ]
+      ; _ } ->
+        fmt_docstring c ~standalone ~pro (Some [({txt; loc}, standalone)])
+    | attr -> space $ fmt_attribute c ~key attr
+  in
+  list attrs "" aux
+
+and fmt_attributes_and_docstrings =
+  fmt_attributes_and_docstrings_aux ~key:Attr.Key.Regular
+
+and fmt_floating_attributes_and_docstrings =
+  fmt_attributes_and_docstrings_aux ~key:Attr.Key.Floating
 
 and fmt_payload c ctx pld =
   protect c (Pld pld)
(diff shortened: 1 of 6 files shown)


test/passing/tests/comments_in_record.ml.ref
```diff
@@ -48,7 +48,9 @@ type t =
   | Struct of
       { name: string
       ; elts: t vector (* possibly cyclic, name unique *)
-            [@compare.ignore] [@equal.ignore] [@sexp_drop_if fun _ -> true]
+            [@compare.ignore]
+            [@equal.ignore]
+            [@sexp_drop_if fun _ -> true]
```
Julow, line 53: This is a new behavior for attributes. I think the wrapping should be re-implemented (only on `@` attributes).

test/passing/tests/attributes.ml
```diff
@@ -11,7 +11,9 @@ val x : ?x:unit (** not dropped *) -> unit
 type t =
   { a: int
   ; b: int [@default 1] [@drop_if]
-  ; c: int [@default 1] [@drop_if]
+  ; c: int
+        [@default 1]
+        [@drop_if]
         (** docstring that is long enough to break *) }
```
Julow, line 17: I think docstrings shouldn't interfere with attributes. This box shouldn't be removed.

Julow (commented):
This is a nice code improvement :) However, I wonder what are the diffs on other projects ?

test/passing/tests/comments_in_record.ml.ref
gpetiot: reverted

test/passing/tests/attributes.ml
gpetiot: reverted

gpetiot:
Here is test_branch's diff with conventional profile:
```diff
diff --git a/testsuite/tests/parsing/attributes.ml b/testsuite/tests/parsing/attributes.ml
index df13f572a..b290d2650 100644
--- a/testsuite/tests/parsing/attributes.ml
+++ b/testsuite/tests/parsing/attributes.ml
@@ -20,8 +20,7 @@ module M = struct
   type t = { l : (t[@foo]) [@foo] } [@@foo] [@@foo]
 
   [@@@foo]
-end
-[@foo]
+end [@foo]
 [@@foo]
 
 module type S = sig
diff --git a/testsuite/tests/typing-deprecated/deprecated.ml b/testsuite/tests/typing-deprecated/deprecated.ml
index a05f222c0..b892b828a 100644
--- a/testsuite/tests/typing-deprecated/deprecated.ml
+++ b/testsuite/tests/typing-deprecated/deprecated.ml
@@ -214,8 +214,7 @@ module M : sig val x : X.t end
 
 module M = struct
   let x = X.x
-end
-[@ocaml.warning "-3"]
+end [@ocaml.warning "-3"]
 
 [%%expect {|
 module M : sig val x : X.t end
@@ -264,8 +263,7 @@ module rec M : sig
   val x : X.t
 end [@ocaml.warning "-3"] = struct
   let x = X.x
-end
-[@ocaml.warning "-3"]
+end [@ocaml.warning "-3"]
```

lib/Fmt_ast.ml
```diff
@@ -594,7 +594,39 @@ and fmt_attributes = fmt_attributes_aux ~key:Attr.Key.Regular
 
 and fmt_item_attributes = fmt_attributes_aux ~key:Attr.Key.Item
 
-and fmt_floating_attributes = fmt_attributes_aux ~key:Attr.Key.Floating
+and fmt_attributes_and_docstrings_aux c ~key attrs =
+  let standalone, pro, space =
+    match key with
+    | Attr.Key.Regular | Attr.Key.Item ->
+        (false, break c.conf.doc_comments_padding 0, break 1 0)
```
Julow, line 601: `doc_comments_padding` is now applied to more docs, are we sure it is not done in new places ? I couldn't find any.

Julow (approved):
Looks good :)

lib/Fmt_ast.ml
gpetiot: I think I already checked it

PR #1934, opened 2021-11-29
Remove unnecessary surrounding parentheses for immediate objects (OCaml 4.14 feature)

gpetiot:
Handle https://github.com/ocaml/ocaml/pull/10441

### DIFF
--- a/test/passing/tests/object_expr.ml.ref
+++ b/test/passing/tests/object_expr.ml.ref
@@ -1,0 +1,16 @@
+(object
+   method one = 1
+end )
+  #one
+;;
+
+Some
+  (object
+     method one = 1
+  end )
+;;
+
+ignore
+  (object
+     method one = 1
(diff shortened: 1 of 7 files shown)


test/passing/tests/object_expr.ml.ref
```diff
@@ -0,0 +1,16 @@
+(object
+   method one = 1
+end )
```
Julow, line 3: Unrelated, there's a bug in these parentheses: there should either be no spacing or a space both at the opening and closing.

Julow (approved):
Looks good.

test/passing/tests/object_expr.ml.ref
gpetiot: Noted, but I think there is already an issue open about that. It might cause a big diff though, we can take care of this for the next release.

PR #1935, opened 2021-11-29
ocamlformat-rpc v2: Add a command to the rpc server to format in the style of a file

panglesd:
Ocamlformat_rpc is used in editors to format code quickly by avoiding the starting time of ocamlformat. This is especially interesting for snippets of code that need to be reactive, such as showing type on hover.

However, currently ocamlformat_rpc does not take into account any of the rules of the project given by the dot configuration files such as the `.ocamlformat` files. This prevents using ocamlformat_rpc for formatting the whole file, and is not ideal for snippets as well.

This PR adds a command to the ocamlformat_rpc server and client. When the server receive a canonical sexp `("Combine", (path, conf, x))`, the server adopts the configuration it would have for formatting the file at `path`. Then, it overrides the options given in the `(key, value)` list `conf`. Finally, it formats the string `x` using the computed configuration, and send it back to the client as a canonical sexp `("Combine", (path, conf, formatted_x))`.

The man page and tests are updated.

Unfortunately, as ocamlformat_rpc uses ocamlformat as an internal library and not an external one (there is still no `ocamlformat_lib`), only users of the latest ocamlformat will be able to benefit from this feature.


bin/ocamlformat-rpc/main.ml
```diff
@@ -58,63 +106,36 @@ let rec rpc_main = function
...
-            ; format Module_type
-            ; format Expression
-            ; format Use_file ] ;
+          ( match run_format conf x with
+          | `Format formatted ->
+              ignore (Format.flush_str_formatter ()) ;
+              V1.Command.output stdout (`Format formatted)
+          | `Error e -> handle_format_error e ) ;
           rpc_main state
+      | `Combined (file, c, x) ->
```
Julow, line 115: This isn't much more extensible than the previous format command. What are your thoughts on:

- Add a "config_for_path" and "override_config" commands, that sets states within the server, the sequence of commands would be `("config_for_path", "path")` then `("format", "buffer")`.

- Take extra arguments to the format command that is a "sexp record" so it's extensible:
  ```sexp
  ( "format" ("config_for_path" "path") ("override_config" (config ...)) )
  ```

panglesd: I think that @rgrinberg whished to be able to configure and format in a single command for vs-code interaction with ocamlformat_rpc. So the goal of the Combined command wasn't really for extensibility. However, I agree that it would make sense to add a `Config_for_path` command. So that the `Combined` command is a combination of three commands: `Config_for_path`, `Config` and `Format` (instead of only the two latter + config for path but not available as a command).

I also see the advantage of being able to override the config, without modifying it permanently, in the format command. However, I'm not sure how much we can modify it, without bumping the protocol version...

Following your comment, I can add a `Config_for_path` command, and make the `Combined` command also takes an `overriding_config` that overrides but no modify the global config. I don't think that modifying the `Format` command is a good idea, as bumping the protocol version is a good idea, as I think it requires further thinking to design a better protocol.

What do you think?

Julow: Modifying the "format" command can be done in a backward compatible way, I think.

It's currently matched as `List [Atom "Format"; Atom x]`: `(Format "path")`.
It is safe to add more arguments: `List (Atom "Format" :: Atom x :: args)`: `(Format "path" option1 option2)` where each options could be arbitrary sexps.

gpetiot: We could also imagine a new command "Path" that could modify the local config similarly to Config. Vscode/ocamllsp would only call this command when the new file to reformat is in a different directory, it would be more efficient than recomputing the whole config every time a new file is loaded.

lib-rpc/ocamlformat_rpc_lib.ml
```diff
@@ -151,9 +215,19 @@ module V1 :
...
+      with
+      | `Format (x, _args) -> Ok x
+      | `Error msg -> Error (`Msg msg)
+      | _ -> Error (`Msg "failing to format input: unknown error")
+
+    let path path t =
+      match query (`Path path) t with
+      | `Path _path -> Ok ()
       | `Error msg -> Error (`Msg msg)
       | _ -> Error (`Msg "failing to format input: unknown error")
```
gpetiot, line 232: Should be `failing to set path`

CHANGES.md
```diff
@@ -104,6 +104,10 @@
   + Handle let operator punning uniformly with other punning forms.
     Normalizes let operator to the punned form where possible, if output syntax version is at least OCaml 4.13.0. (#1834, #1846, @jberdine)
 
+  + Added optionnal arguments to the format command to set or override the config and path. (#1935, @panglesd)
+
+  + Added a command to ask the server to take the configuration given by the ocamlformat’s dot files for a specific path. (#1935, @panglesd)
```
gpetiot, line 109: Maybe we should add a separate section for the RPC (both server and client) in the changelog

panglesd: I added a `RPC changelog` section

lib-rpc/ocamlformat_rpc_lib.ml
```diff
@@ -37,7 +37,16 @@ module type Client_S = sig
   val config :
     (string * string) list -> t -> (unit, [> `Msg of string]) result
 
-  val format : string -> t -> (string, [> `Msg of string]) result
+  val format :
+       ?path:string
+    -> ?config:(string * string) list
+    -> ?overriding_path:string
+    -> ?overriding_config:(string * string) list
```
Julow, line 44: Why not take `format_args` here ?

Julow (commented):
What's the point in having the arguments on the `Format` command at the same time as having the `Path` and `Config` commands ?
Both ways seem as powerful but I'm not sure it's a good idea to have both wrt to maintenance.

lib-rpc/ocamlformat_rpc_lib.ml
gpetiot: Or maybe only the overriding_path and overriding_config here, I think it's useful for the "formatting the tooltip" workflow.

For the rest I agree that it would be easier to maintain more atomic commands, as long as sending 2 or 3 commands instead of 1 isn't too much overhead:
- sending a Path command when a file from a new dir is formatted (up to vscode to cache the path of the last dir requested)
- sending a Config command if needed (I guess this is mostly needed for the margin)
- sending a Format command

rgrinberg: > sending a Path command when a file from a new dir is formatted (up to vscode to cache the path of the last dir requested)

How does the client know what's the argument to the Path command? How should vscode decide when to expire the cache of the last dir requested?

> sending a Config command if needed (I guess this is mostly needed for the margin)

In general, how should clients know whether a config is needed? My impression is every formatting needs to use the most recent configuration.

Julow: > sending 2 or 3 commands instead of 1 isn't too much overhead

Sending many commands is fine, the problem I point out is that there's two workflows: passing arguments to the `Format` command or the `Path` and `Config` commands.
One of the two workflow isn't necessary. Both ways can implement the lib interface but I think the `Format` command with argument is more consistent wrt to the API of `ocamlformat_rpc_lib`.

bin/ocamlformat-rpc/main.ml
Julow: The per-directory caching could pretty much be implemented in the server.

rgrinberg: Hi, I'd like a single command not because of performance (although I'll take always take the extra perf) but because I'd like to assume a little as possible about how ocamlformat works. I don't want to do any of:

* Load/Unload configuration from files
* Think about the lifetime of configuration values
* Be aware that configuration even exists

You might say I'm rejecting a more flexible API and limiting myself in the future. But the way I see it is that I'm reserving your project the maximum degrees of freedom to change the implementation without consulting me. Consequently, the lesser coupling of our projects grants our users more freedom to pick components without worry. That is valuable.

rgrinberg:
By the way, what was the motivation in making the RPC stateful? I think this introduces all sorts of footguns for clients.

panglesd:
I removed the two arguments `config` and `path` to the format command, so that there is now only one workflow to change the state of the server.

I also followed the suggestion of making it an optional argument of type `format_args` instead of two (formerly four) optional arguments. However, I think it makes it less concise for the end user. For instance:
```ocaml
format
    ~overriding_path:"small_margin/foo.ml"
    "aaa -> bbb -> ccc -> ddd -> eee -> fff -> ggg" 
```
becomes the more verbose:
```ocaml
format
    ~format_args:
      {Ocf.empty_args with overriding_path= Some "small_margin/foo.ml"}
    "aaa -> bbb -> ccc -> ddd -> eee -> fff -> ggg" 
```
I can revert or keep this change, depending on what you prefer!

Also, if there is an agreement to switch to the stateless server in #1953, I can implement these changes on top of this PR. (However, as I mentioned in the #1953 PR thread, the approach of this PR (#1935) has the advantage of offering both the stateless and stateful workflow)

gpetiot:
Regarding your comment on https://github.com/ocaml-ppx/ocamlformat/pull/1953 I have updated this PR to implement your changes as V2 API, V1 should still be compatible with previous releases.

Let me know what you think (personally I still think that removing the whole API-versioning would be easier to maintain, APIs change and users are used to that :D)

panglesd:
I guess the API versioning was to allow new versions of `ocamlformat_rpc_lib` to speak with old versions of `ocamlformat_rpc`?
But I'm not sure it makes sense to uncouple the `ocamlformat_rpc_lib` and the `ocamlformat_rpc` versions.

I think it would be more important to uncouple the `ocamlformat` and the `ocamlformat_rpc` versions, so that new versions of `ocamlformat_rpc` can be used to format in the style of old versions of ocamlformat, for project that uses them

But I do not have a strong opinion, I'm in favor of what seems the best for you :) If you prefer the #1953 way, I'll be happy to update ocaml-lsp accordingly (that was the original issue I was trying to solve: make ocaml-lsp use ocamlformat_rpc to format files).

CHANGES.md
```diff
@@ -10,6 +10,12 @@
 
 #### New features
 
+#### RPC
+
+  + RPC v2 (#1935, @panglesd):
+    - Define a 'Format' command parameterized with optionnal arguments to set or override the config and path.
```
panglesd, line 16: Following Julow suggestions I only kept arguments to override config and path
```suggestion
    - Define a 'Format' command parameterized with optional arguments to override the config and path.
```

gpetiot (approved):
I rebased this PR and it looks good to me now, any feedback @panglesd @rgrinberg @Julow @voodoos @emillon before I merge it?

rgrinberg:
It looks good to me, but is the plan to release the V2 API in its current form? Perhaps you could consider my (hopefully) small feature request first? https://github.com/ocaml-ppx/ocamlformat/issues/1966

lib-rpc/ocamlformat_rpc_lib.ml
```diff
@@ -159,10 +165,171 @@ module V1 :
...
-type client = [`V1 of V1.Client.t]
+module V2 : sig
+  module Command :
+    Command_S
+      with type t =
+        [ `Halt
+        | `Unknown
+        | `Error of string
+        | `Path of string
+        | `Config of (string * string) list
```
Julow, line 176: Now that we have a version bump, we might as well get rid of these commands now that they can be passed as option to the `format` command.

panglesd: They can be passed to the `format` command but in this case they are not persistent, so they cannot be "emulated" by the `format` command.
I'm OK to remove them, as I guess we always want to recompute the options from the `path`, in case the `.ocamlformat` file (or other relevant file/environment variable) have been modified.

panglesd:
Looks good to me!

lib-rpc/ocamlformat_rpc_lib.ml
Julow: We can certainly cache the `.ocamlformat` lookups. I think most clients would unconditionally send the `path` command before every `format` anyway, the `path` command is not an optimization.

gpetiot: I'm fine with removing `Config` and `Path` as well

gpetiot: I'm fine with removing `Config` and `Path` as well

CHANGES.md
```diff
@@ -21,6 +21,10 @@
   + ocamlformat-rpc-lib is now functorized over the IO (#1975, @gpetiot).
     Now handles `Csexp.t` types instead of `Sexplib0.Sexp.t`.
 
+  + RPC v2 (#1935, @panglesd):
+    - Define a 'Format' command parameterized with optionnal arguments to set or override the config and path.
+    - Define a 'Path' command to ask the server to take the configuration given by the `.ocamlformat` files for a specific path.
```
Julow, line 26: This changed.

lib-rpc/ocamlformat_rpc_lib.ml
```diff
@@ -9,12 +9,21 @@
 (*                                                                        *)
 (**************************************************************************)
 
+type format_args =
+  { overriding_path: string option
+  ; overriding_config: (string * string) list option }
```
Julow, line 14: We can drop the `overriding_` prefix.

ocamlformat-rpc-help.txt
```diff
@@ -41,6 +45,16 @@ COMMANDS
        by OCamlFormat, the formatted output is sent as a reply of the same
        form Format CSEXP
 
+       Specific commands supported on version v2 are:
+
+       - Format CSEXP: submits a list as canonical s-expression CSEXP, where
+       the first element of the list is a string to be formatted by
+       OCamlFormat. The other arguments are (key, value) pairs, where key can
+       be either "Path", "Config", "OverridingPath" or "OverridingConfig".
```
Julow, line 53: There's no `Path` and `Config` key anymore. I think the `Overriding` prefix can be removed everywhere.

gpetiot: done

CHANGES.md

bin/ocamlformat-rpc/main.ml
```diff
@@ -143,7 +156,22 @@ let rec rpc_main = function
...
+        | Error e -> handle_config_error e ; rpc_main state ) )
+    | V2 -> (
+      match V2.Command.read_input stdin with
+      | `Halt -> Ok ()
+      | `Unknown | `Error _ -> rpc_main state
+      | `Format (x, format_args) ->
+          let conf =
+            match run_format_with_args format_args conf x with
+            | Ok (`Format formatted) ->
+                ignore (Format.flush_str_formatter ()) ;
```
Julow, line 168: Is it needed ? If it is, perhaps it should be done in `Translation_unit`.

Julow (approved):
Looks good :)

bin/ocamlformat-rpc/main.ml
gpetiot: Apparently it's not, I'm removing this.

PR #1941, opened 2021-12-01
[WIP] Parse and format REPL phrases in doc-comments blocks

gpetiot:


lib/Extended_ast.ml
```diff
@@ -111,32 +115,63 @@ module Parse = struct
...
     | Expression -> Parse.expression lexbuf
+    | Repl_file ->
+        let str = String.strip (Bytes.to_string lexbuf.lex_buffer) in
+        let phrases = Astring.String.cuts ~sep:"\n# " str in
+        let phrases =
+          List.mapi phrases ~f:(fun i p ->
+              let p = String.strip p in
+              match i with
+              | 0 ->
+                  if String.is_prefix p ~prefix:"# " then
```
hhugo, line 126: `String.drop_prefix` should do what you want

hhugo: sorry, I meant `String.chop_prefix `, combining both `is_prefix` and `drop_prefix`

lib/Fmt_ast.ml
```diff
@@ -4471,7 +4482,19 @@ let fmt_code ~debug =
...
+        if String.is_prefix (String.strip s) ~prefix:"# " then
+          match
+            Parse_with_comments.parse Parse.ast Repl_file conf ~source:s
+          with
+          | {ast; comments; source; prefix= _} ->
+              let cmts = Cmts.init Repl_file ~debug source ast comments in
+              let ctx = Top in
+              Ok
+                (fmt_file ~ctx ~debug Repl_file source cmts conf ast
+                   ~fmt_code )
```
Julow, line 4495: This code is mostly the same as the "structure" case (the only difference are the "kind" and `ctx`). I think it'd be worth having a "is_repl_block" function that can be shared with the normalization (it's important that it's using the exact same).

Julow (commented):
The normalization is missing and the result is always a "Formatting of doc-comment is unstable" error.

lib/Extended_ast.ml
gpetiot: done

gpetiot:
> The normalization is missing and the result is always a "Formatting of doc-comment is unstable" error.

Yes I am also working on another PR to improve the normalize step before I tackle this one.

lib/Fmt_ast.ml
gpetiot: I added a function `Docstring.is_repl_block`

I'm not sure it's possible to factorize these 2 parts of code because of the gadt, I tried a few things but the first instantiation of the fragment makes it a `Structure Extended_ast.t` instead of the generic `a Extended_ast.t`.

emillon:
Thanks. This is not a comment about the implementation, but before this is merged we have to think about whether to gate this behind an option so that it plays well with old mdx versions.

lib/Fmt_ast.ml
Julow: I've remove some of the duplicated code here: https://github.com/gpetiot/ocamlformat/pull/6
But I also attempted to remove the parsing attempts, let me know what you think.

gpetiot: merged, thanks!

gpetiot:
The feature looks ready to me, any feedback @emillon @julow?

lib/Conf.ml
```diff
@@ -1243,8 +1251,15 @@ let kind : Syntax.t option ref =
   let intf = (Some Syntax.Signature, Arg.info ["intf"] ~doc ~docs) in
   let doc = "Deprecated. Same as $(b,impl)." in
   let use_file = (Some Syntax.Use_file, Arg.info ["use-file"] ~doc ~docs) in
+  let doc =
+    "Parse file with unrecognized extension as toplevel phrases with their \
```
Julow, line 1255: I think the "unrecognized extension" part is misleading because this option takes priority even if the extension would be recognized as something else. (same remark for `impl` and `intf`)

lib/Conf.ml
```diff
@@ -971,6 +972,13 @@ module Formatting = struct
       (fun conf x -> {conf with parse_docstrings= x})
       (fun conf -> conf.parse_docstrings)
 
+  let parse_toplevel_phrases =
+    let doc = "Parse and format toplevel phrases and their output." in
+    let names = ["parse-toplevel-phrases"] in
+    C.flag ~default:false ~names ~doc ~kind
```
Julow, line 978: I think this should be on by default, when `parse-docstrings` is enabled. Especially if we support the case without `;;`.

lib/Extended_ast.ml
```diff
@@ -101,6 +107,26 @@ module Parse = struct
...
+          let p = String.strip p in
+          match i with
+          | 0 -> Option.value (String.chop_prefix p ~prefix:"# ") ~default:p
+          | _ -> p )
+    in
+    List.map phrases ~f:(fun p ->
+        let p, prepl_output =
+          match Astring.String.cut ~sep:";;\n" p with
+          | Some (p, o) -> (p ^ ";;", o)
+          | None -> (p, "")
```
Julow, line 124: If the rule above is not implemented, this isn't a good default to return. The top-level output part will certainly cause a syntax error.

test/passing/tests/cinaps.ml
```diff
@@ -18,10 +18,12 @@ let x = 1
...
 
-(*$ ;; #use "import.cinaps"
+(*$
+  #use "import.cinaps";;
 
-  ;; List.iter all_fields ~f:(fun (name, type_) -> printf "\nexternal get_%s
-  : unit -> %s = \"get_%s\"" name type_ name) *)
+  List.iter all_fields ~f:(fun (name, type_) ->
+      printf "\nexternal get_%s\n  : unit -> %s = \"get_%s\"" name type_ name )
+*)
```
Julow, line 26: This works because of the smarter parsing ? I think we should add a `.ref` file to keep the original.

Julow (commented):
Nice :)

test/passing/tests/cinaps.ml
gpetiot: done

lib/Conf.ml
gpetiot: True, I will fix this in another PR in the meantime.

gpetiot: done

lib/Extended_ast.ml
```diff
@@ -101,6 +107,26 @@ module Parse = struct
...
+    let phrases =
+      List.mapi phrases ~f:(fun i p ->
+          let p = String.strip p in
+          match i with
+          | 0 -> Option.value (String.chop_prefix p ~prefix:"# ") ~default:p
+          | _ -> p )
+    in
+    List.map phrases ~f:(fun p ->
+        let p, prepl_output =
+          match Astring.String.cut ~sep:";;\n" p with
```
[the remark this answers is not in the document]
gpetiot: Hmm maybe we could print a warning in such a case? I'm not sure making the parsing more complex to handled old versions is the way to go, we should encourage users to make their code "nice"

lib/Extended_ast.ml
gpetiot: If an exception is raised when parsing as a toplevel_phrase, the code block will be printed verbatim, which is the least surprising outcome for the user I think. If there is no `;;` then we assume that the whole input is the toplevel_phrase, we can't really guess if there is the output in the block and where it starts.

Julow: I think it's better to raise an exception at this point and to not attempt to parse.

lib/Extended_ast.ml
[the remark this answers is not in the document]
Julow: I'm not a fan of warnings. I'm not sure that would make the parsing more complex than the current.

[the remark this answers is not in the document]
gpetiot: I modified this function to raise a `Syntaxerr.Error Expecting` exception, to use the same interface as the existing parsing functions from `Parse`, otherwise we would need to wrap a result type around everything, and we would just print the input verbatim and not display any error/warning anyway.

lib/Conf.ml
gpetiot: I am not sure we should support the case without `;;`, I am in favor of having this feature opt-in for one or two releases in case we find more bugs, before enabling it by default

lib/Extended_ast.ml
gpetiot: I replied in the other comment, I think it's better to mimic the other standard parsing functions (thus raising exceptions from Syntaxerr), all exceptions are caught anyway and the intput = output in this case (we don't do anything specific in the error case)

lib/Extended_ast.ml
```diff
@@ -101,6 +107,28 @@ module Parse = struct
 
   let normalize fg x = map fg fix_letop_locs @@ map fg normalize_lists @@ x
 
+  let repl_file (lx : Lexing.lexbuf) =
+    let str = String.strip (Bytes.to_string lx.lex_buffer) in
```
Julow, line 111: This only works by chance and might break in the future because that's not how this API is supposed to be used.
I think we should use Mdx's lexer directly and remove this potential bug and the string manipulations. https://github.com/realworldocaml/mdx/blob/1.11.1/lib/lexer_top.mll

gpetiot: done

lib/Toplevel_lexer.mll
```diff
@@ -0,0 +1,67 @@
...
+  List.fold_left (fun acc -> function
+      | `Command (lines, _loc) ->
+          { prepl_phrase=
+              String.concat "\n" lines
+              |> Lexing.from_string |> Parse.toplevel_phrase
+          ; prepl_output= "" }
+          :: acc
+      | `Output line -> (
+          match acc with
+          | [] -> raise (Syntaxerr.Error (Expecting (Location.curr lx, "#")))
```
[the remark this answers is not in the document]
Julow: I've pushed a failing test related to this in https://github.com/gpetiot/ocamlformat/pull/7

Julow (commented):
I've pushed some fixes to the lexer here: https://github.com/gpetiot/ocamlformat/pull/7 (simplification and fix locations)

lib/Toplevel_lexer.mll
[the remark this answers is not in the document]
gpetiot: Merged, thanks!

NathanReb (commented):
It looks great! Thanks for this!

I'm mostly reviewing as an MDX maintainer.

I think the `.mli` test case would benefit from explicit descprition of the expected behaviour in the comments. These are just raw blocks right now and I think in the long run it will be hard to maintain those tests once we have forgotten what the spec was. Having the spec in the test file will help preserving the right behavior, i.e. the one that plays well with MDX.

I think things that should appear here are:
- how to deal with block delimiters: they must live on their own line
- how to indent the content of blocks, as a reminder this was our original spec:
  *The indentation of the first line of code within the block will be used as the base indentation level for the whole block.
  That includes phrases and text between them in toplevel blocks.
  If the code does not start on a newline, the <block header indentation> + 1 will be used.*
- how to deal with phrases evaluations: do not parse, preserve their formatting
- how phrases must be delimited (start with a `#`, end with `;;`, anything in between is evaluation)
- how phrases and evaluations must separated in the output. Here's the original spec as a reminder:
*The text between toplevel phrases should start on a new line. In other words, the phrase terminating ;; should be followed by a newline. If it’s not, the formatter will insert one.*

I think it would also be nice to add a test case showing that an ocaml block in a doc comment that doesn't parse will be left untouched, I don't think there's one here atm and this is important to work with MDX blocks in some cases (e.g. import of ocaml code fragments that don't parse as ocaml on their own, a couple pattern matching cases for instance).

The full spec is available here if you need it: https://hackmd.io/cEq-BCkETjeekJZC_kUMFw#MDX-blocks-in-mli-files

test/passing/tests/repl.mli.ref
```diff
@@ -0,0 +1,39 @@
+(** Blocks such as this one:
+
+    {[
+      let x = 2 in
+      x + x
+    ]}
+
+    Or, as of odoc 2.1, this one:
+
+    {@ocaml[ let x = 2 ]}*)
```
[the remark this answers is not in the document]
Julow: It's better if Mdx and OCamlformat agree on this ! (especially top-level blocks, which are re-written by Mdx)
These blocks are supposed to be big, so breaking the line is more appropriate I think.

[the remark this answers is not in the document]
gpetiot: I agree it's better to put the block delimiters on their own line, I will even extract that in a separate PR because it should have been the case for master.
I diversified the tests to showcase what is allowed or not, but I still have to handle the "base indentation level" part of the spec.

lib/Toplevel_lexer.mll
```diff
@@ -0,0 +1,67 @@
...
+  | ";;" eol { newline lexbuf; List.rev ((Buffer.contents buf ^ ";;") :: acc) }
+  | _ as c   { Buffer.add_char buf c; phrase acc buf lexbuf }
+
+{
+let repl_file lx =
+  let x =
+    try
+      newline lx;
+      token lx
+    with _ -> raise (Syntaxerr.Error (Expecting (Location.curr lx, ";;")))
```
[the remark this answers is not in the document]
Julow: I believe this `try` isn't catching anything.
The test `missing_semisemi.repl` in `test/cli/repl_file_errors.t/run.t` is actually failing, we expect an ";; expected" message.

lib/Toplevel_lexer.mll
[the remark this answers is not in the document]
Julow: It's still failing. In `test/cli/repl_file_errors.t/run.t` it's `empty_line_begin.repl`.

vendor/ocaml-4.13-extended/parsetree.mli
```diff
@@ -993,3 +993,9 @@ and directive_argument_desc =
   | Pdir_int of string * char option
   | Pdir_ident of Longident.t
   | Pdir_bool of bool
+
+type repl_phrase =
+  {
+    prepl_phrase : toplevel_phrase;
+    prepl_output : string;
+  }
```
Julow, line 1001: Is this record useful at all ? It's causing problems if a block starts with something that's not a phrase.
I think it could be a `` [ `Command of string | `Output of string ] list ``.

NathanReb:
> I believe this try isn't catching anything.
> The test missing_semisemi.repl in test/cli/repl_file_errors.t/run.t is actually failing, we expect an ";; expected" message.

Quick note on this: from MDX's perspective I think it's okay if ocamlformat simply refuses to format those blocks since it will add the `;;` in the output itself so running ocamlformat again after that will properly format the block. I don't think there's a very good reason for ocamlformat to error out here, a warning might be better as it would format the most blocks then. I think this decision is yours to make but I wanted to clarfiy that it's fine to simply skip such blocks, as it should with blocks that don't parse to proper OCaml, since MDX won't emit those anymore.

test/passing/tests/repl.mli
```diff
@@ -0,0 +1,99 @@
...
+    # let x = 2;;
+    val x : int = 2
+    # x + 2;;
+    - : int = 4
+    # let x = 2 and y = 3 in x+y;;
+    ]}
+
+    Output are printed afer a newline:
+    {[
+    # let x = 2;; val x : int = 2
```
NathanReb, line 46: I think it would be fair if ocamlformat didn't handle this as MDX won't produce such blocks and simply left output untouched but putting on its own line is fine as well, whatever is easier for you!

NathanReb (commented):
It looks good, I think the tests have quite improved since last time!

I left a couple comments for things that are probably not strictly necessary.

For future reference: We discussed the indentation preservation part of the spec with @gpetiot and established that it would be hard to accomplish with the output of odoc-parser and is not that important especially for the first iteration so we can leave it out for now.

vendor/ocaml-4.13-extended/parsetree.mli
gpetiot: I'm not sure we should allow this kind of blocks (maybe we should idk), I have the feeling `mdx` is keeping this representation because it doesn't need something more precise and doesn't need to maintain an AST like ocamlformat.

Julow: This is specifically causing problems with repl-files starting with an empty line. Perhaps we can at least skip empty lines ?

lib/Toplevel_lexer.mll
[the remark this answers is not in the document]
gpetiot: I think havint just `Syntax error` printed comes from the location not being properly set in the exception, I improved the situation a bit but it looks like it's not fixed yet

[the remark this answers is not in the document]
gpetiot: @julow I tried a few things, but the locations are invalid anyway, since we have locations relative to the opening of the block instead of the start of the file. `odoc-parser` is attaching a `Loc.span` to the code block but it's not enough to get back a `Lexing.position`.
Since we don't modify invalid blocks it's not really a must-have to print a warning. We can open another PR if there is a way to get a position in the future.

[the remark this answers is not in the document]
Julow: That's not ideal but it might be possible to craft a `Lexing.position` value from odoc's location.
Otherwise, reporting the location of the entire comment is more useful than a completely wrong location, I think.

vendor/ocaml-4.13-extended/parsetree.mli
gpetiot: @NathanReb I didn't find anything about preserving empty lines and such in the spec, what is your take on this?

NathanReb: I'm unsure what MDX allows here but I think we don't take a huge risk by doing by accepting whitespaces before the first phrase in toplevel blocks in the input but stripping them out in the output.

If there is anything else than this before the first phrase I would consider the block invalid and simply not format it.

Does that sound like a reasonable behaviour to you?

Note that I would also be perfectly happy if the blocks were required to start with the first phrase on the first line without any extra whitespace!

lib/Toplevel_lexer.mll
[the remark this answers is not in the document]
gpetiot: I added a warning when the parsing of toplevel block failed

lib/Fmt_ast.ml
```diff
@@ -4467,6 +4467,17 @@ let fmt_toplevel c ctx itms =
   let ast x = Tli x in
   fmt_item_list c ctx update_config ast fmt_item itms
 
+let fmt_repl_phrase c ctx {prepl_phrase; prepl_output} =
+  str "# "
+  $ fmt_toplevel c ctx [prepl_phrase]
+  $ fmt ";;"
```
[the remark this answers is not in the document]
Julow: Also, `fmt_toplevel` already have some logic for printing `;;`, are we sure it's not clashing in some edge cases ? Anyway, it'll be better to use it from here.

Julow (commented):
Very nice :)

test/cli/repl_file_errors.t/run.t
```diff
@@ -0,0 +1,34 @@
...
+  $ ocamlformat --repl-file line7.repl
+  ocamlformat: ignoring "line7.repl" (syntax error)
+  File "line7.repl", line 7, characters 5-5:
+  Error: Syntax error
+  [1]
+
+  $ ocamlformat --repl-file missing_semisemi.repl
+  ocamlformat: ignoring "missing_semisemi.repl" (syntax error)
+  File "missing_semisemi.repl", line 2, characters 3-3:
+  Error: Syntax error
```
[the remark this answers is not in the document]
gpetiot: fixed

test/passing/tests/repl.mli
```diff
@@ -0,0 +1,99 @@
...
+    Multiple phrases without `;;` at the end:
+    {[
+    # let x = 2 in x+1
+    # let x = 4 in x+1
+    ]}
+
+    Linebreak after `#`:
+    {[
+    #
+      let x = 2 in x+1;;
```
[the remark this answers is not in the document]
gpetiot: done

test/cli/repl_file_errors.t/run.t
```diff
@@ -0,0 +1,34 @@
...
+  [1]
+
+  $ ocamlformat --repl-file empty_line_begin.repl
+  ocamlformat: ignoring "empty_line_begin.repl" (syntax error)
+  File "empty_line_begin.repl", line 4, characters 0-0:
+  Error: Syntax error: "".
+  Hint: A toplevel block must start with a toplevel phrase starting with `# `. not expected.
```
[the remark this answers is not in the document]
gpetiot: fixed

test/passing/tests/repl.mli
```diff
@@ -0,0 +1,99 @@
...
+
+    ]}
+*)
+
+(** INVALID BLOCKS: The formatting of invalid blocks is preserved.
+
+    Invalid toplevel phrase/ocaml block:
+    {[
+    - : int =
+     4
```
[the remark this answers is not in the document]
gpetiot: This one does not contain `#` so it should not even be parsed as a toplevel phrase, that's weird. But the error message is still helpful.

lib/Fmt_ast.ml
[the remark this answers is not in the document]
Julow: I could find a way using a new `ctx` constructor: https://github.com/gpetiot/ocamlformat/pull/8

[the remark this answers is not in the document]
gpetiot: it's better indeed, thanks!

gpetiot:
> Thanks. This is not a comment about the implementation, but before this is merged we have to think about whether to gate this behind an option so that it plays well with old mdx versions.

Last thing I wonder before merging: @NathanReb do you agree about adding an option to not change the output for old mdx versions? I struggle to think of a case where it would be bad for the user.

NathanReb:
> Last thing I wonder before merging: @NathanReb do you agree about adding an option to not change the output for old mdx versions? I struggle to think of a case where it would be bad for the user.

Having a way to disable this would be useful indeed be it an option or something else. I can't think of any problem it could cause and I agree we should not bother making it compatible by default.

Either way I don't think there's any very good reason for anyone not to upgrade to mdx.2.0 so if it turns out there's difficulties with disabling this behaviour, I'd say we can just ask that users do not upgrade ocamlformat until they also updated MDX or that they use existing ocamlformat feature not to format the problematic files or file parts.

PR #1942, opened 2021-12-02
Fix stabilization of comments inside attributes

### LINKED ISSUE #1940: Bug: comment inside attribute fails to stabilize
**Describe the bug**
The following code fails to stabilize:
```
let _ =
  match x with
  | _
    when f ~f:((function
      | _ -> .) [@ocaml.warning (* ....................................... *) "-4"]) -> y
;;
```

**How to Reproduce**
Format with ocamlformat built from 6620833caa19858d429cddfa08274d6d85edf28e and with `--profile=janestreet`.

gpetiot:
Fix #1940, no diff with test_branch.sh so I added a few tests.

@ceastlund please let us know if there are a few cases that are not fixed by this patch.


Julow (approved):
Looks good.

PR #1951, opened 2021-12-06
Remove the ocamlformat-bench opam package

gpetiot:
This package should not be released on opam-repo (so no changelog needed), and I didn't find a way to generate the opam files for each package except this one (from the dune-project file)


gpetiot:
It seems the .opam file is needed for current-bench, so this is the wrong approach. I will try to pass only the packages I need to `dune-release` the next time I release ocamlformat (and will update the release procedure accordingly)

PR #1953, opened 2021-12-07
[RFC] Make RPC server stateless: specify version for each command

gpetiot:
Following up #1935, here is what the RPC could look like if we removed the internal state keeping the version of the RPC and the configuration, the goal is to ease the maintenance and make it less error prone for users.
cc @rgrinberg @Julow @panglesd @voodoos

By making the red_input functions return a result type, we could get rid of the Unknown and Error commands, but I kept them for now for retro-compatibility.

Todo: the tests have to be updated once the config is included in the Format command.

### DIFF
--- a/bin/ocamlformat-rpc/main.ml
+++ b/bin/ocamlformat-rpc/main.ml
@@ -18,105 +18,80 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter);;
 
 Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 
-module V = struct
-  type t = V1
-
-  let handshake = function
-    | "v1" | "V1" -> `Handled V1
-    | _ -> `Propose_another V1
-
-  let to_string = function V1 -> "v1"
-end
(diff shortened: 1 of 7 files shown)


panglesd:
In my point of view, the advantage with #1935 and the signature:
```ocaml
val format :
       ?path:string
    -> ?config:(string * string) list
    -> ?overriding_path:string
    -> ?overriding_config:(string * string) list
    -> string
    -> t
    -> (string, [> `Msg of string]) result
```
is that it allows both a stateless use (passing the path and special configuration at each request), or the previous approach without modification: there is no breaking change in the API. (Although maybe only two optional arguments should be kept)

It can be documented somewhere that the good way of using it is by always passing the path and conf, and in a later release, deprecate the stateful version.

Also, with a stateless version we can imagine watching the configuration files to avoid reloading the configuration every time. Until that is implemented, I don't see a necessity to change the API.

(This is just a very humble opinion, I don't have much experience in these kinds of things!)

PR #1967, opened 2022-01-03
Add missing parentheses around variant class arguments

### LINKED ISSUE #1964: Bug: ast changed
**Describe the bug**

I get the following error message when I try to format `test.ml`:
```
ocamlformat: Cannot process "test.ml".
  BUG: ast changed.
```

**Formatted file content**

```ocaml

class virtual roman (`I i)  =
  object
  end
```

**Selected options**
Just launching `ocamlformat test.ml`

**`--print-config` output**

```
profile=default
quiet=false
max-iters=10
comment-check=true
wrap-fun-args=true
wrap-comments=false
type-decl-indent=2
type-decl=compact
stritem-extension-indent=0
space-around-variants=true
space-around-records=true
space-around-lists=true
space-around-arrays=true
single-case=compact
sequence-style=terminator
parse-docstrings=false
parens-tuple=always
parens-ite=false
ocp-indent-compat=false
nested-match=wrap
module-item-spacing=sparse
max-indent=68
match-indent-nested=never
match-indent=0
margin=80
line-endings=lf
let-module=compact
let-binding-spacing=compact
let-binding-indent=2
let-and=compact
infix-precedence=indent
indent-after-in=0
if-then-else=compact
function-indent-nested=never
function-indent=2
field-space=loose
extension-indent=2
exp-grouping=parens
dock-collection-brackets=true
doc-comments-tag-only=default
doc-comments-padding=2
doc-comments=after-when-possible
disable=false
cases-exp-indent=4
break-struct=force
break-string-literals=auto
break-sequences=true
break-separators=after
break-infix-before-func=false
break-infix=wrap
break-fun-sig=wrap
break-fun-decl=wrap
break-cases=fit
assignment-operator=end-line
align-variants-decl=false
align-constructors-decl=false
align-cases=false
```

gpetiot:
Fix #1964, no diff with test_branch.sh

There is no similar bug for class types as there is no equivalent syntax.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 #### Bug fixes
 
+  + Add missing parentheses around variant class arguments (#<PR_NUMBER>, @gpetiot)
+
 #### Changes
 
 #### New features
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 #### Bug fixes
 
-  + Add missing parentheses around variant class arguments (#<PR_NUMBER>, @gpetiot)
+  + Add missing parentheses around variant class arguments (#1967, @gpetiot)
 
 #### Changes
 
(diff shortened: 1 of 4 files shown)


Julow (approved):
Looks good :)

Issue #1938, opened 2021-11-29
Bug: Wrapping of lambda arguments near nested constructors


cwong-ocaml:
**Describe the bug**
The wrapping of lambda arguments changes in the presence of nested constructors.

**How to Reproduce**
At 4812abd2ae, run `ocamlformat --profile=janestreet` on the following text:

```ocaml
let x =
  foo (`A b) ~f:(fun thing ->
    something that reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaally needs wrapping)
;;
```

This is reasonable, but if we change the `b` argument to be a polymorphic variant:

```ocaml
let x =
  foo
    (`A `b)
    ~f:(fun thing ->
      something that reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaally needs wrapping)
;;
```

then the format changes (the location of the `fun` changes lines), which is strange and
inconsistent. This isn't strictly an issue with polymorphic variants, either, contrast

```ocaml
let x =
  foo [ A; B ] ~f:(fun thing ->
    something that reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaally needs wrapping)
;;
```

with

```ocaml
let x =
  foo
    [ [ A ]; B ]
    ~f:(fun thing ->
      something that reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaally needs wrapping)
;;
```

It also doesn't matter whether or not the lambda argument is named.

PR #1968, opened 2022-01-03
Variant expressions with no argument are considered "simple"

gpetiot:
Taking care of the first part of #1938: the variants, since it's not controversial. #1943 will be rebased to only take care of the lists and arrays.


lib/Ast.ml
```diff
@@ -313,6 +313,7 @@ module Exp = struct
     | Pexp_apply
         ({pexp_desc= Pexp_ident {txt= Lident "not"; _}; _}, [(_, e1)]) ->
         is_trivial e1
+    | Pexp_variant (_, None) -> true
```
Julow, line 316: What about variant with a trivial arguments like for regular variants ? (like for `Pexp_construct`)

PR #1969, opened 2022-01-04
Fix indentation of module binding RHS

gpetiot:
See the tests for the diff (test_branch's diff is big, but that's because some files had huge misindented modules, that are now indented properly). Other constructs should be good.

### DIFF
--- a/test/passing/tests/module.ml
+++ b/test/passing/tests/module.ml
@@ -71,3 +71,12 @@ let helper ?x =
   match x with Some (module X : X_typ) -> X.f | None -> X_add_one.f
 
 let helper ?x:((module X) = (module X_add_one : X_typ)) = X.f
+
+module GZ : functor (X : sig end) () (Z : sig end) -> sig end =
+  (val Mooooooooooooooooooo)
+
+module GZ : functor (X : sig end) () (Z : sig end) -> sig end =
+  [%foooooooooooooooo]
+
+module GZZZZZZZZZZZZZZ : functor (X : sig end) () (Z : sig end) -> sig end =
+  _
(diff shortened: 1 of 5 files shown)


Julow (commented):
I wonder if that wasn't on purpose ? Especially `Pmod_extension` and `Pmod_functor`, which might have multi-line content and where there's nothing after the end that might be ambiguous without indentation.

```ocaml
module M ... =
functor () -> struct
  (* many lines *)
end
```

vs
```ocaml
module M ... =
  functor () -> struct
    (* many lines *)
  end
```

At least, this formatting allows to add a type annotation to a functor without indenting everything:
```ocaml
module M () = struct
  (* many lines *)
end
```

test/passing/tests/module.ml
```diff
@@ -71,3 +71,12 @@ let helper ?x =
...
 let helper ?x:((module X) = (module X_add_one : X_typ)) = X.f
+
+module GZ : functor (X : sig end) () (Z : sig end) -> sig end =
+  (val Mooooooooooooooooooo)
+
+module GZ : functor (X : sig end) () (Z : sig end) -> sig end =
+  [%foooooooooooooooo]
+
+module GZZZZZZZZZZZZZZ : functor (X : sig end) () (Z : sig end) -> sig end =
+  _
```
Julow, line 82: I think we definitely want this patch for `Pmod_hole` and `Pmod_unpack`.

gpetiot: Alright I will only keep these 2 cases then, and we will wait until someone complains about the other 2 cases to see what we do.

PR #1975, opened 2022-01-07
Functorize ocamlformat-rpc-lib over the IO

### LINKED ISSUE #1966: ocamlformat-rpc: asynchronous IO
The current API is married to synchronous IO and requires using stdlib's channels. It would be preferable if the library was functorized to allow for users to insert their own (usually monadic) IO.

gpetiot:
@rgrinberg Fix #1966, this is a first draft, it seems there is a mismatch over the channels used somewhere (the tests don't terminate). It couldn't hurt to get the feedback of someone more familiar with this kind of architecture.

### DIFF
--- a/bin/ocamlformat-rpc/main.ml
+++ b/bin/ocamlformat-rpc/main.ml
@@ -38,8 +38,6 @@ module IO = struct
     List.iter lx ~f:(Csexp.to_channel oc) ;
     Out_channel.flush oc ;
     return ()
-
-  let close oc = Out_channel.close oc
 end
 
 module V = struct
@@ -171,7 +169,9 @@ let info =
         "Once the client and the server agree on a common version, the \
          requests you can send may differ from one version to another."
     ; `P "On version $(b,v1), the supported RPC commands are:"
-    ; `P "- $(b,Halt) to close the connection to the RPC"
+    ; `P
+        "- $(b,Halt) to end the communication with the RPC server. The \
+         caller must close the input and output channels."
     ; `P
         "- $(b,Config) $(i,CSEXP): submits a list of (key, value) pairs (as \
          a canonical s-expression) to update OCamlFormat's configuration \
--- a/bin/ocamlformat-rpc/main.ml
+++ b/bin/ocamlformat-rpc/main.ml
@@ -18,6 +18,26 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter);;
 
 Caml.at_exit (Format_.pp_print_flush Format_.err_formatter)
 
+module IO = struct
+  type 'a t = 'a
+
+  type ic = In_channel.t
+
+  type oc = Out_channel.t
+
+  let ( >>= ) x f = f x
+
+  let return x = x
+
+  let read_line ic = In_channel.input_line ic
+
+  let read _ _ = failwith "unused"
+
+  let write = Out_channel.output_string
+
+  let flush = Out_channel.flush
+end
+
 module V = struct
   type t = V1
 
@@ -29,6 +49,8 @@ module V = struct
 end
 
 type state = Waiting_for_version | Version_defined of (V.t * Conf.t)
+
+include Make (IO)
 
 let format fg conf source =
   let input_name = "<rpc input>" in
--- a/bin/ocamlformat-rpc/main.ml
+++ b/bin/ocamlformat-rpc/main.ml
@@ -29,13 +29,19 @@ module IO = struct
 
   let return x = x
 
-  let read_line ic = In_channel.input_line ic
+  module Csexp = Csexp.Make (Sexp)
 
-  let read _ _ = failwith "unused"
+  let read ic =
+    match Csexp.input ic with
+    | Ok x -> return (Some x)
+    | Error _ -> return None
 
-  let write = Out_channel.output_string
+  let write oc lx =
+    List.iter lx ~f:(Csexp.to_channel oc) ;
+    Out_channel.flush oc ;
+    return ()
 
-  let flush = Out_channel.flush
+  let close oc = Out_channel.close oc
 end
 
 module V = struct
--- a/bin/ocamlformat-rpc/main.ml
+++ b/bin/ocamlformat-rpc/main.ml
@@ -29,8 +29,6 @@ module IO = struct
 
   let return x = x
 
-  module Csexp = Csexp.Make (Sexp)
-
   let read ic =
     match Csexp.input ic with
     | Ok x -> return (Some x)
--- a/lib-rpc/IO.ml
+++ b/lib-rpc/IO.ml
@@ -1,0 +1,37 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                              OCamlFormat                               *)
+(*                                                                        *)
+(*            Copyright (c) Facebook, Inc. and its affiliates.            *)
+(*                                                                        *)
+(*      This source code is licensed under the MIT license found in       *)
+(*      the LICENSE file in the root directory of this source tree.       *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** The [IO] module defines the blocking interface for reading and writing to
+    Cohttp streams *)
+
+module type S = sig
+  (** ['a t] represents a blocking monad state *)
+  type +'a t
+
+  val ( >>= ) : 'a t -> ('a -> 'b t) -> 'b t
+  (** [a >>= b] will pass the result of [a] to the [b] function. This is a
+      monadic [bind]. *)
+
+  val return : 'a -> 'a t
+  (** [return a] will construct a constant IO value. *)
+
+  (** [ic] represents an input channel *)
+  type ic
+
+  (** [oc] represents an output channel *)
+  type oc
+
+  val read : ic -> Sexplib0.Sexp.t option t
+
+  val write : oc -> Sexplib0.Sexp.t list -> unit t
+
+  val close : oc -> unit
+end
--- a/lib-rpc/IO.ml
+++ b/lib-rpc/IO.ml
@@ -32,6 +32,4 @@ module type S = sig
   val read : ic -> Csexp.t option t
 
   val write : oc -> Csexp.t list -> unit t
-
-  val close : oc -> unit
 end
--- a/lib-rpc/IO.ml
+++ b/lib-rpc/IO.ml
@@ -29,9 +29,9 @@ module type S = sig
   (** [oc] represents an output channel *)
   type oc
 
-  val read : ic -> Sexplib0.Sexp.t option t
+  val read : ic -> Csexp.t option t
 
-  val write : oc -> Sexplib0.Sexp.t list -> unit t
+  val write : oc -> Csexp.t list -> unit t
 
   val close : oc -> unit
 end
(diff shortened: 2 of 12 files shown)


bin/ocamlformat-rpc/main.ml
```diff
@@ -18,6 +18,26 @@ Caml.at_exit (Format.pp_print_flush Format.err_formatter);;
...
+
+  type ic = In_channel.t
+
+  type oc = Out_channel.t
+
+  let ( >>= ) x f = f x
+
+  let return x = x
+
+  let read_line ic = In_channel.input_line ic
```
rgrinberg, line 32: This primitive is a little suspicious. What if there's a new line inside the Csexp.t?

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,115 @@
...
+      condition is reached. *)
+
+  val read : ic -> int -> string t
+  (** [read ic len] will block until a maximum of [len] characters are read
+      from the input channel [ic]. It returns an empty string if EOF or some
+      other error condition occurs on the input channel, and can also return
+      fewer than [len] characters if input buffering is not sufficient to
+      satisfy the request. *)
+
+  val write : oc -> string -> unit t
```
rgrinberg, line 43: An alternative design is to make the primitives:

```
val read : ic -> Csexp.t option t
val write : oc -> Csexp.t list -> unit t (* the list allows us to control flushing a little better *)
val close : oc -> unit
```

The advantage of this scheme is that your code becomes simpler and users will be able to provide a more performant implementation than something that is byte based. The disadvantage of course is that you're leaking the serialization protocol.

gpetiot: I changed the module signature and fixed the tests, it looks better indeed.

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -28,25 +28,11 @@ module type IO = sig
...
-      satisfy the request. *)
-
-  val write : oc -> string -> unit t
-  (** [write oc s] will block until the complete [s] string is written to the
-      output channel [oc]. *)
-
-  val flush : oc -> unit t
-  (** [flush oc] will return when all previously buffered content from
-      calling {!write} have been written to the output channel [oc]. *)
+  val read : ic -> Sexplib0.Sexp.t option t
```
rgrinberg: Could you use the types from `csexp` instead if you don't mind? E.g. `Csexp.t`.

gpetiot: Done.

lib-rpc/IO.ml
```diff
@@ -0,0 +1,37 @@
...
+  type ic
+
+  (** [oc] represents an output channel *)
+  type oc
+
+  val read : ic -> Csexp.t option t
+
+  val write : oc -> Csexp.t list -> unit t
+
+  val close : oc -> unit
```
Julow, line 36: No `open` but there's `close`. I suggest removing this function, and let the user handle the closing. Especially that there's no close for `ic` and that it might be easier to do it safely in the caller code.

To make sure the close is not missed, the `halt` function can be documented about the lifetime of the `client` values.

lib-rpc/ocamlformat_rpc_lib.ml
```diff
@@ -9,194 +9,199 @@
...
+      let query command t =
+        let open IO in
+        Command.output t.output command
+        >>= fun () -> Command.read_input t.input
+
+      let halt t =
+        let open IO in
+        match Command.output t.output `Halt with
+        | exception _ ->
+            return (Error (`Msg "failing to close connection to server"))
```
Julow, line 140: This error might be ignored by the caller but the `close` must happen anyway.

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,75 @@
...
+    val to_sexp : t -> Csexp.t
 
-  val output : Stdlib.out_channel -> t -> unit
-end
+    val output : IO.oc -> t -> unit IO.t
+  end
 
-module type Client_S = sig
-  type t
+  module type Client_S = sig
```
Julow, line 23: This type is not useful outside of the module, right ? I think it can be inlined into `V.Client` and only the `type t` left. Same for `V.Command`.
Otherwise, that's two way of accessing the same functions.

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,75 @@
...
-      | `Config of (string * string) list
-      | `Format of string ]
+  module V1 :
+    V
+      with type Command.t =
+        [ `Halt
+        | `Unknown
+        | `Error of string
+        | `Config of (string * string) list
+        | `Format of string ]
```
Julow, line 59: Why expose the commands ? It's not used in the API below.

Julow (approved):
Looks good :)

lib-rpc/IO.ml
gpetiot: Done, and I modified the doc.

lib-rpc/ocamlformat_rpc_lib.mli
gpetiot: If we remove this we can't match on the command sent/received on the server side, maybe we could redesign the whole thing in a future PR

lib-rpc/ocamlformat_rpc_lib.ml
gpetiot: Taken care of by the last change

lib-rpc/ocamlformat_rpc_lib.mli
gpetiot: I inlined Client_S but Command_S is used twice.

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,78 @@
...
-  -> (client, [`Msg of string]) result
-(** [pick_client ~pid in out versions] returns the most-fitting client
-    according to a list of [versions], that is a list ordered from the most
-    to the least wished version. *)
+  val pid : client -> int
 
-val pid : client -> int
+  val halt : client -> (unit, [> `Msg of string]) result IO.t
+  (** The caller must close the input and output channels after calling
+      [halt]. *)
```
Julow, line 78: I would rather talk about the `client` type here.
```suggestion
  (** Tell the server to close the connection. No more commands can be sent
      using the same `client` value. *)
```

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,78 @@
...
-type client = [`V1 of V1.Client.t]
+  val pick_client :
+       pid:int
+    -> IO.ic
+    -> IO.oc
+    -> string list
+    -> (client, [`Msg of string]) result IO.t
+  (** [pick_client ~pid in out versions] returns the most-fitting client
+      according to a list of [versions], that is a list ordered from the most
+      to the least wished version. *)
```
Julow, line 72:
```suggestion
      to the least wished version. The given `IO.ic` and `IO.oc` values are
      referenced by the `client` value and must be kept open until `halt` is
      called. *)
```

I'm not sure about my english here but I think this is worth documenting.

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,78 @@
...
+  end
 
-module type Client_S = sig
-  type t
+  module type V = sig
+    module Command : Command_S
 
-  type cmd
+    module Client : sig
+      type t
```
Julow, line 27: All but `type t` can be removed from `Client` ?

Except if `pick_client` isn't suitable for every users, but in this case I would suggest having an other constructor function for the other use cases.

lib-rpc/ocamlformat_rpc_lib.mli
```diff
@@ -9,70 +9,78 @@
...
+      val format : string -> t -> (string, [> `Msg of string]) result IO.t
+    end
+    with type cmd = Command.t
+  end
 
-  module Client : Client_S with type cmd = Command.t
-end
+  (** Version used to set the protocol version *)
+  module Init :
+    Command_S with type t = [`Halt | `Unknown | `Version of string]
```
Julow, line 51: This `Init` module isn't useful for users of the library. The server uses it but I think it should be defined there. (same for `Command_S` and `V1.Command`.

To be sure the server and the lib are using the same types, there can be a module dedicated to that, which doesn't pretend to be part of the API:
```ocaml
  (** Types used for communicating between the client and the server. *)
  module Commands : sig
    module V1 : sig
      type init = [`Halt | `Unknown | `Version of string]
      type t =
        [ `Halt
        | `Unknown
        | `Error of string
        | `Config of (string * string) list
        | `Format of string ]
    end
  end
```

lib-rpc/ocamlformat_rpc_lib.mli
gpetiot: Sounds good to me

lib-rpc/ocamlformat_rpc_lib.mli
gpetiot: `Init` is useful for the server `bin/ocamlformat-rpc/main.ml` that is also a server of the library unfortunately. Each type of command also needs the read/output/to_sexp functions, that are parameterized by IO, so I don't find a better architecture than the current one for now.

lib-rpc/ocamlformat_rpc_lib.mli
gpetiot: If we scrap these functions from the `Client` module it doesn't compile anymore

rgrinberg:
Before you declare this interface as stable, let me try it out in ocamllsp. So if you want o make a release, mark this interface as unstable for now please.

gpetiot:
Sure, we can wait for your feedback before making a release.

rgrinberg:
I gave the new api a try. Much better this time. Thanks.

A few things I'd like to point:

1. I see V1 and V2 modules but no versioning contract is fleshed out anywhere. What are the backwards compatibility guarantees provided?
2. The public api for clients is too large. Do the clients really need to know that `Command_S` functions are available? I understand this is for the server, but you should separate the public api for clients so that we know what we can rely on.
3. The documentation around the various commands is poor. Can you please add some?
4. `IO.S` should either be included in the `ocamlformat_rpc_lib.mli` or inlined in the function argument. Otherwise you're referring to a module that isn't actually accessible to users. In particular, this makes it impossible to browse via odoc generated docs.

gpetiot:
Thanks for your feedback, I will improve the docs in the mli.
We didn't write the compatibility guarantees anywhere that's true, it's more of a best effort to not change the signature of the mli (we obviously had to make an exception for the IO functorization) too much, and preserve the original behavior of the versioned clients as much as possible. I will add this in the docs.

I also agree that the API looks complicated for end users, I don't know how to organize this better. On the other hand, I'm not sure this library has a lot of users beyond ocamllsp.
I just generated the doc with odoc and the doc of the IO signature is generated: _build/default/_doc/_html/ocamlformat-rpc-lib/Ocamlformat_rpc_lib/Make/argument-1-IO/index.html, but we can indeed makes it more accessible as part of a new user-friendly API.

cc @Julow @emillon maybe we can make a cleaner API, I don't think it's blocking for a release though, what do you think?

gpetiot:
@rgrinberg we improved the documentation thanks to your comments.

Is everything alright for you so we can make a release of ocamlformat?

rgrinberg:
Sure go for it. I plan to send some changes to your library, but they should be backwards compatible.

PR #1980, opened 2022-01-17
odoc: Print code block delimiters on their own line

gpetiot:


test/passing/tests/doc_comments.mli.ref
```diff
@@ -525,12 +539,34 @@ val k : int
...
   -> fooooooooooooooooooooooooo
 
-type x = {a: t  (** {[ (* a b *) ]} *); b: [`A  (** {[ (* a b *) ]} *)]}
-
-type x = A of a  (** {[ (* a b *) ]} *) | B of b  (** {[ (* a b *) ]} *)
+type x =
+  { a: t
+        (** {[
+              (* a b *)
+            ]} *)
```
Julow, line 557: Isn't this comment over indented ?

gpetiot: There is a +4 indentation for a comment attached to a record type field (`label_declaration`), the +2 indentation is used when the field breaks on the `:` for example, I'm not a fan of it either but I think there was some backlash the last time I tried to change it.

PR #1981, opened 2022-01-17
odoc: Align code block contents under the opening character

gpetiot:
Reduce the diff of #1941

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@
 
   + Print odoc code block delimiters on their own line (#1980, @gpetiot)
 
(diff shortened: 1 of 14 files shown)


Julow (commented):
I'm not sure this is reducing the diff caused by repl blocks, in the end. This is a pretty big change.

Perhaps the repl blocks and regular blocks can be indented differently ? Makes sense to me because repl blocks are kind of indented with `# ` and `- ` prefixes.

gpetiot:
You are right, apparently `ocp-indent` indents +2 the contents of the blocks and it makes sense to keep the same indentation. It would be different than the indentation inside markdown blocks (`\`` ... ```) but that's not an issue.
I'm closing this issue and reverting this change in #1941

PR #1985, opened 2022-01-20
Fix position of ':=' when 'assignment-operator=end-line'

gpetiot:
Reduce the diff of #1667, no diff with test_branch.sh

The formatting could still be perfected, but it's still a slight improvement compared to the current formatting.


Julow (approved):
Looks good :)

PR #1988, opened 2022-01-21
Do not wrap docstrings, 'wrap-comments' should only impact non-doc comments

### LINKED ISSUE #1984: Bug: Doc comments changed because wrap-comments
This example from the `Format` module causes a crash:

```ocaml
(** - Item 1
    - Item 2, that contains two block elements: {v v} *)
```

It formats to a different comment (and crash on purpose):

```ocaml
(** - Item 1 - Item 2, that contains two block elements: {v v} *)
```

The second list item triggers a warning in `odoc-parser` because of the verbatim not starting on its own line (it doesn't alter the AST this time but we can't be sure).
So ocamlformat decides to give up parsing and to "wrap" the content of the comment instead.

We shouldn't wrap doc-strings.

This happens with `0.19.0` and perhaps before. It happens in these files, with options `--profile=conventional --parse-docstrings --wrap-comments`:
+ `test-extra/code/ocaml/stdlib/printf.mli`
+ `test-extra/code/ocaml/stdlib/random.ml`
+ `test-extra/code/ocaml/stdlib/format.mli`
+ `test-extra/code/ocaml/middle_end/flambda/simple_value_approx.mli`
+ `test-extra/code/ocaml/testsuite/tests/tool-ocamldoc/t01.ml`
+ `test-extra/code/ocaml/testsuite/tests/tool-ocamldoc/t04.ml`
+ `test-extra/code/ocaml/testsuite/tests/tool-ocamldoc/Paragraph.mli`
+ `test-extra/code/ocaml/utils/diffing_with_keys.ml`

gpetiot:
Fix #1984, test_branch.sh shows that files that were previously not formatted now are formatted.


Julow (approved):
Looks good!

PR #1993, opened 2022-01-26
Also check equality regarding Extended_ast

gpetiot:
Reduce diff of #1941

### DIFF
--- a/lib/Normalize_extended_ast.ml
+++ b/lib/Normalize_extended_ast.ml
@@ -38,18 +38,23 @@ let dedup_cmts fragment ast comments =
   in
   Set.(to_list (diff (of_list (module Cmt) comments) (of_ast ast)))
 
+let normalize_comments dedup fmt comments =
+  let comments = dedup comments in
+  List.sort comments ~compare:(fun {Cmt.loc= a; _} {Cmt.loc= b; _} ->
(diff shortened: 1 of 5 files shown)


lib/Normalize_extended_ast.ml
```diff
@@ -182,15 +187,13 @@ let diff_cmts (conf : Conf.t) x y =
             in
             let len = String.length str - chars_removed in
             let source = String.sub ~pos:1 ~len str in
-            match
-              Parse_with_comments.parse Parse.ast Structure conf ~source
-            with
-            | exception _ -> norm_non_code z
```
gpetiot, line 188: @julow we used to call `norm_non_code` on the source before, now we don't, but there is no change on the tests. And since this makes the check "stronger" maybe it's a good idea, was it intentional or not?

Julow: This is when a block of code fails to parse. We format these exactly as they were in the input, and with Odoc's indentation corrections we don't mess the indentation either.
The stronger check works :)

Julow (approved):

lib/Normalize_extended_ast.ml
gpetiot: Noted, thanks!

PR #1994, opened 2022-01-26
tests: Setting the ocaml-version after it has been overriden by a profile

gpetiot:
Fix #1991 (@kit-ty-kate can you confirm?)

@Julow We could also move `ocaml-version` from `Conf.t` to `Conf.opts` but that means rewriting all the functions updating the config (this would also allow to set `debug` and `margin-check` in .ocamlformat files), but that's a much bigger change. But I think eventualy we should do that.
And also move `comment-check`, `max-iters` and `quiet` out of `Conf.t` in the same way.

PR #1995, opened 2022-01-27
Do not override the values of non-formatting options when a profile is set

### LINKED ISSUE #1991: The testsuite fails with OCaml 4.14
```
$ dune runtest -p ocamlformat
# File "test/passing/tests/js_source.ml.ref", line 1, characters 0-0:
#          git (internal) (exit 1)
# (cd _build/default && /usr/bin/git --no-pager diff --no-index --color=always -u test/passing/tests/js_source.ml.ref test/passing/js_source.ml.stdout)
# diff --git a/test/passing/tests/js_source.ml.ref b/test/passing/js_source.ml.stdout
# index d81f988..c68d3a8 100644
# --- a/test/passing/tests/js_source.ml.ref
# +++ b/test/passing/js_source.ml.stdout
# @@ -542,9 +542,9 @@ let _ = Obj.extension_constructor 7 (* Invald_arg *)
#  
#  let _ =
#    Obj.extension_constructor
# -    (object
# -       method m = 3
# -    end)
# +    object
# +      method m = 3
# +    end
#  ;;
#  
#  (* Invald_arg *)
# @@ -1398,27 +1398,27 @@ and (_, _) ty_case =
#  
#  let ty_abc : (([ `A of int | `B of string | `C ] as 'a), 'e) ty =
#    Sum
# -    (object
# -       method proj =
# -         function
# -         | `A n -> "A", Some (Tdyn (Int, n))
# -         | `B s -> "B", Some (Tdyn (String, s))
# -         | `C -> "C", None
# -
# -       method cases =
# -         [ "A", TCarg (Thd, Int)
# -         ; "B", TCarg (Ttl Thd, String)
# -         ; "C", TCnoarg (Ttl (Ttl Thd))
# -         ]
# -
# -       method inj
# -           : type c.  (int -> string -> noarg -> unit, c) ty_sel * c
# -                     -> [ `A of int | `B of string | `C ] =
# -         function
# -         | Thd, v -> `A v
# -         | Ttl Thd, v -> `B v
# -         | Ttl (Ttl Thd), Noarg -> `C
# -    end)
# +    object
# +      method proj =
# +        function
# +        | `A n -> "A", Some (Tdyn (Int, n))
# +        | `B s -> "B", Some (Tdyn (String, s))
# +        | `C -> "C", None
# +
# +      method cases =
# +        [ "A", TCarg (Thd, Int)
# +        ; "B", TCarg (Ttl Thd, String)
# +        ; "C", TCnoarg (Ttl (Ttl Thd))
# +        ]
# +
# +      method inj
# +          : type c.  (int -> string -> noarg -> unit, c) ty_sel * c
# +                    -> [ `A of int | `B of string | `C ] =
# +        function
# +        | Thd, v -> `A v
# +        | Ttl Thd, v -> `B v
# +        | Ttl (Ttl Thd), Noarg -> `C
# +    end
#  ;;
#  
#  type 'a vlist =
# @@ -1431,19 +1431,19 @@ let ty_list : type a e. (a, e) ty -> (a vlist, e) ty =
#    let tcons = Pair (Pop t, Var) in
#    Rec
#      (Sum
# -       (object
# -          method proj =
# -            function
# -            | `Nil -> "Nil", None
# -            | `Cons p -> "Cons", Some (Tdyn (tcons, p))
# -
# -          method cases = [ "Nil", TCnoarg Thd; "Cons", TCarg (Ttl Thd, tcons) ]
# -
# -          method inj : type c. (noarg -> a * a vlist -> unit, c) ty_sel * c -> a vlist =
# -            function
# -            | Thd, Noarg -> `Nil
# -            | Ttl Thd, v -> `Cons v
# -       end))
# +       object
# +         method proj =
# +           function
# +           | `Nil -> "Nil", None
# +           | `Cons p -> "Cons", Some (Tdyn (tcons, p))
# +
# +         method cases = [ "Nil", TCnoarg Thd; "Cons", TCarg (Ttl Thd, tcons) ]
# +
# +         method inj : type c. (noarg -> a * a vlist -> unit, c) ty_sel * c -> a vlist =
# +           function
# +           | Thd, Noarg -> `Nil
# +           | Ttl Thd, v -> `Cons v
# +       end)
#  ;;
#  
#  (*
```
cc @gpetiot

gpetiot:
Fix #1991
Replacing #1994


bin/ocamlformat-rpc/main.ml
```diff
@@ -127,7 +126,18 @@ let rec rpc_main = function
       | `Handled v ->
           Init.output stdout (`Version vstr) ;
           Out_channel.flush stdout ;
-          rpc_main (Version_defined (v, Conf.default_profile))
+          rpc_main
+            (Version_defined
+               ( v
+               , { fmt_opts= Conf.default_profile
+                 ; opr_opts=
```
Julow: We need a `default_opts` somewhere.

lib/Translation_unit.ml
```diff
@@ -291,31 +292,35 @@ let format (type a b) (fg : a Extended_ast.t) (std_fg : b Std_ast.t)
...
     in
-    if opts.debug then
+    if conf.opr_opts.debug then
       format ~box_debug:true |> fst
       |> dump_formatted ~suffix:".boxes"
       |> (ignore : string option -> unit) ;
     let fmted, cmts_t = format ~box_debug:false in
-    let conf = if opts.debug then conf else {conf with quiet= true} in
+    let conf =
+      if conf.opr_opts.debug then conf
```
Julow, line 317: This `if` is not necessary, the else branch covers all the cases.

Julow (approved):
Looks good :)

But all this to be able to specify `comment_check`, `ocaml_version`, etc.. in attributes ? I think the only non-formatting option we still want in attributes is `disable` and I don't think it needs to be in `Conf.t` either.

gpetiot:
> But all this to be able to specify `comment_check`, `ocaml_version`, etc.. in attributes ? I think the only non-formatting option we still want in attributes is `disable` and I don't think it needs to be in `Conf.t` either.

Told you this would be a big change! :p

From the options in `opr_opts` only `disable` can be set in attributes because operational options have `allow_inline=false` by default. Having these options in `Conf.t` allows to set them in `.ocamlformat` files, this was already the case for the ones in the old `Conf.t` but not for the ones that were built with `mk ...`, we could only set them on cmdline.

bin/ocamlformat-rpc/main.ml
gpetiot: done

kit-ty-kate:
OCaml-CI confirms this fixes the tests on 4.14

lib/Translation_unit.ml
Julow: It is not! Ignore this.

PR #1998, opened 2022-01-31
ocamlformat-rpc-lib: remove covariance requirement on IO

rgrinberg:
it is not used, so let's not require it.


gpetiot (approved):
Thanks, looks good!

PR #2002, opened 2022-02-01
Remove incorrect parentheses around polymorphic type constraint

### LINKED ISSUE #2000: Bug: bad paren placement around polymorphic functions
In some contexts, ocamlformat inserts an extra pair of parens around polymorphic function definitions, generating invalid syntax:
```ocaml
class c =
  let id : 'a . 'a -> 'a = fun x -> x in
  object end;;
```
yields `BUG: generating invalid ocaml syntax.`, because it generates the following invalid AST:
```ocaml
class c =
  let (id : 'a. 'a -> 'a) = fun x -> x in
  object end
```
I'm not sure which other contexts the bug appears in. It reproduces in `0.20.1` and `main` but not in `0.19.0`.

gpetiot:
Fix #2000, was broken by #1734.
It looks like this change is correct for any context, I didn't find any issue. No change with test_branch.sh.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,8 @@
 
   + Do not override the values of the following non-formatting options when a profile is set: `comment-check`, `disable`, `max-iters`, `ocaml-version`, and `quiet` (#1995, @gpetiot).
 
+  + Remove incorrect parentheses around polymorphic type constraint (#<PR_NUMBER>, @gpetiot)
+
 #### Changes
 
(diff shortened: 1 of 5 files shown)


Julow (approved):
Looks good.

This is indeed broken since https://github.com/ocaml-ppx/ocamlformat/pull/1734 but it's really unclear why at first glance. Some of these patterns really do more than what I expected.

PR #2003, opened 2022-02-01
emacs: Remove temp files in the event of an error

### LINKED ISSUE #2001: Bug: At emacs startup, ocamlformat creates a lot of temporary files
**Describe the bug**
OCamlformat creates a lot of files in /tmp/ named ocamlformat[random ascii*].ml when starting emacs

**How to Reproduce**
- Create an empty directory with a file.ml file
- Download the file [plain-ocamlformat.zip](https://github.com/ocaml-ppx/ocamlformat/files/7979040/plain-ocamlformat.zip) and unzip it
- Start emacs in the directory like this:
  - emacs -q -l plain-ocamlformat.el file.ml
- You should now see a lot of files being created and a max-lisp-eval-depth bug appear

gpetiot:
@mattiasdrp do you still experience the bug with this patch? I still can't reproduce the error with emacs 27.2 so I need someone to try this.

But based on `unwind-protect`'s documentation (https://www.gnu.org/software/emacs/manual/html_node/elisp/Cleanups.html) it should do the job.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 #### Bug fixes
 
+  + emacs: Remove temp files in the event of an error (#2003, @gpetiot)
+
 #### Changes
 
   + Qtest comments are not re-formatted (#2034, @gpetiot)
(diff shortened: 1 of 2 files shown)


mattiasdrp:
Sorry, no, the bug persists :-(

gpetiot:
Even though it doesn't fix the bug, using unwind-protect looks like an improvement. I made some tests and it looks like nothing broke with this change, so we could merge it.

Julow (commented):
Sorry I don't know enough about emacs to review this. The use of `unwind-protect` seems right.

PR #2004, opened 2022-02-02
Apply 'sequence-style' to add a space before ';;' between toplevel items

gpetiot:
As suggested by https://github.com/facebook/infer/pull/1546#discussion_r737611709 I agree it adds more consistency.
This only impacts `sequence-style = separator` and the `ocamlformat` profile (cc @jberdine @jvillard).


Julow (approved):
No strong opinion but I don't think `;;` needs to be consistent with `;`. Due to the small impact on known code, I'd say let's go if users of the `ocamlformat` profile want this.

gpetiot:
I don't have any strong opinion either, but the added consistency makes sense to me.
@jvillard is this kind of change really important to you? It would be nice to have feedback from users of the `profile = ocamlformat` on the facebook repositories.

jvillard:
I like this change, thanks! My rationale for this is that `;;` is a stronger separator than `;` so should be visually stronger too, i.e. be separated by the same amount or more of spaces and linebreaks than `;`.

PR #2005, opened 2022-02-02
test_branch.sh: Include janestreet and community projects to the list

gpetiot:
The goal is to add all the projects from https://github.com/janestreet/opam-repository/tree/master/packages to the list curated for test_branch.sh, but I would like to generate it automatically since it's kinda big.
The `grep+sed` commands are doing what they're supposed to do, the whole script is not working yet (WIP), should work but doesn't (as usual).

@Julow let's replace this Makefile by some OCaml codesome day? we're the only ones running this script anyway!

### DIFF
--- a/test-extra/Makefile
+++ b/test-extra/Makefile
@@ -9,10 +9,12 @@
 #                                                                    #
 ######################################################################
 
+JS_DIRS=code/base code/base_bigstring code/base_quickcheck code/core code/core_bench code/core_compat code/core_extended code/core_kernel code/core_profiler code/core_unix
+
 # To test all source files below a directory
 #       make DIRS=<directory> test
 # By default, test projects used as regression tests
-DIRS=code/ocamlformat code/infer code/js_of_ocaml code/dune
+DIRS=code/ocamlformat code/infer code/js_of_ocaml code/dune $(JS_DIRS)
 
 # Extra test directories, for which looser checking is done
 XDIRS=code/ocaml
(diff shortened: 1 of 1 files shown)


test-extra/Makefile
```diff
@@ -53,8 +57,21 @@ code/dune: URI = https://github.com/ocaml/dune.git
 # code/alcotest: URI = https://github.com/mirage/alcotest
 # code/citty: URI = https://github.com/ocurrent/citty
 
+# include similar rules from the janestreet repositories
+-include js.mak
+
+js-repo:
+	@git clone https://github.com/janestreet/opam-repository.git js-repo
```
Julow, line 64: This is too big to be practical. I'd prefer if there were a few important projects rather than all of them.

Julow (approved):
> let's replace this Makefile

Absolutely ! A shell script would be better because it's just running lots of commands in sequences.

gpetiot:
I've added the `base` and `core` family of projects instead.
Replacing the Makefile with a proper script can be done in a separate PR.

test-extra/Makefile
```diff
@@ -39,19 +41,23 @@ code/infer: URI = https://github.com/facebook/infer.git
 code/js_of_ocaml: URI = https://github.com/ocsigen/js_of_ocaml.git
 code/ocaml: URI = https://github.com/ocaml/ocaml.git
 code/dune: URI = https://github.com/ocaml/dune.git
-
-# DIRS += code/owl code/irmin code/index code/duniverse code/dune-release \
-# 	code/coq code/mirage code/alcotest code/citty
```
Julow, line 44: This was commented on purpose to avoid spending too much time and power unless needed. This will make the script a lot slower.

test-extra/Makefile
```diff
@@ -9,10 +9,12 @@
 #                                                                    #
 ######################################################################
 
+JS_DIRS=code/base code/base_bigstring code/base_quickcheck code/core code/core_bench code/core_compat code/core_extended code/core_kernel code/core_profiler code/core_unix
+
 # To test all source files below a directory
 #       make DIRS=<directory> test
 # By default, test projects used as regression tests
-DIRS=code/ocamlformat code/infer code/js_of_ocaml code/dune
+DIRS=code/ocamlformat code/infer code/js_of_ocaml code/dune code/owl code/irmin code/index code/dune-release code/mirage $(JS_DIRS)
```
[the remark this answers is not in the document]
gpetiot: coq and citty are too far behind so I don't really check them anymore when doing preview PRs, I can add alcotest back though

test-extra/Makefile
gpetiot: I don't see a huge difference on my side, this script is not supposed to be run as often as `dune runtest`, but we could group these repos in categories: community, mirage, janestreet for example and add an argument to decide on which groups to run the tests?

Julow: It's still possible to remove packages before running, so it's still possible to do small runs.

PR #2008, opened 2022-02-07
Format .mld files as odoc documentation files

### LINKED ISSUE #1965: Feature request: run ocamlformat on mld files
I'm starting to write readmes and manuals for my opam packages written in mld, that odoc/odig pick up and render nicely. The documents typically contain embedded OCaml code in the form of snippets or examples:

```
{@ocaml[
let the = fun () -> begin
  (* ... *)
  end
]}
```

Could it be possible to have ocamlformat run on the embedded OCaml snippets? It would make them more readable for the end-reader of the documentation and for the author as well. Even better, ocamlformat could have full support of mld files since I believe is able to reformat documentation in OCaml source code, and do simple stuff such as break lines and reformat code snippets.

gpetiot:
Closes #1965 
Just reusing the docstring parsing&formatting for .mld files.

### DIFF
--- a/lib/Docstring.ml
+++ b/lib/Docstring.ml
@@ -20,6 +20,9 @@ let parse ~loc text =
   | [] -> Ok (Odoc_parser.ast v)
   | warnings -> Error warnings
 
+let parse_file location text =
+  Odoc_parser.ast (Odoc_parser.parse_comment ~location ~text)
+
 let warn fmt warning =
   Format.fprintf fmt "Warning: Invalid documentation comment:@,%s\n%!"
     (Odoc_parser.Warning.to_string warning)
@@ -155,3 +158,7 @@ let normalize ~parse_docstrings ~normalize_code text =
     let parsed = Odoc_parser.parse_comment ~location ~text in
     let c = {normalize_code} in
     Format.asprintf "Docstring(%a)%!" (odoc_docs c) (Odoc_parser.ast parsed)
+
+let dump fmt x =
+  let c = {normalize_code= Fn.id} in
+  Format.fprintf fmt "Docstring(%a)%!" (odoc_docs c) x
--- a/lib/Normalize_extended_ast.ml
+++ b/lib/Normalize_extended_ast.ml
@@ -50,7 +50,8 @@ let normalize_parse_result ast_kind ast comments =
     comments
 
 let normalize_code conf (m : Ast_mapper.mapper) txt =
-  match Parse_with_comments.parse_toplevel conf ~source:txt with
+  let input_name = "<output>" in
+  match Parse_with_comments.parse_toplevel conf ~input_name ~source:txt with
   | First {ast; comments; _} ->
       normalize_parse_result Use_file
         (List.map ~f:(m.toplevel_phrase m) ast)
--- a/test/passing/tests/doc.mld.ref
+++ b/test/passing/tests/doc.mld.ref
@@ -1,0 +1,164 @@
+{0 Parent/Child Specification}
+
+This parent/child specification allows more flexible output support, e.g.,
+per library documentation. See
+{{:https://v3.ocaml.org/packages} v3.ocaml.org/packages}.
+
+The rules are;
+
+- [.mld] files may or may not have a parent [.mld].
+- Compilation units must have a parent [.mld].
+- The parent [.mld] file must be compiled before any of its children, and the
+  children must be specified at the parent's compilation time.
+- The output paths of [.mld] files and compilation units are subdirectories
+  of their parent's output directory.
+- The output directory of a [.mld] file [x.mld] with children is
+  [<parent_output_directory>/x], and its file name is [index.html]. That is
+  to say, [<parent_output_directory>/x/index.html]
+- The output directory of a [.mld] file [x.mld] without children is
+  [<parent_output_directory> /x.html]
+- The output directory of a compilation unit [X] is
+  [<parent_output_directory>/X/index.html]
+
+{b Note:} The [--pkg <package>] option is still supported for backward
+compatibility in [odoc >= v2.0.0], although it's now equivalent to specifying
+a parent [.mld] file.
+
+For example, let's consider [John] whose is [Doe] and [Mark]'s father. [Doe]
+has children, [Max], and page [foo], whereas [Mark] has no children. That is
+to say, [john.mld], [doe.mld], [mark.mld], [max.mld], [foo.ml] respectively.
+For instance;
+
+[john.mld]
+
+{v
+{0 About John}
+
+I'm John the father to {{!page-doe}Doe} and {{!page-mark}Mark}.
+v}
+
+[doe.mld]
+
+{v
+{0 About Doe}
+
+I'm Doe, the;
+- son to {{!page-john}John}
+- brother to {{!page-mark}Mark}
+- father to {{!page-max}Max}
+
+I also own page {{!page-foo}foo}
+v}
+
+[mark.mld]
+
+{v
+{0 About Mark}
+
+I'm Mark {{!page-doe}Doe}'s brother and I have no children.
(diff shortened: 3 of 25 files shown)


lib/Docstring.ml
```diff
@@ -22,6 +20,11 @@ let parse ~loc text =
   | [] -> Ok (Odoc_parser.ast v)
   | warnings -> Error warnings
 
+let parse_file lx =
+  let location = Lexing.lexeme_start_p lx in
+  let text = Bytes.to_string lx.lex_buffer in
```
Julow, line 25: This is not respecting the API and works only by chance, because the lexbuf is created with `Lexing.from_string`. This stops working however when the buffer is advanced, for example the shbang is skipped in `Parse_with_comments`:

```
#!/usr/bin/truc

{0 Foo}

bar
```

This is unstable because each iterations will duplicate the first line.

The ideal solution would be to modify `odoc-parser`, for example by vendoring it. Otherwise, I see two simpler solutions:

- Changing `Parse.ast` to take `location` and `source` rather than `lexbuf`. The shbang logic could be made by manual string manipulations rather than a lexer.
- Make a "lexbuf to string" function that complies with the API of `Lexing`. This is a bit wasteful as the string will be turned into a lexbuf immediately inside `odoc-parser` but that's the simplest.

Julow (commented):
There's a lexbuf problem but otherwise looks good.

lib/Docstring.ml
gpetiot: Out of curiosity would it be difficult to make odoc-parser take a lexbuf instead of string?

I tried the "change the signature" option but couldn't make it work, I will give a shot to the "lexbuf to string" option next time.

Julow: It wouldn't be difficult at all. odoc-parser is immediately creating a lexbuf out of that string. We still need a workaround until this is done and released (the next release might have a new syntax so we need strong version constraint anyway).

Julow (commented):
I get this error, which seems to come from OCaml's parser:
```
File "doc/getting_started.mld", line 98, characters 31-32:
98 | - Does not trigger warning 50 (“Unexpected documentation comment.”). For code that triggers warning 50, it is unlikely that OCamlFormat will happen to preserve the documentation
 string attachment.
                                    ^
Alert deprecated: ISO-Latin1 characters in identifiers
ocamlformat: ignoring "doc/getting_started.mld" (syntax error)
File "doc/getting_started.mld", line 98, characters 32-33:
98 | - Does not trigger warning 50 (“Unexpected documentation comment.”). For code that triggers warning 50, it is unlikely that OCamlFormat will happen to preserve the documentation
 string attachment.
                                     ^
Error: Illegal character (\128)
```

or

```
ocamlformat: ignoring "doc/howtos.mld" (syntax error)
File "doc/howtos.mld", line 12, characters 0-40:
12 | 2ceaf76b9f84cb632327c1479d0f30acfa3eeba2
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Invalid literal 2ceaf76b9f84cb632327c1479d0f30acfa3eeba2
```

I'm not sure why this happens but I think calling the wrong parser shouldn't be attempted.

Also, I think formatting of links should be improved:
```
the accepted language from {{:https://github.com/ocaml/odoc} odoc} (detailed
in {{:https://ocaml.github.io/odoc/odoc_for_authors.html} odoc's
documentation}).
```
with boxing and a fits-breaks before, it would look like this: (but I'm not sure it's the best formatting)
```
the accepted language from {{:https://github.com/ocaml/odoc} odoc} (detailed
in
{{:https://ocaml.github.io/odoc/odoc_for_authors.html}
 odoc's documentation}).
```

smorimoto:
The goal of this PR is pretty fantastic. What is the status of this?

gpetiot:
> What is the status of this?

I'm stuck addressing Jules' comment, there are too many things to do to adapt odoc-parser to feed it a lexbuf (@Julow it expects more parameters that rely on the string being provided, offset_to_location, I couldn't get the correct locations in the error messages).

Julow:
Sorry if my suggestion was wrong. Have you also tried modifying `Parse.ast` to avoid the string to lexbuf to string conversions ?

The problem with attempting to call the wrong parser remains.

gpetiot:
I finally went with your first suggestion (parser taking a string instead of a lexbuf), let me know wat you think!

lib/Normalize_extended_ast.ml
```diff
@@ -50,7 +50,8 @@ let normalize_parse_result ast_kind ast comments =
     comments
 
 let normalize_code conf (m : Ast_mapper.mapper) txt =
-  match Parse_with_comments.parse_toplevel conf ~source:txt with
+  let input_name = "<input>" in
```
Julow, line 53: Normalize is not called on user's input but on the formatting step's output, which is also the output to the user. What about `<output>` ?

test/passing/tests/doc.mld.ref
```diff
@@ -0,0 +1,164 @@
+{0 Parent/Child Specification}
+
+This parent/child specification allows more flexible output support, e.g.,
+per library documentation. See {{:https://v3.ocaml.org/packages}
+v3.ocaml.org/packages}.
```
Julow, line 5: Not really related to this PR but unanswered: Formatting of links could be improved.

test/passing/tests/doc.mld.ref
```diff
@@ -0,0 +1,164 @@
...
+
+[odoc compile-deps foo.cmt]
+
+and we shall get
+
+{[
+  Stdlib aea3513d44d604b62eaff79ad12007b3
+  Foo x5ab79b5411a3c3476029260eda0b4a26
+  CamlinternalFormatBasics f562e7b79dbe1bb1591060d6b4e854cf
+]}
```
Julow, line 162: Comment about formatting: I don't like the extra indentation in code blocks. I don't think it's needed to distinguish blocks, which already have visible delimiters (on their own line) and are surrounded by empty lines (block element).

Also, verbose blocks are not indented.

test/passing/tests/doc_repl.mld.err
```diff
@@ -0,0 +1,22 @@
+Warning: Invalid documentation comment:
+File "tests/doc_repl.mld", line 28, character 2 to line 31, character 0:
+invalid code block: not expecting: unexpected character 'v'.
+Hint: did you forget a space after the '#' at the start of the line?
```
Julow, line 4: These warnings should be silenced. Code blocks can contain many things that don't parse, for example here, MDX toplevel syntax. They spam the output and users might not want to fix them.

This can show unexpected output too, for example the `Invalid literal 2ceaf76b9f84cb632327c1479d0f30acfa3eeba2` error in my comment above, which is printed with "error" in red.

My suggestion: Remove warnings when the language of the block is not specified. (of course keep parsing)

Julow (commented):
Looks really nice :)

lib/Normalize_extended_ast.ml
gpetiot: done

test/passing/tests/doc_repl.mld.err
gpetiot: Done in #2151

test/passing/tests/doc.mld.ref
gpetiot: I agree on the principle, but I think users would complain about a diff like this. I would rather wait until a specific issue is opened to see how people respond to that.

lib/Docstring.ml
gpetiot: done

test/passing/tests/doc.mld.ref
gpetiot: done in #2152

PR #2009, opened 2022-02-08
Replace 'wrap-comments' with 'comments'

gpetiot:
I did some experiments to help the upgrade of the `tezos` repository to a more recent version. It turns out using `comments = verbatim` fixes all changes in the comments that they don't want.
Reformatting `tezos` with this patch gives:  304 files changed, 972 insertions(+), 961 deletions(-) which is better than for the 0.19.0 upgrade.

cc @emillon @raphael-proust


lib/Cmts.ml
```diff
@@ -564,10 +564,10 @@ module Unwrapped = struct
     | _ -> wrap "(*" "*)" @@ str s
 end
 
-let fmt_cmt (cmt : Cmt.t) ~wrap:wrap_comments ~ocp_indent_compat ~fmt_code
-    pos =
+let fmt_cmt (cmt : Cmt.t) ~style ~ocp_indent_compat ~fmt_code pos =
   let mode =
     match cmt.txt with
+    | txt when Poly.(style = `Verbatim) -> `Verbatim ("(*" ^ txt ^ "*)")
```
Julow, line 570: Slightly unrelated, the `ocp_indent_compat` branch of `Unwrapped.fmt` looks a lot like `` `Verbatim ``. Perhaps it can be moved to this function (there's an extra break depending on `pos`) and the `ocp_indent_compat` flag checked here.

Julow (commented):
What's the difference between `default` and `verbatim` ? The only I can find is that it breaks indentation:

```ocaml
  (* Foo

     bar *)
```

becomes:

```ocaml
(* Foo               

     bar *)
```

Do you have an example where this is useful ?

Comments are generally indented relative to the opening `(*` and OCamlformat is allowed to change its position even in verbatim mode.

lib/Conf.ml
```diff
@@ -1109,18 +1120,12 @@ module Formatting = struct
...
-       aligned are not wrapped either."
+  let ( (* wrap_comments *) ) =
+    let msg =
+      "`wrap-comments=true` is replaced with `comments=wrap, and \
+       `wrap-comments=false` is replaced with `comments=default`."
     in
-    C.flag ~default:false ~names:["wrap-comments"] ~doc ~kind
-      (fun conf x -> update conf ~f:(fun f -> {f with wrap_comments= x}))
-      (fun conf -> conf.fmt_opts.wrap_comments)
+    C.removed_option ~since:V0_21_0 ~names:["wrap-comments"] ~msg
```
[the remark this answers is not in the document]
gpetiot: done, setting up the old option would now set the new one.

gpetiot:
> What's the difference between `default` and `verbatim` ? The only I can find is that it breaks indentation:
> 
> ```ocaml
>   (* Foo
> 
>      bar *)
> ```
> 
> becomes:
> 
> ```ocaml
> (* Foo               
> 
>      bar *)
> ```
> 
> Do you have an example where this is useful ?
> 
> Comments are generally indented relative to the opening `(*` and OCamlformat is allowed to change its position even in verbatim mode.

The goal is to not have diffs like this one: https://gitlab.com/tezos/tezos/-/merge_requests/4414/diffs

I will do more tests (test_branch still running) but so far it looks good.

ocamlformat-help.txt
```diff
@@ -372,14 +372,6 @@ OPTIONS (CODE FORMATTING STYLE)
            between constructors and record fields. The default value is
            compact.
 
-       --wrap-comments
-           Wrap comments and docstrings. Comments and docstrings are divided
-           into paragraphs by open lines (two or more consecutive newlines),
-           and each paragraph is wrapped at the margin. Multi-line comments
-           with vertically-aligned asterisks on the left margin are not
-           wrapped. Consecutive comments with both left and right margin
-           aligned are not wrapped either. The flag is unset by default.
```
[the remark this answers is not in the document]
gpetiot: done

lib/Cmts.ml
gpetiot: Indeed I remember that we needed something similar for the Jane Street codebase.
@ceastlund in your experiments could you try to use my branch `verbatim-comments-janestreet` and see if it improves the results? (in this branch I'm not relying on ocp_indent_compat, but instead on comments=vrebatim to get the same behaviorm and the janestreet profile sets comments=verbatim)

Julow: This looks like what I had in mind.

Julow:
The indentation is normalized like this:
- Ignore the first line below, strip all of its indentation
- Find the line with the smallest indent
- Remove that indentation in every lines

Perhaps, not ignoring the first line can help in this case ? By using the location + the length of `(*` to determine the indentation of the first line. The case looks like this:

```ocaml
(* Foo
Very long unindented comment *)
```

Julow (commented):
I don't mind renaming the option but the `verbatim` case still has this problem: https://github.com/ocaml-ppx/ocamlformat/pull/2009#pullrequestreview-877256546
The indentation is messed up when the whole comment is re-indented. I'd expect the indentation to still be normalized but in a different way (for example allow lines to be less indented than the opening of the comment).

gpetiot:
Replaced by #2028

PR #2010, opened 2022-02-09
Use cmdliner.1.1.0

gpetiot:
Setting the dependency on version 1.1.0 since the api changed and it's better to stay up to date.


Julow (approved):
Nice.

PR #2011, opened 2022-02-11
Prevent rpc to crash on version mismatch with `.ocamlformat`

panglesd:
When loading the configuration for a path, `ocamlformat_lib` raises a `Conf_error` if it encounters a `version = ...` that does not match its version.
This made the rpc server crash, when a V2 command encountered such exception.

This PR fixes this by making the `Conf_error` public in `ocamlformat_lib`, and catching it in the `rpc`.
The error handling is also refactored.

Finally, in the `V2` protocol, the `V1.Command.output` function was used to return the errors. Both `V1.Command.output` and `V2.Command.output` are currently the same, but I still fixed this for consistency.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,8 @@
   + RPC v2 (#1935, @panglesd):
     Define a 'Format' command parameterized with optionnal arguments to set or override the config and path, to format in the style of a file.
 
+  + Prevent RPC to crash on version mismatch with `.ocamlformat` (#2011, @panglesd, @Julow)
+
 ### 0.20.1 (2021-12-13)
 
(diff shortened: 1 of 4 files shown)


panglesd:
Thanks @Julow! I think you are right that it is a better approach.

I hesitated to change the type of `build_config` to return a `result`, but I though I'd rather not change the type of something public (even though it is supposed to be public only internally).

Julow (approved):
It was only half public if the exception wasn't exposed.

gpetiot (approved):
Looks good to me, thanks!
I added an entry in the changelog since this is a bugfix.

PR #2015, opened 2022-02-21
Do not use dune 3.0

gpetiot:
See https://github.com/ocaml/opam-repository/pull/20765

### DIFF
--- a/ocamlformat.opam
+++ b/ocamlformat.opam
@@ -8,11 +8,11 @@ authors: ["Josh Berdine <jjb@fb.com>"]
 homepage: "https://github.com/ocaml-ppx/ocamlformat"
 bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
 depends: [
-  "dune" {>= "2.8"}
   "ocaml" {>= "4.08" & < "4.15"}
   "alcotest" {with-test}
   "base" {>= "v0.12.0" & < "v0.15"}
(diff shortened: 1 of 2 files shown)


ocamlformat.opam
```diff
@@ -8,11 +8,11 @@ authors: ["Josh Berdine <jjb@fb.com>"]
 homepage: "https://github.com/ocaml-ppx/ocamlformat"
 bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
 depends: [
-  "dune" {>= "2.8"}
   "ocaml" {>= "4.08" & < "4.15"}
   "alcotest" {with-test}
   "base" {>= "v0.12.0" & < "v0.15"}
   "cmdliner" {>= "1.1.0"}
+  "dune" {>= "2.8" & with-test & < "3.0"}
```
Julow, line 15: I think you meant this:
```suggestion
  "dune" {>= "2.8"}
  "dune" {with-test & < "3.0"}
```

Julow (approved):

ocamlformat.opam
gpetiot: Yes but this is generated from the `dune-project` file

Julow: Oups! You are right. I've found a way to fix this: https://github.com/gpetiot/ocamlformat/pull/10

gpetiot: Thanks!

PR #2018, opened 2022-02-21
Vendor odoc-parser.1.0.0

gpetiot:
Vendored with https://github.com/brettlangdon/git-vendor

The goal is to then modify odoc-parser according to https://github.com/ocaml-ppx/ocamlformat/pull/2008


gpetiot:
Apparently this makes it impossible to rebase so this is not gonna work, I think it's better to wait for the API of odoc-parser to work with a lexbuf.

Julow:
> Apparently this makes it impossible to rebase

Why is that ?

PR #2020, opened 2022-02-22
Define a diff-friendly profile

### LINKED ISSUE #1958: Feature request: Update default profile to use indicate-multiline-delimiters = closing-on-separate-line

The `indicate-multiline-delimiters = closing-on-separate-line` setting was removed in ocamlformat 0.20.0. I've personally found this feature very useful and had good feedback from other developers on the same codebase after it was introduced. In particular, it makes modifying anything with a closing delimiter easier because the last line of the delimited body is no longer special. This also makes diffs easier to understand as the `)`/`]`/`}` no longer clutters diffs when the delimited body changes.

I would like to have `indicate-multiline-delimiters = closing-on-separate-line` as part of the `default` profile. As @gpetiot pointed out to me, this will cause a lot of churn for users of the `default` profile! So this seems like it's worth considering before a 1.0.0 release.

**Additional context**
 
Issues:
* Existing request for closing parens on a separate line: https://github.com/ocaml-ppx/ocamlformat/issues/1578
* Other issues mentioning this feature which may indicate interest in it - leaving out my own to avoid too much bias!:
    * https://github.com/ocaml-ppx/ocamlformat/issues/1762
    * https://github.com/ocaml-ppx/ocamlformat/issues/1596
    * https://github.com/ocaml-ppx/ocamlformat/issues/927

PRs:
* Deprecation of the feature: https://github.com/ocaml-ppx/ocamlformat/pull/1900
* Original implementation: https://github.com/ocaml-ppx/ocamlformat/pull/876

gpetiot:
Closes #1958

This is a new profile I wanted to work on for quite some time. It doesn't impact existing profiles.

New values are added to a few existing options, that users can set individually, although I advise to just use this profile if they want this behavior.
- `if-then-else = vertical` (#2174)
- `break-cases = vertical` (#2176)

It also adds new fields to the `Conf.fmt_opts` type that should not become new options, this is just how we implement behavior switches in ocamlformat instead of matching on the profile.
- `unfold-let-binding : bool`

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -180,6 +180,72 @@ let ocamlformat_profile from =
   ; wrap_docstrings= elt true
   ; wrap_fun_args= elt true }
 
+let diff_friendly_profile from =
+  let elt content = Elt.make content from in
+  { align_symbol_open_paren= elt false
+  ; assignment_operator= elt `End_line
+  ; break_before_in= elt `Fit_or_vertical
+  ; break_cases= elt `Vertical
+  ; break_collection_expressions= elt `Fit_or_vertical
+  ; break_colon= elt `After
+  ; break_infix= elt `Fit_or_vertical
+  ; break_infix_before_func= elt true
+  ; break_fun_decl= elt `Fit_or_vertical
+  ; break_fun_sig= elt `Fit_or_vertical
+  ; break_separators= elt `After
+  ; break_sequences= elt true
+  ; break_string_literals= elt `Auto
+  ; break_struct= elt true
+  ; cases_exp_indent= elt 4
+  ; cases_matching_exp_indent= elt `Compact
+  ; disambiguate_non_breaking_match= elt false
+  ; doc_comments= elt `Before_except_val
+  ; doc_comments_padding= elt 2
+  ; doc_comments_tag_only= elt `Default
+  ; dock_collection_brackets= elt true
+  ; exp_grouping= elt `Parens
+  ; extension_indent= elt 2
+  ; field_space= elt `Tight
+  ; function_indent= elt 2
+  ; function_indent_nested= elt `Never
+  ; if_then_else= elt `Vertical
+  ; indent_after_in= elt 0
+  ; indicate_multiline_delimiters= elt `Closing_on_separate_line
+  ; indicate_nested_or_patterns= elt `Unsafe_no
+  ; infix_precedence= elt `Indent
+  ; leading_nested_match_parens= elt false
+  ; let_and= elt `Sparse
+  ; let_binding_indent= elt 2
+  ; let_binding_deindent_fun= elt true
+  ; let_binding_spacing= elt `Compact
+  ; let_module= elt `Sparse
+  ; line_endings= elt `Lf
+  ; margin= elt 80
+  ; match_indent= elt 0
+  ; match_indent_nested= elt `Never
+  ; max_indent= elt None
+  ; module_item_spacing= elt `Sparse
+  ; nested_match= elt `Wrap
+  ; ocp_indent_compat= elt false
+  ; parens_ite= elt false
+  ; parens_tuple= elt `Always
+  ; parens_tuple_patterns= elt `Multi_line_only
+  ; parse_docstrings= elt false
+  ; parse_toplevel_phrases= elt false
+  ; sequence_blank_line= elt `Compact
+  ; sequence_style= elt `Separator
+  ; single_case= elt `Sparse
+  ; space_around_arrays= elt false
+  ; space_around_lists= elt false
+  ; space_around_records= elt false
+  ; space_around_variants= elt false
+  ; stritem_extension_indent= elt 0
+  ; type_decl= elt `Sparse
+  ; type_decl_indent= elt 2
+  ; wrap_comments= elt false
+  ; wrap_docstrings= elt true
+  ; wrap_fun_args= elt false }
+
 let janestreet_profile from =
   let elt content = Elt.make content from in
   { align_symbol_open_paren= elt false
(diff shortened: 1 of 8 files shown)


lib/Conf.ml
```diff
@@ -1571,6 +1574,70 @@ let conventional_profile =
...
+  { align_cases= false
+  ; align_constructors_decl= false
+  ; align_variants_decl= false
+  ; assignment_operator= `End_line
+  ; break_before_in= `Fit_or_vertical
+  ; break_cases= `All (* should always unfold *)
+  ; break_collection_expressions= `Fit_or_vertical
+  ; break_infix= `Fit_or_vertical
+  ; break_infix_before_func= true
+  ; break_fun_decl= `Fit_or_vertical (* should always unfold *)
```
gpetiot, line 1587: todo

lib/Conf.ml
```diff
@@ -1571,6 +1574,70 @@ let conventional_profile =
...
+  ; break_fun_sig= `Fit_or_vertical (* should always unfold *)
```
gpetiot, line 1588: todo

lib/Conf.ml
```diff
@@ -1571,6 +1574,70 @@ let conventional_profile =
 
 let default_profile = conventional_profile
 
+let diff_friendly_profile =
+  { align_cases= false
+  ; align_constructors_decl= false
+  ; align_variants_decl= false
+  ; assignment_operator= `End_line
+  ; break_before_in= `Fit_or_vertical
+  ; break_cases= `All (* should always unfold *)
```
gpetiot, line 1583: todo

test/passing/tests/source-diff-friendly.ml.ref
```diff
@@ -0,0 +1,11226 @@
...
+
+type 'a with_default =
+  ?size:int (** default [42] *) -> ?resizable:bool (** default [true] *) -> 'a
+
+type obj =
+  < meth1: int -> int  (** method 1 *) ; meth2: unit -> float  (** method 2 *) >
+
+type var =
+  [ `Foo  (** foo *)
+  | `Bar of int * string  (** bar *) ]
```
smondet, line 35: isn't this more diff-friendly:

```ocaml
type var = [
  | `Foo  (** foo *)
  | `Bar of int * string  (** bar *)
]
```

it would also be more cconsistent with records and regular variants


e.g. when adding or reordering cases

test/passing/tests/source-diff-friendly.ml.ref
```diff
@@ -0,0 +1,11226 @@
...
+let _ =
+  Obj.extension_constructor 7 (* Invald_arg *)
+
+let _ =
+  Obj.extension_constructor
+    (object
+       method m = 3
+    end
+    )
+(* Invald_arg *)
```
smondet, line 654: Should this be something like:

```ocaml
let _ =
  Obj.extension_constructor
    (
     object
       method m = 3
     end
    )
```

test/passing/tests/source-diff-friendly.ml.ref
```diff
@@ -0,0 +1,11226 @@
...
+;;
+
+print_endline
+  (string_of_int
+     (f
+        (Some
+           ( module struct
+             let x =
+               false
+           end
```
smondet, line 1040: the `end` does not match the `module`

test/passing/tests/source-diff-friendly.ml.ref
```diff
@@ -0,0 +1,11226 @@
...
+        raise VariantMismatch ;
+      let builder =
+        create_builder ()
+      in
+      List.iter2
+        (fun (Field {label; field_type; set}) (lab, v) ->
+          if label <> lab then
+            raise VariantMismatch ;
+          set builder (devariantize field_type v)
+        )
```
smondet, line 1268: this a place were ``exp_grouping= `Preserve`` could make more sense (?)

smondet: (I meant letting users put `begin .. end`)

tmattio:
I just tried this on some code and **really** like it 🙂 

Some notes:

```diff
-  val ( and+ )
-    :  'a Cmdliner.Term.t
-    -> 'b Cmdliner.Term.t
-    -> ('a * 'b) Cmdliner.Term.t
+  val ( and+ ) :
+    'a Cmdliner.Term.t -> 'b Cmdliner.Term.t -> ('a * 'b) Cmdliner.Term.t
```

The former seems more diff friendly?

```diff
-  Fmt_tty.setup_std_outputs ();
-  Logs.set_level log_level;
-  Logs.set_reporter (Logs_fmt.reporter ~app:Fmt.stdout ());
+  Fmt_tty.setup_std_outputs () ;
+  Logs.set_level log_level ;
+  Logs.set_reporter (Logs_fmt.reporter ~app:Fmt.stdout ()) 
```

Why add spaces there? It doesn't seem to improve the diff. I'm assuming this is something coming from the sparse profile, but should we use the default config here?

```diff
 let handle_errors = function
-  | Ok () -> if Logs.err_count () > 0 then 3 else 0
+  | Ok () ->
+      if Logs.err_count () > 0 then
+        3
+      else
+        0
   | Error err ->
-      Logs.err (fun m -> m "%s" (Error.to_string err));
+      Logs.err (fun m -> m "%s" (Error.to_string err)) ;
       error_to_code err
```

This does seem to be a regression, we don't need two level of indentation here I think.

```diff
 let term =
   let open Common.Syntax in
-  let+ _term = Common.term
+  let+ _term =
+    Common.term
   and+ name =
-    let doc = "The name to greet." in
-    let docv = "NAME" in
+    let doc =
+      "The name to greet."
+    in
+    let docv =
+      "NAME"
+    in
```

That indeed could improve the diff, but perhaps breaking every let binding into three lines is a bit much? 😅 

My feedback aside, thanks a lot for working on this @gpetiot, I'm very happy to see a profile that uses objective guidelines as an alternative to the other profiles.

PR #2021, opened 2022-02-23
README.md: Latest version of ocamlformat

OscarLahaie:
The newest version of ocamlformat is no longer the 0.20.0 but the 0.20.1.
I suggest to replace in .ocamlformat, the part where the version of is indicated.

`version = 0.20.0`

to 

`version = 0.20.1`


gpetiot (approved):
Thanks!

PR #2022, opened 2022-02-23
Expose ocamlformat libraries

tmattio:
This allows users to use ocamlformat as a library.

We give a public name `ocamlformat` to the library and each library is exposed as a sublibrary of `ocamlformat` (e.g. `ocamlformat.parse_wyc` and `ocamlformat.parser_shims`).

I also cleanup the dune-project file to remove the duplicated dependencies. `ocamlformat-rpc-lib` and `ocamlformat-bench` now depends on `ocamlformat {= version}`. I can remove this part if needed.

### DIFF
--- a/ocamlformat-rpc.opam
+++ b/ocamlformat-rpc.opam
@@ -10,24 +10,9 @@ bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
 depends: [
   "dune" {>= "2.8"}
   "ocaml" {>= "4.08" & < "4.15"}
-  "ocamlformat-rpc-lib"
+  "ocamlformat" {= version}
+  "ocamlformat-rpc-lib" {= version}
   "alcotest" {with-test}
-  "base" {>= "v0.12.0" & < "v0.15"}
-  "cmdliner" {>= "1.1.0"}
-  "dune-build-info"
-  "either"
-  "fix"
-  "fpath"
-  "menhir" {>= "20201216"}
-  "menhirLib" {>= "20201216"}
-  "menhirSdk" {>= "20201216"}
-  "ocaml-version" {>= "3.3.0"}
-  "ocp-indent"
-  "odoc-parser" {>= "1.0.0"}
-  "re" {>= "1.7.2"}
-  "stdio" {< "v0.15"}
-  "uuseg" {>= "10.0.0"}
-  "uutf" {>= "1.0.1"}
   "odoc" {with-doc}
 ]
 build: [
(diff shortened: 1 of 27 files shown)


ocamlformat.opam
```diff
@@ -28,6 +25,8 @@ depends: [
   "stdio" {< "v0.15"}
   "uuseg" {>= "10.0.0"}
   "uutf" {>= "1.0.1"}
+  "alcotest" {with-test}
+  "dune" {>= "2.8" & with-test & < "3.0"}
```
Julow, line 29: That's why we had two `dune` in the depends for this package. The first one would be updated by Dune while generating the opam file and the other would be kept.

tmattio: Thanks, I added it back 🙂

gpetiot (approved):
Looks good to me, thanks!
Feel free to merge when ready or let us know and we will.

tmattio:
Thanks @gpetiot!
It's ready - I don't have the permission to merge though, so feel free to do it 🙂

ocamlformat-rpc.opam
```diff
@@ -10,24 +10,9 @@ bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
...
-  "menhirLib" {>= "20201216"}
-  "menhirSdk" {>= "20201216"}
-  "ocaml-version" {>= "3.3.0"}
-  "ocp-indent"
-  "odoc-parser" {>= "1.0.0"}
-  "re" {>= "1.7.2"}
-  "stdio" {< "v0.15"}
-  "uuseg" {>= "10.0.0"}
-  "uutf" {>= "1.0.1"}
+  "ocamlformat" {= version}
```
panglesd, line 14: Why should `ocamlformat-rpc` have the same version as `ocamlformat`? This is making some dependencies problem... (see https://github.com/ocaml/ocaml-lsp/pull/599#issuecomment-1083375199)

Julow: `ocamlformat-rpc` was the rpc server, it is specific to the version of ocamlformat. Since recently, the RPC server is part of the `ocamlformat` package, no more confusion.
I think you meant to use `ocamlformat-rpc-lib`, which is the library for defining a client and doesn't depends on ocamlformat.

panglesd: Okay. I did meant `ocamlformat-rpc`, as this breaks the `ocaml-lsp` CI (`ocaml-lsp` CI needs `ocamlformat-rpc.0.21.0` (for testing) and `ocamlformat.0.19.0` (for checking formatting). But I guess that's logic to bound those version, I'm sure it can be worked out in ocaml-lsp CI :)

PR #2028, opened 2022-02-28
New syntax (*= ... *) for verbatim comments

gpetiot:
Follow-up of #1810 and #2009

Introduces a new syntax `(*= ... *)` to print comments "verbatim".

Here is the preview on tezos' code: https://gitlab.com/tezos/tezos/-/merge_requests/4587

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -533,34 +533,39 @@ module Unwrapped = struct
     in
     vbox 0 ~name:"multiline" (list_fl unindented fmt_line $ fmt_opt epi)
 
-  let fmt ~ocp_indent_compat Cmt.{txt= s; loc} (pos : Cmt.pos) =
+  let fmt Cmt.{txt= s; loc} =
     let open Fmt in
     let is_sp = function ' ' | '\t' -> true | _ -> false in
     match String.split_lines (String.rstrip s) with
     | first_line :: (_ :: _ as tl) when not (String.is_empty first_line) ->
-        if ocp_indent_compat then
-          (* Not adding artificial breaks and keeping the comment contents
-             verbatim will not interfere with ocp-indent. *)
-          match pos with
-          | Before -> wrap "(*" "*)" @@ str s
-          | Within -> wrap "(*" "*)" @@ str s
-          | After -> break_unless_newline 1000 0 $ wrap "(*" "*)" @@ str s
-        else
-          let epi =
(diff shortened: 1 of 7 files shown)


raphael-proust:
What's the interaction with mli docstrings? Specifically,

- Do I ever need this new syntax for special comments (`(**`-`*)`) in an mli?
- Do they show up in the generated html documentation?
- Does it break other markup features of the html documentation generation? What does `(**= {1 …} *)` render to?

gpetiot:
> What's the interaction with mli docstrings? Specifically,
> 
>     * Do I ever need this new syntax for special comments (`(**`-`*)`) in an mli?

No, by default docstrings are not formatted unless `--parse-docstrings` is set, which is not the case on tezos codebase.

> 
>     * Do they show up in the generated html documentation?

I think only docstrings comments appear in the html doc, not the regular comments.

> 
>     * Does it break other markup features of the html documentation generation? What does `(**= {1 …} *)` render to?

It would break the odoc parser, we don't plan to add this syntax for docstrings since we do not format them by default (see 1st point).

lib/Cmts.ml
```diff
@@ -533,34 +533,39 @@ module Unwrapped = struct
...
+    let open Fmt in
+    match String.split_lines (String.rstrip s) with
+    | first_line :: _ :: _ when not (String.is_empty first_line) -> (
+      (* Not adding artificial breaks and keeping the comment contents
+         verbatim will not interfere with ocp-indent. *)
+      match pos with
+      | Before -> wrap "(*" "*)" @@ str s
+      | Within -> wrap "(*" "*)" @@ str s
+      | After -> break_unless_newline 1000 0 $ wrap "(*" "*)" @@ str s )
     | _ -> wrap "(*" "*)" @@ str s
```
Julow, line 569: This can be simplified a lot: `wrap` and `str` are common to every cases, the only special case would be `fmt_if (pos = After && String.contains s '\n' && not (String.starts_with ~prefix:"\n" s)) (break_unless_newline 1000 0)`. I'm not even sure the newline checks are necessary.

Julow (approved):
I think we need a syntax like this.

lib/Cmts.ml
gpetiot: done

gpetiot:
I'm about to merge this PR, @raphael-proust is everything alright with you?

raphael-proust:
Yep. That's fine.

PR #2034, opened 2022-03-01
Do not re-format Qtest comments

### LINKED ISSUE #2033: bug: ocamlformat modifies non-docstring comments
**Describe the bug**
Some comments like:

```
(*$inject
…
*)
```

are modified into:

```
(*$
  inject
*)
```
which breaks tooling that relies on these, such as https://github.com/vincent-hugot/qtest/ .

gpetiot:
Fix #2033

It would add a lot of specific code to format the 5 different kinds of comments or so, but it doesn't cost much to ignore these comments, which seems to be enough. What do you think?

### DIFF
--- a/lib/Cmts.ml
+++ b/lib/Cmts.ml
@@ -574,6 +574,8 @@ let fmt_cmt (cmt : Cmt.t) ~wrap:wrap_comments ~ocp_indent_compat ~fmt_code
     | "*" when Location.width cmt.loc = 4 -> `Verbatim ""
     | "*" -> `Verbatim "*"
     | "$" -> `Verbatim "$"
+    | str when Char.(str.[0] = '$' && not (is_whitespace str.[1])) ->
+        `Qtest cmt
     | str when Char.equal str.[0] '$' -> (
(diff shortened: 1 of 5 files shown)


Julow (commented):
Perhaps the code can be improved a bit if rebased on top of the refactors in https://github.com/ocaml-ppx/ocamlformat/pull/2028 ?

gpetiot:
@c-cube is this change working for you?

lib/Cmts.ml
```diff
@@ -574,6 +574,8 @@ let fmt_cmt (cmt : Cmt.t) ~wrap:wrap_comments ~ocp_indent_compat ~fmt_code
     | "*" when Location.width cmt.loc = 4 -> `Verbatim ""
     | "*" -> `Verbatim "*"
     | "$" -> `Verbatim "$"
+    | str when Char.(str.[0] = '$' && not (is_whitespace str.[1])) ->
+        `Qtest cmt
```
Julow, line 578: Could be `` `Verbatim str `` ?

Julow (approved):

c-cube:
Yes, it seems to do the trick! Thank you :)

PR #2036, opened 2022-03-10
Prevent comments from moving after a keyword inside a sub-expression

### LINKED ISSUE #1582: Feature request: consistent placement of comments before nested let
The following example is OCamlformatted:

```ocaml
;;
assert (
  (* [a] is true *)
  let a = true in
  a )
```

This placement of the comment before the `let` binding looks correct to me, but the equivalents for `let module` and `let exception` do not behave in the same way. In particular, the comments get put somewhere inside the `let` binding itself:

```ocaml
;;
assert (
  let module (* [A.v] is true *)
      A = struct
    let v = true
  end in
  A.v || b )

;;
assert (
  let exception (* [Foo] is raised *)
                  Foo in
  raise Foo )
```

These last two examples look broken to me. Note that the same issue doesn't happen for `let`-bindings not inside a function parameter:

```ocaml
let () =
  (* [a] is true *)
  let a = true in
  (* [A.v] is true *)
  let module A = struct
    let v = true
  end in
  (* [Foo] is raised *)
  let exception Foo in
  ()
```

gpetiot:
Fix #1582

No bug found with test_branch

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
   + emacs: Remove temp files in the event of an error (#2003, @gpetiot)
 
+  + Prevent comments from moving after a keyword inside a sub-expression (#2036, @gpetiot)
+
 #### Changes
 
   + Qtest comments are not re-formatted (#2034, @gpetiot)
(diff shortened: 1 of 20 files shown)


gpetiot:
Wrong fix, let's re-evaluate this issue when the AST has become more concrete (we will have more precise locations to attach comments to).

PR #2037, opened 2022-03-11
Doc-comments code blocks with a language other than 'ocaml' (set in metadata) are not parsed as OCaml

### LINKED ISSUE #1597: Feature request: An option to allow code-blocks in docstrings to pass through unmolested, but still parse/wrap the *rest* of the docstring
In writing a parser for a non-OCaml language, `--parse-docstrings` attempts to treat my codeblocks as valid OCaml source-code, thus changing the semantics in their *actual* language. i.e.:

```diff
@@ -70,7 +70,7 @@ and word = string or_subexpr list
     A single shell "word" can actually be composed of multiple quotations or
     subexpressions, as long as they are not separated by whitespace:
 
-    {[ echo "this""is""only""a""single"(echo word)(echo also) ]}
+    {[ echo "this" "is" "only" "a" "single" (echo word) (echo also) ]}
 
     Such a composition is represented in this type, by a series of {!or_subexpr}s in a
     [list]. *)
```

I understand (and in other circumstances, appreciate!) the purpose behind this behaviour. Unfortunately, it's really problematic in some circumstances (like mine, here.)

Ideally, I'd love it if "is this OCaml and should it be formatted" were determined *per-codeblock*, as it is in Markdown (i.e. `{[ocaml Here is some code ]}`), but such an extension seems rather unlikely.

In the absence of that, I'd at least like to be able to have my docstrings wrapped, and just disable all codeblock-formatting across the entire project; perhaps `--no-parse-docstring-codeblocks`?

(See, relatedly, Prettier's [`--embedded-language-formatting`](https://prettier.io/docs/en/options.html#embedded-language-formatting) option)

gpetiot:
Fix #1597

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 
   + ocamlformat-rpc is now distributed through the ocamlformat package (#2035, @Julow)
 
+  + Doc-comments code blocks with a language other than 'ocaml' (set in metadata) are not parsed as OCaml (#2037, @gpetiot)
+
 #### New features
 
   + New syntax `(*= ... *)` for verbatim comments (#2028, @gpetiot)
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -75,25 +75,31 @@ let fmt_code_block conf s1 s2 =
   let wrap_code x =
     str "{" $ opt s1 fmt_metadata $ fmt "[@;<1000 2>" $ x $ fmt "@ ]}"
   in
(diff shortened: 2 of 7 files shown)


Julow (commented):
Yes, we need to match this !
However, I think we should continue to format code blocks that don't have this field. Odoc will continue to highlight them as ocaml code.

gpetiot:
> Yes, we need to match this ! However, I think we should continue to format code blocks that don't have this field. Odoc will continue to highlight them as ocaml code.

Done!

PR #2039, opened 2022-03-14
Do not merge global .ocamlformat config file with project configuration

### LINKED ISSUE #1459: Feature request: don't merge global config files with project ones
I recently added a global `.ocamlformat` config file (inside `$XDG_CONFIG_HOME`) and set ` '(ocamlformat-enable 'enable-outside-detected-project)`, intending to use those styles for files / scripts that don't live inside repositories. I was surprised to find this caused my _projects_ to be reformatted with the global styles, even though those have their own `.ocamlformat` files. Looking at the documentation in the README, this behaviour is expected:

> .ocamlformat files in the containing and all ancestor directories for each input file are used, as well as the global .ocamlformat file defined in `$XDG_CONFIG_HOME/ocamlformat`. The global .ocamlformat file has the lowest priority, then the closer the directory is to the processed file, the higher the priority.

I find this behaviour problematic, since it means that using a global config file makes project formatting non-reproducible across machines. We've recently had some trouble with external contributors accidentally reformatting projects with a different style, and I suspect this is the cause.

Ideally, I would expect the behaviour to be:

 - When inside a detected project, configs are applied in top-down starting from the project root (i.e. excluding the global config file)
 - Otherwise, when `--enable-outside-detected-project` is set, apply the global config file.

Note that it's possible to get around the inheriting behaviour by setting `profile = conventional` in project `.ocamlformat` files, but we recently stopped doing that when it became the default. If the inheritance behaviour is to be kept the way it is, I suggest changing the [recommended `.ocamlformat` file](https://github.com/ocaml-ppx/ocamlformat#what-configuration-should-i-use) to include `profile = conventional` in the interests of reproducibility.

CC @Julow

gpetiot:
Fix #1459 (cc @CraigFe)

The doc should reflect the changes, the behavior is what has been discussed in #1459 

<details>

- if a project root is found:
  * apply the local .ocamlformat files, not the global conf
- else
  * if `enable-outside-detected-project` is set
    + apply .ocamlformat files up to `/`
    + if no file has been found then apply the global conf if it exists

</details>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 - ocamlformat-rpc is now distributed through the ocamlformat package (#2035, @Julow)
 - Doc-comments code blocks with a language other than 'ocaml' (set in metadata) are not parsed as OCaml (#2037, @gpetiot)
 - More comprehensible error message in case of version mismatch (#2042, @gpetiot)
+- The global configuration file (`$XDG_CONFIG_HOME` or `$HOME/.config`) is only applied when no project is detected, `--enable-outside-detected-project` is set, and no applicable `.ocamlformat` file has been found. Global and local configurations are no longer used at the same time. (#2039, @gpetiot)
 
 ### New features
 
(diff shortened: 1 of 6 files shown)


gpetiot:

PR #2040, opened 2022-03-15
Break after 'fun ->' following a monadic op

gpetiot:
No diff for the janestreet and ocamlformat profiles.

cc @samoht for the changes on the conventional profile.

In addition to what we can see in the tests, test_branch.sh exhibits changes like:

```diff
iff --git a/iobuf/test/test_iobuf.ml b/iobuf/test/test_iobuf.ml
index 8eb8a0e..092e5f3 100644
--- a/iobuf/test/test_iobuf.ml
+++ b/iobuf/test/test_iobuf.ml
@@ -290,9 +290,10 @@ struct
     assert (n = 10);
     Poke.char t ~pos:0 'a';
     Poke.char t ~pos:1 'b';
-    ( sub_shared t |> fun t ->
-      assert (length t = n);
-      assert (Consume.char t = 'a') );
+    (sub_shared t
+    |> fun t ->
+    assert (length t = n);
+    assert (Consume.char t = 'a'));
     assert (Consume.char t = 'a');
     assert (Consume.char t = 'b')
 
@@ -698,7 +699,8 @@ struct
             assert (advance_fails (-1));
             assert (advance_fails (len + 1)));
           for amount = 0 to len do
-            sub_shared t |> fun t ->
+            sub_shared t
+            |> fun t ->
             advance t amount;
             assert (length t = len - amount)
```

There is no special handling for `|> fun ... ->` since `|>` is not considered to be a monadic operator anymore (using the same definition as ocp-indent). The changes of spacing between the parentheses is unfortunate, it would need to be normalized, but if we do we would also get a huge diff on code using this profile.

Here is the diff on irmin:

<details>

```diff
diff --git a/bench/irmin-pack/_layers.ml b/bench/irmin-pack/_layers.ml
index a3c564dde6..f740839752 100644
--- a/bench/irmin-pack/_layers.ml
+++ b/bench/irmin-pack/_layers.ml
@@ -131,7 +131,9 @@ let first_5_cycles config repo =
   let rec aux i c =
     add_min c;
     if i > 4 then Lwt.return c
-    else write_cycle config repo c >>= fun c -> aux (i + 1) c
+    else
+      write_cycle config repo c >>= fun c ->
+      aux (i + 1) c
   in
   aux 0 c
 
diff --git a/bench/irmin-pack/bench_common.ml b/bench/irmin-pack/bench_common.ml
index 52bd500a8e..58b39a4de0 100644
--- a/bench/irmin-pack/bench_common.ml
+++ b/bench/irmin-pack/bench_common.ml
@@ -33,7 +33,8 @@ let reporter ?(prefix = "") () =
         Fmt.(styled `Magenta string)
         (Logs.Src.name src)
     in
-    msgf @@ fun ?header ?tags fmt -> with_stamp header tags k fmt
+    msgf @@ fun ?header ?tags fmt ->
+    with_stamp header tags k fmt
   in
   { Logs.report }
 
diff --git a/bench/irmin-pack/trace_replay.ml b/bench/irmin-pack/trace_replay.ml
index 8d45d932f9..f610a2e049 100644
--- a/bench/irmin-pack/trace_replay.ml
+++ b/bench/irmin-pack/trace_replay.ml
@@ -321,7 +321,9 @@ module Make (Store : Store) = struct
 
     let rec aux commit_seq i =
       match commit_seq () with
-      | Seq.Nil -> on_end () >|= fun () -> i
+      | Seq.Nil ->
+          on_end () >|= fun () ->
+          i
       | Cons (ops, commit_seq) ->
           let* () = add_operations t repo ops i stats check_hash empty_blobs in
           let len0 = Hashtbl.length t.contexts in
diff --git a/bench/irmin-pack/trace_stat_summary_cb.ml b/bench/irmin-pack/trace_stat_summary_cb.ml
index 7e30a70c84..d86b8141d7 100644
--- a/bench/irmin-pack/trace_stat_summary_cb.ml
+++ b/bench/irmin-pack/trace_stat_summary_cb.ml
@@ -175,7 +175,8 @@ let create_raw_results2 s : result list =
       ("inode_decode_bin", s.pack.inode_decode_bin);
       ("inode_encode_bin", s.pack.inode_encode_bin);
     ]
-    |> List.map @@ fun (name, lbs) -> (name, [ lbs.value_after_commit.diff ])
+    |> List.map @@ fun (name, lbs) ->
+       (name, [ lbs.value_after_commit.diff ])
   in
   [ { name; metrics } ]
 
@@ -195,7 +196,8 @@ let create_raw_results3 s : result list =
       ("node_val_find", s.tree.node_val_find);
       ("node_val_list", s.tree.node_val_list);
     ]
-    |> List.map @@ fun (name, lbs) -> (name, [ lbs.value_after_commit.diff ])
+    |> List.map @@ fun (name, lbs) ->
+       (name, [ lbs.value_after_commit.diff ])
   in
   [ { name; metrics } ]
 
@@ -223,7 +225,8 @@ let create_raw_results5 s : result list =
       ("promoted words", s.gc.promoted_words);
       ("compactions", s.gc.compactions);
     ]
-    |> List.map @@ fun (name, lbs) -> (name, [ lbs.value_after_commit.diff ])
+    |> List.map @@ fun (name, lbs) ->
+       (name, [ lbs.value_after_commit.diff ])
   in
   let metrics' =
     [
diff --git a/src/irmin-chunk/irmin_chunk.ml b/src/irmin-chunk/irmin_chunk.ml
index 11bff712a4..4e4a0b6fb1 100644
--- a/src/irmin-chunk/irmin_chunk.ml
+++ b/src/irmin-chunk/irmin_chunk.ml
@@ -178,7 +178,9 @@ struct
               else List.length l
             in
             match list_partition n l with
-            | [ i ] -> AO.add t.db key (index t i) >|= fun () -> key
+            | [ i ] ->
+                AO.add t.db key (index t i) >|= fun () ->
+                key
             | l -> Lwt_list.map_p (fun i -> CA.add t.db (index t i)) l >>= aux)
       in
       aux l
@@ -222,7 +224,9 @@ struct
     | Some bufs -> (
         let buf = String.concat "" bufs in
         match value_of_bin_string buf with
-        | Ok va -> check_hash key va >|= fun () -> Some va
+        | Ok va ->
+            check_hash key va >|= fun () ->
+            Some va
         | Error _ -> Lwt.return_none)
 
   let list_range ~init ~stop ~step =
diff --git a/src/irmin-fs/irmin_fs.ml b/src/irmin-fs/irmin_fs.ml
index eca29ff453..eb2df18ad6 100644
--- a/src/irmin-fs/irmin_fs.ml
+++ b/src/irmin-fs/irmin_fs.ml
@@ -81,7 +81,8 @@ struct
 
   let v config =
     let path = get_path config in
-    IO.mkdir path >|= fun () -> { path }
+    IO.mkdir path >|= fun () ->
+    { path }
 
   let close _ = Lwt.return_unit
   let cast t = (t :> read_write t)
@@ -204,7 +205,10 @@ struct
     in
     { t; w }
 
-  let close t = W.clear t.w >>= fun () -> RO.close t.t
+  let close t =
+    W.clear t.w >>= fun () ->
+    RO.close t.t
+
   let find t = RO.find t.t
   let mem t = RO.mem t.t
   let list t = RO.list t.t
@@ -230,7 +234,10 @@ struct
     let+ w = W.watch t.w ?init f in
     (w, stop)
 
-  let unwatch t (id, stop) = stop () >>= fun () -> W.unwatch t.w id
+  let unwatch t (id, stop) =
+    stop () >>= fun () ->
+    W.unwatch t.w id
+
   let raw_value = Irmin.Type.(unstage (to_bin_string V.t))
 
   let set t key value =
diff --git a/src/irmin-git/backend.ml b/src/irmin-git/backend.ml
index cb334d6a55..6016c60782 100644
--- a/src/irmin-git/backend.ml
+++ b/src/irmin-git/backend.ml
@@ -110,7 +110,9 @@ struct
       let+ b = Branch.v ~head ~bare g in
       { g; b; closed = ref false; config = (conf :> Irmin.config) }
 
-    let close t = Branch.close t.b >|= fun () -> t.closed := true
+    let close t =
+      Branch.close t.b >|= fun () ->
+      t.closed := true
   end
 
   module Remote = struct
diff --git a/src/irmin-git/irmin_git.ml b/src/irmin-git/irmin_git.ml
index fe12218290..d5dbc0427c 100644
--- a/src/irmin-git/irmin_git.ml
+++ b/src/irmin-git/irmin_git.ml
@@ -330,7 +330,8 @@ struct
         let close t =
           Contents.CA.close t.contents >>= fun () ->
           Node.CA.close (snd t.nodes) >>= fun () ->
-          Commit.CA.close (snd t.commits) >>= fun () -> Branch.close t.branch
+          Commit.CA.close (snd t.commits) >>= fun () ->
+          Branch.close t.branch
       end
     end
 
diff --git a/src/irmin-git/remote.ml b/src/irmin-git/remote.ml
index 2faba8bf21..6ecf10cbae 100644
--- a/src/irmin-git/remote.ml
+++ b/src/irmin-git/remote.ml
@@ -74,7 +74,8 @@ struct
         G.Ref.write t br value >|= reword_error (msgf "%a" G.pp_error)
         >>? fun () ->
         G.Ref.write t reference value >|= reword_error (msgf "%a" G.pp_error)
-        >>? fun () -> Lwt.return (Ok (Some hash))
+        >>? fun () ->
+        Lwt.return (Ok (Some hash))
     | _ -> assert false
 
   let push t ?depth:_ (ctx, e) br =
diff --git a/src/irmin-graphql/server.ml b/src/irmin-graphql/server.ml
index 6ffefc9a59..e523872da7 100644
--- a/src/irmin-graphql/server.ml
+++ b/src/irmin-graphql/server.ml
@@ -495,7 +495,9 @@ struct
                 let* t = mk_branch s branch in
                 let* remote = remote in
                 Sync.fetch t remote >>= function
-                | Ok (`Head d) -> Store.Head.set t d >|= fun () -> Ok (Some d)
+                | Ok (`Head d) ->
+                    Store.Head.set t d >|= fun () ->
+                    Ok (Some d)
                 | Ok `Empty -> Lwt.return (Ok None)
                 | Error (`Msg e) -> Lwt.return (Error e));
             io_field "push" ~typ:(Lazy.force commit)
@@ -779,7 +781,8 @@ struct
             Store.Commit.of_hash s commit >>= function
             | Some commit ->
                 let* t = mk_branch s branch in
-                Store.Head.set t commit >|= fun () -> Ok (Some commit)
+                Store.Head.set t commit >|= fun () ->
+                Ok (Some commit)
             | None -> Lwt.return (Ok None));
       ]
 
diff --git a/src/irmin-http/irmin_http_server.ml b/src/irmin-http/irmin_http_server.ml
index ae0412aa9d..8a7b9d606d 100644
--- a/src/irmin-http/irmin_http_server.ml
+++ b/src/irmin-http/irmin_http_server.ml
@@ -280,7 +280,8 @@ module Make (HTTP : Cohttp_lwt.S.Server) (S : Irmin.S) = struct
                 Lwt.return_unit)
           in
           Lwt.async (fun () ->
-              Lwt_stream.closed stream >>= fun () -> S.unwatch db w);
+              Lwt_stream.closed stream >>= fun () ->
+              S.unwatch db w);
           push (Some "[");
           `Stream stream
 
@@ -316,7 +317,8 @@ module Make (HTTP : Cohttp_lwt.S.Server) (S : Irmin.S) = struct
                 Lwt.return_unit)
           in
           Lwt.async (fun () ->
-              Lwt_stream.closed stream >>= fun () -> S.unwatch db w);
+              Lwt_stream.closed stream >>= fun () ->
+              S.unwatch db w);
           push (Some "[");
           `Stream stream
 
@@ -346,8 +348,13 @@ module Make (HTTP : Cohttp_lwt.S.Server) (S : Irmin.S) = struct
       (struct
         include B.Contents
 
-        let unsafe_add t k v = unsafe_add t k v >|= fun _ -> ()
-        let batch t f = B.Repo.batch t @@ fun x _ _ -> f x
+        let unsafe_add t k v =
+          unsafe_add t k v >|= fun _ ->
+          ()
+
+        let batch t f =
+          B.Repo.batch t @@ fun x _ _ ->
+          f x
       end)
       (B.Contents.Key)
       (B.Contents.Val)
@@ -357,8 +364,13 @@ module Make (HTTP : Cohttp_lwt.S.Server) (S : Irmin.S) = struct
       (struct
         include B.Node
 
-        let unsafe_add t k v = unsafe_add t k v >|= fun _ -> ()
-        let batch t f = B.Repo.batch t @@ fun _ x _ -> f x
+        let unsafe_add t k v =
+          unsafe_add t k v >|= fun _ ->
+          ()
+
+        let batch t f =
+          B.Repo.batch t @@ fun _ x _ ->
+          f x
       end)
       (B.Node.Key)
       (B.Node.Val)
@@ -368,8 +380,13 @@ module Make (HTTP : Cohttp_lwt.S.Server) (S : Irmin.S) = struct
       (struct
         include B.Commit
 
-        let unsafe_add t k v = unsafe_add t k v >|= fun _ -> ()
-        let batch t f = B.Repo.batch t @@ fun _ _ x -> f x
+        let unsafe_add t k v =
+          unsafe_add t k v >|= fun _ ->
+          ()
+
+        let batch t f =
+          B.Repo.batch t @@ fun _ _ x ->
+          f x
       end)
       (B.Commit.Key)
       (B.Commit.Val)
diff --git a/src/irmin-mirage/git/irmin_mirage_git.ml b/src/irmin-mirage/git/irmin_mirage_git.ml
index f4bfbb676c..a935d1f162 100644
--- a/src/irmin-mirage/git/irmin_mirage_git.ml
+++ b/src/irmin-mirage/git/irmin_mirage_git.ml
@@ -171,16 +171,30 @@ module KV_RO (G : Git.S) = struct
     (S.find_tree t.t t.root >|= function
      | None -> S.Tree.empty ()
      | Some tree -> tree)
-    >|= fun tree -> { Tree.repo; tree }
+    >|= fun tree ->
+    { Tree.repo; tree }
 
-  let exists t k = tree t >>= fun t -> Tree.exists t k
-  let get t k = tree t >>= fun t -> Tree.get t k
-  let list t k = tree t >>= fun t -> Tree.list t k
-  let digest t k = tree t >>= fun t -> Tree.digest t k
+  let exists t k =
+    tree t >>= fun t ->
+    Tree.exists t k
+
+  let get t k =
+    tree t >>= fun t ->
+    Tree.get t k
+
+  let list t k =
+    tree t >>= fun t ->
+    Tree.list t k
+
+  let digest t k =
+    tree t >>= fun t ->
+    Tree.digest t k
 
   let get t k =
     match Key.segments k with
-    | [ "HEAD" ] -> head_message t >|= fun v -> Ok v
+    | [ "HEAD" ] ->
+        head_message t >|= fun v ->
+        Ok v
     | _ -> get t k
 end
 
@@ -242,10 +256,21 @@ module KV_RW (G : Irmin_git.G) (C : Mirage_clock.PCLOCK) = struct
     | Store t -> RO.tree t
     | Batch b -> Lwt.return { Tree.tree = b.tree; repo = repo t }
 
-  let digest t k = tree t >>= fun t -> Tree.digest t k
-  let exists t k = tree t >>= fun t -> Tree.exists t k
-  let get t k = tree t >>= fun t -> Tree.get t k
-  let list t k = tree t >>= fun t -> Tree.list t k
+  let digest t k =
+    tree t >>= fun t ->
+    Tree.digest t k
+
+  let exists t k =
+    tree t >>= fun t ->
+    Tree.exists t k
+
+  let get t k =
+    tree t >>= fun t ->
+    Tree.get t k
+
+  let list t k =
+    tree t >>= fun t ->
+    Tree.list t k
 
   type write_error = [ RO.error | Mirage_kv.write_error | RO.Sync.push_error ]
 
@@ -303,7 +328,9 @@ module KV_RW (G : Irmin_git.G) (C : Mirage_clock.PCLOCK) = struct
           let repo = S.repo s.t in
           (* get the tree origin *)
           get_store_tree s >>= function
-          | None -> f t >|= fun x -> Ok x (* no transaction is needed *)
+          | None ->
+              f t >|= fun x ->
+              Ok x (* no transaction is needed *)
           | Some (origin, old_tree) -> (
               let batch = { repo; tree = old_tree; origin } in
               let b = Batch batch in
diff --git a/src/irmin-pack/_layered/IO_layers.ml b/src/irmin-pack/_layered/IO_layers.ml
index 33f69b53a8..333008d3b2 100644
--- a/src/irmin-pack/_layered/IO_layers.ml
+++ b/src/irmin-pack/_layered/IO_layers.ml
@@ -73,10 +73,12 @@ module IO = struct
           Lwt_unix.openfile file Lwt_unix.[ O_CREAT; O_RDWR; O_CLOEXEC ] 0o644
         in
         let t = { file; fd } in
-        write_flip true t >|= fun () -> t
+        write_flip true t >|= fun () ->
+        t
     | true ->
         Lwt_unix.openfile file Lwt_unix.[ O_EXCL; O_RDWR; O_CLOEXEC ] 0o644
-        >|= fun fd -> { file; fd }
+        >|= fun fd ->
+        { file; fd }
 end
 
 module Lock = struct
@@ -93,5 +95,8 @@ module Lock = struct
 
   let test = Sys.file_exists
   let unlink = Lwt_unix.unlink
-  let close { fd; file } = Lwt_unix.close fd >>= fun () -> unlink file
+
+  let close { fd; file } =
+    Lwt_unix.close fd >>= fun () ->
+    unlink file
 end
diff --git a/src/irmin-pack/_layered/checks.ml b/src/irmin-pack/_layered/checks.ml
index b7721ed458..2a5d60be50 100644
--- a/src/irmin-pack/_layered/checks.ml
+++ b/src/irmin-pack/_layered/checks.ml
@@ -62,7 +62,8 @@ module Make (M : Maker) (Store : S.Store) = struct
           | true -> `Upper1
           | false -> `Upper0
         in
-        IO_layers.IO.close t >|= fun () -> Some a
+        IO_layers.IO.close t >|= fun () ->
+        Some a
 
   module Stat = struct
     module Layer_stat = Simple.Stat
diff --git a/src/irmin-pack/_layered/ext_layered.ml b/src/irmin-pack/_layered/ext_layered.ml
index 1a9c33f79f..0fec04c2d4 100644
--- a/src/irmin-pack/_layered/ext_layered.ml
+++ b/src/irmin-pack/_layered/ext_layered.ml
@@ -618,7 +618,8 @@ module Maker' (Config : Conf.Pack.S) (Schema : Irmin.Schema.Extended) = struct
         on_next_upper t (fun u ->
             iter_copy u ?cancel ~skip_commits ~skip_nodes:(mem_node_next t)
               ~skip_contents:(mem_contents_next t) t newies)
-        >>= fun () -> X.Branch.copy_newies_to_next_upper t.branch
+        >>= fun () ->
+        X.Branch.copy_newies_to_next_upper t.branch
 
       (** Repeatedly call [copy_newies] as long as there are many newies (more
           than newies_limit bytes added). *)
diff --git a/src/irmin-pack/_layered/layered_store.ml b/src/irmin-pack/_layered/layered_store.ml
index d2d7ddcfa4..b01b7669b2 100644
--- a/src/irmin-pack/_layered/layered_store.ml
+++ b/src/irmin-pack/_layered/layered_store.ml
@@ -441,7 +441,8 @@ struct
       >>= function
       | None -> U.test_and_set current k ~test:None ~set
       | Some v ->
-          U.set current k v >>= fun () -> U.test_and_set current k ~test ~set
+          U.set current k v >>= fun () ->
+          U.test_and_set current k ~test ~set
     in
     (U.mem current k >>= function
      | true -> U.test_and_set current k ~test ~set
diff --git a/src/irmin-pack/unix/atomic_write.ml b/src/irmin-pack/unix/atomic_write.ml
index 7c4c3343d1..96d5137088 100644
--- a/src/irmin-pack/unix/atomic_write.ml
+++ b/src/irmin-pack/unix/atomic_write.ml
@@ -182,7 +182,9 @@ module Make_persistent (K : Irmin.Type.S) (V : Value.S) = struct
   let test_and_set t k ~test ~set =
     [%log.debug "[branches] test-and-set %a" pp_key k];
     unsafe_test_and_set t k ~test ~set >>= function
-    | true -> W.notify t.w k set >|= fun () -> true
+    | true ->
+        W.notify t.w k set >|= fun () ->
+        true
     | false -> Lwt.return_false
 
   let list t =
diff --git a/src/irmin-pack/unix/ext.ml b/src/irmin-pack/unix/ext.ml
index af6a54c569..131c94a2e9 100644
--- a/src/irmin-pack/unix/ext.ml
+++ b/src/irmin-pack/unix/ext.ml
@@ -175,7 +175,8 @@ module Maker (Config : Conf.S) = struct
           Index.close t.index;
           Contents.CA.close (contents_t t) >>= fun () ->
           Node.CA.close (snd (node_t t)) >>= fun () ->
-          Commit.CA.close (snd (commit_t t)) >>= fun () -> Branch.close t.branch
+          Commit.CA.close (snd (commit_t t)) >>= fun () ->
+          Branch.close t.branch
 
         let v config =
           Lwt.catch
diff --git a/src/irmin-test/irmin_bench.ml b/src/irmin-test/irmin_bench.ml
index c729cf7696..5fca8fe5d9 100644
--- a/src/irmin-test/irmin_bench.ml
+++ b/src/irmin-test/irmin_bench.ml
@@ -134,7 +134,8 @@ struct
           times ~n:t.tree_add ~init:tree (fun n tree ->
               Store.Tree.add tree paths.(n) "init"))
     in
-    Store.set_tree_exn v ~info [] tree >|= fun () -> Fmt.epr "[init done]\n%!"
+    Store.set_tree_exn v ~info [] tree >|= fun () ->
+    Fmt.epr "[init done]\n%!"
 
   let run t config size =
     let* r = Store.Repo.v config in
@@ -155,14 +156,17 @@ struct
           Store.set_tree_exn v ~info [] tree >|= fun () ->
           if t.clear then Store.Tree.clear tree)
     in
-    Store.Repo.close r >|= fun () -> Fmt.epr "\n[run done]\n%!"
+    Store.Repo.close r >|= fun () ->
+    Fmt.epr "\n[run done]\n%!"
 
   let main t config size =
     let root = "_build/_bench" in
     let config = config ~root in
     let size () = size ~root in
     let t = { t with root } in
-    Lwt_main.run (init t config >>= fun () -> run t config size)
+    Lwt_main.run
+      ( init t config >>= fun () ->
+        run t config size )
 
   let main_term config size = Term.(const main $ t $ pure config $ pure size)
 
diff --git a/src/irmin-test/store.ml b/src/irmin-test/store.ml
index 497acfcc90..71d655b4f5 100644
--- a/src/irmin-test/store.ml
+++ b/src/irmin-test/store.ml
@@ -486,7 +486,8 @@ module Make (S : Generic_key) = struct
       let bindings1 = [ ([ "a" ], "x"); ([ "b" ], "y") ] in
       check_hash "1 level" bindings1 >>= fun () ->
       let bindings2 = [ ([ "a"; "b" ], "x"); ([ "a"; "c" ], "y") ] in
-      check_hash "2 levels" bindings2 >>= fun () -> S.Repo.close repo
+      check_hash "2 levels" bindings2 >>= fun () ->
+      S.Repo.close repo
     in
     run x test
 
@@ -647,7 +648,8 @@ module Make (S : Generic_key) = struct
       in
       let assert_last_modified msg ?depth ~n t key expected =
         let+ last = S.last_modified ?depth ~n t key in
-        S.repo t |> fun repo ->
+        S.repo t
+        |> fun repo ->
         let msg = Printf.sprintf "%s [n=%d]" msg n in
         checks (S.commit_t repo) msg expected last
       in
@@ -984,8 +986,10 @@ module Make (S : Generic_key) = struct
       let* ty = S.clone ~src:tx ~dst:tagy in
       let* vx' = S.find ty xy in
       check_val "update tag" (Some vx) vx';
-      S.status tx |> fun tagx' ->
-      S.status ty |> fun tagy' ->
+      S.status tx
+      |> fun tagx' ->
+      S.status ty
+      |> fun tagy' ->
       check (S.Status.t repo) "tagx" (`Branch tagx) tagx';
       check (S.Status.t repo) "tagy" (`Branch tagy) tagy';
       let* t = S.main repo in
@@ -1134,7 +1138,8 @@ module Make (S : Generic_key) = struct
       Alcotest.(check stats_t) "empty stats" empty_stats s;
       S.set_tree_exn t ~info:(infof "empty tree") [] v1 >>= fun () ->
       let* head = S.Head.get t in
-      S.Commit.key head |> fun head ->
+      S.Commit.key head
+      |> fun head ->
       let* commit = B.Commit.find (ct repo) head in
       let node = B.Commit.Val.node (get commit) in
       let* node = B.Node.find (n repo) node in
@@ -1380,7 +1385,8 @@ module Make (S : Generic_key) = struct
       in
       S.Head.set t c >>= fun () ->
       let* h = S.Head.get t in
-      S.Commit.info h |> fun i ->
+      S.Commit.info h
+      |> fun i ->
       check S.Info.t "commit info" i0 i;
       let* tt = S.of_commit h in
       let* g = S.history tt in
@@ -1846,7 +1852,8 @@ module Make (S : Generic_key) = struct
           match B.Node.Val.find v "500000" with
           | None -> Lwt.return_unit
           | Some _ -> Alcotest.fail "value 500000 should not be found"))
-      >>= fun () -> B.Repo.close repo
+      >>= fun () ->
+      B.Repo.close repo
     in
     run x test
 
@@ -2040,7 +2047,11 @@ module Make (S : Generic_key) = struct
 
   let rec read fn check = function
     | 0 -> []
-    | i -> (fun () -> fn i >|= fun v -> check i v) :: read fn check (i - 1)
+    | i ->
+        (fun () ->
+          fn i >|= fun v ->
+          check i v)
+        :: read fn check (i - 1)
 
   let test_concurrent_low x () =
     let test_branches repo =
@@ -2090,7 +2101,8 @@ module Make (S : Generic_key) = struct
           (fun _ -> S.get t k)
           (fun i -> check S.contents_t (Fmt.str "update: one %d" i) v)
       in
-      perform (write t1 10 @ write t2 10) >>= fun () -> perform (read t1 10)
+      perform (write t1 10 @ write t2 10) >>= fun () ->
+      perform (read t1 10)
     in
     let test_multi repo =
       let k i = [ "a"; "b"; "c"; string_of_int i ] in
@@ -2106,11 +2118,13 @@ module Make (S : Generic_key) = struct
           (fun i -> S.get t (k i))
           (fun i -> check S.contents_t (Fmt.str "update: multi %d" i) (v i))
       in
-      perform (write t1 10 @ write t2 10) >>= fun () -> perform (read t1 10)
+      perform (write t1 10 @ write t2 10) >>= fun () ->
+      perform (read t1 10)
     in
     run x (fun repo ->
         test_one repo >>= fun () ->
-        test_multi repo >>= fun () -> B.Repo.close repo)
+        test_multi repo >>= fun () ->
+        B.Repo.close repo)
 
   let test_concurrent_merges x () =
     let test repo =
@@ -2134,7 +2148,8 @@ module Make (S : Generic_key) = struct
       in
       S.set_exn t1 ~info:(infof "update") (k 0) (v 0) >>= fun () ->
       perform (write t1 1 10 @ write t2 2 10) >>= fun () ->
-      perform (read t1 10) >>= fun () -> B.Repo.close repo
+      perform (read t1 10) >>= fun () ->
+      B.Repo.close repo
     in
     run x test
 
@@ -2248,7 +2263,8 @@ module Make (S : Generic_key) = struct
               Alcotest.(check string) "merge z" a "4" );
           ]
       in
-      set () >>= test_and_set >>= merge >>= fun () -> B.Repo.close repo
+      set () >>= test_and_set >>= merge >>= fun () ->
+      B.Repo.close repo
     in
     run x test
 
@@ -2278,7 +2294,8 @@ module Make (S : Generic_key) = struct
                 let* m = S.clone ~src:t ~dst:tag in
                 S.set_exn m ~info:(infof "update") (k i) (v i) >>= fun () ->
                 let* set = S.Head.find m in
-                Lwt.pause () >>= fun () -> S.Head.test_and_set t ~test ~set))
+                Lwt.pause () >>= fun () ->
+                S.Head.test_and_set t ~test ~set))
       in
       let read t =
         read
@@ -2287,7 +2304,8 @@ module Make (S : Generic_key) = struct
       in
       S.set_exn t1 ~info:(infof "update") (k 0) (v 0) >>= fun () ->
       perform (write t1 1 5 @ write t2 2 5) >>= fun () ->
-      perform (read t1 5) >>= fun () -> B.Repo.close repo
+      perform (read t1 5) >>= fun () ->
+      B.Repo.close repo
     in
     run x test
 
@@ -2463,12 +2481,18 @@ module Make (S : Generic_key) = struct
       S.Backend.Repo.close repo
     in
     (* Test collisions with the empty node (and its commit), *)
-    run x (test @@ fun () -> S.Tree.empty () |> Lwt.return);
+    run x
+      ( test @@ fun () ->
+        S.Tree.empty () |> Lwt.return );
     (* with a length one node, *)
-    run x (test @@ fun () -> add_entries (S.Tree.empty ()) 1);
+    run x
+      ( test @@ fun () ->
+        add_entries (S.Tree.empty ()) 1 );
     (* and with a length >256 node (which is the threshold for unstable inodes
        in irmin pack). *)
-    run x (test @@ fun () -> add_entries (S.Tree.empty ()) 260)
+    run x
+      ( test @@ fun () ->
+        add_entries (S.Tree.empty ()) 260 )
 end
 
 let suite' l ?(prefix = "") (_, x) =
diff --git a/src/irmin-test/store_graph.ml b/src/irmin-test/store_graph.ml
index b0e315b524..a2651d248c 100644
--- a/src/irmin-test/store_graph.ml
+++ b/src/irmin-test/store_graph.ml
@@ -201,7 +201,8 @@ module Make (S : Generic_key) = struct
       in
       test1 () >>= fun () ->
       test2 () >>= fun () ->
-      test3 () >>= fun () -> B.Repo.close repo
+      test3 () >>= fun () ->
+      B.Repo.close repo
     in
     run x test
 
diff --git a/src/irmin-test/store_watch.ml b/src/irmin-test/store_watch.ml
index 8b47d5ca57..4d7491a7f7 100644
--- a/src/irmin-test/store_watch.ml
+++ b/src/irmin-test/store_watch.ml
@@ -22,7 +22,8 @@ module Make (Log : Logs.LOG) (S : Generic_key) = struct
 
   let sleep ?(sleep_t = 0.01) () =
     let sleep_t = min sleep_t 1. in
-    Lwt.pause () >>= fun () -> Lwt_unix.sleep sleep_t
+    Lwt.pause () >>= fun () ->
+    Lwt_unix.sleep sleep_t
 
   let now_s () = Mtime.Span.to_s (Mtime_clock.elapsed ())
 
@@ -63,10 +64,14 @@ module Make (Log : Logs.LOG) (S : Generic_key) = struct
         | Some h -> if eq (`Updated (h, h2)) x then incr r
       in
       let* u =
-        S.watch ?init:h t (fun v -> check v >|= fun () -> failwith "test")
+        S.watch ?init:h t (fun v ->
+            check v >|= fun () ->
+            failwith "test")
       in
       let* v =
-        S.watch ?init:h t (fun v -> check v >>= fun () -> Lwt.fail_with "test")
+        S.watch ?init:h t (fun v ->
+            check v >>= fun () ->
+            Lwt.fail_with "test")
       in
       let* w = S.watch ?init:h t (fun v -> check v) in
       S.set_exn t ~info:(infof "update") key v1 >>= fun () ->
@@ -237,7 +242,8 @@ module Make (Log : Logs.LOG) (S : Generic_key) = struct
               [%log.debug "[waiting for] %s" (msg `Post (n - i) post_w post)];
               fn (n - i) >>= fun () ->
               (* check post-condition *)
-              check `Post (n - i) post_w post >>= fun () -> aux post (i - 1)
+              check `Post (n - i) post_w post >>= fun () ->
+              aux post (i - 1)
         in
         aux s n
     end in
@@ -362,7 +368,8 @@ module Make (Log : Logs.LOG) (S : Generic_key) = struct
       update true (0, 0, 0) 10 ~first:true >>= fun () ->
       S.unwatch u >>= fun () ->
       update false (0, 10, 0) 10 >>= fun () ->
-      B.Repo.close repo >>= fun () -> B.Repo.close repo1
+      B.Repo.close repo >>= fun () ->
+      B.Repo.close repo1
     in
     run x test
 
diff --git a/src/irmin-unix/cli.ml b/src/irmin-unix/cli.ml
index 9b31804da6..51bb5a67a4 100644
--- a/src/irmin-unix/cli.ml
+++ b/src/irmin-unix/cli.ml
@@ -946,7 +946,10 @@ let log =
                 in
                 incr num_count
             in
-            let* max = S.Head.get t >|= fun x -> [ `Commit (S.Commit.key x) ] in
+            let* max =
+              S.Head.get t >|= fun x ->
+              [ `Commit (S.Commit.key x) ]
+            in
             let iter ~commit ~max repo =
               Lwt.catch
                 (fun () ->
diff --git a/src/irmin-unix/fs.ml b/src/irmin-unix/fs.ml
index 8d26ae4987..01948675d0 100644
--- a/src/irmin-unix/fs.ml
+++ b/src/irmin-unix/fs.ml
@@ -91,7 +91,8 @@ module IO = struct
         let* is_stale = is_stale max_age file in
         if is_stale then (
           [%log.err "%s is stale, removing it." file];
-          unlock file >>= fun () -> aux 1)
+          unlock file >>= fun () ->
+          aux 1)
         else
           let create () =
             let pid = Unix.getpid () in
@@ -102,7 +103,8 @@ module IO = struct
                 0o600
             in
             let oc = Lwt_io.of_fd ~mode:Lwt_io.Output fd in
-            Lwt_io.write_int oc pid >>= fun () -> Lwt_unix.close fd
+            Lwt_io.write_int oc pid >>= fun () ->
+            Lwt_unix.close fd
           in
           Lwt.catch create (function
             | Unix.Unix_error (Unix.EEXIST, _, _) ->
@@ -112,7 +114,8 @@ module IO = struct
                        (let i = float i in
                         i *. i)
                 in
-                Lwt_unix.sleep (sleep *. backoff) >>= fun () -> aux (i + 1)
+                Lwt_unix.sleep (sleep *. backoff) >>= fun () ->
+                aux (i + 1)
             | e -> Lwt.fail e)
       in
       aux 1
@@ -120,7 +123,9 @@ module IO = struct
     let with_lock file fn =
       match file with
       | None -> fn ()
-      | Some f -> lock f >>= fun () -> Lwt.finalize fn (fun () -> unlock f)
+      | Some f ->
+          lock f >>= fun () ->
+          Lwt.finalize fn (fun () -> unlock f)
   end
 
   type path = string
@@ -202,10 +207,12 @@ module IO = struct
                 else
                   let* exists = file_exists file in
                   if exists && Sys.is_directory file then
-                    remove_dir file >>= fun () -> aux (i + 1)
+                    remove_dir file >>= fun () ->
+                    aux (i + 1)
                   else (
                     [%log.debug "Got EACCES, retrying in %.1fs" delays.(i)];
-                    Lwt_unix.sleep delays.(i) >>= fun () -> aux (i + 1))
+                    Lwt_unix.sleep delays.(i) >>= fun () ->
+                    aux (i + 1))
             | e -> Lwt.fail e)
       in
       aux 0
diff --git a/src/irmin-unix/resolver.ml b/src/irmin-unix/resolver.ml
index 57cd78a528..2afd48a62e 100644
--- a/src/irmin-unix/resolver.ml
+++ b/src/irmin-unix/resolver.ml
@@ -602,7 +602,9 @@ let build_irmin_config config root opts (store, hash, contents) branch commit :
         | None -> invalid_arg "unknown commit"
         | Some c -> S.of_commit c)
     | None, None -> S.Repo.v config >>= S.main
-    | Some b, None -> S.Repo.v config >>= fun repo -> S.of_branch repo b
+    | Some b, None ->
+        S.Repo.v config >>= fun repo ->
+        S.of_branch repo b
   in
   S (impl, spec, remote)
 
diff --git a/src/irmin/commit.ml b/src/irmin/commit.ml
index 02f6304524..d0541a42d8 100644
--- a/src/irmin/commit.ml
+++ b/src/irmin/commit.ml
@@ -578,7 +578,8 @@ module History (S : Store) = struct
             let rec aux acc = function
               | [] -> Merge.ok (Some acc)
               | old :: olds ->
-                  three_way_merge t ~info acc old >>=* fun acc -> aux acc olds
+                  three_way_merge t ~info acc old >>=* fun acc ->
+                  aux acc olds
             in
             aux old olds
       in
diff --git a/src/irmin/content_addressable.ml b/src/irmin/content_addressable.ml
index f0ab9d8173..28b9ea1480 100644
--- a/src/irmin/content_addressable.ml
+++ b/src/irmin/content_addressable.ml
@@ -40,7 +40,8 @@ module Make (AO : Append_only.Maker) (K : Hash.S) (V : Type.S) = struct
 
   let add t v =
     let k = hash v in
-    add t k v >|= fun () -> k
+    add t k v >|= fun () ->
+    k
 end
 
 module Check_closed (CA : Maker) (K : Hash.S) (V : Type.S) = struct
diff --git a/src/irmin/contents.ml b/src/irmin/contents.ml
index f9e0b68210..f7c68eb1cc 100644
--- a/src/irmin/contents.ml
+++ b/src/irmin/contents.ml
@@ -118,19 +118,23 @@ module Json_value = struct
     let m =
       Merge.(alist Type.string t (fun _key -> option (v t merge_value)))
     in
-    Merge.(f m ~old x y) >>=* fun x -> Merge.ok (`O x)
+    Merge.(f m ~old x y) >>=* fun x ->
+    Merge.ok (`O x)
 
   and merge_float ~old x y =
     let open Merge.Infix in
-    Merge.(f float ~old x y) >>=* fun f -> Merge.ok (`Float f)
+    Merge.(f float ~old x y) >>=* fun f ->
+    Merge.ok (`Float f)
 
   and merge_string ~old x y =
     let open Merge.Infix in
-    Merge.(f string ~old x y) >>=* fun s -> Merge.ok (`String s)
+    Merge.(f string ~old x y) >>=* fun s ->
+    Merge.ok (`String s)
 
   and merge_bool ~old x y =
     let open Merge.Infix in
-    Merge.(f bool ~old x y) >>=* fun b -> Merge.ok (`Bool b)
+    Merge.(f bool ~old x y) >>=* fun b ->
+    Merge.ok (`Bool b)
 
   and merge_array ~old x y =
     let open Merge.Infix in
diff --git a/src/irmin/indexable.ml b/src/irmin/indexable.ml
index f62d4f0586..940f503c83 100644
--- a/src/irmin/indexable.ml
+++ b/src/irmin/indexable.ml
@@ -40,7 +40,10 @@ struct
   end
 
   let index _ h = Lwt.return_some h
-  let unsafe_add t h v = unsafe_add t h v >|= fun () -> h
+
+  let unsafe_add t h v =
+    unsafe_add t h v >|= fun () ->
+    h
 end
 
 module Check_closed (CA : Maker) (Hash : Hash.S) (Value : Type.S) = struct
diff --git a/src/irmin/lock.ml b/src/irmin/lock.ml
index 754944a51b..b1729f2914 100644
--- a/src/irmin/lock.ml
+++ b/src/irmin/lock.ml
@@ -62,5 +62,6 @@ module Make (K : Type.S) = struct
   let with_lock t k fn =
     let* lock = Lwt_mutex.with_lock t.global (lock t k) in
     let* r = Lwt_mutex.with_lock lock fn in
-    Lwt_mutex.with_lock t.global (unlock t k) >>= fun () -> Lwt.return r
+    Lwt_mutex.with_lock t.global (unlock t k) >>= fun () ->
+    Lwt.return r
 end
diff --git a/src/irmin/mem/irmin_mem.ml b/src/irmin/mem/irmin_mem.ml
index 4f784b1170..c987617430 100644
--- a/src/irmin/mem/irmin_mem.ml
+++ b/src/irmin/mem/irmin_mem.ml
@@ -82,7 +82,10 @@ module Atomic_write (K : Irmin.Type.S) (V : Irmin.Type.S) = struct
     let* t = RO.v config in
     Lwt.return { t; w = watches; lock }
 
-  let close t = W.clear t.w >>= fun () -> RO.close t.t
+  let close t =
+    W.clear t.w >>= fun () ->
+    RO.close t.t
+
   let find t = RO.find t.t
   let mem t = RO.mem t.t
   let watch_key t = W.watch_key t.w
@@ -130,7 +133,9 @@ module Atomic_write (K : Irmin.Type.S) (V : Irmin.Type.S) = struct
     let+ () = if updated then W.notify t.w key set else Lwt.return_unit in
     updated
 
-  let clear t = W.clear t.w >>= fun () -> RO.clear t.t
+  let clear t =
+    W.clear t.w >>= fun () ->
+    RO.clear t.t
 end
 
 module Conf = struct
diff --git a/src/irmin/merge.ml b/src/irmin/merge.ml
index e9af71f5e4..718d131371 100644
--- a/src/irmin/merge.ml
+++ b/src/irmin/merge.ml
@@ -128,7 +128,8 @@ let option (type a) ((a, t) : a t) : a option t =
                     [%log.debug "option old=%a" pp o];
                     ok o
               in
-              t ~old v1 v2 >|=* fun x -> Some x
+              t ~old v1 v2 >|=* fun x ->
+              Some x
           | Some x, None | None, Some x -> (
               let open Infix in
               old () >>=* function
@@ -151,7 +152,8 @@ let pair (da, a) (db, b) =
           let o1 = map_promise fst old in
           let o2 = map_promise snd old in
           a ~old:o1 a1 a2 >>=* fun a3 ->
-          b ~old:o2 b1 b2 >|=* fun b3 -> (a3, b3) )
+          b ~old:o2 b1 b2 >|=* fun b3 ->
+          (a3, b3) )
 
 let triple (da, a) (db, b) (dc, c) =
   let dt = Type.triple da db dc in
@@ -168,7 +170,8 @@ let triple (da, a) (db, b) (dc, c) =
           let o3 = map_promise (fun (_, _, x) -> x) old in
           a ~old:o1 a1 a2 >>=* fun a3 ->
           b ~old:o2 b1 b2 >>=* fun b3 ->
-          c ~old:o3 c1 c2 >|=* fun c3 -> (a3, b3, c3) )
+          c ~old:o3 c1 c2 >|=* fun c3 ->
+          (a3, b3, c3) )
 
 exception C of string
 
@@ -225,7 +228,8 @@ let alist_merge_lwt compare_k f l1 l2 =
         l3 := (key, v) :: !l3;
         return_unit
   in
-  alist_iter2_lwt compare_k f l1 l2 >>= fun () -> return !l3
+  alist_iter2_lwt compare_k f l1 l2 >>= fun () ->
+  return !l3
 
 let alist dx dy merge_v =
   let pair = Type.pair dx dy in
@@ -322,7 +326,11 @@ struct
   let iter2 f m1 m2 =
     let m3 = ref [] in
     iter2 (fun key data -> m3 := f key data :: !m3) m1 m2;
-    Lwt_list.iter_p (fun b -> b >>= fun () -> Lwt.return_unit) (List.rev !m3)
+    Lwt_list.iter_p
+      (fun b ->
+        b >>= fun () ->
+        Lwt.return_unit)
+      (List.rev !m3)
 
   let merge_maps f m1 m2 =
     let l3 = ref [] in
@@ -384,7 +392,8 @@ let like_lwt (type a b) da (t : b t) (a_to_b : a -> b Lwt.t)
                 let+ b = a_to_b a in
                 Ok (Some b))
       in
-      bind ((f t) ~old b1 b2) @@ fun b3 -> b_to_a b3 >>= ok
+      bind ((f t) ~old b1 b2) @@ fun b3 ->
+      b_to_a b3 >>= ok
     with Not_found -> conflict "biject'"
   in
   seq [ default da; (da, merge) ]
diff --git a/src/irmin/object_graph.ml b/src/irmin/object_graph.ml
index 6f9202617e..303e030ddb 100644
--- a/src/irmin/object_graph.ml
+++ b/src/irmin/object_graph.ml
@@ -119,7 +119,9 @@ struct
     let* min =
       Lwt_list.fold_left_s
         (fun acc -> function
-          | `Branch _ as x -> pred x >|= fun c -> (x :: c) @ acc
+          | `Branch _ as x ->
+              pred x >|= fun c ->
+              (x :: c) @ acc
           | x -> Lwt.return (x :: acc))
         [] min
     in
@@ -196,7 +198,8 @@ struct
       else (
         [%log.debug "VISIT %a" Type.(pp X.t) key];
         mark key level;
-        treat key >>= fun () -> visit_predecessors key level)
+        treat key >>= fun () ->
+        visit_predecessors key level)
     in
     let rec pop () =
       match Queue.pop todo with
@@ -219,7 +222,8 @@ struct
       Lwt.return_unit
     in
     let skip _ = Lwt.return_false in
-    iter ~depth ~pred ~min ~max ~node ~edge ~skip ~rev:false () >|= fun () -> g
+    iter ~depth ~pred ~min ~max ~node ~edge ~skip ~rev:false () >|= fun () ->
+    g
 
   let min g =
     G.fold_vertex
diff --git a/src/irmin/store.ml b/src/irmin/store.ml
index 15ac7e298b..df58bce21e 100644
--- a/src/irmin/store.ml
+++ b/src/irmin/store.ml
@@ -854,7 +854,8 @@ module Make (B : Backend.S) = struct
     [%log.debug "test-and-set %a" pp_path k];
     retry ~retries @@ fun () ->
     update t k ?allow_empty ?parents ~info (test_and_set_tree_once ~test)
-    @@ fun _tree -> Lwt.return set
+    @@ fun _tree ->
+    Lwt.return set
 
   let test_and_set_tree_exn ?retries ?allow_empty ?parents ~info t k ~test ~set
       =
@@ -894,13 +895,33 @@ module Make (B : Backend.S) = struct
   let merge_exn ?retries ?allow_empty ?parents ~info ~old t k v =
     merge ?retries ?allow_empty ?parents ~info ~old t k v >>= fail "merge_exn"
 
-  let mem t k = tree t >>= fun tree -> Tree.mem tree k
-  let mem_tree t k = tree t >>= fun tree -> Tree.mem_tree tree k
-  let find_all t k = tree t >>= fun tree -> Tree.find_all tree k
-  let find t k = tree t >>= fun tree -> Tree.find tree k
-  let get t k = tree t >>= fun tree -> Tree.get tree k
-  let find_tree t k = tree t >>= fun tree -> Tree.find_tree tree k
-  let get_tree t k = tree t >>= fun tree -> Tree.get_tree tree k
+  let mem t k =
+    tree t >>= fun tree ->
+    Tree.mem tree k
+
+  let mem_tree t k =
+    tree t >>= fun tree ->
+    Tree.mem_tree tree k
+
+  let find_all t k =
+    tree t >>= fun tree ->
+    Tree.find_all tree k
+
+  let find t k =
+    tree t >>= fun tree ->
+    Tree.find tree k
+
+  let get t k =
+    tree t >>= fun tree ->
+    Tree.get tree k
+
+  let find_tree t k =
+    tree t >>= fun tree ->
+    Tree.find_tree tree k
+
+  let get_tree t k =
+    tree t >>= fun tree ->
+    Tree.get_tree tree k
 
   let key t k =
     find_tree t k >|= function
@@ -916,9 +937,17 @@ module Make (B : Backend.S) = struct
     | None -> None
     | Some tree -> Some (Tree.hash tree)
 
-  let get_all t k = tree t >>= fun tree -> Tree.get_all tree k
-  let list t k = tree t >>= fun tree -> Tree.list tree k
-  let kind t k = tree t >>= fun tree -> Tree.kind tree k
+  let get_all t k =
+    tree t >>= fun tree ->
+    Tree.get_all tree k
+
+  let list t k =
+    tree t >>= fun tree ->
+    Tree.list tree k
+
+  let kind t k =
+    tree t >>= fun tree ->
+    Tree.kind tree k
 
   let with_tree ?(retries = 13) ?allow_empty ?parents
       ?(strategy = `Test_and_set) ~info t key f =
@@ -1068,7 +1097,8 @@ module Make (B : Backend.S) = struct
       | `Commit k ->
           Commits.parents (commit_store t) k
           >>= Lwt_list.filter_map_p (Commit.of_key t.repo)
-          >|= fun parents -> List.map (fun x -> `Commit x.key) parents
+          >|= fun parents ->
+          List.map (fun x -> `Commit x.key) parents
       | _ -> Lwt.return_nil
     in
     let* max = Head.find t >|= function Some h -> [ h ] | None -> max in
diff --git a/src/irmin/sync.ml b/src/irmin/sync.ml
index 55bc835bf3..471982afd8 100644
--- a/src/irmin/sync.ml
+++ b/src/irmin/sync.ml
@@ -156,7 +156,9 @@ module Make (S : Store.Generic_key.S) = struct
     | Error e -> Lwt.return (Error (e :> pull_error))
     | Ok (`Head k) -> (
         match kind with
-        | `Set -> S.Head.set t k >|= fun () -> Ok (`Head k)
+        | `Set ->
+            S.Head.set t k >|= fun () ->
+            Ok (`Head k)
         | `Merge info -> (
             S.Head.merge ~into:t ~info k >>= function
             | Ok () -> Lwt.return (Ok (`Head k))
diff --git a/src/irmin/tree.ml b/src/irmin/tree.ml
index cfd26d8985..b0f618209d 100644
--- a/src/irmin/tree.ml
+++ b/src/irmin/tree.ml
@@ -53,7 +53,11 @@ let alist_iter2 compare_k f l1 l2 =
 let alist_iter2_lwt compare_k f l1 l2 =
   let l3 = ref [] in
   alist_iter2 compare_k (fun left right -> l3 := f left right :: !l3) l1 l2;
-  Lwt_list.iter_s (fun b -> b >>= fun () -> Lwt.return_unit) (List.rev !l3)
+  Lwt_list.iter_s
+    (fun b ->
+      b >>= fun () ->
+      Lwt.return_unit)
+    (List.rev !l3)
 
 exception Backend_invariant_violation of string
 exception Assertion_failure of string
@@ -1606,7 +1610,8 @@ module Make (P : Backend.S) = struct
                       | `Contents _ -> ok None
                       | `Node n -> ok (Some n))
                 in
-                Merge.(f m ~old x y) >>=* fun n -> Merge.ok (`Node n))
+                Merge.(f m ~old x y) >>=* fun n ->
+                Merge.ok (`Node n))
         | _ -> Merge.conflict "add/add values"
       in
       k (Merge.seq [ Merge.default elt_t; Merge.v elt_t f ])
@@ -1766,7 +1771,10 @@ module Make (P : Backend.S) = struct
     | None -> err_not_found "get" k
     | Some v -> Lwt.return v
 
-  let get t k = get_all t k >|= fun (c, _) -> c
+  let get t k =
+    get_all t k >|= fun (c, _) ->
+    c
+
   let mem t k = find t k >|= function None -> false | _ -> true
   let mem_tree t k = find_tree t k >|= function None -> false | _ -> true
 
diff --git a/test/irmin-fs/test_fs.ml b/test/irmin-fs/test_fs.ml
index 04b9e783af..e9ca3762c6 100644
--- a/test/irmin-fs/test_fs.ml
+++ b/test/irmin-fs/test_fs.ml
@@ -18,7 +18,11 @@ open Lwt.Infix
 module IO = Irmin_fs.IO_mem
 
 let test_db = Filename.concat "_build" "test-db"
-let init () = IO.clear () >|= fun () -> IO.set_listen_hook ()
+
+let init () =
+  IO.clear () >|= fun () ->
+  IO.set_listen_hook ()
+
 let config = Irmin_fs.config test_db
 let clean () = Lwt.return_unit
 
diff --git a/test/irmin-git/test_git.ml b/test/irmin-git/test_git.ml
index 3089177cac..cb73131726 100644
--- a/test/irmin-git/test_git.ml
+++ b/test/irmin-git/test_git.ml
@@ -54,7 +54,9 @@ module Mem (C : Irmin.Contents.S) = struct
 
   let init () =
     Git.v (Fpath.v test_db) >>= function
-    | Ok t -> S.Git.reset t >|= fun _ -> ()
+    | Ok t ->
+        S.Git.reset t >|= fun _ ->
+        ()
     | _ -> Lwt.return_unit
 end
 
diff --git a/test/irmin-http/test_http.ml b/test/irmin-http/test_http.ml
index ce14d097e8..0540a78f98 100644
--- a/test/irmin-http/test_http.ml
+++ b/test/irmin-http/test_http.ml
@@ -69,7 +69,8 @@ let check_connection id =
         (fun _ -> Lwt.return ())
         (function
           | Unix.Unix_error (Unix.ECONNREFUSED, _, _) ->
-              Lwt_unix.sleep (float n *. 0.1) >>= fun () -> loop (n + 1)
+              Lwt_unix.sleep (float n *. 0.1) >>= fun () ->
+              loop (n + 1)
           | e ->
               Alcotest.failf "Cannot connect to server %a: %a" pp id Fmt.exn e)
   in
@@ -86,10 +87,12 @@ let wait_for_the_server_to_start id =
       let pid = int_of_string line in
       [%logs.debug "read PID %d from %s" pid pid_file];
       Unix.unlink pid_file;
-      check_connection id >|= fun () -> pid)
+      check_connection id >|= fun () ->
+      pid)
     else (
       [%logs.debug "waiting for the server to start..."];
-      Lwt_unix.sleep (float n *. 0.1) >>= fun () -> aux (n + 1))
+      Lwt_unix.sleep (float n *. 0.1) >>= fun () ->
+      aux (n + 1))
   in
   aux 1
 
@@ -123,11 +126,16 @@ let rec lock id =
       if len <> pid_len then
         Lwt_unix.close fd >>= fun () ->
         Lwt.fail_with "Unable to write PID to lock file"
-      else Lwt_unix.rename pid_file_tmp pid_file >|= fun () -> fd)
+      else
+        Lwt_unix.rename pid_file_tmp pid_file >|= fun () ->
+        fd)
     (function
       | Unix.Unix_error (Unix.EAGAIN, _, _) ->
-          Lwt_unix.close fd >>= fun () -> lock id
-      | e -> Lwt_unix.close fd >>= fun () -> Lwt.fail e)
+          Lwt_unix.close fd >>= fun () ->
+          lock id
+      | e ->
+          Lwt_unix.close fd >>= fun () ->
+          Lwt.fail e)
 
 let unlock fd = Lwt_unix.close fd
 
diff --git a/test/irmin-pack/multiple_instances.ml b/test/irmin-pack/multiple_instances.ml
index 37e0a27b67..b67d1d7d9a 100644
--- a/test/irmin-pack/multiple_instances.ml
+++ b/test/irmin-pack/multiple_instances.ml
@@ -92,7 +92,8 @@ let ro_sync_after_add () =
   in
   S.sync ro;
   check ro c2 "a" "y" >>= fun () ->
-  S.Repo.close ro >>= fun () -> S.Repo.close rw
+  S.Repo.close ro >>= fun () ->
+  S.Repo.close rw
 
 let ro_sync_after_close () =
   let binding f = f [ "a" ] "x" in
@@ -103,7 +104,8 @@ let ro_sync_after_close () =
   let* c1 = S.Commit.v rw ~parents:[] ~info:(info ()) tree in
   S.Repo.close rw >>= fun () ->
   S.sync ro;
-  binding (check_binding ro c1) >>= fun () -> S.Repo.close ro
+  binding (check_binding ro c1) >>= fun () ->
+  S.Repo.close ro
 
 let tests =
   let tc name test =
diff --git a/test/irmin-pack/test_existing_stores.ml b/test/irmin-pack/test_existing_stores.ml
index 624453a109..2d178b476c 100644
--- a/test/irmin-pack/test_existing_stores.ml
+++ b/test/irmin-pack/test_existing_stores.ml
@@ -171,7 +171,8 @@ module Test_reconstruct = struct
     [%log.app
       "Checking old bindings are still reachable post index reconstruction)"];
     let* r = S.Repo.v conf in
-    check_repo r archive >>= fun () -> S.Repo.close r
+    check_repo r archive >>= fun () ->
+    S.Repo.close r
 end
 
 module Test_corrupted_stores = struct
diff --git a/test/irmin-pack/test_inode.ml b/test/irmin-pack/test_inode.ml
index 4f86a0098d..a1d99e791e 100644
--- a/test/irmin-pack/test_inode.ml
+++ b/test/irmin-pack/test_inode.ml
@@ -91,7 +91,8 @@ struct
 
     let close t =
       Index.close t.index;
-      Inode.close t.store >>= fun () -> Contents_store.close t.store_contents
+      Inode.close t.store >>= fun () ->
+      Contents_store.close t.store_contents
   end
 end
 
@@ -457,16 +458,22 @@ let test_truncated_inodes () =
     Inode.Val.of_list [ (s00, normal foo); (s10, normal foo) ] |> to_truncated
   in
   Inode.Val.list v1 |> ignore;
-  (iter_steps @@ fun step -> Inode.Val.find v1 step);
-  (iter_steps @@ fun step -> Inode.Val.add v1 step (normal bar));
-  (iter_steps @@ fun step -> Inode.Val.remove v1 step);
+  ( iter_steps @@ fun step ->
+    Inode.Val.find v1 step );
+  ( iter_steps @@ fun step ->
+    Inode.Val.add v1 step (normal bar) );
+  ( iter_steps @@ fun step ->
+    Inode.Val.remove v1 step );
   (* v2 is just a Truncated inode of tag Tree. The pointers are built after
      the call to [to_truncated], they are [Intact]. *)
   let v2 = Inode.Val.add v1 s01 (normal foo) in
   Inode.Val.list v2 |> ignore;
-  (iter_steps @@ fun step -> Inode.Val.find v1 step);
-  (iter_steps @@ fun step -> Inode.Val.add v1 step (normal bar));
-  (iter_steps @@ fun step -> Inode.Val.remove v1 step);
+  ( iter_steps @@ fun step ->
+    Inode.Val.find v1 step );
+  ( iter_steps @@ fun step ->
+    Inode.Val.add v1 step (normal bar) );
+  ( iter_steps @@ fun step ->
+    Inode.Val.remove v1 step );
   (* v3 is just a Truncated inode of tag Tree. The pointers are built before
      the call to [to_truncated], they are [Broken]. *)
   let v3 =
@@ -474,10 +481,14 @@ let test_truncated_inodes () =
       [ (s00, normal foo); (s10, normal bar); (s01, normal bar) ]
     |> to_truncated
   in
-  (with_failure @@ fun () -> Inode.Val.list v3 |> ignore);
-  (iter_steps_with_failure @@ fun step -> Inode.Val.find v3 step);
-  (iter_steps_with_failure @@ fun step -> Inode.Val.add v3 step (normal bar));
-  (iter_steps_with_failure @@ fun step -> Inode.Val.remove v3 step);
+  ( with_failure @@ fun () ->
+    Inode.Val.list v3 |> ignore );
+  ( iter_steps_with_failure @@ fun step ->
+    Inode.Val.find v3 step );
+  ( iter_steps_with_failure @@ fun step ->
+    Inode.Val.add v3 step (normal bar) );
+  ( iter_steps_with_failure @@ fun step ->
+    Inode.Val.remove v3 step );
   Context.close t
 
 let test_intermediate_inode_as_root () =
@@ -491,7 +502,10 @@ let test_intermediate_inode_as_root () =
     Inode.Val.of_list
       [ (s000, normal foo); (s001, normal bar); (s010, normal foo) ]
   in
-  let* h_depth0 = Inode.batch t.store @@ fun store -> Inode.add store v0 in
+  let* h_depth0 =
+    Inode.batch t.store @@ fun store ->
+    Inode.add store v0
+  in
   let (`Inode h_depth1) =
     match Inode.Val.pred v0 with
     | [ (_, (`Inode _ as pred)) ] -> pred
@@ -513,7 +527,10 @@ let test_intermediate_inode_as_root () =
     Alcotest.fail "Failed to list entries of loaded inode";
   let _ = Inode.Val.remove v s000 in
   let _ = Inode.Val.add v s000 (normal foo) in
-  let* _ = Inode.batch t.store @@ fun store -> Inode.add store v in
+  let* _ =
+    Inode.batch t.store @@ fun store ->
+    Inode.add store v
+  in
 
   (* On inode with depth=1 *)
   let* v =
diff --git a/test/irmin-pack/test_pack.ml b/test/irmin-pack/test_pack.ml
index c7ed55b298..bb714ffad0 100644
--- a/test/irmin-pack/test_pack.ml
+++ b/test/irmin-pack/test_pack.ml
@@ -208,7 +208,8 @@ module Pack = struct
     test t.pack >>= fun () ->
     let* pack2 = t.clone_pack ~readonly:false in
     test pack2 >>= fun () ->
-    Context.close t.index t.pack >>= fun () -> Pack.close pack2
+    Context.close t.index t.pack >>= fun () ->
+    Pack.close pack2
 
   let test_readonly_pack () =
     let* t = Context.get_pack () in
@@ -245,7 +246,8 @@ module Pack = struct
       Lwt.return_unit
     in
     test t.pack >>= fun () ->
-    Context.close t.index t.pack >>= fun () -> Context.close i r
+    Context.close t.index t.pack >>= fun () ->
+    Context.close i r
 
   let test_reuse_index () =
     (* index and pack with different names. However, this behaviour is not exposed by irmin_pack.*)
@@ -288,7 +290,8 @@ module Pack = struct
     let* i3, w3 = t.clone_index_pack ~readonly:true in
     let* y1 = Pack.find w3 k1 >|= get in
     Alcotest.(check string) "x1.3" x1 y1;
-    Context.close i2 w2 >>= fun () -> Context.close i3 w3
+    Context.close i2 w2 >>= fun () ->
+    Context.close i3 w3
 
   let test_close_pack () =
     let* t = Context.get_pack () in
@@ -369,7 +372,8 @@ module Pack = struct
       Alcotest.(check (option string)) "sync after flush" (Some x2) y2
     in
     test t.pack >>= fun () ->
-    Context.close t.index t.pack >>= fun () -> Context.close i r
+    Context.close t.index t.pack >>= fun () ->
+    Context.close i r
 
   let readonly_find_index_flush () =
     let* t = Context.get_pack () in
@@ -409,7 +413,8 @@ module Pack = struct
       check k3 x3 "find after flush new values"
     in
     test t.pack >>= fun () ->
-    Context.close t.index t.pack >>= fun () -> Context.close i r
+    Context.close t.index t.pack >>= fun () ->
+    Context.close i r
 
   let tests =
     [
@@ -502,7 +507,8 @@ module Branch = struct
     let* t1 = Branch.v ~fresh:true name in
     let* t2 = Branch.v ~fresh:false name in
     add t1 >>= fun () ->
-    Branch.close t1 >>= fun () -> test t2
+    Branch.close t1 >>= fun () ->
+    test t2
 
   let tests =
     [
diff --git a/test/irmin-pack/test_pack_version_bump.ml b/test/irmin-pack/test_pack_version_bump.ml
index 718e625f9a..18b38d62e2 100644
--- a/test/irmin-pack/test_pack_version_bump.ml
+++ b/test/irmin-pack/test_pack_version_bump.ml
@@ -22,7 +22,8 @@ module Util = struct
   let copy_dir src dst =
     assert (not (Sys.file_exists dst));
     (* don't check if it is empty; perhaps we should *)
-    Filename.quote_command "cp" [ "-R"; src; dst ] |> fun cmd ->
+    Filename.quote_command "cp" [ "-R"; src; dst ]
+    |> fun cmd ->
     exec_cmd cmd |> function
     | Ok () -> ()
     | Error n ->
diff --git a/test/irmin-unix/test_unix.ml b/test/irmin-unix/test_unix.ml
index 244b0255d6..49978871f7 100644
--- a/test/irmin-unix/test_unix.ml
+++ b/test/irmin-unix/test_unix.ml
@@ -55,7 +55,9 @@ module Git = struct
     let+ () =
       if Sys.file_exists test_db then
         Git_unix.Store.v (Fpath.v test_db) >>= function
-        | Ok t -> Git_unix.Store.reset t >|= fun _ -> ()
+        | Ok t ->
+            Git_unix.Store.reset t >|= fun _ ->
+            ()
         | Error _ -> Lwt.return_unit
       else Lwt.return_unit
     in
```

</details>

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -198,6 +198,16 @@ module String_id = struct
     && Option.is_none
          (String.lfindi s ~pos:3 ~f:(fun _ c -> not (Char_id.is_kwdop c)))
 
+  let is_monadic s =
+    let is_monadop s =
+      match String.sub s ~pos:0 ~len:(min 2 (String.length s)) with
+      | ">>" | ">|" | "@@" | "@>" -> true
+      | _ -> false
+    in
+    is_monadop s
+    (* "*>>=", "+>>>", "/>>|", etc. *)
+    || (String.length s > 3 && is_monadop (String.sub s ~pos:1 ~len:2))
+
   let is_infix i =
     if Char_id.is_infixop i.[0] then true
     else
@@ -226,6 +236,8 @@ module Longident = struct
   let is_prefix = test ~f:String_id.is_prefix
 
   let is_monadic_binding = test ~f:String_id.is_monadic_binding
+
+  let is_monadic = test ~f:String_id.is_monadic
 
   let is_infix = test ~f:String_id.is_infix
 
@@ -293,6 +305,8 @@ module Exp = struct
   let is_index_op = test_id ~f:Longident.is_index_op
 
   let is_monadic_binding = test_id ~f:Longident.is_monadic_binding
+
+  let is_monadic = test_id ~f:Longident.is_monadic
 
   let is_symbol = test_id ~f:Longident.is_symbol
 
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -198,16 +198,6 @@ module String_id = struct
     && Option.is_none
          (String.lfindi s ~pos:3 ~f:(fun _ c -> not (Char_id.is_kwdop c)))
 
-  let is_monadic s =
-    let is_monadop s =
-      match String.sub s ~pos:0 ~len:(min 2 (String.length s)) with
-      | ">>" | ">|" | "@@" | "@>" -> true
-      | _ -> false
-    in
-    is_monadop s
-    (* "*>>=", "+>>>", "/>>|", etc. *)
-    || (String.length s > 3 && is_monadop (String.sub s ~pos:1 ~len:2))
-
   let is_infix i =
     if Char_id.is_infixop i.[0] then true
     else
@@ -236,8 +226,6 @@ module Longident = struct
   let is_prefix = test ~f:String_id.is_prefix
 
   let is_monadic_binding = test ~f:String_id.is_monadic_binding
-
-  let is_monadic = test ~f:String_id.is_monadic
 
   let is_infix = test ~f:String_id.is_infix
 
@@ -305,8 +293,6 @@ module Exp = struct
   let is_index_op = test_id ~f:Longident.is_index_op
 
   let is_monadic_binding = test_id ~f:Longident.is_monadic_binding
-
-  let is_monadic = test_id ~f:Longident.is_monadic
 
   let is_symbol = test_id ~f:Longident.is_symbol
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 - Doc-comments code blocks with a language other than 'ocaml' (set in metadata) are not parsed as OCaml (#2037, @gpetiot)
 - More comprehensible error message in case of version mismatch (#2042, @gpetiot)
 - The global configuration file (`$XDG_CONFIG_HOME` or `$HOME/.config`) is only applied when no project is detected, `--enable-outside-detected-project` is set, and no applicable `.ocamlformat` file has been found. Global and local configurations are no longer used at the same time. (#2039, @gpetiot)
+- Break after `fun ->` following a monadic op (#2040, @gpetiot)
 
 ### New features
 
--- a/lib/Ast.mli
+++ b/lib/Ast.mli
@@ -63,6 +63,8 @@ module String_id : sig
   (** [is_monadic_binding id] returns whether [id] is a monadic binding
       operator of the form [let**] or [and**] where [**] can be 1 or more
       operator chars. *)
+
+  val is_monadic : string -> bool
 end
 
 module Longident : sig
@@ -81,6 +83,8 @@ module Longident : sig
       operator of the form [let**] or [and**] where [**] can be 1 or more
       operator chars. *)
 
+  val is_monadic : t -> bool
+
   val field_alias_str : field:t -> string -> bool
 
   val field_alias : field:t -> t -> bool
@@ -97,6 +101,8 @@ module Exp : sig
   (** [is_monadic_binding id] returns whether [id] is a monadic binding
       operator of the form [let**] or [and**] where [**] can be 1 or more
       operator chars. *)
+
+  val is_monadic : expression -> bool
 
   val exposed_left : expression -> bool
   (** [exposed_left exp] holds if the left-most subexpression of [exp] is a
--- a/lib/Ast.mli
+++ b/lib/Ast.mli
@@ -63,8 +63,6 @@ module String_id : sig
   (** [is_monadic_binding id] returns whether [id] is a monadic binding
       operator of the form [let**] or [and**] where [**] can be 1 or more
       operator chars. *)
-
-  val is_monadic : string -> bool
 end
 
 module Longident : sig
@@ -83,8 +81,6 @@ module Longident : sig
       operator of the form [let**] or [and**] where [**] can be 1 or more
       operator chars. *)
 
-  val is_monadic : t -> bool
-
   val field_alias_str : field:t -> string -> bool
 
   val field_alias : field:t -> t -> bool
@@ -101,8 +97,6 @@ module Exp : sig
   (** [is_monadic_binding id] returns whether [id] is a monadic binding
       operator of the form [let**] or [and**] where [**] can be 1 or more
       operator chars. *)
-
-  val is_monadic : expression -> bool
 
   val exposed_left : expression -> bool
   (** [exposed_left exp] holds if the left-most subexpression of [exp] is a
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1797,7 +1797,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         ; ( Nolabel
           , ({pexp_desc= Pexp_fun _; pexp_loc; pexp_attributes; _} as r) ) ]
       )
-    when Longident.is_infix id
+    when Longident.is_monadic id
          && (not (Longident.is_monadic_binding id))
          && not c.conf.fmt_opts.break_infix_before_func ->
       (* side effects of Cmts.fmt c.cmts before Sugar.fun_ is important *)
@@ -1809,18 +1809,6 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
       let fmt_cstr, xbody = type_constr_and_body c xbody in
       let indent_wrap = if parens then -2 else 0 in
       let pre_body, body = fmt_body c ?ext xbody in
-      let followed_by_infix_op =
-        match xbody.ast.pexp_desc with
-        | Pexp_apply
-            ( { pexp_desc= Pexp_ident {txt= id; loc= _}
-              ; pexp_attributes= []
-              ; _ }
-            , [ (Nolabel, _)
-              ; (Nolabel, {pexp_desc= Pexp_fun _ | Pexp_function _; _}) ] )
-          when Longident.is_infix id ->
-            true
-        | _ -> false
-      in
       wrap_fits_breaks_if c.conf parens "(" ")"
         ( hovbox 0
             (wrap_if has_attr "(" ")"
@@ -1840,8 +1828,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
                                (fmt_fun_args c xargs $ fmt_opt fmt_cstr)
                            $ fmt "@ ->" ) )
                    $ pre_body )
-               $ fmt_or followed_by_infix_op "@;<1000 0>" "@ "
-               $ body $ fmt_if parens_r ")" $ cmts_after ) )
+               $ break 1000 0 $ body $ fmt_if parens_r ")" $ cmts_after ) )
         $ fmt_atrs )
   | Pexp_apply
       ( ( {pexp_desc= Pexp_ident {txt= id; loc= _}; pexp_attributes= []; _}
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1797,7 +1797,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         ; ( Nolabel
           , ({pexp_desc= Pexp_fun _; pexp_loc; pexp_attributes; _} as r) ) ]
       )
-    when Longident.is_monadic id
+    when Longident.is_infix id
          && (not (Longident.is_monadic_binding id))
          && not c.conf.fmt_opts.break_infix_before_func ->
       (* side effects of Cmts.fmt c.cmts before Sugar.fun_ is important *)
--- a/test/passing/tests/infix_bind-break.ml.ref
+++ b/test/passing/tests/infix_bind-break.ml.ref
@@ -238,3 +238,10 @@ let _ = f (( let* ) x (fun y -> z))
 let _ = f (( let* ) x (function y -> z))
 
 let _ = (x >>= fun () -> ()) [@a]
+
+let _ =
+  x
+  >>= (* xxx *)
+  fun _ ->
+  (* yyy *)
+  x
--- a/test/passing/tests/infix_bind-fit_or_vertical-break.ml.ref
+++ b/test/passing/tests/infix_bind-fit_or_vertical-break.ml.ref
@@ -244,3 +244,10 @@ let _ = f (( let* ) x (fun y -> z))
 let _ = f (( let* ) x (function y -> z))
 
 let _ = (x >>= fun () -> ()) [@a]
+
+let _ =
+  x
+  >>= (* xxx *)
+  fun _ ->
+  (* yyy *)
+  x
--- a/test/passing/tests/infix_bind-fit_or_vertical.ml.ref
+++ b/test/passing/tests/infix_bind-fit_or_vertical.ml.ref
@@ -17,15 +17,19 @@ f x >>= function
         g y >>= function _ -> y () ) )
 ;;
 
-eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x ;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x ->
+x
+;;
 
 eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
-|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
+|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx ->
+x
 ;;
 
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
   eeeeeeeeeeee eeeeeeeeee
-|> fun x -> x
+|> fun x ->
+x
 ;;
 
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
@@ -34,20 +38,26 @@ eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
 xxxxxxxxxxx xxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxx xxxxxxxxxxx
 ;;
 
-eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x ;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x ->
+x
+;;
 
 eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
-|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
+|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx ->
+x
 ;;
 
-eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxx xxxxxxxxxxxxx -> x ;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxx xxxxxxxxxxxxx ->
+x
+;;
 
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee
-|> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx -> x
+|> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx ->
+x
 ;;
 
-eeeeeeeeeeeee eeeeeeeeeeeeeeeeee
-|> fun xxxxxxxxxxxxx -> xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxxxxxx ->
+xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
 ;;
 
 eeeeeeeeeeeee eeeeeeeeee
@@ -136,7 +146,8 @@ let end_gen_implementation ?toplevel ~ppf_dump
   ( clambda
   ++ Profile.record "cmm" (Cmmgen.compunit ~ppf_dump)
   ++ Profile.record "compile_phrases" (List.iter (compile_phrase ~ppf_dump))
-  ++ fun () -> () ) ;
+  ++ fun () ->
+    () ) ;
   fooooooooooooooo
 
 let foo =
@@ -208,7 +219,9 @@ let f =
 (** The tests below are testing a dropped comment with
     `--no-break-infix-before-func` *)
 
-let _ = x |> fun y -> y (* *)
+let _ =
+  x |> fun y ->
+  y (* *)
 
 let _ = x |> function y -> y (* *)
 
@@ -229,11 +242,10 @@ let encoder f =
   field_encode (f.fget x) k
 
 let default =
-  command ## hasPermission #= (fun ctx -> foooooooooooooooooo fooooooooooo) ;
-  command
-  ## hasPermission
-  #= (fun ctx ->
-  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo ) ;
+  command##hasPermission #= (fun ctx ->
+  foooooooooooooooooo fooooooooooo) ;
+  command##hasPermission #= (fun ctx ->
+  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo) ;
   foo
 
 let _ = ( let* ) x (fun y -> z)
--- a/test/passing/tests/infix_bind-fit_or_vertical.ml.ref
+++ b/test/passing/tests/infix_bind-fit_or_vertical.ml.ref
@@ -3,7 +3,8 @@ g y >>= fun () ->
 f x >>= fun y ->
 g y >>= fun () ->
 f x >>= fun y ->
-g y >>= fun () -> y ()
+g y >>= fun () ->
+y ()
 ;;
 
 f x >>= function
@@ -45,8 +46,8 @@ eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee
 |> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx -> x
 ;;
 
-eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxxxxxx ->
-xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee
+|> fun xxxxxxxxxxxxx -> xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
 ;;
 
 eeeeeeeeeeeee eeeeeeeeee
@@ -141,7 +142,9 @@ let end_gen_implementation ?toplevel ~ppf_dump
 let foo =
   (* get the tree origin *)
   get_store_tree s >>= function
-  | None -> f t >|= fun x -> Ok x (* no transaction is needed *)
+  | None ->
+      f t >|= fun x ->
+      Ok x (* no transaction is needed *)
   | Some (origin, old_tree) ->
       let batch = {repo; tree= old_tree; origin} in
       let b = Batch batch in
@@ -154,7 +157,9 @@ let _ =
   | Afoooooooooooooooooo fooooooooo -> false
   | Bfoooooooooooooooooooooo fooooooooo -> true
 
-let _ = foo >>= fun [@warning "-4"] x -> fooooooooooooooooooooooo
+let _ =
+  foo >>= fun [@warning "-4"] x ->
+  fooooooooooooooooooooooo
 
 let _ =
   foo >>= fun [@warning "-4"] x y ->
@@ -170,7 +175,9 @@ let _ =
   foo >>= fun (* foo before *) [@warning "-4"] (* foo after *) x ->
   fooooooooooooooooooooooo
 
-let f = Ok () >>= (* *) fun _ -> Ok ()
+let f =
+  Ok () >>= (* *) fun _ ->
+  Ok ()
 
 let f =
   (* fooooooooooooooo foooooooooooooooo *)
@@ -213,16 +220,20 @@ let _ = () (* This is needed here to avoid the comment above from moving *)
 
 let encoder f =
   let field_encode = unstage (t f.ftype) in
-  stagged @@ fun x k : t -> field_encode (f.fget x) k
+  stagged @@ fun x k : t ->
+  field_encode (f.fget x) k
 
 let encoder f =
   let field_encode = unstage (t f.ftype) in
-  stagged @@ fun x k : t -> field_encode (f.fget x) k
+  stagged @@ fun x k : t ->
+  field_encode (f.fget x) k
 
 let default =
-  command##hasPermission #= (fun ctx -> foooooooooooooooooo fooooooooooo) ;
-  command##hasPermission #= (fun ctx ->
-  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo) ;
+  command ## hasPermission #= (fun ctx -> foooooooooooooooooo fooooooooooo) ;
+  command
+  ## hasPermission
+  #= (fun ctx ->
+  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo ) ;
   foo
 
 let _ = ( let* ) x (fun y -> z)
@@ -233,4 +244,12 @@ let _ = f (( let* ) x (fun y -> z))
 
 let _ = f (( let* ) x (function y -> z))
 
-let _ = ((x >>= fun () -> ()) [@a])
+let _ =
+  ( (x >>= fun () ->
+    ())
+  [@a] )
+
+let _ =
+  x >>= (* xxx *) fun _ ->
+  (* yyy *)
+  x
--- a/test/passing/tests/infix_bind.ml
+++ b/test/passing/tests/infix_bind.ml
@@ -17,15 +17,19 @@ f x >>= function
         g y >>= function _ -> y () ) )
 ;;
 
-eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x ;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x ->
+x
+;;
 
 eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
-|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
+|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx ->
+x
 ;;
 
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
   eeeeeeeeeeee eeeeeeeeee
-|> fun x -> x
+|> fun x ->
+x
 ;;
 
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
@@ -34,20 +38,26 @@ eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeee
 xxxxxxxxxxx xxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxx xxxxxxxxxxx
 ;;
 
-eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x -> x ;;
+eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee |> fun x ->
+x
+;;
 
 eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
-|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx -> x
+|> fun xxxxxx xxxxxxxxxx xxxxxxxx xxxxxxxx ->
+x
 ;;
 
-eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxx xxxxxxxxxxxxx -> x ;;
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxx xxxxxxxxxxxxx ->
+x
+;;
 
 eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee
-|> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx -> x
+|> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx ->
+x
 ;;
 
-eeeeeeeeeeeee eeeeeeeeeeeeeeeeee
-|> fun xxxxxxxxxxxxx -> xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxxxxxx ->
+xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
 ;;
 
 eeeeeeeeeeeee eeeeeeeeee
@@ -131,7 +141,8 @@ let end_gen_implementation ?toplevel ~ppf_dump
   ( clambda
   ++ Profile.record "cmm" (Cmmgen.compunit ~ppf_dump)
   ++ Profile.record "compile_phrases" (List.iter (compile_phrase ~ppf_dump))
-  ++ fun () -> () ) ;
+  ++ fun () ->
+    () ) ;
   fooooooooooooooo
 
 let foo =
@@ -203,7 +214,9 @@ let f =
 (** The tests below are testing a dropped comment with
     `--no-break-infix-before-func` *)
 
-let _ = x |> fun y -> y (* *)
+let _ =
+  x |> fun y ->
+  y (* *)
 
 let _ = x |> function y -> y (* *)
 
@@ -224,10 +237,10 @@ let encoder f =
   field_encode (f.fget x) k
 
 let default =
-  command ## hasPermission #= (fun ctx -> foooooooooooooooooo fooooooooooo) ;
-  command ## hasPermission
-  #= (fun ctx ->
-  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo ) ;
+  command##hasPermission #= (fun ctx ->
+  foooooooooooooooooo fooooooooooo) ;
+  command##hasPermission #= (fun ctx ->
+  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo) ;
   foo
 
 let _ = ( let* ) x (fun y -> z)
--- a/test/passing/tests/infix_bind.ml
+++ b/test/passing/tests/infix_bind.ml
@@ -3,7 +3,8 @@ g y >>= fun () ->
 f x >>= fun y ->
 g y >>= fun () ->
 f x >>= fun y ->
-g y >>= fun () -> y ()
+g y >>= fun () ->
+y ()
 ;;
 
 f x >>= function
@@ -45,8 +46,8 @@ eeeeeeeeeeeee eeeeeeeeeeeeeeeeee eeeeeeeeeeeeeeeeeee
 |> fun xxxxxxxx xxxxxxxxx xxxxxxxxxxxxx -> x
 ;;
 
-eeeeeeeeeeeee eeeeeeeeeeeeeeeeee |> fun xxxxxxxxxxxxx ->
-xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
+eeeeeeeeeeeee eeeeeeeeeeeeeeeeee
+|> fun xxxxxxxxxxxxx -> xxxxxxxx xxxxxxxxxx xxxxxxxxx xxxxxxxxxxxxx
 ;;
 
 eeeeeeeeeeeee eeeeeeeeee
@@ -136,7 +137,9 @@ let end_gen_implementation ?toplevel ~ppf_dump
 let foo =
   (* get the tree origin *)
   get_store_tree s >>= function
-  | None -> f t >|= fun x -> Ok x (* no transaction is needed *)
+  | None ->
+      f t >|= fun x ->
+      Ok x (* no transaction is needed *)
   | Some (origin, old_tree) ->
       let batch = {repo; tree= old_tree; origin} in
       let b = Batch batch in
@@ -149,7 +152,9 @@ let _ =
   | Afoooooooooooooooooo fooooooooo -> false
   | Bfoooooooooooooooooooooo fooooooooo -> true
 
-let _ = foo >>= fun [@warning "-4"] x -> fooooooooooooooooooooooo
+let _ =
+  foo >>= fun [@warning "-4"] x ->
+  fooooooooooooooooooooooo
 
 let _ =
   foo >>= fun [@warning "-4"] x y ->
@@ -165,7 +170,9 @@ let _ =
   foo >>= fun (* foo before *) [@warning "-4"] (* foo after *) x ->
   fooooooooooooooooooooooo
 
-let f = Ok () >>= (* *) fun _ -> Ok ()
+let f =
+  Ok () >>= (* *) fun _ ->
+  Ok ()
 
 let f =
   (* fooooooooooooooo foooooooooooooooo *)
@@ -208,16 +215,19 @@ let _ = () (* This is needed here to avoid the comment above from moving *)
 
 let encoder f =
   let field_encode = unstage (t f.ftype) in
-  stagged @@ fun x k : t -> field_encode (f.fget x) k
+  stagged @@ fun x k : t ->
+  field_encode (f.fget x) k
 
 let encoder f =
   let field_encode = unstage (t f.ftype) in
-  stagged @@ fun x k : t -> field_encode (f.fget x) k
+  stagged @@ fun x k : t ->
+  field_encode (f.fget x) k
 
 let default =
-  command##hasPermission #= (fun ctx -> foooooooooooooooooo fooooooooooo) ;
-  command##hasPermission #= (fun ctx ->
-  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo) ;
+  command ## hasPermission #= (fun ctx -> foooooooooooooooooo fooooooooooo) ;
+  command ## hasPermission
+  #= (fun ctx ->
+  foooooooooooooooooo fooooooooooo foooooo fooooooooo foooooooooo ) ;
   foo
 
 let _ = ( let* ) x (fun y -> z)
@@ -228,4 +238,12 @@ let _ = f (( let* ) x (fun y -> z))
 
 let _ = f (( let* ) x (function y -> z))
 
-let _ = ((x >>= fun () -> ()) [@a])
+let _ =
+  ( (x >>= fun () ->
+    ())
+  [@a] )
+
+let _ =
+  x >>= (* xxx *) fun _ ->
+  (* yyy *)
+  x


gpetiot:

lib/Ast.ml
```diff
@@ -198,6 +198,16 @@ module String_id = struct
...
 
+  let is_monadic s =
+    let is_monadop s =
+      match String.sub s ~pos:0 ~len:(min 2 (String.length s)) with
+      | ">>" | ">|" | "@@" | "@>" -> true
+      | _ -> false
+    in
+    is_monadop s
+    (* "*>>=", "+>>>", "/>>|", etc. *)
+    || (String.length s > 3 && is_monadop (String.sub s ~pos:1 ~len:2))
```
Julow, line 209: It seems that you really want to check whether `s` contains `">>"`. Do you have a list of the operators that should be considered a monadic bind ?

Julow (commented):
What's wrong with the previous style ?

lib/Ast.ml
gpetiot: I just used the same definition as `ocp-indent` (https://github.com/OCamlPro/ocp-indent/blob/master/src/indentBlock.ml#L440-L450) for consistency.

gpetiot:
> What's wrong with the previous style ?

The idea was to only have this behavior for intended cases like `>>=` instead of `|>` or `+` or whatever operator was allowed by `is_infix`, and I propose to use the same check as `ocp-indent` for consistency. (btw I think a similar check would have its place in the stdlib).
I also find it easier to read to always break after the `->` and not only if followed by another monadic expression.

lib/Ast.ml
Julow: ocp-indent does that to indent differently some operators. OCamlformat doesn't do that, it formats specifically the `_ OP fun _ -> EXP` case. (and even fore specifically when `EXP` is an operator apply)
To be more consistent with ocp-indent you would need to format every operators that way (not just when it's followed by a fun), which would be a big change. 
I don't think we need consistency with ocp-indent here.

gpetiot: I reverted this change, on the other hand what do you think of always breaking after the `->` ? if not then we can close this PR.

Julow: Breaking after the `->` isn't nice if the expression fits on one line. For example:
```
  f (g >>= fun _ -> h)
```
Perhaps breaking after the `->` if there's a chain of more than one operator ?

PR #2042, opened 2022-03-16
More comprehensible error message in case of version mismatch

gpetiot:
cc @tmattio @samoht for user friendliness.


gpetiot:

PR #2043, opened 2022-03-17
Do not use extension nodes for sugared extension forms

### LINKED ISSUE #2081: Bug: match and match%ext are not treated the same by nested-match=align
**Describe the bug**

`nested-match=align` doesn't treat a "normal" `match` and one with a ppx extension (such as `match%lwt`) the same way. The normal ones are correctly aligned. While the ones with extension are nested.

**How to Reproduce**

Minimal version

```
let () =
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false -> match foo with true -> false | false -> false

let () =
  match%lwt foo with
  | true -> false
  | false -> (
      match%lwt foo with
      | true -> false
      | false -> (
          match%lwt foo with
          | true -> false
          | false -> (
              match%lwt foo with
              | true -> false
              | false -> ( match%lwt foo with true -> false | false -> false))))
```

```
version=0.21.0
profile=default

nested-match=align # deprecated
```

With our profile

```
let () =
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false ->
  match foo with
  | true -> false
  | false -> false

let () =
  match%lwt foo with
  | true -> false
  | false ->
    ( match%lwt foo with
    | true -> false
    | false ->
      ( match%lwt foo with
      | true -> false
      | false ->
        ( match%lwt foo with
        | true -> false
        | false ->
          ( match%lwt foo with
          | true -> false
          | false -> false
          )
        )
      )
    )
```

```
version=0.21.0
quiet=true
profile=default

break-cases=toplevel # deprecated
break-infix-before-func=false
break-infix=fit-or-vertical
break-struct=natural # deprecated
cases-exp-indent=2
doc-comments=before
exp-grouping=preserve
indicate-multiline-delimiters=closing-on-separate-line # deprecated
let-and=sparse
m=120
max-indent=2
module-item-spacing=preserve # deprecated
nested-match=align # deprecated
ocp-indent-compat=true
parens-ite=true
parens-tuple=multi-line-only
type-decl=sparse
```

gpetiot:
Fix #2081

- Preserves more of the sugared extension forms (wasn't the case when we formatted a function in an Apply when `break-infix-before-func=false` for example)
- Behaves the same with/without extension (for example regarding `module-item-spacing` (only diff we get when running test_branch.sh))
- Removes code duplication and the whole extension-node layer in Fmt_ast.ml and Ast.ml
- Removes `Source.extension_using_sugar` that relied on the input anyway


lib/Fmt_ast.ml
```diff
@@ -3554,7 +3480,7 @@ and fmt_functor_arg c {loc; txt= arg} =
 
 and fmt_module_type c ({ast= mty; _} as xmty) =
   let ctx = Mty mty in
-  let {pmty_desc; pmty_loc; pmty_attributes} = mty in
+  let {pmty_desc; pmty_loc; pmty_attributes; pmty_ext= _} = mty in
```
hhugo, line 3483: Why is it ok to ignore pmty_ext here ?

test/passing/tests/extensions-indent.ml.ref
```diff
@@ -387,11 +391,11 @@ let _ = f (if%ext x then y else z)
 
 let _ = f [%ext if x then y else z]
 
-let _ = (if%ext x then y else z) [@attr]
+let _ = if%ext [@attr] x then y else z
```
hhugo, line 394: Is the diff regarding attributes position related?

vendor/parse-wyc/lib/parser.mly
```diff
@@ -2352,7 +2338,17 @@ simple_expr:
   | BEGIN ext_attributes seq_expr END
       { Pexp_beginend $3, $2 }
   | BEGIN ext_attributes END
-      { Pexp_construct (mkloc (Lident "()") (make_loc $sloc), None), $2 }
+      { let ext, attrs = $2 in
```
hhugo, line 2341: This PR would be a good opportunity to preserve begin end in this case.

vendor/ocaml-4.13-extended/parsetree.mli
```diff
@@ -90,6 +90,7 @@ and core_type =
      ptyp_loc: Location.t;
      ptyp_loc_stack: location_stack;
      ptyp_attributes: attributes; (* ... [@id1] [@id2] *)
+     ptyp_ext: string loc option;
```
hhugo, line 93: It is weird to have ext attached to core_type instead of core_type_desc. The same comment apply to other kind. Attaching to "desc" would make it easier to not forget any cases in fmt.ml

lib/Fmt_ast.ml
gpetiot: It should not have been added to module_type since no extensions are allowed, I will remove it.

vendor/parse-wyc/lib/parser.mly
gpetiot: Done

PR #2046, opened 2022-03-23
Fix unstable comment formatting around prefix op

### LINKED ISSUE #1264: Bug: unstable comment formatting for prefix op
Hi,

    !(*a*) (*b*)x

gpetiot:
Fix #1264, no diff with test_branch.sh

I'm not a fan of this fix, but having the `@,` break unconditionally may look surprising for users, I like it even less.


Julow (approved):
Fine to me but I'd be curious to see the original code.

PR #2047, opened 2022-03-24
Do not check output of '--help' command in CI

### LINKED ISSUE #2027: Do not expose cmdliner output in ocamlformat tests (at least in CI)
> I would be grateful if people could refrain from putting expect tests on cmdliner’s man page and error message output in the test suites that are run by default on the opam repository.
> Cmdliner’s man page and error messages are not part of its interface and subject to change from version to version. Having packages fail in the CI because of this is unhelpful and time consuming.

Wished by dbuenzli (cf. https://discuss.ocaml.org/t/ann-cmdliner-1-1-0/9295/2)

gpetiot:
Fix #2027

### DIFF
--- a/dune
+++ b/dune
@@ -22,7 +22,7 @@
  (package ocamlformat))
 
 (rule
- (alias runtest)
+ (alias help)
  (package ocamlformat)
  (action
   (diff ocamlformat-help.txt ocamlformat-help.actual)))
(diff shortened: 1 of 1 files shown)


Julow:
Why ? We at least need a CI check for making sure the help file is up to date.

gpetiot:
In https://discuss.ocaml.org/t/ann-cmdliner-1-1-0/9295/2 dbuenzli insists that we should not have such tests in CI as it's making the CI fail every time cmdliner changes its output (see PR in opam-repo for cmdliner 1.1.0 and 1.1.1 more recently).

Julow:
I think we should have the check in our own CI anyway. I've started https://github.com/ocaml-ppx/ocamlformat/pull/2055

PR #2048, opened 2022-03-24
bench: Add upper bound for dune (< 3.0)

gpetiot:


gpetiot:
Apparently the `lint-doc` CI target fails because of `bechamel-js`, so I'm closing this PR.

PR #2055, opened 2022-03-28
CI: add Github Action workflow for MacOS and Windows

Julow:
It builds and runs the tests on macos, ubuntu and windows and also run the `@help` alias.
It doesn't try different version of OCaml because that's already done by OCaml-CI.


Julow:
The new check shows that the helps files weren't uptodate and that the tests fail on Windows (due to end of lines perhaps?)

1972chris (approved):

test/cli/conf.t
```diff
@@ -5,39 +5,39 @@ Invalid version:
 
 Exit code is printed by hand because sed succeeding would hide the error.
 
-  $ (<a.ml ocamlformat --impl -; echo [$?]) 2>&1 | sed 's/version is "[^"]*"/version is "..."/g'
-  ocamlformat: Error while parsing $TESTCASE_ROOT/.ocamlformat:
+  $ (<a.ml ocamlformat --impl -; echo [$?]) 2>&1 | sed 's/version is "[^"]*"/version is "..."/g' | make_deterministic
+  ocamlformat: Error while parsing <testcase_root>/.ocamlformat:
```
Julow, line 9: Why is this needed ?

gpetiot: Because dune doesn't pick up Windows paths and doesn't do the rewrite with TESTCASE_ROOT

Julow: Perhaps we should disable all the tests on Windows ? We could add a windows-specific formatting test to be notified of formatting changes and that would be already better than before.

gpetiot: Done, I've also added some tests for the Eol_compat lib. Everything looks good to me now.

gpetiot (approved):
@Julow if it looks good to you we can merge!

test/unit/test_indent.ml
```diff
@@ -24,11 +24,13 @@ let x =
 y
 in
 2|}
-        ~expected:{|let f =
+        ~expected:
+          (Eol_compat.normalize_eol ~line_endings:`Lf
```
Julow, line 28: Would be better inside the `test` function.

vendor/eol-compat/dune
```diff
@@ -0,0 +1,3 @@
+(library
+ (name eol_compat)
```
Julow, line 2: Why is this placed in the `vendor` directory ? Would be better in `lib`.

Julow (commented):
I have a small comment but otherwise looks good.

test/unit/test_indent.ml
gpetiot: done

vendor/eol-compat/dune
gpetiot: done

PR #2059, opened 2022-03-30
Bump JS dependencies to include 0.15

cgaebel:
As far as I can tell, ocamlformat works fine with the latest
(and recently released) Jane Street libraries, 0.15. This PR
bumps ocamlformat's dependencies to include that latest release.

I'd like to push for releasing this change into opam sooner rather
than later. Is there a process for that?

### DIFF
--- a/ocamlformat-bench.opam
+++ b/ocamlformat-bench.opam
@@ -14,7 +14,7 @@ depends: [
   "bechamel" {>= "0.2.0"}
   "bechamel-js" {>= "0.2.0"}
   "ocamlformat" {= version}
-  "stdio" {< "v0.15"}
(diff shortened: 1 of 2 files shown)


cgaebel:
This is failing the Changelog check. Do changes of this nature deserve a changelog entry?

gpetiot:
Thanks for this change, we shouldn't need to add a changelog entry for that.

To apply this change on opam-repository I think you can directly open a PR there with that change applied to ocamlformat opam files.

gpetiot (approved):
LGTM, thanks!

kit-ty-kate:
Could the upper-bound constraint be totally removed? It makes it much easier for next time

PR #2064, opened 2022-04-19
Set ocaml-version to a fixed version by default to avoid reproducibility issues and surprising behaviours

kit-ty-kate:
See for example https://github.com/ocaml/opam-repository/pull/21236#discussion_r853088590

To me, ocamlformat should always aim to give a reproducible output regardless of the version of the compiler you currently have. To this end I’m setting the default value of `ocaml-version` to the oldest version of the compiler to ever be compatible with ocamlformat.


Julow (approved):
This is the most correct behavior. 

> to the oldest version of the compiler to ever be compatible with ocamlformat.

It should be the oldest version of the compiler compatible with the programs that are formatted, this has never been specified but is perhaps close to 4.02 or 4.00. But that doesn't matter because this value is only compared against 4.13 and 4.14 for now.

PR #2067, opened 2022-04-28
Fix build for OCaml 5

samoht:


ocamlformat.opam
```diff
@@ -8,7 +8,7 @@ authors: ["Josh Berdine <jjb@fb.com>"]
 homepage: "https://github.com/ocaml-ppx/ocamlformat"
 bug-reports: "https://github.com/ocaml-ppx/ocamlformat/issues"
 depends: [
-  "ocaml" {>= "4.08" & < "4.15"}
+  "ocaml" {>= "5.00"}
```
Julow, line 11:
```suggestion
  "ocaml" {>= "4.08"}
```
Compatible with 4.08, no needs to restraint the range.

gpetiot (approved):
It looks good to me, thanks.

I will setup a local 5.0 switch for me but do we know when ocaml-ci will offer a 5.0 target?

PR #2069, opened 2022-04-30
Suggest a simple reformat on save setup

benjamin-thomas:
Having a quick look at Ocaml this morning, the official vscode extension does not handle this common behavior.

I wasted a bit of time trying to find alternatives so I'm suggesting a quick workaround for those who want to get something going quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,15 @@ The most common usecase involves using the `dune` build system, once your projec
 dune build @fmt
 ```
 
+### Format on save with your editor
+
+Alternatively, you may want to see your files reformatted on save. If you're running on Linux or macOS, you may enable this workflow easily regardless of your editor:
+
+```bash
+# apt install entr or brew install entr
+find ./bin ./test -name "*.ml" | entr -p ocamlformat --inplace /_
+```
+
 ## Community
 
 * forum: <https://discuss.ocaml.org/tags/ocamlformat>


benjamin-thomas:
Please let me know if you want me to sign this CLA (I don't think it would be required for such a simple comment on a README)

gpetiot:
Hi, thank you for your contribution. No need to sign the CLA for documentation indeed.

I think this kind of information would be better in the manual though: `doc/editor_setup.mld`, in a new section specific to VScode maybe.

benjamin-thomas:
Sounds good to me.

Believe it or not, my eyes went straight to the `Formatting code!` section, I missed the smaller documentation link above!

Would you be ok if I remove the `Getting started` section from the README, and point straight to the online docs (for which I'll add a section with my suggestion).

It would be better IMO since both the README and the docs have a `Getting started` section and they differ slightly (which is a little confusing).

gpetiot:
> Would you be ok if I remove the `Getting started` section from the README, and point straight to the online docs (for which I'll add a section with my suggestion).

I like having a minimal doc to get started in the README so that people don't have to go to the manual straight away, it should be enough for most usecases (most people just need to create a .ocamlformat and run dune build @fmt). And then the manual is there for the more specific cases. But I'm not opposed to improvements.

gpetiot:
Replaced by #2132

benjamin-thomas:
> Replaced by #2132

Thanks :)

Sorry I forgot to get back to you in may, looks all good ^^

PR #2084, opened 2022-05-19
Display a warning for redundant option definitions when using '--print-config'

### LINKED ISSUE #2079: Feature request: offer an option to see only non standard values in a configuration

We have a .ocamlformat configuration that grew organically over time. It has many options (that interact with each other). And changing that file is always tricky (creates noise, makes existing PR harder to merge, requires us to go through the diff to make sure that we didn't introduce some ugly formatting, ...). But it actually contains values that are the default, which could be removed from our config, and would make the whole thing less scary. And it's kind of hard to know that without reading help.txt for each option we are using.

I'd like to have an option like `--print-config` (or maybe a new flag that would affect `--print-config`) to show only the minimum .ocamlformat required for our project. Or maybe which would show a warning for every option we use that is also the default value

**Additional context**

Our .ocamlformat

```
version=0.21.0
quiet=true
break-cases=toplevel
break-infix-before-func=false
break-infix=fit-or-vertical
break-separators=after
space-around-lists=false
dock-collection-brackets=true
break-sequences=true
break-struct=natural
cases-exp-indent=2
doc-comments=before
exp-grouping=preserve
field-space=loose
indicate-multiline-delimiters=closing-on-separate-line
let-and=sparse
m=120
max-indent=2
nested-match=align
ocp-indent-compat=true
parens-ite=true
parens-tuple=multi-line-only
sequence-style=terminator
space-around-arrays=true
space-around-lists=true
space-around-records=true
space-around-variants=true
type-decl=sparse
wrap-fun-args=true
module-item-spacing=preserve
```

would become
```
version=0.21.0
quiet=true
profile=default

break-cases=toplevel
break-infix-before-func=false
break-infix=fit-or-vertical
break-struct=natural
cases-exp-indent=2
doc-comments=before
exp-grouping=preserve
indicate-multiline-delimiters=closing-on-separate-line
let-and=sparse
m=120
max-indent=2
module-item-spacing=preserve
nested-match=align
ocp-indent-compat=true
parens-ite=true
parens-tuple=multi-line-only
type-decl=sparse
```

This is part of an effort to try to preserve the usage of ocamlformat in our codebase despite the fact that some deprecated options look critical to us.

gpetiot:
Fix #2079

It looks like a valuable feature to me.
Systematically (not only when using `--print-config` I mean) emitting a warning produces too much noise from what I observed, so I chose to modify the output of `--print-config` instead.


Julow (approved):
Looks good :)

PR #2086, opened 2022-05-20
github actions: Check the version number of Windows binaries

gpetiot:


gpetiot:
I expected the line I added to be present in the log produced by the github action, but apparently it's not.

PR #2088, opened 2022-05-24
Add a script generating a formatting preview of the upcoming release

gpetiot:
To automatize the preview process, as I've been doing it by hand for longer than I initially expected.

Does not open PR, as we have to manually check if the generated diff would be the same as upgrading to a previous version (in case a project is several versions behind), and also every way to authenticate using `gh` fails for me.


tools/preview_new_release.sh
```diff
@@ -0,0 +1,157 @@
...
+        git branch -D $preview_branch --quiet;
+    fi;
+    git checkout -b $preview_branch --quiet;
+
+    if [ "$project" == "ocaml/dune" ]; then
+        make release;
+        dune=_build/default/bin/dune.exe;
+    fi;
+
+    ocamlformat_file=`readlink -f .ocamlformat`;
```
Julow, line 149: Is here really a project doing that ? Otherwise `sed` can handle this, all the seds can be written as:
```
  sed -i --follow-symlinks -e "1iversion = $version" -e '/^version/d' .ocamlformat
```
What's the point of doing the formatting while the `version` line is removed ? It would be better to leave it as a check.

tools/preview_new_release.sh
```diff
@@ -0,0 +1,157 @@
...
+            prefix="$1";
+            prefix_set=1;
+            shift 1;
+            ;;
+        *)
+            usage;
+            exit 1;
+            ;;
+    esac
+done;
```
Julow, line 57: There's a builtin function for parsing options but it can only work with single character options. It would look like that:
```suggestion
function usage()
{
    echo "usage: $0 -u <GITHUB_USERNAME> -v <RELEASE> [-U] [-p <PREFIX>]"
}

while getopts ":u:v:Up:" opt; do
    case "$opt" in
        u)
            user=$OPTARG
            user_set=1
            ;;
        v)
            version=$OPTARG
            version_set=1
            ;;
        U)
            update_repos=1;
            ;;
        p)
            prefix=$OPTARG
            prefix_set=1
            ;;
        *)
            usage;
            exit 1;
            ;;
    esac
done;
```

Julow (approved):

tools/preview_new_release.sh
gpetiot: Yes `infer` is doing that.
We need to remove the version because this script's purpose it to make a preview of unreleased versions (right now 0.21.0-XX-YYYYYYYYY, I don't want to rely on having a local tag, I'm too clumsy and would get it pushed by mistake), so we need to disable the version check (we cannot pass --no-version-check to `dune build @fmt`).

Julow: I see. Perhaps we should have an environment variable for this option ? Otherwise, the script can be improved anyway with `-i --follow-symlinks`.

gpetiot: done

tools/preview_new_release.sh
```diff
@@ -0,0 +1,157 @@
...
+mkdir --parents $log_dir;
+
+for project in "${projects[@]}"; do
+    cd $prefix;
+
+    namespace=`echo $project | cut -d "/" -f 1`;
+    dir=`echo $project | cut -d "/" -f 2`;
+    echo "=> Checking $project";
+
+    if [ ! -d "$dir" ] ; then
```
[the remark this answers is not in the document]
gpetiot: done

PR #2089, opened 2022-05-25
Add dependency to ocamlformat-rpc-lib

gpetiot:
Upstreaming https://github.com/ocaml/opam-repository/pull/21421/commits/2f13dc613a6d131f4a60b2930fa767eaf617811a

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -56,6 +56,11 @@
    (>= 20201216))
   (ocaml-version
    (>= 3.3.0))
+  (ocamlformat-rpc-lib
+   (and
+    :with-test
+    :post
+    (= :version)))
   ocp-indent
   (odoc-parser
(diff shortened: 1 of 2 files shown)


Julow (commented):
I don't think this is the right fix. The issue seems to be a test that is not associated to the right package.
This will break the development environment.

gpetiot:
I think it's fine, as `test/rpc/dune` depends on `ocamlformat-rpc-lib` and the tests are enabled for `(package ocamlformat)` ()previously it was for ocamlformat-rpc when this package was distinct from ocamlformat.

Julow:
This will install the released version of `ocamlformat-rpc-lib` when we do `opam install --deps-only -t`, this is probably not what we want. I think it is associated to the wrong package.

PR #2094, opened 2022-05-26
Remove option 'ocp-indent-compat'

gpetiot:
This follows a request from @ceastlund

Jane Street is dropping ocp-indent so ocamlformat does not need to tweak its output to play nice with ocp-indent anymore, the `ocp-indent-compat` option will be removed.
#2099, #2102 and #2149 are behaviors previously taken care of by ocp-indent-compat, that are now preserved, the rest can be removed.


gpetiot:
@ceastlund IIRC all the features that you wanted to preserve will be preserved after this PR is merged and the `ocp-indent-compat` option is removed. Let me know if more work is needed on our side to improve your diff :)

gpetiot:
Would need to be started again from scratch, the attack angle changed regarding ocp-indent compatibility.

PR #2101, opened 2022-06-07
Rename 'bench' binary to 'ocamlformat-bench'

### LINKED ISSUE #2100: ocamlformat ships a public `bench` executable
It seems that in recent versions, ocamlformat started shipping a binary called `bench`.

If this is intentionally public, it would be good to prefix its name to avoid conflicts. If it's not it would be good to make it private.

This has been causing troubles in projects using `opam-monorepo` as ocamlformat is not the only one doing so, meaning vendoring it alongside those other packages causes dune to have multiple rules to build the `bench` executable.

I suspect this should be a private executable, used for benchmarking in ocamlformat only. There are a few tricks you can use in dune to be able to refer to a private executable as if it was public. If that may help you can see how we do this in opam-monorepo:
https://github.com/ocamllabs/opam-monorepo/blob/main/test/bin/dune#L9.

Let me know if I can help with this in any way!

gpetiot:
Fix #2100

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@
 
 ### Changes
 
-- `bench` binary renamed to `ocamlformat-bench` to avoid conflicts (#<PR_NUMBER>, @gpetiot)
+- `bench` binary renamed to `ocamlformat-bench` to avoid conflicts (#2101, @gpetiot)
 - JaneStreet profile: set `max-indent = 2` (#2099, @gpetiot)
 
 ### New features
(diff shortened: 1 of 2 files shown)


Julow:
Will the benchmarks still recognize it and run it ? Is it still used at all ?

gpetiot:
Yes it's still working: https://autumn.ocamllabs.io/ocaml-ppx/ocamlformat/pull/2101

gpetiot:
I thought it might be useful to keep providing this binary but maybe I'm too optimistic and no one actually uses it.

PR #2114, opened 2022-06-29
Preserve normalization of cons-patterns into bracket lists

gpetiot:
Fixing a small regression caused by #2113, all is good now.
Adding some tests.

### DIFF
--- a/lib/Extended_ast.ml
+++ b/lib/Extended_ast.ml
@@ -52,22 +52,17 @@ module Parse = struct
     Ast_mapper.{default_mapper with binding_op}
 
   let list_pat pat =
-    let rec list_pat_ pat acc =
-      match pat.ppat_desc with
-      | Ppat_construct ({txt= Lident "[]"; loc= _}, None) -> (
-        (* Empty lists are always represented as Lident [] *)
-        match acc with [] -> None | _ -> Some (List.rev acc) )
-      | Ppat_construct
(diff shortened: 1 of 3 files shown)


lib/Extended_ast.ml
```diff
@@ -52,22 +52,17 @@ module Parse = struct
...
-              , { ppat_desc= Ppat_tuple [hd; ({ppat_attributes= []; _} as tl)]
-                ; ppat_attributes= []
-                ; _ } ) ) ->
-          list_pat_ tl (hd :: acc)
-      | _ -> None
-    in
-    list_pat_ pat []
+    match pat.ppat_desc with
+    (* Do not normalize (x :: []) *)
+    | Ppat_cons (_ :: _ :: _ :: _ as l) -> (
```
hhugo, line 57: I don't understand why we only do this for list of size at least 3 ?

gpetiot: That's actually for a size of 2 elements, the last element in the list is actually `Lident []`:
- `_ :: _ :: _` matches `[x ;  Lident[] ]` (that we don't want to normalize (*))
- `_ :: _ :: _ :: _` matches `[x ; y ; Lident[] ]` but not `[x; Lident []]`


(*) I don't remember who asked for this to be preserved but that's he current behavior

hhugo: I still don't understand why the special case here

PR #2119, opened 2022-07-07
Fix byte-compile warnings

syohex:
- Use `...' instead of '...'
- Remove needless quotes from choice values
- Escape single quote in docstring

```
ocamlformat.el:49:2: Warning: custom-declare-variable `ocamlformat-command'
    docstring has wrong usage of unescaped single quotes (use \= or different
    quoting)
ocamlformat.el:84:2: Warning: custom-declare-variable `ocamlformat-file-kind'
    docstring has wrong usage of unescaped single quotes (use \= or different
    quoting)
ocamlformat.el:84:12: Warning: defcustom for ‘ocamlformat-file-kind’ has
    syntactically odd type ‘'(choice (const :tag implementation
    'implementation) (const :tag interface 'interface) (const :tag none nil))’

In ocamlformat-before-save:
ocamlformat.el:96:2: Warning: docstring has wrong usage of unescaped single
    quotes (use \= or different quoting)
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 - Preserve comments around object open/close flag (#2097, @trefis, @gpetiot)
 - Preserve comments around private/mutable/virtual keywords (#2098, @trefis, @gpetiot)
 - Closing parentheses of local open now comply with `indicate-multiline-delimiters` (#2116, @gpetiot)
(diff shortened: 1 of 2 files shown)


gpetiot:
Looks good to me, thanks! I've added a changelog entry as well.

PR #2124, opened 2022-07-13
Tests: reset .ocamlformat using profile=default

emillon:
Dune 2 and dune 3 implement cram tests slightly differently. In dune 3, a `.git` directory is added so that executables that rely on git work properly.

Without this (in dune 2), ocamlformat walks up the directory structure and applies 2 files: the `.ocamlformat` file at the project root (the one used for ocamlformat itself) and the empty ones that are touched in each cram test. Merging this is equivalent to the exterior one.

In dune 3, only the empty one is used. This makes a difference in the `repl_file_errors.t` test.

The dune 3 behavior is the intended one. This commit fixes the "reset" strategy to use instead `profile=default` (which will properly override the external file) and makes the test suite compatible with dune 2 and dune 3.

### DIFF
--- a/test/cli/bad_kind.t
+++ b/test/cli/bad_kind.t
@@ -1,4 +1,4 @@
-  $ touch .ocamlformat
+  $ echo profile=default > .ocamlformat
 
   $ echo 'module X : S' > a.mli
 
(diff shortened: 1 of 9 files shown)


gpetiot:
Looks good, thanks!

PR #2131, opened 2022-07-20
Reset max-indent when 'max-indent' is not set

### LINKED ISSUE #2127: Bug: calling ocamlformat on one or multiple files does not yield the same result
Calling ocamlformat on one or multiple files does not yield the same result, i.e. `find ocaml -regex ".*\.mli?$"|xargs ocamlformat -i` does not indent like `for i in $(find ocaml -regex ".*\.mli?$"); do ocamlformat -i $i; done`. More info on request ;)

gpetiot:
Fix #2127

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ### Bug fixes
 
 - Fix indentation when ocamlformat is disabled on an expression (#2129, @gpetiot)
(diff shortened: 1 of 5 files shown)


Julow (commented):
Sorry for the late review. Looks good!

PR #2133, opened 2022-07-28
Bump ocaml-re dependency to 1.10.3

### LINKED ISSUE #1234: Unable to define recursive paths in .ocamlformat-ignore
I am trying to use a `.ocamlformat-ignore` file to specify directories that should be ignored by `ocamlformat`. 
I have a project in which I would like to ignore a folder (call it `lib/tests`) that contains many subfolders and subsubfolders, and I am currently struggling to specify this correctly in the `.ocamlformat-ignore`.

I expected this to have the same semantics as `.gitignore` files, i.e. ideally I should only have to specify `lib/tests`. Unfortunately, this is not working. I also tried multiple combinations of `*` and `**`, without success.
The only solution I found is to add one line per folder, which is not really possible in my scenario as there are too many, potentially changing, and that is not really handy anyway.

Did I miss an easy solution here, and if not, would it be possible to make that file be interpreted the same way `.gitignore` files are?

gpetiot:
Closes #1234


Julow (approved):
Looks good.

PR #2150, opened 2022-09-01
Simplify parsing and formatting of indexop-access expressions

gpetiot:
Extracted from ocamlformat-ng's concrete AST in our long-running effort to make the AST closer to the original source (removing sugaring/desugaring).

The more noticeable change for the users is that the indexop-access having a RHS will break a sequence, which is more readable). That's the only diff visible when checking test_branch.sh on the preset profiles.


Julow (approved):
Looks good and I like the formatting changes.

However, the patches on top of ocaml's parser are increasing in size faster than I expected. Updating to new versions of ocaml is harder after every such changes.

gpetiot:
> However, the patches on top of ocaml's parser are increasing in size faster than I expected. Updating to new versions of ocaml is harder after every such changes.

Yes indeed, by the time the migration to a concrete syntax tree will be complete I don't think `vendor/diff-parsers-std-ext.patch` will be useful anymore. But hopefully the new parser will be much easier to read (no sugaring/normalization) and parsing rules would match the CST nodes almost on 1:1 ratio.

PR #2159, opened 2022-09-19
Remove abusive normalization in docstrings references

EmileTrotignon:
This fixes the following failure : formatting `(** {! refref [ } *)` into `(** {! refref \[ } *)`

### DIFF
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -121,7 +121,7 @@ let fmt_math_block s =
   in
   hvbox 2 (wrap "{math@;" "@;<0 -2>}" (list_fl lines fmt))
 
-let fmt_reference = ign_loc ~f:str_normalized
+let fmt_reference = ign_loc ~f:str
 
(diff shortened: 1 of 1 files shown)


Julow (commented):
Looks good to me :) For context, this goes with https://github.com/ocaml-doc/odoc-parser/pull/12.

The CI is failing because the test `test/failing/tests/unstable_docstrings.mli` was expecting this reference to be escaped in the middle: `{!serve_with_handler]/{!serve}}`. I think the whole test can be removed.

gpetiot:
> The CI is failing because the test `test/failing/tests/unstable_docstrings.mli` was expecting this reference to be escaped in the middle: `{!serve_with_handler]/{!serve}}`. I think the whole test can be removed.

Should this error be handled by odoc-parser instead? If not we can move the test from `failing/` to `passing/`

Julow:
Adding such a test to `passing/` would be nice but only the interesting part: A reference that contains `[]` and `{}`. The rest doesn't seem useful.

There's no error from the pov of odoc-parser.

PR #2160, opened 2022-09-23
Replace `_ ref` by `() -> _` to access Cmdliner options in conf

EmileTrotignon:
Remove export of references by the `mk` function. Now you get a `unit -> 'a` that can be viewed as a read-only reference. A reference is still used internally.


gpetiot (approved):
Looks good to me, thanks!
I rebased and changed the title of the PR.

PR #2162, opened 2022-09-27
Remove abusive normalization in docstrings references

EmileTrotignon:
Same as : https://github.com/ocaml-ppx/ocamlformat/pull/2159 but with tests, and in its own branch.


gpetiot (approved):
LGTM, thanks!

I rebased and added an entry in the changelog.

PR #2168, opened 2022-10-02
Create ocamlformat Dockerfile

### LINKED ISSUE #1447: Feature request: Publish docker image for CI usage
Running ocamlformat in CI is expensive for the moment because it does require to use ocaml/opam2 docker image + installing ocamlformat using opam.

It would help speeding up CI jobs by publishing a docker image on ocaml/ocamlformat (for example).
I started [a project on GitLab](https://gitlab.com/dannywillems/docker-ocamlformat) doing the job, but that would be better to have official and maintained images.

jooola:
Attempt to fix #1447

Adds a Dockerfile used to build from the source, using a multistage build and alpine base image to keep the final image size as small as possible.

A build job has also been added in the CI to test the build process, and maybe publish the image.

I am unsure about a few details:
- When should such image be published (on tags ? releases ?) ? And where should the image be published (for now I have set ghcr.io/ocaml-ppx/ocamlformat) ?
- Which version of ocaml and ocamlformat do you wish to build ? Maybe you have multiple development branches ?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,3 +36,28 @@ jobs:
 
       - name: Check help files
         run: opam exec -- dune build @help --auto-promote
+
+  docker:
+    runs-on: ubuntu-latest
+
+    env:
+      REGISTRY: ghcr.io
+      NAMESPACE: ${{ github.repository_owner }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: docker/setup-buildx-action@v2
+
+      - uses: docker/login-action@v2
+        with:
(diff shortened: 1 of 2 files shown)


gpetiot:
That's great, thank you for having a look at this!

I think it's enough to build an image for the last releases (0.23.0 and 0.24.1). We only use tags for releases, and we only consider the `main` branch. We should be able to run the workflow after each release like what is done for mingw64 binaries (https://github.com/ocaml-ppx/ocamlformat/blob/main/.github/workflows/build-mingw64.yml), so we can just attach the image as an asset on the releases published on GitHub.

PR #2172, opened 2022-10-05
Remove diff between standard and extended parsers

gpetiot:
For context, there are 4 parsers:
1. **upstream** (from ocaml)
2. **standard** (slight modification of the upstream parser to make it compatible to a few nodes that don't have an equivalent in the upstream parser, like AST holes)
3. **extended** (heavy modification if the standard parser, that is the base for ocamlformat formatting)
4. **recovery** (ex-parsewyc) (slight modification of the extended parser used for the recovery mode)

The diff between the standard and the extended parsers (2 --> 3) becomes less and less readable as we come closer to a more concrete AST. It just adds up some noise to these PRs. Once the transition to the CST is made a big part of this diff will be a huge deletion of lines followed by a huge addition of lines, not really useful.

OTOH the diff between the upstream and the standard parsers (1 --> 2) is still useful, same for the diff between the extended and the recovery parsers (3 --> 4).


Julow (approved):
The diff can be computed when we need it so it's not useful at all.

It was useful to have a sense of how big the diff is growing but now it's so big it's a lost battle.

PR #2175, opened 2022-10-06
Remove access to irrelevant parts of the config to the library

EmileTrotignon:

PR #2181, opened 2022-10-14
Add a test for option 'disable-conf-attrs'

gpetiot:
Reduce diff of #2175


Caoxokhuyen:
Check

fbf72cd98d177ff153ce20c362a5cf88cbdcb5a9

Caoxokhuyen (commented):
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

PR #2187, opened 2022-10-17
Fix alignment inconsistency between let-binding and let-open

### LINKED ISSUE #2186: Bug: Inconsistency between let and local open
Hi,
In the following situation, there's an inconsistency depending on whether the first `let` is a `let in` or a local open (`let open in`).

The following are formatted outputs:

```ocaml
let _ =              
  f
    (let open M in
    let x = 1 in
    () )
```

```ocaml
let _ =              
  f
    (let x = 1 in
     let open M in
     () )
```

This means that when refactoring and adding a local open, unrelated code gets reformatted. I'd expect the first code snippet to have an extra space like:

```ocaml
let _ =              
  f
    (let open M in
     let x = 1 in
     () )
```

Thanks!

gpetiot:
Fix #2186

What do you think @samoht (https://github.com/ocaml-ppx/ocamlformat/actions/runs/3267004322/jobs/5371567574), @jberdine (https://github.com/ocaml-ppx/ocamlformat/actions/runs/3267004322/jobs/5371567691), @ceastlund (https://github.com/ocaml-ppx/ocamlformat/actions/runs/3267004322/jobs/5371567786) ? The diff is quite large but it looks like the improvement is worth it.
(unfold the `Test [...] profile` task to see the diff)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@
 - Fix parentheses around symbols in if-then-else branches (#2169, @gpetiot)
 - Preserve position of comments around variant identifiers (#2179, @gpetiot)
 - Fix parentheses around symbol identifiers (#2185, @gpetiot)
-- Fix alignment inconsistency between let-binding and let-open (#<PR_NUMBER>, @gpetiot)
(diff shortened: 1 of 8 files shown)


ceastlund:
The results look good.

PR #2188, opened 2022-10-18
Fix reporting of operational settings origin in presence of profiles

EmileTrotignon:
The `from` metadata is now stored alongside the option's value, instead of with the option declaration.


lib/Config_option.ml
```diff
@@ -329,8 +371,7 @@ module Make (C : CONFIG) = struct
       ; allow_inline
       ; default
       ; to_string
-      ; get_value
-      ; from
+      ; get_value (* ; from *)
```
Julow, line 374: It's gone or not ? :p

lib/bin_conf/Bin_conf.ml
```diff
@@ -293,19 +293,21 @@ let update_from_ocp_indent c (oic : IndentConfig.t) =
     | Never -> `Never
     | Auto -> `Auto
   in
+  let elt v = Conf.Elt.make v `Default in
```
Julow, line 296: We should have a "from" for ocp-indent config.

Julow (approved):
Looks good :) Thanks for refactoring these uneasy parts.

Julow:
The CI is asking for changelog. Perhaps this can be added to a "library" section in the changelog ?

lib/Config_option.ml
EmileTrotignon: Good catch, I forgot to remove it !

lib/bin_conf/Bin_conf.ml
EmileTrotignon: That sounds reasonable. I have no idea how this feature is supposed to work though, could you point me to an example ?

EmileTrotignon:
This also fixes a bug, so bugfixes section would not be crazy.

lib/Config_option.ml
gpetiot: done

lib/bin_conf/Bin_conf.ml
gpetiot: I had a go at it in my last commit https://github.com/ocaml-ppx/ocamlformat/pull/2188/commits/49d99ac4ae325300866c88c57a8636911b2563f1 (there is no redundancy check as it would require more heavy lifting and it's not really a must-have for that), let me know what you think

Julow: Nice :) Do you think we can get rid of `` `Default `` ?

gpetiot: I don't think we can

gpetiot (approved):
LGTM!
@EmileTrotignon if it's ready to go on your side we can merge.

PR #2189, opened 2022-10-20
Vendor odoc-parser

gpetiot:
See https://github.com/ocaml/odoc/issues/898, cc @Julow @jonludlam

I have the feeling the CI fails because some files are cached.


Julow (approved):
Looks good :) Fixing https://github.com/ocaml/odoc/issues/898 for future versions of ocamlformat.

PR #2191, opened 2022-10-21
Simplify representation and formatting of mod-unpack

gpetiot:
Extracted from ocamlformat-ng's concrete AST in our long-running effort to make the AST closer to the original source.
Removing intermediate artificial expression and core_type nodes that complicate the comments and the pattern-matching.


Julow (approved):
Looks good.

These things should really be proposed upstream. It's general maintenance of the parser: Simplifying the code, allowing more usecases.

PR #2193, opened 2022-10-26
Make the declaration of options into a regular module instead of a functor

EmileTrotignon:


lib/Conf_t.mli
```diff
@@ -0,0 +1,138 @@
+(**************************************************************************)
```
Julow, line 1: This module doesn't need a mli. It is 90% type definition, so it is 90% repeated code. The function in this modules are either trivial (eg. `to_string`) or shouldn't be there (eg. `warn_deprecated`).

Julow (approved):
You are not afraid to make long diffs ! That's good, we need to dust off the code base sometimes.

Looks good :)

lib/Conf_t.mli
EmileTrotignon: I agree that `warn_deprecated` is out of place here, but `Conf_decl` needs it. I will try to move it anyway.

gpetiot: I think the diff would be smaller if we kept `Config_option` instead of renaming it into `Conf_decl` as it looks to me most of its content comes from `Config_option`.

EmileTrotignon: Yes, its the same module, thats why I put the renaming in a separate commit. 
I think the renaming is a good idea because it is a whole lot clearer. I had a lot of trouble understanding this part of the codebase because of this naming choice. If you dont think its a good idea to rename at all, we could remove the commit renaming it, or if the diff concerns you more, I could extract that commit to a separate pull request.

gpetiot: For me `Config_option` made sense as the `'a Config_option.t` refers to an individual option and not the the configuration itself, and it obviously can't name it `Option`, I think we lose a bit of this information by using `Conf_decl`.

But I don't mind that much actually, I'm more concerned with the size of the diff. By renaming the file in a different PR, this PR will get much clearer.

gpetiot (approved):
LGTM, thanks!

I don't think the changelog entry was necessary so I removed it. I rebased on top of the main branch and force-pushed. I will merge as soon as the CI is green.

Julow:
I think we should start writing changelogs for the library, especially a patch that makes it slightly usable.

gpetiot:
> I think we should start writing changelogs for the library, especially a patch that makes it slightly usable.

Alright I've put it back!

PR #2198, opened 2022-10-27
Remove unnecessary parentheses around partially applied infix operators with attributes

gpetiot:
More consistency with how partially applied infix operators are handled, this change was already made for if-then-else


Julow (approved):
Looks good.

PR #2200, opened 2022-11-03
Apply 'if-then-else=vertical' to ocamlformat

gpetiot:
ocamlformat will become the first project to adopt the future diff-friendly profile (#2020) before it's made available to other projects. Let's switch one option at a time to reduce the future diff.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -4,3 +4,4 @@ margin = 77
 parse-docstrings = true
 wrap-comments = true
 line-endings = lf
+if-then-else = vertical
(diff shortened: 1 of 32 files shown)


Julow (commented):
Let's switch the profile at once. It's possible to ask `git blame` to ignore some commits ([see in odoc](https://github.com/ocaml/odoc/blob/master/.git-blame-ignore-revs)), it's easier to keep track if there's fewer of them.

gpetiot:
Yes but we may have to modify 20 options to get the profile ready, I think it would be nice to benefit from these improvements as soon as we can on our own code.

gpetiot:
Alright let's wait to be able to apply the whole profile at once to reduce the noise.

Issue #2202, opened 2022-11-03
Feature request: change alignment within parentheses for the janestreet profile


ceastlund:
A number of expressions are aligned with an opening parenthesis rather than with the code inside the parenthesis. Examples:

```
let x =
  f
    (if some_condition_______________________
    then consequence________________________
    else alternative________________________)
;;
```

```
let x =
  f
    ("A string _____________________"
    ^ "Another string _____________"
    ^ "Yet another string _________")
;;
```

In the first case, we would prefer the `then` and `else` to be aligned with the `if`. In the second case, we would prefer the `^` operators to be aligned with the opening `"A string`.

Julow:
I confirm this was fixed by #2203 and #2204, the output looks like that:
```ocaml
let x =                                
  f
    (if some_condition_______________________
     then consequence________________________
     else alternative________________________)
;;

let x =
  f
    ("A string _____________________"
     ^ "Another string _____________"
     ^ "Yet another string _________")
;;
```

PR #2203, opened 2022-11-04
Fix alignment inconsistency of if-then-else in apply

gpetiot:
First half of #2202

Make the formatting consistent with let-binding and let-open, so should be changed across all profiles. This should make a big diff when people update but usually people don't complain about more consistency.


Julow (approved):
Yes ! The previous formatting always felt broken to me.

The test_branch is huge. We might want to collect feedback before doing a release.

tmattio:
This is fixing inconsistencies (so fixing a bug) and blocking for some users, let's merge now!

Julow:
This PR is not blocked in any way. I'll merge now.

PR #2204, opened 2022-11-04
JaneStreet profile: doesn't align infix ops with open paren

gpetiot:
Second half of #2202

Re-using the hidden option that was used for aligning the `|` with the beginning of the pattern matching instead of the opening paren. Difference with #2203 is that here we don't want to apply this change everywhere but only with the janestreet profile.


Julow (approved):
Looks good to me.

PR #2205, opened 2022-11-05
Fix mingw64 build script, automate Windows release asset

### LINKED ISSUE #2068: Windows binary release wrongly versioned
The binary release for Windows does not show the right version, eg for 0.21.0 I get:
```
$ ./ocamlformat.exe --version
0.21.0-4-gddca4a3-dirty
```
This makes it inconvenient to use with `.ocamlformat` files that specify a `version`.

nojb:
The first commit fixes a problem (#2068) wherein the Windows binary built in the GH action does not have the right `--version` string.

Moreover, I noticed that the Windows binary has not been uploaded for the last couple of releases. I propose to automate this part as well by hooking the GH action so that the Windows binary is automatically uploaded and recorded as an asset when a release happen.

An example of the machinery can be seen in my fork:

- the log of the GH action triggered by a release https://github.com/nojb/ocamlformat/actions/runs/3401060180/jobs/5655903936
- the release itself https://github.com/nojb/ocamlformat/releases/tag/bar, where you can see the asset `ocamlformat-bar.exe` that was built and uploaded automatically.

Fixes #2068

### DIFF
--- a/.github/workflows/build-mingw64.yml
+++ b/.github/workflows/build-mingw64.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       ref:
         required: true
-        default: 'master'
+        default: 'main'
 
 env:
   CYGWINSETUP: setup-x86_64.exe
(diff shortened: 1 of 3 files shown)


gpetiot (approved):
It looks good to me, thanks!

nojb:
> It looks good to me, thanks!

Thanks for fixing the conflicts!

PR #2206, opened 2022-11-05
Fix interpretation of glob patterns in .ocamlformat-ignore under Windows

### LINKED ISSUE #1773: Bug: .ocamlformat-ignore glob patterns do not work on Windows
```bash
$ ls -a
.  ..  .ocamlformat  .ocamlformat-ignore  a.ml
$ cat .ocamlformat
$ cat .ocamlformat-ignore
*.ml
$ cat a.ml
let _ =
1
$ ocamlformat a.ml
let _ = 1 # BUG: a.ml should be ignored!
```
Note that this is only about glob patterns, literal filenames work fine (they were fixed in #1752).

nojb:
Fixes #1773 

`ocaml-re` has a flag `~match_backslashes:true` which means that forward backslashes in the pattern will match both forward and backward backslashes in the matched filename. But it gets a bit tricky because _backward_ slashes in the pattern will still act as normal backslashes (escaping the following character), so we need to make sure there are no backward slashes in the pattern before building the regexp.

All in all, an ugly solution, but am not sure we can do much better with `ocaml-re`'s current API.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@
 - Fix automated Windows build (#2205, @nojb)
 - Fix spacing between recursive module bindings and recursive module declarations (#2217, @gpetiot)
 - ocamlformat-rpc: use binary mode for stdin/stdout (#2218, @rgrinberg)
+- Fix interpretation of glob pattern in `.ocamlformat-ignore` under Windows (#2206, @nojb)
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


gpetiot (approved):
It looks good to me, but do you have a minimal example we could add to the tests to try this behavior on windows?

Issue #2208, opened 2022-11-14
Bug: multiple kinds of formatting apparently broken at tip of main


ceastlund:
If I format `(<) x y` or `{x=x}`, I used to get back `x < y` and `{x}`, whitespace notwithstanding. Now ocamlformat leaves the original forms as-is. This is true at tip of `main` but was not true at `0.24.1`. I'm not sure when the regression got introduced.

gpetiot:
This should be fixed by #2210 so I'm closing this issue, but feel free to reopen @ceastlund if we missed some cases.

PR #2170, opened 2022-10-04
Desugar type constraints in record fields

gpetiot:
Extracted from ocamlformat-ng's concrete AST in our long-running effort to make the AST closer to the original source.
This removes the sugaring of type constraints that the parser does for record fields that over-complicates the pattern matching in Fmt_ast.ml. The formatting is also simplified.

The diff of test_branch.sh is a fix:
```diff
diff --git a/core/src/univ_map.ml b/core/src/univ_map.ml
index 82193a7..18552b8 100644
--- a/core/src/univ_map.ml
+++ b/core/src/univ_map.ml
@@ -24,10 +24,9 @@ struct
       [%sexp
         { name= (Type_equal.Id.name type_id : string)
         ; uid=
-            (( if am_running_inline_test then Sexp.Atom "<uid>"
-             else Type_equal.Id.Uid.sexp_of_t (Type_equal.Id.uid type_id) ) : Sexp
-                                                                              .t)
-        }]
+            ( if am_running_inline_test then Sexp.Atom "<uid>"
+              else Type_equal.Id.Uid.sexp_of_t (Type_equal.Id.uid type_id)
+              : Sexp.t ) }]
```

PR #2210, opened 2022-11-21
Closes Issue #2208

EmileTrotignon:
Fixes regression pointed out in #2208 

Works by adding an Ast_mapper that transforms the concrete syntax into the simpler version.

Commits that introduced the regression are the following : 33679b4933d873416ace0aa402e9954083cb1e24 for `{x = x}`and 78f1e72de0d1c1bda318589382f6dee201de2c3c for `( + ) 1 2`.


Julow (commented):
You need to run `dune build @fmt`. Also, feel free to add a change entry, even thought the bug is not released yet.

Julow (approved):
Looks good to me.
You need to rebase on top of https://github.com/ocaml-ppx/ocamlformat/pull/2212 and add a change entry.

gpetiot:
I rebased the PR.

Regarding the changelog, the bug that was introduced was not logged in the changelog as #2165 and #2170 were supposed to not be noticeable to users, and it concerns the unreleased version, I'm not sure that's worth mentioning we fixed a regression that was not even released.

gpetiot (approved):

Julow:
Indeed but it would be nice to at least mention that the PR number in the change entry for 2165 and 2170

PR #2213, opened 2022-12-06
[WIP] Replace AST with CST

gpetiot:
Please don't merge this!
The purpose is to visualize the diff between the current state of the ast+parser of ocamlformat and the cst+parser of ocamlformat-ng prototype (https://github.com/tarides/ocamlformat-ng), the files in this branch have been made closer to the current main branch of ocamlformat, but this doesn't build.

To display the diff: `git diff main cst -- vendor/parser-extended/`

This is a suggestion of roadmap (non-exhaustive) of the changes left to report in ocamlformat, it should be easier to start with the type changes before working on the ext_attributes and the parentheses. Refer to parsetree.mli:

### 1. Enrich existing type constructors
- [x] `ptyp_alias` (#2239)
- [x] `pexp_pack` (#2234)
- [x] `pval_prim` (#2238)
- [x] `pcstr_record` (#2237)
- [x] `pcsig_self` (#2216)
- [x] `pcstr_self` (#2236)
- [x] new type `fun_param` (#2466)

### 2. Improve representation of poly/newtype
- [x] redesign the whole representation of `pexp_poly`, `pexp_newtype`, `ptyp_newtype_poly` (#2529)

### 3. Add location in Longident.t type
- [ ] replace occurrences of`Longident.t. loc` with `Longident.t`

### 4. Add extension and attributes
- [x] add fields `*_ext_attributes` (#2247), as these fields are kept together in the parser.

### 5. Preserve parentheses
- [ ] `Pexp_parens`
- [ ] `Ppat_parens`
- [ ] `Ptyp_parens`
- [ ] `Pcl_parens`
- [ ] `Pmty_parens`
- [ ] `Pmod_parens`

### Bonus
- [ ] types `module_expr` and `module_type` probably need to be reworked too
- [ ] undo the work made in `Sugar.ml` by reverting the normalization made in the parser

While reporting some of the changes into ocamlformat, if design choices make the files diverge from ocamlformat-ng please update the files in this branch.
Also don't forget to propagate the changes in `parser-recovery/lib/parser.mly` and run `dune build @all` to regenerate the diffs.

PR #2216, opened 2022-12-14
Make pcsig_self optionnal

EmileTrotignon:
Please do not merge, this for visualizing a diff.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -758,8 +758,9 @@ and fmt_core_type c ?(box = true) ?pro ?(pro_space = true) ?constraint_ctx
   Cmts.fmt c ptyp_loc
   @@ (fun k -> k $ fmt_docstring c ~pro:(fmt "@ ") doc)
   @@ ( if List.is_empty atrs then Fn.id
-     else fun k ->
-       hvbox 0 (Params.parens c.conf (k $ fmt_attributes c ~pre:Cut atrs)) )
+       else fun k ->
+         hvbox 0 (Params.parens c.conf (k $ fmt_attributes c ~pre:Cut atrs))
+     )
   @@
   let parens = parenze_typ xtyp in
   hvbox_if box 0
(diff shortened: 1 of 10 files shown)


vendor/parser-extended/parsetree.mli
```diff
@@ -684,15 +684,21 @@ and class_type_desc =
   | Pcty_extension of extension  (** [%id] *)
   | Pcty_open of open_description * class_type  (** [let open M in CT] *)
 
+and class_sig_self =
+  {
+   pcss_desc:core_type option;
+   pcss_loc : Location.t;
+  }
```
gpetiot, line 691: Is there really a location if `pcss_desc` is None? I think it's enough to just use `core_type option`, the loc of the type (if any) is the loc of the class_sig_self, as in #2213

vendor/parser-extended/ast_mapper.ml
```diff
@@ -316,7 +316,7 @@ module CT = struct
 
   let map_signature sub {pcsig_self; pcsig_fields} =
     Csig.mk
-      (sub.typ sub pcsig_self)
+      ({pcsig_self with pcss_desc = map_opt (sub.typ sub) pcsig_self.pcss_desc})
```
gpetiot, line 319: Personally I like having a function for each type (eg. `map_class_sig_self`), and if we keep the loc (see my comment on parsetree, I don't think we should), the loc needs to be mapped as well with `sub.location`.

lib/Fmt_ast.ml
```diff
@@ -2662,12 +2662,7 @@ and fmt_class_signature c ~ctx ~parens ?ext self_ fields =
     | Pctf_attribute atr -> update_config c [atr]
     | _ -> c
   in
-  let cmts_after_self = Cmts.fmt_after c self_.ptyp_loc in
-  let self_ =
-    match self_ with
-    | {ptyp_desc= Ptyp_any; ptyp_attributes= []; _} -> None
-    | s -> Some s
-  in
+  let cmts_after_self = Cmts.fmt_after c self_.pcss_loc in
```
gpetiot, line 2665: The issue you might see here is that ocamlformat will attach comments to the closest location, so some comments might be previously attached to `self_.ptyp_loc` and moved to `self_.pcss_loc`.

And you don't print the comments located *before* the location `self_.pcss_loc`. It was not an issue with `self_.ptyp_loc` because when calling `fmt_core_type` we would print the remaining comments attached to this type anyway, not the case with the new loc field.

I dont't think adding a new location makes much sense here, as I've said in the parsetree comment, when the type is None the location should not exist anyway, so imho the location of the existing type is enough.

vendor/parser-extended/parsetree.mli
EmileTrotignon: The loc was used in a few places, I will try to remove these uses

gpetiot (approved):
LGTM!

PR #2217, opened 2022-12-15
Fix spacing between recursive module bindings and recursive module declarations

### LINKED ISSUE #2215: Bug: Formatting recursive modules removes disambiguating new lines
**Describe the bug**
When a recursive module can be written on a single line (so is of the form `module rec X : S`), it is displayed in a "compact" form, removing newlines between modules. However, this happens even when the module has doc comments, resulting in an ambiguous doc comment:

**How to Reproduce**
Attempting to format the following `.mli` file will produce a "BUG: generating invalid ocaml syntax." error.

```ocaml
module type S = sig end

(** A *)
module rec A : S

(** B *)
and B : S
```

If we disable that check, we can see that this file is formatted to the following:

```ocaml
module type S = sig end

(** A *)
module rec A : S
(** B *)
and B : S
```

gpetiot:
Fix #2215 

@SquidDev let us know how does that work for you!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@
 - Fix reporting of operational settings origin in presence of profiles (#2188, @EmileTrotignon)
 - Fix alignment inconsistency of if-then-else in apply (#2203, @gpetiot)
 - Fix automated Windows build (#2205, @nojb)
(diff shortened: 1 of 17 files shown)


SquidDev:
That works perfectly, thank you!

PR #2230, opened 2023-01-06
Distribute the library in a separate ocamlformat-lib package

### LINKED ISSUE #2190: Distributing the library in a different package
Distributing the library in the main package pose some problems:

- The package, once compiled, is bigger and has more dependencies.
  The binary and the documentation don't have many dependencies (there's no dynamic ocaml libraries).
  The library has huge dependencies: ocaml + all the ocaml dependencies.
  For example, on nixpkgs, `ocamlformat_0_21_0` weights 120Mb and the closure size is 660Mb. With the library stripped, both number could be around 30Mb.

- ocaml-platform-installer also wants to strip libraries from the tools, to be able to cache and relocate compiled binaries.
  We rename the package to `ocamlformat+bin+platform` to distribute the stripped package but this workaround is compromised with the new `--with-dev-setup` in Opam 2.2.
  https://github.com/tarides/ocaml-platform-installer/issues/128

- Some packages depend on `ocamlformat` but we don't know if they depend on the binary or the library. A separate package would help analyzing how it's used.

gpetiot:
Split the current ocamlformat package in 2:
- ocamlformat-lib: the library
- ocamlformat: the binary

This can be useful for the ocaml platform, and maybe some users will prefer using the lib instead of the rpc (already possible but at least that won't add a dependance on the binary).

Closes #2190

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,7 @@
 - Remove unnecessary parentheses around partially applied infix operators with attributes (#2198, @gpetiot)
 - JaneStreet profile: doesn't align infix ops with open paren (#2204, @gpetiot)
 - Re-use the type let_binding from the parser instead of value_binding, improve the spacing of let-bindings regarding of having extension or comments (#2219, @gpetiot)
+- The `ocamlformat` package now only contains the binary, the library is available through the `ocamlformat-lib` package (#2230, @gpetiot)
 
 ### New features
 
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -11,7 +11,7 @@
 
 open Bechamel
 open Toolkit
-open Ocamlformat
+open Ocamlformat_lib
 
 type range = int * int
 
--- a/bench/dune
+++ b/bench/dune
@@ -11,7 +11,7 @@
 
 (executable
  (name bench)
- (libraries bechamel bechamel-js ocamlformat stdio yojson))
+ (libraries bechamel bechamel-js ocamlformat-lib stdio yojson))
 
 (rule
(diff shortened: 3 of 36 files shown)


hhugo:
I don't understand the split, the lib is already installed an available for people to use. What do we gain by splitting ? except higher maintenance..

Julow:
The library is problematic for https://github.com/tarides/ocaml-platform-installer which wants to distribute the binary alone. It's also increasing the installed size of ocamlformat by a lot (from 168Mb to 70Mb on Opam, 660Mb to 60Mb on Nix).

I doubt the library has many users. And if it does, a separate package would be useful to signal what project depends on it.

Issue: https://github.com/ocaml-ppx/ocamlformat/issues/2190

hhugo:
> The library is problematic for https://github.com/tarides/ocaml-platform-installer which wants to distribute the binary alone. It's also increasing the installed size of ocamlformat by a lot (from 168Mb to 70Mb on Opam, 660Mb to 60Mb on Nix).

I don't really understand how splitting packages help here. Compiling the binary still requires all deps to be there somehow.

> I doubt the library has many users. And if it does, a separate package would be useful to signal what project depends on it.

ok

Julow:
In the case of Opam, the library itself is taking half the place. In Nix, the deps are not needed if the binary is downloaded from the remote cache.

Julow:
The `ocamlformat` package depends on `ocamlformat-lib` but this shouldn't be necessary for an executable. This doesn't help marking which packages depend on the library since installing the main package also installs the library.

PR #2231, opened 2023-01-07
docs: update FAQ to use recursive glob for ignoring

glennsl:
Also updates the project's own `.ocamlformat-ignore` to use recursive glob. I have not tested this locally however, since I haven't been able to find instructions for how to build the project.


gpetiot (approved):
LGTM, thanks!

PR #2232, opened 2023-01-09
Conf renaming

EmileTrotignon:
The biggest change in this PR is the renaming of `Config_option` into `Conf_decl`. The name `Config_option` was highly confusing to me.
I believe my change is a lot clearer because the `Config_option` contains functions and type to declare config options, not the options themselves, which are in `Conf`. The fact that one module has the declaration machinery and the other has the instances is very poorly expressed by one having an extra `-option` suffix. I believe the `-decl` suffix expresses this difference better.
I also dropped the `-ig` in `Config` for the sake of consistency.

The type `'a option_decl` is also renamed into `'a declarator` because it is not a declaration, but a declator, and having only `decl` does not disambiguate between the two. 

In various places in the code, an alias was made : `module C = Config_option`. This is a relic of the time when `Config_option` was a functor. I felt this name was way too obscure, but `C` was used a lot, so we still needed a short alias.
I replaced it with `Decl`.


Purello:
> Looks good to me, thanks!

Can you please help me recover my Facebook account

PR #2233, opened 2023-01-09
Remove conf mutability, and correctly display the conventional profile when using print-config

EmileTrotignon:
Remove the two refs over lists over existentials that were used to store option declarations.
Catches a bug in UI : newer operational options were missing.
Add a record to Bin_conf.
The default profile is now explicit, its declared as any other profile, and not built from default values given at each option declaration.


gpetiot (approved):
Looks good to me, thanks!

I took the liberty of cleaning up some bits, fixing the output of print-config (conventional was printed as default), and adding a changelog entry. If everything is alright we can merge soon.

EmileTrotignon:
Seems good to me, thank you for the cleanup !

PR #2234, opened 2023-01-10
Add optional package_type to pexp_pack

EmileTrotignon:
Has a few bugs that should be fixed by adding parentheses nodes.


vendor/diff-parsers-ext-parsewyc.patch
```diff
@@ -605,9 +605,9 @@
...
        { Pexp_new($3), $2 }
    | LPAREN MODULE ext_attributes module_expr RPAREN
-       { Pexp_pack $4, $3 }
-   | LPAREN MODULE ext_attributes module_expr COLON package_core_type RPAREN
-       { Pexp_constraint (ghexp ~loc:$sloc (Pexp_pack $4), $6), $3 }
+       { Pexp_pack ($4, None), $3 }
+   | LPAREN MODULE ext_attributes module_expr COLON package_type RPAREN
+       { Pexp_pack ($4, Some $6), $3 }
 -  | LPAREN MODULE ext_attributes module_expr COLON error
 -      { unclosed "(" $loc($1) ")" $loc($6) }
```
Julow, line 612: Why remove this rule ? (and the others below)

Julow (approved):
Looks like a good idea :)

vendor/diff-parsers-ext-parsewyc.patch
gpetiot: All the error cases have been removed from the recovery parser because if we exit on error we don't recover (not done in this PR, what we see is the lines around the diff).
That was done by Frederic when he implemented it at first, and I hadn't had a look at it since, I don't know if we can modify the algorithm to work while keeping the error cases.

vendor/parser-extended/parser.mly
```diff
@@ -2201,8 +2201,11 @@ expr:
 ;
 
 simple_expr:
-  | LPAREN seq_expr RPAREN
-      { reloc_exp ~loc:$sloc $2 }
+  | LPAREN e = seq_expr RPAREN
+      { match e.pexp_desc with
+        | Pexp_pack _ ->
+            mkexp ~loc:$sloc (Pexp_parens e)
+        | _ -> reloc_exp ~loc:$sloc e }
```
gpetiot, line 2208: I made this dirty fix (introducing a new node Pexp_parens only in this case) to preserve the location of comments here. As this is the direction we are taking anyway.
@Julow  @EmileTrotignon what do you think?

If you are okay with introducing this new node in the ast now, we can do it in another PR to not pollute this one.

EmileTrotignon: I am fine with this

PR #2236, opened 2023-01-18
Make pcstr_self optional

EmileTrotignon:
Makes pcstr_self an option. Similar to #2216.


Julow (approved):
The CI is failing with probably unrelated diffs but I fail to reproduce them locally. 
Otherwise the code looks good :)

PR #2238, opened 2023-01-18
Add locations to pval_prim strings

EmileTrotignon:
Add locations to pval_prim.


Julow (commented):
It's hard to understand the diffs in the `.patch` files. Have you diffed on top of a new version of the compiler's parser or is this a change that should have been committed before ?

EmileTrotignon:
> It's hard to understand the diffs in the `.patch` files. Have you diffed on top of a new version of the compiler's parser or is this a change that should have been committed before ?

I honestly don't really know what these files are and how they are generated. My switch is on ocaml 4.14.0 though. And the diff on the parser itself and not the parser patch is really small.

PR #2239, opened 2023-01-18
Add location to ptyp_alias

EmileTrotignon:
Add a location to the string in ptyp_alias.
Ast_mapper doesn't call its location on it : I do not really know if it should.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -780,7 +780,8 @@ and fmt_core_type c ?(box = true) ?pro ?(pro_space = true) ?constraint_ctx
       hvbox 0
         (parenze_constraint_ctx
            ( fmt_core_type c (sub_typ ~ctx typ)
-           $ fmt "@ as@ " $ fmt_type_var str.txt ) )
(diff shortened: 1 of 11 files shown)


lib/Fmt_ast.ml
```diff
@@ -776,11 +776,11 @@ and fmt_core_type c ?(box = true) ?pro ?(pro_space = true) ?constraint_ctx
     | _ -> Fn.id
   in
   match ptyp_desc with
-  | Ptyp_alias (typ, txt) ->
+  | Ptyp_alias (typ, str) ->
       hvbox 0
         (parenze_constraint_ctx
            ( fmt_core_type c (sub_typ ~ctx typ)
-           $ fmt "@ as@ " $ fmt_type_var txt ) )
+           $ fmt "@ as@ " $ fmt_type_var str.txt ) )
```
gpetiot, line 783: We need to check the comments at the location `str.loc` here

EmileTrotignon: I have pushed a commit that I believe fixes and tests this issue.

PR #2241, opened 2023-01-23
Fix lint-lower-bounds

gpetiot:
(this is ocaml-ci's last linting check)

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -87,7 +87,8 @@
   (ocamlformat
    (= :version))
   stdio
-  yojson))
+  (yojson
+   (>= 1.6.0))))
 
 (package
  (name ocamlformat-rpc-lib)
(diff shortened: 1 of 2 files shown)


test/unit/test_unit.ml
```diff
@@ -122,4 +122,4 @@ let tests =
   ; ("Range", Test_range.tests)
   ; ("Translation_unit", Test_translation_unit.tests) ]
 
-let () = Alcotest.run "ocamlformat" tests
+let () = Alcotest.run ~argv:[|""; "--compact"|] "ocamlformat" tests
```
Julow, line 125: I don't know the consequences of overriding the argv here but there's a `~compact:true` argument that seems better.

gpetiot: Oh I missed this arg, good catch!

PR #2242, opened 2023-01-24
Keep the location of 'in' keywords in local bindings

### LINKED ISSUE #2096: Bug: Crash on `[@ocamlformat disabled]` on local function
**Describe the bug**
`ocamlformat` crashes on a local function with `[@ocamlformat disabled]` specified.

**How to Reproduce**
In `ocamlformat.ml`:
```ocaml
let f () =
  let [@ocamlformat "disable"] g () = ()
  in
  ()
```

```
$ ocamlformat ocamlformat.ml --enable-outside-detected-project -o ocamlformat2.ml 
ocamlformat: Cannot process "ocamlformat.ml".
```

The produced `invalid-ast.ml` file appears to be missing the `in`:

```ocaml
let f () =
  let [@ocamlformat "disable"] g () = ()
  ()
```

Can be reproduced with default options, but was originally found with:

```
profile=default
quiet=false
max-iters=10
comment-check=true
wrap-fun-args=true
wrap-comments=true (file middle_end/flambda2/.ocamlformat:13)
type-decl-indent=2
type-decl=sparse (file middle_end/flambda2/.ocamlformat:12)
stritem-extension-indent=0
space-around-variants=false (file middle_end/flambda2/.ocamlformat:11)
space-around-records=true
space-around-lists=false (file middle_end/flambda2/.ocamlformat:10)
space-around-arrays=true
single-case=compact
sequence-style=terminator
parse-docstrings=false
parens-tuple=multi-line-only (file middle_end/flambda2/.ocamlformat:9)
parens-ite=false
ocp-indent-compat=false
nested-match=wrap
module-item-spacing=sparse
max-indent=68
match-indent-nested=never
match-indent=0
margin=80
line-endings=lf
let-module=compact
let-binding-spacing=compact
let-binding-indent=2
let-and=compact
infix-precedence=indent
indent-after-in=0
if-then-else=keyword-first (file middle_end/flambda2/.ocamlformat:8)
function-indent-nested=never
function-indent=2
field-space=loose
extension-indent=2
exp-grouping=preserve (file middle_end/flambda2/.ocamlformat:7)
dock-collection-brackets=false (file middle_end/flambda2/.ocamlformat:6)
doc-comments-tag-only=default
doc-comments-padding=2
doc-comments=before (file middle_end/flambda2/.ocamlformat:5)
disable=false
cases-exp-indent=2 (file middle_end/flambda2/.ocamlformat:4)
break-struct=force
break-string-literals=auto
break-sequences=true
break-separators=after
break-infix-before-func=false
break-infix=wrap
break-fun-sig=wrap
break-fun-decl=wrap
break-cases=fit
assignment-operator=begin-line (file middle_end/flambda2/.ocamlformat:3)
align-variants-decl=false
align-constructors-decl=false
align-cases=false
```

EmileTrotignon:
Fixes #2096 . 
In did this by expanding the let_bindings type to record the loc of the `in` keyword. Then update_config_maybe_disable uses this loc ('s end) to extract the verbatim code in case of disabling.


test/passing/tests/sequence-preserve.ml.ref
```diff
@@ -75,9 +75,8 @@ let foo x y =
 
 let foo x y =
   do_some_setup x ;
-  let () = do_some_setup y in
+  let () = do_some_setup y in (* Empty line after let *)
```
Julow, line 78: This formats like the input, which is nice. However, it's also the output of:
```ocaml
  let () = do_some_setup y in
  (* Empty line after let *)
```

I think it's better to break the line before the comment to make the `in` clearer. (`Cmt.fmt_after` has an optional argument or this)
Also, this case needs some care:

```ocaml
  important_function x ;
  let () = do_some_setup y in
  (* Empty line after let *)
```

Julow (approved):
This is nice! I just have a comment on the placement of comments.

test/passing/tests/sequence-preserve.ml.ref
EmileTrotignon: I agree your proposal is better, however having looked around a bit, I think `fmt_after` is not directly used here, and because of this I am not sure how to proceed.

CHANGES.md
```diff
@@ -58,6 +58,7 @@ profile. This started with version 0.26.0.
 - Fix indentation of trailing double-semicolons (#2295, @gpetiot)
 - Fix formatting of comments in "disable" chunks (#2279, @gpetiot)
 - Fix non-stabilizing comments attached to private/virtual/mutable keywords (#2272, #2307, @gpetiot, @Julow)
+- Fix the formatting of the `in` keyword when `[@ocamlformat disable]` is attached to a let-binding (#2242, @EmileTrotignon)
```
Julow, line 61: That's not in the right section of the changelog anymore.

hhugo:
don't we have the same issue with `let module`, `let exception`, `let open`, .. ?

test/passing/tests/disable_local_let.ml
```diff
@@ -0,0 +1,4 @@
+let f () =
+  let [@ocamlformat "disable"] x = y
+  in
+  ()
```
hhugo, line 4: We should also test the other syntax
```ocaml
let f () =
  let x = y [@@ocamlformat "disable"]
  in
  ()
```

lib/Fmt_ast.ml
```diff
@@ -4344,9 +4352,17 @@ and fmt_let c ~ext ~rec_flag ~bindings ~parens ~fmt_atrs ~fmt_expr ~body_loc
...
-  update_config_maybe_disabled c lb_loc lb_attrs
+  let loc_with_in =
+    match loc_in with
+    | None -> lb_loc
+    | Some loc_in ->
+        { loc_start= lb_loc.loc_start
+        ; loc_end= loc_in.loc_end
+        ; loc_ghost= lb_loc.loc_ghost || loc_in.loc_ghost }
+  in
+  update_config_maybe_disabled c loc_with_in lb_attrs
```
hhugo, line 4365: Another approach would be to still print `in_` and `epi` when disabled

test/passing/tests/disable_local_let.ml
gpetiot: done

lib/Fmt_ast.ml
```diff
@@ -4342,9 +4350,17 @@ and fmt_let c ~ext ~rec_flag ~bindings ~parens ~fmt_atrs ~fmt_expr ~body_loc
        $ hvbox 0 fmt_expr ) )
   $ fmt_atrs
 
-and fmt_value_binding c ~rec_flag ?ext ?in_ ?epi
+and fmt_value_binding c ~rec_flag ?ext ?in_ ?loc_in ?epi
     {lb_op; lb_pat; lb_args; lb_typ; lb_exp; lb_attrs; lb_loc; lb_pun} =
-  update_config_maybe_disabled c lb_loc lb_attrs
+  let loc_with_in =
```
Julow, line 4355: It would be nice to have a comment saying what `loc_with_in` means and in which cases `loc_in` is expected to be `None`.

lib/Fmt_ast.ml
Julow: With the current approach, could the `in_` be dropped if `loc_in` is `None` ? I would be more comfortable with a new function that takes a non-optional `loc_in`.

Julow: This is addressed in the recent commit :tada:

lib/Fmt_ast.ml
```diff
@@ -4393,17 +4397,30 @@ and fmt_let c ~ext ~rec_flag ~bindings ~parens ~fmt_atrs ~fmt_expr ~body_loc
...
+      if last then
+        Some
+          { txt= (fun indent -> fmt_in indent)
+          ; loc=
+              (* I have checked the parser, this should never happen. *)
+              Option.value_exn
+                ~message:
+                  "Bug in the parser : let bindings with an should always \
+                   have a matching loc_in"
+                loc_in }
```
Julow, line 4410: This is not satisfying as `fmt_let` is not specific to an ast node and could be reused for other purposes.
Also, I doubt it as `pvbs_loc_in` is an option.

Julow (commented):
The last refactor is nice but I'm not sure I trust the "I checked" comment. How hard would it be to make `loc_in` not optional ?

lib/Fmt_ast.ml
```diff
@@ -4393,18 +4393,18 @@ and fmt_let c ~ext ~rec_flag ~bindings ~parens ~fmt_atrs ~fmt_expr ~body_loc
   in
   let fmt_binding ~first ~last binding =
     let ext = if first then ext else None in
-    let in_ indent = fmt_if_k last (fmt_in indent) in
+    let in_ =
+      if last then Some {txt= (fun indent -> fmt_in indent); loc= loc_in}
```
Julow, line 4397: Not everyone has seen a `(int -> Fmt.t) loc option` in its life.

To be clear, this is fine.

vendor/parser-extended/parser.mly
```diff
@@ -1828,8 +1828,8 @@ class_expr:
       { $1 }
   | FUN attributes class_fun_def
       { wrap_class_attrs ~loc:$sloc $3 $2 }
-  | let_bindings(no_ext) IN class_expr
-      { class_of_let_bindings ~loc:$sloc $1 $3 }
+  | let_bindings(no_ext) mkrhs(IN) class_expr
+      { class_of_let_bindings ~loc:$sloc ~loc_in:$2.loc $1 $3 }
```
Julow, line 1832: I find it weird to use `mkrhs` but then discard the result. Otherwise:
```suggestion
  | let_bindings(no_ext) IN class_expr
      { class_of_let_bindings ~loc:$sloc ~loc_in:(make_loc $loc($2)) $1 $3 }
```

What do you think ?

vendor/parser-extended/parser.mly
```diff
@@ -2230,14 +2230,17 @@ expr:
         mkexp_attrs ~loc:$sloc desc attrs }
   | mkexp(expr_)
       { $1 }
-  | let_bindings(ext) IN seq_expr
-      { expr_of_let_bindings ~loc:$sloc $1 $3 }
-  | pbop_op = mkrhs(LETOP) bindings = letop_bindings IN body = seq_expr
+  | let_bindings(ext) mkrhs(IN) seq_expr
+      { expr_of_let_bindings ~loc:$sloc ~loc_in:$2.loc $1 $3 }
+  | pbop_op = mkrhs(LETOP) bindings = letop_bindings _in_kw=IN body = seq_expr
```
Julow, line 2235: Why is it prefixed with `_` ?

vendor/parser-extended/parsetree.mli
```diff
@@ -1113,7 +1116,7 @@ and value_bindings =
   {
     pvbs_bindings: value_binding list;
     pvbs_rec: rec_flag;
-    pvbs_extension: string loc option
+    pvbs_extension: string loc option;
```
Julow, line 1119: Unnecessary diff. We are careful with diffs in this file as we regularly have to backport changes to it.

Julow (approved):
This is a nice improvement to the code and a bug fix. I have a few comments but other it's ready to merge.

vendor/parser-extended/parser.mly
EmileTrotignon: Because it is unused : we only access the loc, and the way menhir does its cuisine, asking for the loc does make the variable used.

vendor/parser-extended/parser.mly
EmileTrotignon: Yes thats fine

vendor/parser-extended/parser.mly
Julow: Good to know. Seems like a bug in menhir.

Julow:
Perfect, let's merge!

PR #2245, opened 2023-01-27
CI: test each profile without rebuilding ocamlformat

gpetiot:
Currently every "Test branch" action (3) rebuilds ocamlformat, let's save the planet


.github/workflows/build.yml
```diff
@@ -29,10 +31,25 @@ jobs:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
 
       - name: Opam dependencies
-        run: opam install --deps-only -t .
+        run: opam install --deps-only -t ./ocamlformat.opam
```
Julow, line 34: What was wrong with this? The new command misbehave but I can't explain why.

gpetiot: I tried this because it was this way in the test-branch action, just trying a few things to make it work.

Julow: The previous action managed to build ocamlformat, so I think this shouldn't be changed.

PR #2246, opened 2023-01-29
Remove dependency from ocamlformat to ocamlformat-lib

gpetiot:
Follow up on #2230

### DIFF
--- a/vendored-lib/dune
+++ b/vendored-lib/dune
@@ -1,0 +1,22 @@
+(library
+ (name vendored_lib)
+ (public_name ocamlformat.vendored_lib)
+ (libraries ocaml-version ocp-indent.lib uuseg.string dune-build-info))
+
+(copy_files# ../vendor/odoc-parser)
+
+(copy_files# ../vendor/ocaml-common)
+
+(copy_files# ../vendor/ocamlformat-stdlib)
+
+(copy_files# ../vendor/parser-shims)
+
+(copy_files# ../vendor/parser-upstream)
+
+(copy_files# ../vendor/parser-standard)
+
+(copy_files# ../vendor/parser-extended)
+
+(copy_files# ../vendor/parser-recovery)
+
+(copy_files# ../lib/*.ml*)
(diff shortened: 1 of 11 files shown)


Julow (commented):
This doesn't work, `dune build -p ocamlformat`:
```
File "bin/ocamlformat/dune", line 21, characters 12-27:
21 |  (libraries ocamlformat-lib bin_conf))
                 ^^^^^^^^^^^^^^^
Error: Library "ocamlformat-lib" not found.
```

This seems like it should be possible but I didn't manage to fix it.

gpetiot:
> 

This works for me on this branch.

gpetiot:
I tried to improve the situation with `copy_files` in the dune files without any success, for reference the last error is:
```
File "lib/bin_conf/dune", line 14, characters 41-75:
14 |  (libraries ocamlformat-lib.ocaml_common ocamlformat-lib.ocamlformat_stdlib re))
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Library "ocamlformat-lib.ocamlformat_stdlib" not found.
-> required by library "ocamlformat.bin_conf" in _build/default/lib/bin_conf
etc.
```
The 2 libs mentioned are private, but using `copy_files` to get their sources instead doesn't fix the issue.

The combination of private and public libraries makes it tricky. I'm out of ideas for now.

vendored-lib/dune
```diff
@@ -0,0 +1,22 @@
+(library
+ (name vendored_lib)
+ (public_name ocamlformat.vendored_lib)
```
kit-ty-kate, line 3: I'm not sure to understand the issue, why exporting the vendored library?
```suggestion
```

gpetiot:
I've played around with different solutions without success.  At this point I'm wondering if this wouldn't be better for someone else to just take over this PR (free hand :D), possibly sanitize the whole situation with vendored libs, we probably don't follow `dune`'s best practices.

We don't have a lot of constraints regarding which libs should be private or public as long as we abide by the existing opam files.

For context that's the last issue I wanted to fix before wrapping up the release.

gpetiot:
After discussing with @emillon we cannot do better than what is already done on `main`. Closing this one.

Issue #2109, opened 2022-06-15
Feature request: put inline attributes at top of functor declaration


ceastlund:
Currently, ocamlformat moves `@inline` attributes to the end of a functor declaration:

```ocaml
module Make (M : S) = struct end [@@inline]
```

When the body of the functor is very long, the attribute is no longer apparent when reading the top of the declaration. We would like the attribute to be left at the beginning:

```ocaml
module [@inline] Make (M : S) = struct end
```

PR #2247, opened 2023-02-01
Moves module attributes after the module keyword

EmileTrotignon:
This implements issue #2109, and will be useful for reworking parentheses.
The implementation is a bit naive, as far as I know it is fine, but it almost looks too good to be true.
Maybe the `~pre` argument needs to be changed, I would like someone to get a look into it, its beyond my understanding as of now.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -331,18 +331,19 @@ module Structure_item = struct
      |Pstr_type (_, {ptype_attributes= atrs; _} :: _)
      |Pstr_typext {ptyext_attributes= atrs; _}
      |Pstr_recmodule ({pmb_expr= {pmod_attributes= atrs; _}; _} :: _)
-     |Pstr_extension (_, atrs)
-     |Pstr_class_type ({pci_attributes= atrs; _} :: _)
-     |Pstr_class ({pci_attributes= atrs; _} :: _) ->
+     |Pstr_extension (_, atrs) ->
         List.exists ~f:Attr.is_doc atrs
     | Pstr_exception
         { ptyexn_attributes= atrs1
         ; ptyexn_constructor= {pext_attributes= atrs2; _}
         ; _ } ->
         List.exists ~f:Attr.is_doc atrs1 || List.exists ~f:Attr.is_doc atrs2
-        | Pstr_open
-        {popen_attributes= ea; _}
-    | Pstr_modtype {pmtd_ext_attrs=ea; _} -> Ext_attrs.has_doc ea
+    | Pstr_open {popen_attributes= ea; _}
+     |Pstr_class_type ({pci_attributes= ea; _} :: _)
+     |Pstr_class ({pci_attributes= ea; _} :: _)
+     |Pstr_modtype {pmtd_ext_attrs= ea; _} ->
+      print_endline "hey";
+        Ext_attrs.has_doc ea
     | Pstr_module
         {pmb_ext_attrs= ea; pmb_expr= {pmod_attributes= attrs; _}; _}
      |Pstr_include
@@ -426,14 +427,15 @@ module Signature_item = struct
      |Psig_type (_, {ptype_attributes= atrs; _} :: _)
      |Psig_typesubst ({ptype_attributes= atrs; _} :: _)
      |Psig_typext {ptyext_attributes= atrs; _}
-     |Psig_extension (_, atrs)
-     |Psig_class_type ({pci_attributes= atrs; _} :: _)
-     |Psig_class ({pci_attributes= atrs; _} :: _) ->
-        List.exists ~f:Attr.is_doc atrs
-    (* two attribute list *)
-    | Psig_modtype {pmtd_ext_attrs= ea; _}
+     |Psig_extension (_, atrs) ->
+        List.exists ~f:Attr.is_doc atrs (* two attribute list *)
+    | Psig_class_type ({pci_attributes= ea; _} :: _)
+     |Psig_class ({pci_attributes= ea; _} :: _)
+     |Psig_modtype {pmtd_ext_attrs= ea; _}
      |Psig_modtypesubst {pmtd_ext_attrs= ea; _}
      |Psig_modsubst {pms_ext_attrs= ea; _} ->
+      print_endline "hey";
+
         Ext_attrs.has_doc ea
     | Psig_exception
         { ptyexn_attributes= atrs1
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -326,7 +326,6 @@ module Structure_item = struct
     | Pstr_attribute atr -> Attr.is_doc atr
     (* one attribute list *)
     | Pstr_eval (_, atrs)
-     |Pstr_value {pvbs_bindings= {pvb_attributes= atrs; _} :: _; _}
      |Pstr_primitive {pval_attributes= atrs; _}
      |Pstr_typext {ptyext_attributes= atrs; _}
      |Pstr_recmodule ({pmb_expr= {pmod_attributes= atrs; _}; _} :: _)
@@ -341,7 +340,8 @@ module Structure_item = struct
      |Pstr_class_type ({pci_attributes= ea; _} :: _)
      |Pstr_class ({pci_attributes= ea; _} :: _)
      |Pstr_modtype {pmtd_ext_attrs= ea; _}
-     |Pstr_type (_, {ptype_attributes= ea; _} :: _) ->
+     |Pstr_type (_, {ptype_attributes= ea; _} :: _)
+     |Pstr_value {pvbs_bindings= {pvb_attributes= ea; _} :: _; _} ->
         Ext_attrs.has_doc ea
     | Pstr_module
         {pmb_ext_attrs= ea; pmb_expr= {pmod_attributes= attrs; _}; _}
@@ -507,7 +507,9 @@ module Signature_item = struct
 end
 
 module Lb = struct
-  let has_doc itm = List.exists ~f:Attr.is_doc itm.pvb_attributes
+  let has_doc itm =
+    List.exists ~f:Attr.is_doc itm.pvb_attributes.attrs_before
+    || List.exists ~f:Attr.is_doc itm.pvb_attributes.attrs_after
 
   let is_simple (i, (c : Conf.t)) =
     Poly.(c.fmt_opts.module_item_spacing.v = `Compact)
@@ -706,7 +708,7 @@ let attributes = function
   | Exp x -> x.pexp_attributes
   | Fpe _ | Fpc _ -> []
   | Vc _ -> []
-  | Lb x -> x.pvb_attributes
+  | Lb x -> attrs_of_ext_attrs x.pvb_attributes
   | Bo _ -> []
   | Mb x -> attrs_of_ext_attrs x.pmb_ext_attrs
   | Md x -> attrs_of_ext_attrs x.pmd_ext_attrs
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -432,14 +432,13 @@ module Signature_item = struct
      |Psig_modtypesubst {pmtd_ext_attrs= ea; _}
      |Psig_modsubst {pms_ext_attrs= ea; _} ->
         Ext_attrs.has_doc ea
-
     | Psig_include
         {pincl_mod= {pmty_attributes= atrs; _}; pincl_attributes= ea; _}
      |Psig_exception
         { ptyexn_attributes= ea
         ; ptyexn_constructor= {pext_attributes= atrs; _}
         ; _ }
-    | Psig_recmodule
+     |Psig_recmodule
         ({pmd_type= {pmty_attributes= atrs; _}; pmd_ext_attrs= ea; _} :: _)
      |Psig_module {pmd_ext_attrs= ea; pmd_type= {pmty_attributes= atrs; _}; _}
       ->
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -328,7 +328,6 @@ module Structure_item = struct
     | Pstr_eval (_, atrs)
      |Pstr_value {pvbs_bindings= {pvb_attributes= atrs; _} :: _; _}
      |Pstr_primitive {pval_attributes= atrs; _}
-     |Pstr_type (_, {ptype_attributes= atrs; _} :: _)
      |Pstr_typext {ptyext_attributes= atrs; _}
      |Pstr_recmodule ({pmb_expr= {pmod_attributes= atrs; _}; _} :: _)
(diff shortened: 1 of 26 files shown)


test/passing/tests/protected_object_types.ml
```diff
@@ -25,17 +25,15 @@ module Space_around = struct
   (* Ensure that the protection mechanism does not add extra spaces when
      [--space-around-*] options are sufficient. *)
 
-  module Records = struct
+  module [@ocamlformat "space-around-records = true"] Records = struct
```
Julow, line 28: The new formatting doesn't look great for long attributes or attributes not interpreted by the compiler like this.

Julow (commented):
This is a big change and I don't think the new formatting is better for every attributes or every users.

I think we should "preserve" the location of the attribute instead. This could be done by comparing locations or by storing the number of `@` of each attributes in the CST.

test/passing/tests/protected_object_types.ml
EmileTrotignon: I agree for attributes interpreted by the compiler, it makes sense. I am going to do a preserving version :)

EmileTrotignon:
I added two commit to have the attribute stay were the user put them. Each one handles one case, and there are a few, so a bunch of commit will de needed.

vendor/parser-extended/parsetree.mli
```diff
@@ -914,7 +914,8 @@ and module_declaration =
     {
      pmd_name: string option loc;
      pmd_type: module_type;
-     pmd_attributes: attributes;  (** [... [\@\@id1] [\@\@id2]] *)
+     pmd_attributes_start: attributes;  (** [... [\@\@id1] [\@\@id2]] *)
+     pmd_attributes_end: attributes;  (** [... [\@\@id1] [\@\@id2]] *)
```
gpetiot, line 918: I would like to start using the new field `pmd_ext_attributes` introduced in #2213 and that will be applied to most types (see all the `*_ext_attributes` in the CST) in the long run, even for the cases where there is no extension allowed.

Julow (commented):
This seems like a very intrusive way. It'll make huge diffs on the parsetree and parser (added maintenance).

What do you think of instead modifying the `attribute` type to remember which "attachment" it has ? (the number of `@`)
"start" and "end" attributes are different in the number of `@` they use:

```
module [@foo] M = struct end [@@bar]
```

The printing code would first render only attributes that have one `@` and at a different position render only attributes that have two `@@`.

EmileTrotignon:
The diffs are big thats true, but they were scheduled as part of the tranformation to the CST anyway, and they are also used to have explicit representation of the short form extensions (which make the biggest diff to the parser anyway, because the return type of some rules changes). 
I think it is the way to go forward.

gpetiot:
Will this PR subsume #2043? (meaning extension points will be treated the same as "pre-" attributes)

EmileTrotignon:
It would do the same thing as #2043 , but only on nodes that can take both pre and post attributes.
Maybe we could remove the extension part of this PR, I am not sure it has its place in our attributes record, now that you mention it.

gpetiot:
It has its place in this PR if we keep extensions and attributes together in the `*_ext_attributes`, which makes sense to me for now (I'm open to a different solution if the code is cleaner).

I just wanted to check before closing the other PR.

EmileTrotignon:
I was not sure that shortcut extension always admitted attributes in both position, which is why I was not sure about the choice, but after having checked it is the case so I think you can indeed close #2043

Julow (commented):
This looks mostly good but there's a regression in test_branch.

The changes affect many AST nodes and are huge, which make the review difficult. Could you extract the parts affecting only modules in an other PR ? 

There don't seem to be tests about value bindings and I think it's an important node to test.

lib/Ast.ml
```diff
@@ -508,7 +502,9 @@ module Signature_item = struct
 end
 
 module Lb = struct
-  let has_doc itm = List.exists ~f:Attr.is_doc itm.pvb_attributes
+  let has_doc itm =
+    List.exists ~f:Attr.is_doc itm.pvb_attributes.attrs_before
+    || List.exists ~f:Attr.is_doc itm.pvb_attributes.attrs_after
```
Julow, line 507: That's `Ext_attrs.has_doc`.

lib/Ast.ml
```diff
@@ -550,7 +546,9 @@ module Md = struct
 end
 
 module Td = struct
-  let has_doc itm = List.exists ~f:Attr.is_doc itm.ptype_attributes
+  let has_doc itm =
+    List.exists ~f:Attr.is_doc itm.ptype_attributes.attrs_before
+    || List.exists ~f:Attr.is_doc itm.ptype_attributes.attrs_after
```
Julow, line 551: Also `Ext_attrs.has_doc`.

lib/Exposed.ml
```diff
@@ -59,7 +59,9 @@ module Right = struct
     | {pcd_args= args; _} -> constructor_arguments args
 
   let type_declaration = function
-    | {ptype_attributes= _ :: _; _} -> false
+    | { ptype_attributes= {attrs_before= _ :: _; _} | {attrs_after= _ :: _; _}
```
Julow, line 62: This is written several time and is not clear. I think we need a `Ast.Ext_attrs.is_empty`.

lib/Fmt_ast.ml
```diff
@@ -180,7 +183,8 @@ let update_config_maybe_disabled c loc l f =
 
 let update_config_maybe_disabled_attrs c loc attrs f =
   let l = attrs.attrs_before @ attrs.attrs_after in
-  update_config_maybe_disabled c loc l f
+  let c = update_config c l in
+  maybe_disabled c loc l f
```
Julow, line 187: The previous was equivalent but simpler.

lib/Fmt_ast.ml
```diff
@@ -400,20 +404,23 @@ let docstring_epi ~standalone ~next ~epi ~floating =
       str "\n" $ epi
   | _ -> epi
 
-let fmt_docstring c ?(standalone = false) ?pro ?epi doc =
-  list_pn (Option.value ~default:[] doc)
-    (fun ~prev:_ ({txt; loc}, floating) ~next ->
+let fmt_docstring' c ?(standalone = false) ?pro ?epi doc =
```
Julow, line 407: These changes to `fmt_docstring` and the addition of `extract_doc_attrs` were reverted in #2451.

Julow (commented):
Docstring attachment is broken in some cases, for example in `test/passing/tests/doc_comments.ml`:

```diff
@@ -20,11 +19,8 @@ include Mod
 let x = 2
 (** after *)
 
-(**floatting1*)
-(**floatting2*)
-
 (**before*)
-and y = 2
+and (**floatting1*)  (**floatting2*) y = 2
 (** after *)
 
 (** A *)
```

The reproduction:

```ocaml
(** before *)
let x = 2
(** after *)

(**floatting1*)
(**floatting2*)

(**before*)
and y = 2
(** after *)
```

EmileTrotignon:
I believe this is ready for merging. Could you give it a look @Julow

lib/Fmt_ast.ml
```diff
@@ -2417,6 +2432,8 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
       let override = is_override flag in
       let outer_parens = has_attr && parens in
       let inner_parens = has_attr || parens in
+      (* [let open] extensions are external for now. *)
+      (* let ext = attributes.attrs_extension in *)
```
Julow, line 2436: I don't understand this comment. The two syntaxes for let open extensions seem to work as before.

vendor/parser-extended/parser.mly
```diff
@@ -435,54 +444,48 @@ type let_binding =
...
     lb_loc: Location.t; }
 
 type let_bindings' =
   { lbs_bindings: let_binding list;
     lbs_rec: rec_flag;
-    lbs_extension: string Asttypes.loc option }
+    lbs_has_ext: bool }
 
-let mklb first ~loc (p, args, typ, e, is_pun) attrs =
+let mklb ?(text=[]) ~docs ~loc (p, args, typ, e, is_pun) attrs =
```
Julow, line 455: Note for later: With some work, we could store docstrings into the new `ext_attrs` and avoid encoding them in attributes.

lib/Exposed.ml
```diff
@@ -59,7 +59,8 @@ module Right = struct
     | {pcd_args= args; _} -> constructor_arguments args
 
   let type_declaration = function
-    | {ptype_attributes= _ :: _; _} -> false
+    | {ptype_attributes; _} when Ast.Ext_attrs.has_attrs ptype_attributes ->
```
Julow, line 62: `Ext_attrs.has_attrs` is not right here as this checks whether an attribute will be present on the right of the item. We must only check `attrs_after`, otherwise the item will feel protected when it isn't.

For example, this generates invalid syntax:
```
[%foo
type[@foo] t = < .. >
]
```
(please add it to the testsuite)

vendor/parser-extended/parser.mly
```diff
@@ -1493,32 +1494,30 @@ module_type_declaration:
...
-  {
-    let attrs = attrs1 @ attrs2 in
-    let loc = make_loc $sloc in
-    let docs = symbol_docs $sloc in
-    Opn.mk me ~override ~attrs ~loc ~docs, ext
-  }
+    OPEN
+    override = override_flag
+    ext = ext
+    before = attributes
```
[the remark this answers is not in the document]
EmileTrotignon: I do not think you proposal is good idea, I know a lot of things are made like that in the ast, but i think labels shoulds descriptive in their context, and `before`/`after` is. I am willing to switch back to attrs1 and attrs2 and have `~before:attrs1`. This will also reduce the diff.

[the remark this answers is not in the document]
EmileTrotignon: I also just noticed that this PR uses the proposed convention a lot, so I am switching to it everywhere

lib/Exposed.ml
EmileTrotignon: This does not surprise as I mostly rewrote this file in autopilote. Thanks for noticing this.

vendor/parser-extended/parsetree.mli
```diff
@@ -1108,15 +1108,15 @@ and value_binding =
...
     pvb_is_pun: bool;
-    pvb_attributes: attributes;
+    pvb_attributes: ext_attrs;
     pvb_loc: Location.t;
   }(** [let pat : type_constraint = exp] *)
 
 and value_bindings =
   {
     pvbs_bindings: value_binding list;
+    pvbs_has_ext: bool;
```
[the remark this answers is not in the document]
EmileTrotignon: Nice catch !

lib/Fmt_ast.ml
Julow: I see.

lib/Exposed.ml
```diff
@@ -73,7 +74,9 @@ module Right = struct
         list ~elt:constructor_declaration cdecls
 
   let type_extension = function
-    | {ptyext_attributes= _ :: _; _} -> false
+    | {ptyext_attributes; _} when Ast.Ext_attrs.has_attrs ptyext_attributes
```
Julow, line 77: Same bug as above, this should check only for `attrs_after`. Two other cases below.

I wonder if that causes the small regressions found by test-branch.

lib/Ast.ml
```diff
@@ -699,19 +698,19 @@ let attrs_of_ext_attrs ea = ea.attrs_before @ ea.attrs_after
 let attributes = function
   | Pld _ -> []
   | Typ x -> x.ptyp_attributes
-  | Td x -> x.ptype_attributes
+  | Td x -> x.ptype_attributes.attrs_before @ x.ptype_attributes.attrs_after
```
Julow, line 701:
```suggestion
  | Td x -> attrs_of_ext_attrs x.ptype_attributes
```

EmileTrotignon: Thanks

lib/Exposed.ml
Julow: This bug happens two more times in the same module.

vendor/parser-extended/parser.mly
gpetiot: Do you think that's a reasonable amount of work? The "decoding" code added in Fmt_ast would be greatly reduced if we did it in this PR.

lib/Exposed.ml
EmileTrotignon: This is fixed

vendor/parser-extended/parser.mly
EmileTrotignon: I think we can add later as the "decoding" code is already written.

This PR was quite stale and is very old, I do not have much time to spend on ocamlformat, and I do not want this to go stale again.

Some of that code must also have been merged in #2451

vendor/parser-extended/parser.mly
```diff
@@ -85,15 +89,20 @@ let psig_typext (te, ext) =
...
-  (Psig_type (nr, tys), ext)
-let psig_typesubst ((nr, ext), tys) =
+*)
+
+let psig_type (nr, tys) =
+  Psig_type (nr, tys)
+let psig_typesubst (nr, tys) =
   assert (nr = Recursive); (* see [no_nonrec_flag] *)
-  (Psig_typesubst tys, ext)
+  Psig_typesubst tys
```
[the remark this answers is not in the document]
EmileTrotignon: The use of `psig_typesubst` looks like
```ocaml
| type_subst_declarations
        { psig_typesubst $1 }
```

This is not substitutable with `Psig_typesubst $1` because `$1` is a tuple.

We could comment them but this would result in a larger diff later in this file. This applies to all the "tuple-taking" functions here.

vendor/parser-extended/ast_helper.ml
```diff
@@ -54,8 +54,12 @@ module Attr = struct
     { attr_name = name;
       attr_payload = payload;
       attr_loc = loc }
+
   let ext_attrs ?ext ?(before=[]) ?(after=[]) () =
     {attrs_extension = ext; attrs_before = before; attrs_after = after }
+
+    let empty_ext_attrs =
+   ext_attrs ()
```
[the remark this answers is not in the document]
EmileTrotignon: Thanks, that's done.

vendor/parser-extended/parser.mly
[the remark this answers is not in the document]
Julow: The patch is indeed large and will make backports harder. I'll merge this PR knowing that:
- The new representation for extensions doesn't make sense upstream or in a CST library.
- The next backport will require some more work.

PR #2250, opened 2023-02-10
Fix parens on array literals with attrs when used as function args

ccasin:
Currently, `ocamlformat` is incorrectly formatting array literals with attributes when used as function arguments.  For example:

```
let x = f ([| 1 |] [@foo])
```

is being misformatted as 

```
let x = f [| 1 |] [@foo]
```

(And then it detects the bug and gives a "Cannot process" error)

This PR fixes that.  There are changes in two files (plus a test):
- In `Fmt_ast`, the result of `parenze_exp` was being ignored for array literals.  Now it isn't.
- in `Ast`, the `parenze_exp` function was saying that array literals should be parenthesized too often, but this went unnoticed because its result was being ignored in `Fmt_ast`.  Now it follows the same rules used for list literals.

Credit to @antalsz for noticing this bug.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
 - Preserve position of comments around constructor record (#2237, @EmileTrotignon)
 - Preserve position of comments around external declaration strings (#2238, @EmileTrotignon, @gpetiot)
 - Preserve position of comments around module pack expressions (#2234, @EmileTrotignon, @gpetiot)
+- Correctly parenthesize array literals with attributes in argument positions (#2250, @ccasin)
 
 ### Changes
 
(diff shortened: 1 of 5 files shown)


ccasin:
(I'm contributing this on behalf of Jane Street, and we have signed the corporate CLA, though I don't know how to convince the bot of that)

Julow (approved):
Looks good to me :) Can you add a changelog entry ?

The failure in ocaml-ci seems to be caused by https://github.com/ocaml-ppx/ocamlformat/pull/2230. (ocaml-ci doesn't need to be changed, we should be fixing this in https://github.com/ocaml-ppx/ocamlformat/pull/2246)

ccasin:
Done - thanks!

gpetiot (approved):
Looks good to me, thanks!

PR #2252, opened 2023-02-16
Tests: Generate js_source.ml.ocp using JaneStreet profile and promote diff

Julow:
ocp-indent has a `JaneStreet` profile and that should be used when generating `js_source.ml.ocp`.

The second commit diff the `.ocp` and the `.ref`, which we aim to reduce. The `.ocp` file is no longer useful so it is replaced by the diff, which is a lot smaller.


gpetiot (approved):
LGTM, you can merge when it's ready :)

Julow:
The CI is failing on alpine due to the different diff command. I'm looking into it.

Julow:
Promoting the diff is difficult due to the output not being the same on the different implementations of `diff`. (even when specifying the `--minimal` algorithm. I reverted back to promoting the indented file.

Issue #2257, opened 2023-02-23
Bug: line splitting between named argument and "function" expression


ceastlund:
At tip of `master` (currently 75be51454a52c8b4d7a5b50583e3925f61af7197), if I format this using `--profile=janestreet`:

```
let bind t ~f =
  unfold_step
    ~f:(function
      | Sequence { state = seed; next }, rest ->
        (match next seed with
         | Done ->
           (match rest with
            | Sequence { state = seed; next } ->
              (match next seed with
               | Done -> Done
               | Skip { state = s } -> Skip { state = empty, Sequence { state = s; next } }
               | Yield { value = a; state = s } ->
                 Skip { state = f a, Sequence { state = s; next } }))
         | Skip { state = s } -> Skip { state = Sequence { state = s; next }, rest }
         | Yield { value = a; state = s } ->
           Yield { value = a; state = Sequence { state = s; next }, rest }))
    ~init:(empty, t)
;;
```

... I get this:

```
let bind t ~f =
  unfold_step
    ~f:
      (function
       | Sequence { state = seed; next }, rest ->
         (match next seed with
          | Done ->
            (match rest with
             | Sequence { state = seed; next } ->
               (match next seed with
                | Done -> Done
                | Skip { state = s } ->
                  Skip { state = empty, Sequence { state = s; next } }
                | Yield { value = a; state = s } ->
                  Skip { state = f a, Sequence { state = s; next } }))
          | Skip { state = s } -> Skip { state = Sequence { state = s; next }, rest }
          | Yield { value = a; state = s } ->
            Yield { value = a; state = Sequence { state = s; next }, rest }))
    ~init:(empty, t)
;;
```

The problem here is the newline between `~f:` and `(function`.

PR #2259, opened 2023-03-01
Janestreet profile: Fix indentation of functions passed as labelled argument

Julow:
Fixes https://github.com/ocaml-ppx/ocamlformat/issues/2257

This is the formatting of the default profile but unfortunately, `align_symbol_open_paren` is needed in other cases and can't be removed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## unreleased
+
+### Bug fixes
+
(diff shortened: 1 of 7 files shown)


gpetiot (approved):
Looks good to me, thanks!
Let's wait for the 0.25.0 release to update the changelog and it's good to go.

PR #2263, opened 2023-03-06
Fix dune files for opam-repo checks

kit-ty-kate:
Should unlock https://github.com/ocaml/opam-repository/pull/23407

### DIFF
--- a/test/unit/dune
+++ b/test/unit/dune
@@ -1,4 +1,7 @@
 (test
  (name test_unit)
+ (deps
+  ../passing/tests/partial.ml
+  ../passing/tests/format_invalid_files_with_locations.ml)
  (package ocamlformat-lib)
  (libraries alcotest base ocamlformat-lib))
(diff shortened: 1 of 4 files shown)


test/unit/dune
```diff
@@ -1,4 +1,7 @@
 (test
  (name test_unit)
+ (deps
+  ../passing/tests/partial.ml
+  ../passing/tests/format_invalid_files_with_locations.ml)
```
Julow, line 5: There's a read to `../../lib/Fmt_ast.ml` in `Test_translation_unit` but I think you can remove the read call.

Julow (approved):

test/unit/dune
gpetiot: This patch is minimal to please the CI of https://github.com/ocaml/opam-repository/pull/23407 so I will merge it as it is.

PR #2272, opened 2023-03-07
Fix non-stabilizing comments attached to private/virtual/mutable keywords

### LINKED ISSUE #2269: Bug: comment between "mutable" and field does not stabilize
This code:
```
type t = { mutable
 (* really long comment that doesn't fit on the same line as other stuff *)
 x : int }
```

Yields this error:
```
```

gpetiot:
Fix #2269

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -341,37 +341,37 @@ let fmt_direction_flag = function
   | Upto -> fmt "@ to "
   | Downto -> fmt "@ downto "
 
+let fmt_private c loc = hvbox 0 @@ Cmts.fmt c loc @@ str "private"
+
+let fmt_virtual c loc = hvbox 0 @@ Cmts.fmt c loc @@ str "virtual"
+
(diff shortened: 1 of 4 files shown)


lib/Fmt_ast.ml
```diff
@@ -341,37 +341,37 @@ let fmt_direction_flag = function
...
 
+let opt_flag ~f c x = opt x (fun x -> fmt "@ " $ f c x)
+
 let fmt_mutable_virtual_flag c = function
   | {mv_mut= Some m; mv_virt= Some v} when Location.compare_start v m < 1 ->
-      fmt " "
-      $ Cmts.fmt c v @@ str "virtual"
-      $ fmt " "
-      $ Cmts.fmt c m @@ str "mutable"
+      opt_flag c (Some v) ~f:fmt_virtual $ opt_flag c (Some m) ~f:fmt_mutable
```
Julow, line 366: The `fmt "@ "` of `opt_flag` could be a `?(epi=fmt "@ ")` argument of the `fmt_private/virtual/mutable` functions instead. This removes a falsely general function that is only used for this. `fmt_mutable_flag` can be made more consistent also.

Issue #2275, opened 2023-03-07
Bug: record shorthand no longer enforced


ceastlund:
The following code is no longer shortened:
```
let _ = { error = (error : Error.t) }
```

It used to get shortened:
```
let _ = { error : Error.t }
```

We'd like the shorthand version to be restored.

PR #2280, opened 2023-03-08
Module patterns with a module type constraint are formatted using the short syntax

### LINKED ISSUE #2277: Bug: module patterns + module types no longer shortened
```
let f ((module M) : (module S)) = ()
```

The above code is no longer shortened. It should become:

```
let f (module M : S) = ()
```

gpetiot:
Fix #2277

This is implemented using a "hidden" option in Conf_t, we could use the same trick for #2275.

We could argue having seperate fields like `short_syntax_pat_unpack`, `short_syntax_pat_record` is better than `short_syntax`, I don't have a strong opinion either way as long as these options are hidden to the user (except the changes to `Parser.ast` propagate in a few files so maybe let's not add too many parameters).


Julow (commented):
I'd prefer if this was not specific to the janestreet profile. Record fields have a similar rewriting, that is not an option. I consider this to be a regression.

PR #2282, opened 2023-03-09
Restore short form formatting of record field aliases

### LINKED ISSUE #2275: Bug: record shorthand no longer enforced

gpetiot:
Fix #2275

(I don't think the big pattern matching could be further simplified.)


Julow (approved):
LGTM!

PR #2283, opened 2023-03-09
Internal: Fix debug warning when dropping comments

Julow:
Bug in the debug output that I noticed when reviewing https://github.com/ocaml-ppx/ocamlformat/pull/2279


gpetiot (approved):
Looks good to me, thanks!

PR #2284, opened 2023-03-10
Janestreet: Indent extension payload

Julow:
For consistency with ocp-indent.


CHANGES.md
```diff
@@ -10,9 +10,9 @@
 
 ### Changes
 
-- JaneStreet profile: add extra parens around tuple with type annotation (#2281, @gpetiot)
 - Restore short form formatting of module patterns with a module type constraint (`((module M) : (module S))` formatted as `(module M : S)`) (#2280, @gpetiot)
 - Restore short form formatting of record field aliases (#2282, @gpetiot)
+- Tweaks the JaneStreet profile to be more consistent with ocp-indent (#2281, #2284, @gpetiot, @Julow)
```
Julow, line 15: So far our changelog have been less useful than `git log`. I suggest that we start to editorialize it by grouping related entries, breaking the line before an example, etc..

gpetiot: I agree in principle, but I'm afraid in the long run it will bloat the changelog. That's not something that other projects are doing either (I just checked ocaml and dune).
Maybe if we erase the details of versions older than 5 years in the changelog from now on?

Julow: I don't think that will bloat the changelog too much.

Bug fixes don't need to be developed too much (the previous release had 5 entries for "Preserve position of comments around ..."). On the other hand, the "Changes" and "New features" sections are unreadable because they are too terse.

Let's talk again later about this. For now, I suggest that ocp-indent related fixes for Janestreet are all grouped into one entry.

PR #2289, opened 2023-03-10
JaneStreet profile: add extra parens around pattern-matching with type annotation

### LINKED ISSUE #2266: Feature request: put parens around `match` inside type annotation
The following code:
```
((match foo with | Bar -> bar | Baz -> baz) : string)
```

Is now formatted like this:
```
(match foo with
 | Bar -> bar
 | Baz -> baz
  : string)
```

It used to leave the inner parentheses intact, to clearly associate the type annotation with `match` instead of `baz`.

On its own, the new code would probably be fine, as the indentation makes the association pretty clear. However, with our current use of `ocp-indent`, the `: string` gets indented flush with `baz`. So until we're ready to ditch ocp-indent, we'd like to keep the extra parentheses.

_Edit:_ I only checked this with `--profile=janestreet`, and I'm only requesting the change in that profile.

gpetiot:
Fix #2266

This is a dirty hack in `Sugar.ml`, one day we should rewrite the parsetree to remove the need for `Sugar.Let_binding.type_cstr` (the change is very heavy).

### DIFF
--- a/lib/Sugar.ml
+++ b/lib/Sugar.ml
@@ -313,7 +313,18 @@ module Let_binding = struct
             when Source.type_constraint_is_first typ exp.pexp_loc ->
               Cmts.relocate cmts ~src:body.pexp_loc ~before:exp.pexp_loc
                 ~after:exp.pexp_loc ;
-              (xpat, `Other (xargs, sub_typ ~ctx typ), sub_exp ~ctx exp)
+              let typ_ctx = Exp xbody.ast in
+              let exp_ctx =
+                (* The type constraint is moved to the pattern, so we need to
+                   replace the context from [Pexp_constraint] to [Pexp_fun].
+                   This won't be necessary once the normalization is moved to
+                   [Extended_ast]. *)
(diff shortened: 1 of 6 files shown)


lib/Sugar.ml
```diff
@@ -313,7 +313,18 @@ module Let_binding = struct
...
                 ~after:exp.pexp_loc ;
-              (xpat, `Other (xargs, sub_typ ~ctx typ), sub_exp ~ctx exp)
+              let typ_ctx = Exp xbody.ast in
+              let exp_ctx =
+                (* The type constraint is moved to the pattern, so we need to
+                   replace the context from [Pexp_constraint] to [Pexp_fun].
+                   This won't be necessary once the normalization is moved to
+                   [Extended_ast]. *)
+                let pat = Ast_helper.Pat.any () in
+                Exp (Ast_helper.Exp.fun_ Nolabel None pat exp)
```
Julow, line 323: Hopefully removed by future CST work.

Julow (approved):
LGTM

PR #2292, opened 2023-03-13
Improved errors when comments are not preserved

gpetiot:


Julow (commented):
I get unhandled exceptions for normal comments:
```
ocamlformat: Cannot process "test/passing/tests/let_binding.ml".
(Invalid_argument "length mismatch in zip_exn: 1 <> 0")
```

For docstrings, the output looks nice:
```
ocamlformat: Cannot process "test/passing/tests/let_binding.ml".
  BUG: doc comments moved.
File "_none_", line 1:
Error: Docstring (**  doc x  *) dropped.
```

gpetiot:
Do you have the minimal code that triggered the first exception?

Julow:
No, I modified `Fmt_ast` to drop some comments and ran the tests.

gpetiot:
I generalized the error reporting for comments and doc-comments, unsafe usage of `List.zip` has been removed. Let me know what you think.

Julow (approved):
Looks good!

lib/Normalize_extended_ast.ml
```diff
@@ -159,18 +159,42 @@ let docstring conf =
...
+  List.rev (aux [] l1 l2)
+
+let diff ~f x y =
+  let loc = function Either.First x | Either.Second x -> x.Cmt.loc in
+  Set.symmetric_diff (f x) (f y)
+  |> Sequence.to_list
+  |> List.sort ~compare:(fun x y ->
+         Migrate_ast.Location.compare (loc x) (loc y) )
+  |> List.partition_map ~f:Fn.id
+  |> zip_align
```
Julow, line 182: It would be good to have a comment to explain this operation (partition then align)

gpetiot: done

Julow: That's useful but it would be nice to write why the two lists need to be aligned (and why `symmetric_diff` is not enough).

gpetiot: I realized a lot of the code I added in this function was not needed, I added some comments for clarity.

So [`Modified] errors on regular comments are not being used, I'm leaving the type as is for now, to stay generic.

Julow (approved):
Looks good!

PR #2295, opened 2023-03-14
Fix indentation of trailing double-semicolons

### LINKED ISSUE #2286: Double-semicolon mis-indented
The double-semicolon seems to have an indentation of `-2` starting from the body of the item, which has an indentation of `1` when it starts with a `fun`:

```ocaml
[@@@ocamlformat "let-binding-spacing=double-semicolon"]

module A = struct
  let f : int S.t ab -> float S.t ab -> string =
   fun (l : int S.t ab) (r : float S.t ab) ->
    match (l, r) with A, B -> "f A B"
 ;;
end
```

gpetiot:
Fix #2286


Julow (approved):
Looks good, thanks!

PR #2296, opened 2023-03-14
GA: Reuse build artifacts when running test_branch

Julow:
Avoid building ocamlformat for every test_branch jobs. Artifacts are uploaded after the main build job and downloaded by the test_branch builds.

This is not ready for review. I'm opening a PR to debug more easily (triggers are unreliable on my fork)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,3 +36,10 @@ jobs:
 
       - name: Check help files
         run: opam exec -- dune build @help --auto-promote
+
+      - name: Upload binary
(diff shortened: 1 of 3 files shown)


Julow:
This is an alternative to https://github.com/ocaml-ppx/ocamlformat/pull/2245 that I'm investigating.

Julow:
After a number of random permutations, this works!

- `test_branch` was broken since the `ocamlformat-lib` split, this fixes that.
- it no longer requires ocaml-setup and is a lot faster.

An inconvenient is that it waits on all build jobs, including `macos` and `windows`, which are not needed. This delays the startup time by 15 minutes.

PR #2298, opened 2023-03-15
Unify printing of functional arguments

Julow:
This unifies printing of function passed argument, which has been duplicated in #2291.

The new function `fmt_fun` has a quite complicated signature but that reflects the current state. 

Test_branch finds this diff:

```diff
 let partitioni_tf t ~f =
   (partition_mapi t ~f:(fun ~key ~data ->
-       if f ~key ~data then First data else Second data) [@nontail])
+       if f ~key ~data then First data else Second data)
+  [@nontail])
```

### DIFF
--- a/test/passing/tests/eliom_ext.eliom
+++ b/test/passing/tests/eliom_ext.eliom
@@ -7,8 +7,8 @@ let%client () =
     (* NB The service underlying the server_function isn't available on the
        client before loading the page. *)
     (fun
-      () ->
-      Lwt.async (fun () -> log "Hello from the client to the server!") )
+      ()
+      -> Lwt.async (fun () -> log "Hello from the client to the server!") )
 
 let%client () =
   Eliom_client.onload
--- a/test/passing/tests/eliom_ext.eliom
+++ b/test/passing/tests/eliom_ext.eliom
@@ -21,7 +21,7 @@ let%client () =
   (* NB The service underlying the server_function isn't available on the
      client before loading the page. *)
     ~foo:(fun () ->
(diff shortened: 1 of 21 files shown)


test/passing/tests/issue289.ml
```diff
@@ -33,31 +33,31 @@ let foo =
       ~doc:"Toy ID."
       ~args:[]
       ~typ:(non_null guid)
-      ~resolve:(fun _ctx x -> x.id)
+      ~resolve:(fun _ctx x -> x.id )
```
Julow, line 36: This is a regression.

test/passing/tests/labelled_args-414.ml.ref
```diff
@@ -13,7 +13,7 @@ let () =
 let () =
   very_long_function_name
     ~very_long_argument_label:(* foo *)
-    (fun
+      (fun
```
Julow, line 16: This isn't a regression, this is consistent with fun args that are not the last.

test/passing/tests/eliom_ext.eliom
```diff
@@ -6,8 +6,9 @@ let%client () =
   Eliom_client.onload
     (* NB The service underlying the server_function isn't available on the
        client before loading the page. *)
-    (fun () ->
-      Lwt.async (fun () -> log "Hello from the client to the server!") )
+    (fun
+      ()
+      -> Lwt.async (fun () -> log "Hello from the client to the server!") )
```
Julow, line 11: I couldn't stop the difference in the boxes that causes this regression.

Julow:
This fixes more things than it breaks, I'll merge.

PR #2299, opened 2023-03-15
ocp-indent-compat: Break label if '->' doesn't fit

Julow:
On top of https://github.com/ocaml-ppx/ocamlformat/pull/2298

Ocp-indent will align the argument list with the 'fun' keyword:

    foo ~bar:(fun
               baz ->

We must make sure to break this into:

    foo
      ~bar:
        (fun
          baz ->

Test_branch reports a few fixes related to labels with comments.


Julow:
`test_branch` is happy so I merge.

PR #2301, opened 2023-03-16
Dock 'object .. end' blocks in class signatures

### LINKED ISSUE #2288: Bad formatting of class signatures
```ocaml
class c :
  'a
  -> object
       val x : 'b
     end
```
The `object` part that comes last should be docked. The arguments can then fit on the first line.

Desired output:

```ocaml
class c : 'a -> object
    val x : 'b
  end
```

Julow:
This includes https://github.com/ocaml-ppx/ocamlformat/pull/2297
Fix #2288 

Make class signatures more consistent with class structures. This does not make class signature consistent with function signatures yet but is a step in the right direction.


gpetiot:
Great job!

Do you think we could slightly improve the formatting of `class c : 'a -> let open X in k` for the Jane Street profile? We get:

```
class c : 'a
  -> let open X in
     k
```

Before the change it was formatted as:
```
class c :
  'a
  -> let open X in
     k
```

With the conventional profile we get:
```
class c : 'a ->
  let open X in
  k
```

gpetiot:
Fix #2288

gpetiot:
Fair enough! All good to go then!

gpetiot (approved):

PR #2302, opened 2023-03-16
Janestreet: Do not de-indent top-level 'fun'

Julow:
Reduces the diff between ocamlformat and ocp-indent in the janestreet profile.

This is implemented as a separate, internal option instead of relying on the 'ocp_indent_compat' option to show that the two call sites are related.


gpetiot (approved):
Looks good!

PR #2307, opened 2023-03-16
Avoid breaking after the 'mutable' keyword

Julow:
The mutable keyword is allowed to break due to comment. Make sure this happen only when necessary.


Julow:
This is a recent regression, no changelog needed.

gpetiot:
> This is a recent regression, no changelog needed.

If this has been introduced by #2272 we can add this PR number in the same changelog entry

gpetiot (approved):
Looks good, thanks!

Issue #2278, opened 2023-03-08
Bug: Double parenthesis around tupple


lyrm:
**Describe the bug**
Add extra parentheses around a tupple resulting in something like `( (tuple))`

**How to Reproduce**
Ocamlformat version : I tried `0.23.0` and `0.25.1`.
Option : none  (default)

Here is an example where it happens: 
```ocaml
type 'a bar = A of 'a * 'a option

let foo a =
  match a with A (n, m) -> ( (n, match m with None -> None | Some b -> b))
````

Julow:
Thanks for the report, fixed in https://github.com/ocaml-ppx/ocamlformat/pull/2308

PR #2308, opened 2023-03-16
Remove double parenthesis around tuple in a match

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2278

The formatting for tuples doesn't respect the `~parens` mechanism due to the `parens_tuple` option.
It's not implemented in `Ast` because it depends on an option. Perhaps we need a new mechanism on top ?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 ### Bug fixes
 
+- Remove double parenthesis around tuple in a match (#2308, @Julow)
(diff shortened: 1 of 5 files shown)


gpetiot:
> It's not implemented in `Ast` because it depends on an option.

For this, we already use references on some options in `Ast.ml`, not very pretty but it's working. And maybe it's worth it to simplify `Fmt_ast.ml` to the detriment of `Ast.ml`.

gpetiot (approved):
It looks better, thanks!

PR #2310, opened 2023-03-17
Remove parens around match sugared type constraint

Julow:
fun a : t ->
      match b with

The AST for the code above looks like

    Pexp_fun (<a>, Pexp_constraint (Pexp_match _, <t>))

We must not use it as a context as this would add parenthesis in ocp-indent-compat mode. Instead, create a ghost expression that do not contain `Pexp_constraint` and use that as the context for the `match`.


gpetiot (approved):
The `janestreet` diff looks good to me, thanks!

PR #2311, opened 2023-03-20
ocp-indent-compat: Dock trailing fun after infix

Julow:
Indentation of the trailing fun of an apply should be indented starting
from the infix, not from the beginning of the apply.

The box structure is different depending on the option.


gpetiot (approved):
It looks consistent with the ocp-indent output. LGTM!

PR #2313, opened 2023-03-21
Control indentation of [@@] attributes after value binding

Julow:
Should not be indented.


gpetiot:
It makes sense to not indent the attributes at item-level, I even think it should be the default behavior.

Julow:
Updated to work on `val`s and `external`s.

gpetiot:
There are some diffs on the conventional and ocamlformat profiles (with the test_branch action):
```diff
   let bar : type a. a t -> int32 -> int -> a =
    fun t f i -> match t with IO -> Some i | F -> f
-   [@@inline always]
+    [@@inline always]
```

The janestreet profile has extra diff that doesn't look like it's related to this change though.

Julow (commented):
This is fixed. No more diffs in test_branch for the other profiles. I can't reproduce the diffs in the janestreet profile locally. I'll merge.

PR #2316, opened 2023-03-22
ocp-indent-compat: Docked match after infix

Julow:
This implements the `~epi` mechanism for matches.

Some care is needed to disable alignment and set the right indentation for potentially docked matches.


gpetiot (approved):
> Some care is needed to disable alignment and set the right indentation for potentially docked matches.

It looks like you took care of this.

LGTM, thanks! Feel free to merge when ready!

PR #2320, opened 2023-04-03
[debug] Output the *.boxes.ml filenames on stderr

gpetiot:
I've got a path that is hard to guess (`/var/folders/0w/1cq3286x3953d_fkq269nsv00000gn/T/js2.boxes.ml`) when debugging the box structure. That would be useful to output the filename (it is even clickable in VS code).

In the future, maybe add a dependency to `Logs` for nicer messages?

Edit: it looks like `sed` is not installed on the windows image, I'm disabling the test on windows


Julow (approved):
Looks good!

PR #2314, opened 2023-03-22
Improvements to ocp-indent-compat and the Janestreet profile

Julow:

PR #2321, opened 2023-04-04
ocp-indent-compat: add missing break before attribute at the end of an expression

gpetiot:
Extracted from #2314

### DIFF
--- a/test/passing/tests/js_source.ml.ocp
+++ b/test/passing/tests/js_source.ml.ocp
@@ -83,7 +83,8 @@ let () =
   and[@foo] y = 4 in
   [%foo
     (let module M = M in
-     ()) [@foo]];
+     ())
+    [@foo]];
   [%foo
     (let open M in
      ()) [@foo]];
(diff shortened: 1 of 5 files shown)


test/passing/tests/js_source.ml.ocp
```diff
@@ -98,18 +99,21 @@ let () =
...
       ()
-    done [@foo]];
+    done
+    [@foo]];
   [%foo
     for x = () to () do
       ()
-    done [@foo]];
+    done
+               [@foo]];
```
Julow, line 108: That's not right. The break should be avoided in this case while still breaking if the attribute would cross the margin.

gpetiot: It looks like something we should fix in ocp-indent instead

Julow: Agree. In the meantime, this PR would cause unwanted diffs so it shouldn't be merged.

This is a minor issue among a huge list so I think we can skip fixing it.

PR #2322, opened 2023-04-04
Improve formatting of module arguments

### LINKED ISSUE #1678: Bug: misaligned higher-order functor parameters

Julow:
This fixes the formatting of module arguments, for example:

```diff
@@ -4501,11 +4501,12 @@ let f (module M : S with type t = int) = {M.a= 0}
 
 let flag = ref false
 
-module F (S : sig
-  module type T
-end)
-(A : S.T)
-(B : S.T) =
+module F
+    (S : sig
+      module type T
+    end)
+    (A : S.T)
+    (B : S.T) =
 struct
   module X = (val if !flag then (module A) else (module B) : S.T)
 end
@@ -6914,9 +6915,11 @@ end
 
 type 'a tree = E | N of 'a tree * 'a * 'a tree
 
-module Bootstrap2 (MakeDiet : functor (X : ORD) ->
-  SET with type t = X.t tree and type elt = X.t) : SET with type elt = int =
-struct
+module Bootstrap2
+    (MakeDiet : functor
+      (X : ORD)
+      -> SET with type t = X.t tree and type elt = X.t) :
+  SET with type elt = int = struct
   type elt = int
 
   module rec Elt : sig
```

The code is rewritten to avoid unsafe box operations. The second commit implements ocp-indent compatibility, which changes the docking behavior and indentation.

edit: fix #1678


gpetiot:
Apparently this fixes #1678

lib/Params.mli
```diff
@@ -36,6 +37,15 @@ module Exp : sig
     -> Fmt.t
 end
 
+module Mod : sig
+  type args =
+    { dock: bool  (** Whether each argument's [pro] should be docked. *)
+    ; arg_psp: Fmt.t  (** Break before every arguments. *)
```
gpetiot, line 43: What do you think of using a `Fmt.sp` here?

gpetiot (approved):
The new output looks great!

LGTM, thanks!

lib/Params.mli
Julow: That was my initial idea but it's `str " "` in some cases. I come to the conclusion that `Fmt.t` is generally better than `Fmt.sp`.

gpetiot: You can use `Fmt.Blank` for `" "`

PR #2328, opened 2023-04-06
Adjust indentation of class-expr function body

gpetiot:
The goal is to not increase the indentation when

```ocaml
class x =
  object
  end
```

becomes:

```ocaml
class x =
  fun x ->
  object
  end
```

but also fix the indentation (regarding the beginning of the `fun` keyword instead of the parenthesis) of:

```ocaml
      (fun v ->
         object
           method v : string = v
         end )
```


Julow (commented):
Not implemented for nested `fun` (using attributes to force desugared form):
```ocaml
let x =
  fun [@foo] x ->
  fun [@foo] y ->
  object
    method x = y
  end
;;

class x =
  fun [@foo] x ->
  fun [@foo] y ->
    object
      method x = y
    end
```

PR #2329, opened 2023-04-06
[debug] Output the *.ast filenames on stderr

gpetiot:
Follow-up of #2320

I find it more convenient to jump to the AST structure when using `--debug` instead of running `tools/printast/printast.exe`.

Disjunctions don't work on `sed` apparently, even with `-E`, if it gets too complicated one day we can switch to an ocaml binary to do the substitutions (like we did in dune-release cram tests)


Julow (approved):
Looks good.

PR #2330, opened 2023-04-06
Indent extensions according to [stritem-]extension-indent

gpetiot:
@Julow here is the indentation as we discussed, plus the diff for the non-janestreet profiles


lib/Fmt_ast.ml
```diff
@@ -559,11 +559,17 @@ let rec fmt_extension_aux c ctx ~key (ext, pld) =
...
   | _ ->
-      wrap "[" "]"
-        ( str (Ext.Key.to_string key)
-        $ fmt_str_loc c ext
-        $ fmt_payload c (Pld pld) pld
-        $ fmt_if (Exposed.Right.payload pld) " " )
+      let indent =
+        match key with
+        | Regular -> c.conf.fmt_opts.extension_indent.v
+        | Item -> c.conf.fmt_opts.stritem_extension_indent.v
```
Julow, line 565: That's not right. The option applies when the extension *contains* structure items, not when it is one.

PR #2331, opened 2023-04-11
Remove eol parameter of fmt_class_expr

gpetiot:
Forcing the linebreaks of comments is usually something we avoid. But worse, this parameter was used to determine the indentation before #2328.

These parameters that are passed everywhere for no reason (or the wrong reasons) should be trimmed with time.


Julow (approved):
Looks good!

PR #2333, opened 2023-04-12
Remove '--numeric' option and recovery feature

gpetiot:
This feature is only available from the emacs plugin, and from the feedbck we got it can lead to surprising results, and the running time is not acceptable on big files, that's not something we can use on the default flow.
This code is not worth maintaining anymore.


Julow (approved):
Looks good!

The failure in CI is not related to this PR (fixed in https://github.com/ocaml-ppx/ocamlformat/pull/2334).

PR #2334, opened 2023-04-13
Update to cmdliner 1.2.0

Julow:
The package remains compatible with `cmdliner >= 1.1.0` but the tests now need `>= 1.2.0`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   build:
     strategy:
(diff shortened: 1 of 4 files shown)


Julow:
The windows builder doesn't have the latest version of opam-repository yet so this needs to wait.

Julow:
The failure on `main` prevents `test_branch` from running and needs to be fixed urgently. I made sure that the failure in the windows build doesn't interfere with `test_branch`. I'll merge.

PR #2336, opened 2023-04-14
Don't print warnings related to odoc code-blocks when '--quiet' is set

gpetiot:


Julow (approved):
Looks good! The CI breakage seems unrelated.

PR #2339, opened 2023-04-19
Generate documentation pages for manpages 

Julow:
Odoc pages containing the `--help` of the two commands are added to the documentation.

They are updated by the new `@gen_manpage` target, which is checked in CI. The `.txt` files are removed.


gpetiot (approved):
LGTM, thanks!

But let's fix the build on 4.08 and 4.09 first :)

PR #2342, opened 2023-04-20
Update opam deps in dune-project file

gpetiot:
See https://github.com/ocaml/dune/issues/3653 about the absence of a `not` operator


Julow (commented):
Thanks!

PR #2345, opened 2023-04-21
Preserve functor concrete syntax in the Parsetree

gpetiot:
I was reminded of the state of the functors while working on the new syntax.

This simplifies the formatting, removes the sugaring, the need to relocate comments and the need to lookup the source.


Julow (approved):
Very nice, LGTM!

Issue #2346, opened 2023-04-24
Bug: `( :: )` is not supported in patterns


rafoo:
**Describe the bug**

Calling Ocamlformat on a file containing `( :: )` in a pattern fails with

```
ocamlformat errors:
ocamlformat: Cannot process [..].
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: generating invalid ocaml syntax.
```

**How to Reproduce**

Call Ocamlformat on the following file:

```
let is_empty = function
  | [] -> true
  | ( :: ) _ -> false
```

PR #2347, opened 2023-04-24
Consistently wrap (::)

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2346

The formatting of (::) was handled differently in several places: pattern/expressions, with/without argument.
Take care of parenthesis in the `fmt_longident` function instead.


gpetiot (approved):
LGTM, thanks!

PR #2348, opened 2023-04-24
Handle short syntax for generative functor types

### LINKED ISSUE #2225: Short syntax for generative functor types
Support new short syntax for generative functor types https://github.com/ocaml/ocaml/pull/11694

gpetiot:
Fix #2225


vendor/parser-extended/parsetree.mli
```diff
@@ -862,7 +862,7 @@ and module_type =
 and module_type_desc =
   | Pmty_ident of Longident.t loc  (** [Pmty_ident(S)] represents [S] *)
   | Pmty_signature of signature  (** [sig ... end] *)
-  | Pmty_functor of functor_parameter loc list * module_type
+  | Pmty_functor of syntax_kind * functor_parameter loc list * module_type
       (** [functor (X1 : MT1) ... (Xn : MTn) -> MT] *)
```
Julow, line 866: It's either `functor (X1 : MT1) ... ->` or `() ->`. I think this would be better represented with a new constructor `Pmty_gen of module_type`.

vendor/parser-extended/parser.mly
```diff
@@ -1124,7 +1124,13 @@ module_expr:
...
       { expecting $loc($1) "struct" }
   | FUNCTOR attrs = attributes args = nonempty_functor_args MINUSGREATER me = module_expr
-      { mkmod ~loc:$sloc ~attrs (Pmod_functor (args, me)) }
+      { let me =
+          match attrs, me with
+          | [], {pmod_desc= Pmod_functor (args', me'); pmod_attributes= []; _} ->
+              Pmod_functor (args @ args', me')
+          | _ -> Pmod_functor (args, me)
+        in
+        mkmod ~loc:$sloc ~attrs me }
```
Julow, line 1133: This could have been a different PR.

gpetiot: done in #2350

vendor/parser-extended/parsetree.mli
gpetiot: done

lib/Fmt_ast.ml
```diff
@@ -3433,6 +3438,28 @@ and fmt_module_type c ?(rec_ = false) ({ast= mty; _} as xmty) =
...
             (fits_breaks " " ~hint:(1, 2) "")
             blk.psp }
+  | Pmty_gen (gen_loc, mt) ->
+      let blk = fmt_module_type c (sub_mty ~ctx mt) in
+      let pre =
+        if V.at_least c.conf Ocaml_version.Releases.v5_1_0 then noop
+        else
+          str "functor"
+          $ fmt_attributes c ~pre:Blank pmty_attributes
+          $ fmt "@;<1 2>"
```
Julow, line 3448: Now that there's a new constructor, I think this cause more problems than it solves.

If the syntax was present in the source file, I think it can be safely outputted.

gpetiot: Yes but the ocaml-version option allows the user to parse a file with 5.1.0 and produce an output for 4.08 for example

Julow: Is this the intended purpose of this option ? I've never taken this into account and I don't think we can ensure this is correctly handled.
This is in contradiction with the direction we are taking regarding parsing.

If this is something we want to do in this case (I don't think so) it'll be better handled as a transformation pass on the AST rather than adding duplicated/inconsistent formatting code.

gpetiot: That's useful if people are not all using the same version of ocaml on the same project. I'm not saying it's the best use of ocamlformat, but that's what we have done for new syntaxes in the past (I have punning in mind, but maybe more). I think @hhugo emitted the idea if I'm not mistaken.

Julow: We didn't have a concrete AST at this time and we needed to avoid sugaring existing code into new syntaxes. 

I don't agree with this rewrite being done in the formatting function and I'd prefer to move that into a transformation pass (similarly to how punning is handled now).

I suggest to move this conversation in a new PR/issue and merge this PR without the rewriting feature.

Julow: I pushed a commit to remove the automatic syntax downgrade and I'll merge. Happy to discuss the proposal in a new PR.

This is necessary due to the concrete AST: we can safely output the new syntax because it was also present in the input. The version of ocaml required to parse the output file is identical to the input file.

That's also the direction we are taking with the more concrete AST.

PR #2349, opened 2023-04-24
Fix formatting of cinaps comments with strings

Julow:
Strings containing breaks must carefully avoid changing the content of strings while indenting the whole code.


gpetiot (approved):
LGTM!

Issue #2351, opened 2023-04-25
Bug: Exposed match on right of `fun _ : _ ->`


Julow:
The sugared type annotation on the `fun` is not taken into account when computing parenthesis:

```
let () =
  match () with
  | _ ->
    fun _ : _ ->
    (match () with
     | _ -> ())
  | _ -> ()
;;
```

Julow:
Fixed.

PR #2352, opened 2023-04-25
Protect match after `fun _ : _ ->`

Julow:
Add a missing check for exposed matches.


Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2351

gpetiot (approved):
LGTM, maybe add the test from #2351 for the CI

PR #2358, opened 2023-04-29
Licensing: Facebook CLA does not apply to new contributions

gpetiot:
The Facebook CLA has recently been removed (I've already removed the tag for PRs since their bot is not checking it anymore).

This PR
- removes the script enforcing the Facebook headers in all files.
- uses the CODE_OF_CONDUCT from `ocaml`.

The FB copyright keeps being enforceable on old files but not on new ones.
I've kept the original license for now. We use the MIT license, which is pretty permissive already. It mentions the Facebook copyright, which should stay, I'm not sure it's worth mentioning that their copyright only applies to files that have the corresponding header (sounds obvious but NAL), so the LICENSE is left unchanged.

cc @jberdine @trefis


jberdine:
This looks good, thanks. Should the old mentions of the CLA be replaced by something about using a DCO instead, rather than just removing it?

gpetiot:
> Should the old mentions of the CLA be replaced by something about using a DCO instead, rather than just removing it?

I'm tempted to just mirror the `ocaml` repo for now and not use one (the benefits are not very clear to me, on `dune` for example I don't find it very contributor-friendly as a very occasional contributor).

Julow:
> I'm not sure it's worth mentioning that their copyright only applies to files that have the corresponding header (sounds obvious but NAL)

It's not obvious to me. I think this should be mentioned.

jberdine:
It isn't clear to me which files are covered by the LGPL added in this PR, or what the consequences are for users and contributors. Can you clarify? Or is there some discussion about the pros and cons of the mixed licensing vs just continuing with MIT? I'm curious.

gpetiot:
Maybe it's simpler to keep the MIT license for new files and add a different header for them (no new file for now, but basically just remove the copyright line).

jberdine:
IANAL but I think it would work to change the `Copyright (c) 2017-present, Facebook, Inc.` line in the LICENSE file to replace `present` with `2023`, and then add additional lines such as `Copyright (c) 2023-present, Tarides.` whenever a new file is added with, for example, `Copyright (c) Tarides.` instead of `Copyright (c) Facebook, Inc. and its affiliates.` in the header. That way the overall license file indicates that different parts of the source code have different copyright holders, and the individual files will indicate their copyright holder.

I suspect that multiple licenses is more complicated for users and so on than multiple copyright holders.

gpetiot:
I've reverted the dual-licensing commit, and added the Tarides copyright in the LICENSE (that will be used for new files).

LICENSE.md
```diff
@@ -1,6 +1,7 @@
 MIT License
 
-Copyright (c) 2017-present, Facebook, Inc.
+Copyright (c) 2017-2023, Facebook, Inc.
+Copyright (c) 2023-present, Tarides.
```
gpetiot, line 4: @tmattio @samoht to confirm the exact formula of the Tarides copyright

jberdine (approved):
LGTM assuming Tarides folks are happy with that copyright line.

samoht:
LGTM

PR #2370, opened 2023-06-01
Fix crash with module unpack with constraint

Julow:
The crash was caused by an Ast rule that is outdated since #2191.


gpetiot (approved):
Looks good to me!

PR #2371, opened 2023-06-06
Refactor handling of comments

Julow:
This rewrite part of the code that parses and formats comments to remove repeated or inconsistent code.

One function interprets comments: `Cmt.decode`. It returns a richer type that allow to implement formatting as it was before.
The normalization is changed to use that, no more `is_docstring` and normalization of indentation is now consistent. This fixes normalization problem for comments parsed as doc and cinaps comments.


lib-rpc-server/ocamlformat_rpc.ml
```diff
@@ -82,7 +82,8 @@ let run_format conf x =
     (* The formatting functions are ordered in such a way that the ones
        expecting a keyword first (like signatures) are placed before the more
        general ones (like toplevel phrases). Parsing a file as `--impl` with
-       `ocamlformat` processes it as a use file (toplevel phrases) anyway.
+       `ocamlformat` processes it as a use file (toplevel phrases)
+       anyway.
```
Julow, line 86: This is a regression.

test/passing/tests/break_separators-after_docked.ml.ref
```diff
@@ -322,8 +322,7 @@ let foooooooooooo =
 let foooooooooooo =
   {
     foooooooooooooo with
-    (* foooooooooooooooo fooooooooooooooooooooooooo
-       foooooooooooooooooooooo *)
+    (* foooooooooooooooo fooooooooooooooooooooooooo foooooooooooooooooooooo *)
```
Julow, line 325: This is an unwrapped comment. It's not obvious how the break appeared with the previous formatting code.

test/passing/tests/doc_comments-after.ml.ref
```diff
@@ -266,11 +266,9 @@ module A = struct
...
 end
 
-(* Same with get_pure, except that when we have both "x = t" and "y = t"
-   where t is a primed ident, * we add "x = y" to the result. This is crucial
-   for the normalizer, as it tend to drop "x = t" before * processing "y =
-   t". If we don't explicitly preserve "x = y", the normalizer cannot pick it
-   up *)
+(* Same with get_pure, except that when we have both "x = t" and "y = t" where t is a primed ident,
+ * we add "x = y" to the result. This is crucial for the normalizer, as it tend to drop "x = t" before
+ * processing "y = t". If we don't explicitly preserve "x = y", the normalizer cannot pick it up *)
```
Julow, line 271: It seems that this kind of asterisk prefixed comments were never handled correctly before.

test/passing/tests/js_source.ml.ocp
```diff
@@ -10042,8 +10042,7 @@ let _ =
   | _ -> false
 ;;
 
-(*
-*)
+(* *)
```
Julow, line 10045: This is an improvement.

test/passing/tests/wrap_comments.ml.ref
```diff
@@ -147,6 +159,6 @@ let _ =
 
 let _ =
   (*
-     blah blah
+    blah blah
```
Julow, line 162: This is an improvement. The extra indentation was caused by a trailing space on the first line.

test/passing/tests/sequence-preserve.ml.ref
```diff
@@ -91,10 +91,11 @@ let foo x y =
   another_important_function x y ;
   cleanup x y
 
-(* This test require --max-iter=3 *)
 let _ =
   some statement ;
-  (* comment with an empty line in it tricky *)
+  (* comment with an empty line in it
+
+     tricky *)
```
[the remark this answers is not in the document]
Julow: On `main`, this code:
```ocaml
(* first line

   last line *)

(* first line fooooooooooooooooooooooooooooooooooooooooooooo

   last line fooooooooooooooooooooooooooooooooooooooooooooo *)
```
formats to:
```
(* first line last line *)

(* first line fooooooooooooooooooooooooooooooooooooooooooooo

   last line fooooooooooooooooooooooooooooooooooooooooooooo *)
```
independently of the `wrap-comments` option. I declare this behavior a bug. With this PR, the first code block is well formatted.

lib/Cmt.ml
```diff
@@ -88,25 +88,115 @@ end
...
+  | hd :: tl -> unindent_lines ~opn_offset hd tl
+
+let split_asterisk_prefixed =
+  let line_is_asterisk_prefixed s = String.is_prefix s ~prefix:"*" in
+  let line_is_asterisk_or_space_prefixed s =
+    if String.is_empty s then true
+    else match s.[0] with '*' | ' ' -> true | _ -> false
+  in
+  (* Whether every lines starts with "*" or " ". At least one line must start
+     with a "*" and completely empty lines are allowed. *)
```
gpetiot, line 135: This creates regressions such as:
```diff
                 (*with exn ->
-                    failwiths "binary_search bug"
-                    (exn, `length length, `search_key search_key, `pos pos, `len len)
-                    <:sexp_of< exn * [ `length of int ] * [ `search_key of int ]
+                 *  failwiths "binary_search bug"
+                 *  (exn, `length length, `search_key search_key, `pos pos, `len len)
+                 *  <:sexp_of< exn * [ `length of int ] * [ `search_key of int ]
                  * [ `pos of int ] * [ `len of int ] >>*)
```

Julow: Good catch, I'll make this parsing stricter.

Julow:
The changes caused by this PR are reasonable to me: the original placement of `*)` is respected more often, comments that shouldn't be wrapped are not.
I don't see any regression in `test_branch` on projects that were previously formatted with the default profile.
I would merge this in its current state.

Any opinion @jberdine @gpetiot ?

lib-rpc-server/ocamlformat_rpc.ml
gpetiot: Did you find what causes this regression? (not important, just wondering)

test/passing/tests/break_separators-after_docked.ml.ref
jberdine: Why is it unwrapped? It is neither a banner comment nor is opened with `(*=`, is there some other mechanism to make it unwrapped?

test/passing/tests/break_separators-after.ml.ref
```diff
@@ -371,7 +370,8 @@ let g () =
        hhhhhhhhhh |] ->
       fooooooooo
 
-let () = match x with _, (* line 1 line 2 *) Some _ -> x
+let () = match x with _, (* line 1 line 2 *)
+                         Some _ -> x
```
jberdine, line 374: This is a regression.

test/passing/tests/doc_comments-after.ml.ref
jberdine: Nice. I don't know what was going on before, but it is good to no longer wrap such banner comments and leave stray `*` characters in the middle of the comment.

test/passing/tests/js_to_do.ml.ref
```diff
@@ -14,7 +14,8 @@ let _ =
 (* js-type *)
 
 (* The following tests incorporate several subtle and different indentation
-   ideas. Please consider this only a proposal for discussion, for now.
+   ideas. Please consider this only a proposal for discussion, for
+   now.
```
jberdine, line 18: This is a regression.

test/passing/tests/break_separators-after_docked.ml.ref
Julow: Right! I forgot that the tests were run with `wrap-comments` enabled. The new formatting is fine then.

jberdine: You mean this PR changes the settings used to run the tests? That makes it hard to tell if there are regressions. We also do want to test the comment wrapping functionality, probably more thoroughly than the non-wrapping functionality, since it is nontrivial.

Julow: The tests are run with `wrap-comments` since before this PR. This is something that I frequently forget about. This diff shows a fix and not a regression, my first comment was wrong.

> We also do want to test the comment wrapping functionality, probably more thoroughly than the non-wrapping functionality, since it is nontrivial.

That's what we always did but I don't agree with the reason. The wrapping algorithm is trivial.
Perhaps you are talking about detecting which comments should be wrapped/preserved/asterisk prefixed ? This is now unrelated to the `wrap-comments` option.

test/passing/tests/break_separators-after.ml.ref
Julow: This change was introduced in intermediate state of this PR and goes away if the test file is reverted. The bug exists but is not related to this PR.

lib-rpc-server/ocamlformat_rpc.ml
Julow: After a few testing, it seems not directly caused by the raw `\n` character because the same happens with `@;<1000 -1000>@\n`.

I have not found a way to print an empty reliably.

test/passing/tests/js_to_do.ml.ref
Julow: I'm stuck on this one. The same thing happen in ocamlformat's source: https://github.com/ocaml-ppx/ocamlformat/pull/2371#discussion_r1219329872

The hack required to print an empty line without trailing space is not the same at a point that would break (if the empty line was swapped with a `@ `) and a point that would not break (several small lines).

Julow:
Closes https://github.com/ocaml-ppx/ocamlformat/issues/2468

jberdine:
I had another quick look at this, do I understand correctly that in the current state of this PR there are still a number of regressions where lines are now wrapped before they reach the margin, and a number of regressions where the margin is now exceeded? The plan is to resolve these before merging this, right?

Julow:
I'm still investigating this issue but I'm now considering merging without it fixed. The issue seems rare and I'm clueless on how to fix it.

test/passing/tests/js_to_do.ml.ref
gpetiot: This should be fixed in a400535

lib-rpc-server/ocamlformat_rpc.ml
gpetiot: This should be fixed in a400535

test/passing/tests/break_separators-after_docked.ml.ref
gpetiot: So no regression if I understand correctly?

Julow: Yes, this is fine.

lib-rpc-server/ocamlformat_rpc.ml
Julow: Nice! Thanks a lot :)

Julow:
The unexplained break regression is fixed and the new cases where the margin is exceeded is due to asterisk prefixed comments that do not wrap. This is ready to be merged!

PR #2388, opened 2023-06-28
Disable short syntax for record field on 4.02

Julow:
The new syntax rewrite introduced in #2282 is not compatible with 4.02.
The default value for the `ocaml-version` option is 4.04 but users are still supporting OCaml 4.02.

`test/passing/tests/record-402.ml.ref` parse on 4.02. `test/passing/tests/record.ml.ref` parse on 4.03.
The tests that are specifically about the newer syntax are moved into an other test.


test/passing/tests/record-tight_decl.ml.ref
```diff
@@ -38,75 +38,11 @@ let f ~l:{f; g} = e
...
-let Mmmmmm.
-      { xxxx
-      ; xxxxxxxxx
-      ; xxxxxxxxxxxxxxxxxx
-      ; xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx (* foooooooooooo *) }
-    (* fooooooooo *) =
-  ()
-
-let _ = {a; b : c = (match b with `A -> A | `B -> B | `C -> C); c}
+(* TODO: let _ = {a; b = ((match b with `A -> A | `B -> B | `C -> C) : c);
```
Julow, line 41: In this form, it changes the AST. I'll fix that in a follow-up PR.

PR #2398, opened 2023-07-17
Restore indentation of `->`

Julow:
The second commit restore the indentation of `->` as it was in 0.25.1.
The third commit fixes a bug in the formatting of `fun` after a label.


Julow:
The test-branch diff is large but intersects with recent regressions. The test-branch diff since the previous release seems to be improved slightly. Let's see in the preview PRs.

PR #2402, opened 2023-07-18
CI: Cache test-extra/code when running test_branch

Julow:
I hope to make test-branch 10% faster.


Julow:
Success, test-branch is 50% faster!

PR #2403, opened 2023-07-18
Prepare 0.26.0

Julow:
I edited the changelog slightly as I think it looks too much like `git log`:

- Non-bug entries are moved to the "Changes" section.
- Lists are sorted by PR number.
- Similar entries are merged.
- Some are re-worded.

I think we can do better. For example, changes that are likely to cause regressions are not marked as such.


Julow:
> For example, changes that are likely to cause regressions are not marked as such.

I've changed that.

PR #2404, opened 2023-07-19
CI and opam file fixes after 0.26.0

Julow:


HACKING.org
```diff
@@ -42,9 +42,7 @@ Git is used to manage the script's state:
   =git push --tags=
 
 - Release in automatic mode
-  =dune-release -p ocamlformat,ocamlformat-rpc-lib=
-  Until dune-release.1.5.0 is released the =main= branch of dune-release
-  should be used.
+  =dune-release -p ocamlformat-lib,ocamlformat=
```
gpetiot, line 45: `ocamlformat-rpc-lib` still needs to be released since we check for the version to be the same as ocamlformat-lib and ocamlformat (only for the tests but still), or we need to change the version requirement for this lib

Julow: Fixed, thanks.
Note that this PR is not ready.

Julow:
Windows CI should finally work.

PR #2412, opened 2023-07-25
Backport 5.1.0-beta1 standard parser changes

Julow:
Changes that are backported:

- New `Pmod_apply_unit` module type constructor. We had a similar concept in parse-extended, so this is just a rename.

- New `value_constraint` field in value bindings.
  This makes the parser less permissive about bugs in the formatting of
  type annotation on let-bindings.

- Some changes in `Location` and `Ast_mapper` are not necessary but are
  backported to reduce the diff.

- String assignment operator has been removed upstream but remains
  supported by commenting out the change.

- Dead code in parser-extended have been removed in the past. I reverted that
  and commented-out the code instead. This makes the code more similar with
  parser-standard and helps recognize new code.

- Ast rules have changed as some constructions are no longer special.


test/passing/tests/js_source.ml.ref
```diff
@@ -9417,7 +9417,7 @@ class ['a] c () =
 let f : type a'. a' = assert false
 let foo : type a' b'. a' -> b' = fun a -> assert false
 let foo : type t'. t' = fun (type t') : t' -> assert false
-let foo : type t. t = assert false
+let foo : 't. 't = fun (type t) : t -> assert false
```
Julow, line 9420: This rewrite is no longer done.

test/passing/tests/let_binding-in_indent.ml.ref
```diff
@@ -13,7 +13,7 @@ let _ =
   let (x : int) = x in
   let x : int = x in
   let (_ : int) = x in
-  let (_ : int) = x in
+  let _ : int = x in
```
Julow, line 16: Parentheses around let-binding patterns follow the original source. These are two different AST now.
Shouldn't change existing code.

test/passing/tests/polytypes-janestreet.ml.ref
```diff
@@ -1,26 +1,26 @@
 let t1 : 'a 'b. 'a t -> b t = ()
 
-let t2 :
-      'a 'b.
-      'a t________________________________ -> 'b t_______________________________________
+let t2
+  : 'a 'b.
+  'a t________________________________ -> 'b t_______________________________________
```
Julow, line 5: I do not consider this a regression as it's the formatting used in slightly different cases.

Julow:
`test_branch` shows some regressions but I'm tempted to accept them as improvements.

Some patterns are no longer recognized as `Ppat_constraint` but rather a let-binding with annotation. Showing the inconsistency between the two code paths.

```diff
index f2e9236473..27701537f5 100644
--- a/infer/src/pulse/PulseCallOperations.ml
+++ b/infer/src/pulse/PulseCallOperations.ml
@@ -233,8 +233,8 @@ let apply_callee tenv ({ PathContext.timestamp } as path) ~caller_proc_desc
         let caller_return_var : Pvar.t =
           Procdesc.get_ret_var caller_proc_desc
         in
-        let (astate, caller_return_val_hist)
-              : AbductiveDomain.t * (AbstractValue.t * ValueHistory.t) =
+        let (astate, caller_return_val_hist) :
+            AbductiveDomain.t * (AbstractValue.t * ValueHistory.t) =
           PulseOperations.eval_var path call_loc caller_return_var astate
         in
         let+ (astate : AbductiveDomain.t) =
```

Some patterns were parenthesed because the expression was a `Pexp_constraint`. This bug is gone:

```ocaml
index 2a34ec8b1..faaf464be 100644
--- a/testsuite/tests/typing-poly/poly.ml
+++ b/testsuite/tests/typing-poly/poly.ml
@@ -1888,11 +1888,11 @@ module Polux :
 
 (* PR#5560 *)
 
-let (a, b) = (raise Exit : int * int)
+let a, b = (raise Exit : int * int)
 
 type t = { foo : int }
 
-let ({ foo }) = (raise Exit : t)
+let { foo } = (raise Exit : t)
 
 type s = A of int
```

This might also be an inconsistency between two code paths, though I can't pin it down yet.

```ocaml
index 395387a93..8b7b830aa 100644
--- a/src/memo/memo.ml
+++ b/src/memo/memo.ml
@@ -1051,9 +1051,9 @@ module Exec : sig
   val exec_dep_node : ('i, 'o) Dep_node.t -> 'o Fiber.t
 end = struct
   let rec restore_from_cache :
-            'o.
-            cached_value:'o Cached_value.t ->
-            'o Cached_value.t Cache_lookup.Result.t Fiber.t =
+        'o.
+        cached_value:'o Cached_value.t ->
+        'o Cached_value.t Cache_lookup.Result.t Fiber.t =
    fun ~cached_value ->
     match cached_value.value with
     | Cancelled _dependency_cycle ->
```

CHANGES.md
```diff
@@ -4,6 +4,17 @@ Items marked with an asterisk (`*`) are changes that are likely to format
 existing code differently from the previous release when using the default
 profile. This started with version 0.26.0.
 
+## unreleased
+
+### New features
+
+- Compatible with OCaml 5.1.0-beta1 (#2412, @Julow)
```
Julow, line 11: The regressions are small in size and affect rare code. I would not mark this as breaking.

test/cli/debug.t
```diff
@@ -260,9 +243,7 @@
               before: (* before x binding #2 *)
               Pexp_let Nonrec
               [
-                <def> (a.ml[5,90+2]..[10,210+28])
-                comments
-                   after: (* after unit *)
+                <def>
```
gpetiot, line 246: The comment is not preserved.

gpetiot: So the `(* after unit *)` comment is not attached anywhere?

Julow: My previous comment is not exact. This node has a location and the comment is still attached to it. Printast was not printing it.

Issue #2417, opened 2023-08-28
Inconsistent formatting of class arrow types


Julow:
Class types arrows are not formatted the same way as value types arrows.

With the `default` profile:

```ocaml
class unix_mockup : base_dir:string -> mem_only:bool ->
  mockup_env:Tezos_mockup_registration.Registration.mockup_environment ->
  chain_id:Chain_id.t -> rpc_context:Tezos_protocol_environment.rpc_context ->
  protocol_data:bytes -> Client_context.full

val unix_mockup :
  base_dir:string ->
  mem_only:bool ->
  mockup_env:Tezos_mockup_registration.Registration.mockup_environment ->
  chain_id:Chain_id.t ->
  rpc_context:Tezos_protocol_environment.rpc_context ->
  protocol_data:bytes ->
  Client_context.full
```

With the `ocamlformat` profile:

```ocaml
class unix_mockup : base_dir:string -> mem_only:bool
  -> mockup_env:Tezos_mockup_registration.Registration.mockup_environment
  -> chain_id:Chain_id.t -> rpc_context:Tezos_protocol_environment.rpc_context
  -> protocol_data:bytes -> Client_context.full

val unix_mockup :
     base_dir:string
  -> mem_only:bool
  -> mockup_env:Tezos_mockup_registration.Registration.mockup_environment
  -> chain_id:Chain_id.t
  -> rpc_context:Tezos_protocol_environment.rpc_context
  -> protocol_data:bytes
  -> Client_context.full
```

Julow:
This is fixed by https://github.com/ocaml-ppx/ocamlformat/pull/2422
Related issue: https://github.com/ocaml-ppx/ocamlformat/issues/2423

PR #2422, opened 2023-09-04
Consistent formatting for arrow class types

Julow:
Use the same indentation and breaks for arrows in class types as for
arrows in core types.
The main challenge is that class types contain class signatures, which
are docked after an arrow.

The `fmt_class_type` is rewritten in the "pro" style, which also
changes:

- Remove unecessary space in object poly types
- Indented `object .. end` after a `let open .. in`

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2801,14 +2801,16 @@ and fmt_class_type ?(pro = noop) c ({ast= typ; _} as xtyp) =
           @@ Params.parens_if parens c.conf
                (fmt_extension c ctx ext $ fmt_attributes c atrs) )
   | Pcty_open (popen, cl) ->
-      hvbox 2
-        ( pro
-        $ Cmts.fmt c pcty_loc
-          @@ Params.parens_if parens c.conf
-          @@ ( fmt_open_description c ~keyword:"let open" ~kw_attributes:atrs
-                 popen
-             $ fmt " in@;<1000 0>"
-             $ fmt_class_type c (sub_cty ~ctx cl) ) )
+      let pro =
+        hvbox 2
+          ( pro
+          $ Cmts.fmt c pcty_loc
+            @@ Params.parens_if parens c.conf
+            @@ ( fmt_open_description c ~keyword:"let open"
+                   ~kw_attributes:atrs popen
+               $ fmt " in@;<1000 0>" ) )
+      in
+      fmt_class_type c ~pro (sub_cty ~ctx cl)
       $ fmt_docstring c ~pro:(fmt "@ ") doc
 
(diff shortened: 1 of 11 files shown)


Julow:
This fixes https://github.com/ocaml-ppx/ocamlformat/issues/2417

lib/Fmt_ast.ml
```diff
@@ -715,6 +715,33 @@ and fmt_arrow_param c ctx {pap_label= lI; pap_loc= locI; pap_type= tI} =
...
+        let indent =
+          if Poly.(c.conf.fmt_opts.break_separators.v = `Before) then 2
+          else 0
+        in
+        fits_breaks "" (String.make (Int.max 1 (indent - separator_len)) ' ')
+    | _ ->
+        fmt_if_k
+          Poly.(c.conf.fmt_opts.break_separators.v = `Before)
+          (fmt_or_k c.conf.fmt_opts.ocp_indent_compat.v (fits_breaks "" "")
+             (fits_breaks "" "   ") )
```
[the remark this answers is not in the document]
Julow: While doing that, I found that this code is better here as I don't want it to be used more than once. I've moved the `separator_len` part to where it was before to make this function simpler.

test/passing/tests/class_sig-after.mli.ref
```diff
@@ -22,10 +22,22 @@ class c : object
...
+class unix_mockup :
+  foooo:string ->
+  foooo:string ->
+  foooo:string ->
+  foooo:string ->
+  foooo:string ->
+  foooo:string ->
+  object
   method foo : string
 end
```
Julow: With the new test case we can see that the indentation of the `object` keyword isn't so nice as it aligns with both the argument before and the content of the object.

It would be very bad to indent the whole object here as that would completely re-indent the object if an argument is added.

Perhaps the `object` keyword could be exceptionally de-indented ?
```ocaml
class unix_mockup :
  foooo:string ->
  foooo:string ->
  foooo:string ->
  foooo:string ->
  foooo:string ->
  foooo:string ->
object
  method foo : string
end
```

That's not blocking this PR though, as the object is formatted the same as before.

Julow: For the same reason, I reverted the change I made to the indentation after a class types `let open`. I've opened an issue for this: https://github.com/ocaml-ppx/ocamlformat/issues/2423

gpetiot: It wouldn't shock me to de-indent the `object` keyword, that's done for `struct` in some cases and it looks nice:
```ocaml
module M (A : sig
  module type T
end) (B : sig
  module type T
end) =
struct
  let f : ((module A.T), (module B.T)) t -> string = function B s -> s
end
```

Issue #2423, opened 2023-09-05
Improve the indentation of classes and class types


Julow:
The indentation of the `object` keyword in class types is not ideal as it's aligned with both what came before it and with the content of the object.

```ocaml
class cls : object
  method foo : string
end

class cls :
  foooooooooooooooooooooo:string ->
  baaaaaaaaaaaaaaaaaaaaar:float ->
  object
  method foo : string
end

class cls :
  let open Foo in
  object
  method foo : string
end
```

The indentation of classes is entirely different, the `object` keyword is indented and its content too. This wastes vertical space for no reason as every classes defined with an `object .. end` have the same structure, where the object is at the end.

```ocaml
class cls =
  object
    method foo = bar
  end

class cls ~foooooooooooooooooooooooooooooooo
  ~baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar =
  object
    method foo = bar
  end

class cls =
  let open Foo in
  object
    method foo = bar
  end
```

## Proposal

- [x] (https://github.com/ocaml-ppx/ocamlformat/pull/2425) Change the indentation of class types to be less ambiguous by de-indenting the `object` keyword. This only change one line in some class types:

```ocaml
class cls : object
  method foo : string
end

class cls :
  foooooooooooooooooooooo:string ->
  baaaaaaaaaaaaaaaaaaaaar:float ->
object
  method foo : string
end

class cls :
  let open Foo in
object
  method foo : string
end
```

- [ ] Change the indentation of classes to match class types. This is a major change and should be guarded behind an option:

```ocaml
class cls = object
  method foo = bar
end

class cls ~foooooooooooooooooooooooooooooooo
  ~baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar =
object
  method foo = bar
end

class cls =
  let open Foo in
object
  method foo = bar
end
```

Julow:
The smallest of the two proposed changes is merged: https://github.com/ocaml-ppx/ocamlformat/pull/2425

PR #2425, opened 2023-09-05
De-indent the `object` keyword in class types

Julow:
This is the first suggestion in https://github.com/ocaml-ppx/ocamlformat/issues/2423#issuecomment-1706553799

The previous indentation of the keyword was ambiguous as it was aligned with the previous and next lines.
This is similar to the formatting of module `struct`.

This is does not compose well with `break-separators=before` but I still think it's better than before:

```ocaml
class unix_mockup :
     foooo:string
  -> foooo:string
  -> foooo:string
  -> foooo:string
  -> foooo:string
  -> foooo:string
  ->
object
  method foo : string
end
```


gpetiot (approved):
I love the new output tbh! Even in combination with `break-separators=before` it looks like an improvement to me.
LGTM!

Julow:
Let's include this in the next release.

PR #2428, opened 2023-09-07
Improve compatibility with ocp-indent

Julow:
These are changes from https://github.com/ocaml-ppx/ocamlformat/pull/2314 that I extracted and polished a bit.

This is a work-in-progress, the goal is to cause no regressions on the default profile.

PR #2429, opened 2023-09-08
Remove extra break in constructor decl after a comment

Julow:
A comment just before a constructor in a type declaration was causing
the argument of the constructor to be formatted in vertical style.

The constructor is now formatted the same as if there were no comment.


Julow:
I found this when I had to change the box structure in https://github.com/ocaml-ppx/ocamlformat/pull/2428

gpetiot (approved):
LGTM!

PR #2433, opened 2023-09-14
ocamlformat.el (ocamlformat-before-save): support caml-mode

nojb:
What it says on the tin.


gpetiot:
I'm not using emacs so that's hard for me to fully confirm but the change looks sensible.

Julow (approved):
Also not a emacs user but this looks right.

nojb:
Thanks!

PR #2436, opened 2023-09-14
Fix dropped attributes on begin-end in an if-then-else branch

### LINKED ISSUE #2435: BUG: ast changed.
```
 let _ =
    match x with
    | [ _ ] ->
        foo
    | _ ->
        begin[@landmark "amc_calib"]
          bar
        end
```

gpetiot:
Fix #2435


Julow (approved):
Thanks!

PR #2437, opened 2023-09-14
Reduce the indentation of (polytype) type constraints

### LINKED ISSUE #2143: Indentation feels off for mutually recursive functions with type constraints
The indentation feels off in the following example:
```ocaml
let rec equal_list :
          'a. ('a, 't) gexpr marked list -> ('a, 't) gexpr marked list -> bool =
 fun es1 es2 ->
  try List.for_all2 equal es1 es2 with Invalid_argument _ -> false

and equal : 'a. ('a, 't) gexpr marked -> ('a, 't) gexpr marked -> bool =
  fun (type a) (e1 : (a, 't) gexpr marked) (e2 : (a, 't) gexpr marked) ->
   match Marked.unmark e1, Marked.unmark e2 with
   …
```
* indentation for the type of `equal_list` is huge
* first `fun` is indented 1 and its body 2, it feels weird to have width-1 increments. Why not 2 and 4, or, better, 2 and 2 ? I'd even be ok with 0 and 2
* the second `fun` is indented 2 and its body 3, which is one more than for the first function and doesn't feel consistent

```editor-config
profile = default
margin = 80
exp-grouping = preserve
break-fun-decl = fit-or-vertical
wrap-comments
parse-docstrings
version=0.24.1
cases-exp-indent=2
parens-tuple=multi-line-only
space-around-lists=false
break-infix-before-func
break-infix= fit-or-vertical
```

gpetiot:
Fix #2143

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ profile. This started with version 0.26.0.
 - Documentation comments are now formatted by default (#2390, @Julow)
   Use the option `parse-docstrings = false` to disable.
 - \* Janestreet profile: do not break `fun _ -> function` (#2460, @tdelvecchio-jsc)
+- \* Reduce the indentation of (polytype) type constraints (#2437, @gpetiot)
 
 ### Fixed
 
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -4353,29 +4353,32 @@ and fmt_value_binding c ~rec_flag ?ext ?in_ ?epi
   in
   let doc1, atrs = doc_atrs lb_attrs in
   let doc2, atrs = doc_atrs atrs in
-  let fmt_cstr =
+  let fmt_newtypes, fmt_cstr =
     let fmt_sep x =
       match c.conf.fmt_opts.break_colon.v with
       | `Before -> fmt "@ " $ str x $ char ' '
       | `After -> char ' ' $ str x $ fmt "@ "
     in
     match lb_typ with
-    | `Polynewtype (pvars, xtyp) ->
-        fmt_sep ":"
-        $ hvbox 0
-            ( str "type "
-            $ list pvars " " (fmt_str_loc c)
-            $ fmt ".@ " $ fmt_core_type c xtyp )
+    | `Polynewtype (pvars, xtyp) -> (
+      match c.conf.fmt_opts.break_colon.v with
+      | `Before ->
+          ( noop
+          , fmt_sep ":"
+            $ hvbox 0
+                ( str "type "
+                $ list pvars " " (fmt_str_loc c)
+                $ fmt ".@ " $ fmt_core_type c xtyp ) )
+      | `After ->
+          ( fmt_sep ":"
+            $ hvbox 0 (str "type " $ list pvars " " (fmt_str_loc c) $ str ".")
+          , fmt "@ " $ fmt_core_type c xtyp ) )
     | `Coerce (xtyp1, xtyp2) ->
-        opt xtyp1 (fun xtyp1 -> fmt_sep ":" $ fmt_core_type c xtyp1)
-        $ fmt_sep ":>" $ fmt_core_type c xtyp2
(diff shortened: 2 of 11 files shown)


lib/Fmt_ast.ml
```diff
@@ -4320,7 +4320,7 @@ and fmt_value_binding c ~rec_flag ?ext ?in_ ?epi
   in
   let cstr_indent =
     match lb_typ with
-    | `Other {ast= {ptyp_desc= Ptyp_poly _; _}; _} -> 6
+    | `Other {ast= {ptyp_desc= Ptyp_poly _; _}; _} -> 3
```
Julow, line 4323: I think this should be 4. The indentation of 1 compared to the function body feels weird. Also, that's the indentation of other kinds of type annotations, which would remove the special case entirely.

CHANGES.md
```diff
@@ -25,6 +25,7 @@ Tags:
 - \* Removed extra break in constructor declaration with comment (#2429, @Julow)
 - \* De-indent the `object` keyword in class types (#2425, @Julow)
 - \* Consistent formatting of arrows in class types (#2422, @Julow)
+- Reduce the indentation of (polytype) type constraints (#2437, @gpetiot)
```
Julow, line 28: It's a large change
```suggestion
- \* Reduce the indentation of (polytype) type constraints (#2437, @gpetiot)
```

Julow (commented):
I'm not against this change but this causes huge diffs, I think we can take the time to investigate different solutions.

What do you think of this ?
```ocaml
  let t_of_sexp : 'a 'perms.
      (sexp -> 'a) -> (sexp -> 'perms) -> sexp -> ('a, 'perms) t =
```

lib/Fmt_ast.ml
gpetiot: I changed the indent to 4 in the last commit to have a look, looks good to me as well, but 6 was a bit excessive.

Julow: The match and the variable can be removed entirely now.

gpetiot:
> I'm not against this change but this causes huge diffs, I think we can take the time to investigate different solutions.
> 
> What do you think of this ?
> 
> ```ocaml
>   let t_of_sexp : 'a 'perms.
>       (sexp -> 'a) -> (sexp -> 'perms) -> sexp -> ('a, 'perms) t =
> ```

I tried this solution as well (see my last commit), I like it, but doesn't Jane Street want to have the `:` at the beginning of the line in this case?

Julow:
I like the new change :) It's a good time to make it has the formatting of polytypes is changing anyway. I wonder what other think ?

> but doesn't Jane Street want to have the : at the beginning of the line in this case?

Yes, I believe this is controlled by the `break_colon` option and we must not break that.

test/passing/tests/polytypes-default.ml.ref
```diff
@@ -1,25 +1,25 @@
 let t1 : 'a 'b. 'a t -> b t = ()
 
 let t2 :
-      'a 'b.
-      'a t________________________________ ->
-      'b t_______________________________________ =
+    'a 'b.
```
Julow, line 4: The formatting between locally abstract types and `Ptyp_poly` is different:
```
let foo : type a b c.
    a -> b -> c foooooooo
    =
  ()

let foo :
    'a 'b 'c.
    a -> b -> c foooooooo
    =
  ()
```

I don't know whether this is important or not. The former is really part of the let-binding while the later is part of the type expression.

Julow (approved):
What do others think ?

test/passing/tests/polytypes-default.ml.ref
gpetiot: I forgot how the whole thing was designed, one day let's make the time to rewrite these in the parser directly so we can remove the abomination code in Sugar.ml

gpetiot: Do you agree in merging it as-is? we can still improve the other case in a separate PR, or better redesign the types later.

Issue #2092, opened 2022-05-25
Enable formatting projects without an `.ocamlformat` by default


tmattio:
This is a follow-up on https://github.com/ocaml/dune/pull/5768 to continue the discussion on whether ocamlformat should support formatting projects without an `.ocamlformat` configuration file by default.

@gpetiot mentioned in the PR above that he thinks the current behavior should be preserved.
I still wanted to provide a few arguments in favor of supporting the formatting of projects without an `.ocamlformat` file here nonetheless.

One pain point that is not specific to ocamlformat that I'm hoping to alleviate with this is the number of configuration files needed to have a full development environment.

By full development environment, I mean a project where I can execute common developer workflow easily (e.g. formatting the code base).

Not requiring a configuration file for ocamlformat would remove one of these.

My understanding of why we want to enforce a configuration file is to prevent different people from formatting the code with a different version of ocamlformat. I see a few problems with enforcing the configuration file to work around that:

- The version field is not required in `.ocamlformat`. We can see a lot of `.ocamlformat` files in the wild that don't have this field. And so just requiring an `.ocamlformat` file does not solve the above problem.
- The above is only a problem when multiple people work on the project. It's not the case for a lot of projects, and in particular for a beginner that is working on a hello world project.

And to go further, I think removing the need for an `.ocamlformat` file would improve a few scenarios:

- A new user who generates a project with `dune init` a the moment won't be able to format their code. They would have to search which formatter to use, and how to enable it. We can hope that they will stumble upon the dune formatting documentation or the best practices on ocaml.org quickly, but this will still remain a friction point for beginners.
- `ocamlformat`, even if it provides a great degree of flexibility when it comes to the formatting configuration, doesn't encourage users to configure it. Not requiring an `.ocamlformat` file would be a good way to incentivize people to use the default profile.

Let me know what you think 🙂

gpetiot:
I am still not sure this is the best, but what we could do is to also enable for the formatting if we are in one of these cases, the `dune-project` file must define one of these:
- `(using fmt 1.x)` < dune 2.0
- `(using fmt 1.x (enabled_for ocaml [...]))` < dune 2.0
- `(formatting (enabled_for ocaml [...]))` > dune 2.0
- not `(formatting disabled)` > dune 2.0

This way we still have the constraint "we detect we are in a dune project" checked, we just enable it by relying on an explicit stanza in the `dune-project` file instead of an `.ocamlformat` file (that could be potentially empty).

[...]

PR #2439, opened 2023-09-15
Cleanup enable/disable-outside-detected-project options

gpetiot:
First step towards #2092

edit: options that are not "formatting" options cannot be tagged as deprecated


lib-rpc-server/ocamlformat_rpc.ml
```diff
@@ -65,7 +65,7 @@ let run_config conf c =
 let run_path path =
   match
     Bin_conf.build_config ~enable_outside_detected_project:false ~root:None
-      ~file:path ~is_stdin:false
+      ~file:path ~is_stdin:false ~on_save:true
```
Julow, line 68: We might want an option in the protocol to change this (maybe `true` by default though).

lib/bin_conf/Bin_conf.ml
```diff
@@ -207,6 +211,15 @@ let name =
     ~set:(fun name conf -> {conf with name})
     Arg.(value & opt (some string) default & info ["name"] ~doc ~docs ~docv)
 
+let on_save =
+  let doc =
+    "Make OCamlformat not format if no .ocamlformat file could be found. \
+     For code editors."
```
Julow, line 217: It would be nice to mention "project root" and the way the `.ocamlformat` files are searched in some ways.

And mention the intent: (I'm not sure about the wording, please edit)
```suggestion
     Avoid formatting files that shouldn't be when used by text editors to format on each save."
```

Julow (approved):
I am in favor of this !

lib-rpc-server/ocamlformat_rpc.ml
gpetiot: I'm trying to think of a case where we don't want the RPC client (which are editors, dune doesn't use the rpc but directly calls ocamlformat) to use `on_save = true`, it sounds like a bad idea?

panglesd:
So, if I understand correctly, the goal is to
- rename `enable-outside-detected-project` into `on-save` (and invert it)
- make it so that the old option is true by default (false by default for `on-save`)

I think I prefer option names that are self-describing, than the one whose name is for the "intent". `enable-outside-detected-project` feels more self-describing than `on-save`. I personally want my editor to format on save even if there is no detected configuration file, so I would never expect this option to do this.

Also, it makes it impossible to use the latest `ocamlformat.el` emacs mode with an older ocamlformat version.

lib-rpc-server/ocamlformat_rpc.ml
Julow: I think the RPC client definitely wants `on_save = false`. It might be simpler if the RPC options has the same defaults as the main tool indeed.

Julow:
> I think I prefer option names that are self-describing, than the one whose name is for the "intent".

The idea behind the "intent" name is to allow improvements to the defaults and to the on-save workflow independently as well as to remember the intent of the option for ever. We almost forgot the intent behind the current defaults.

I agree the "intent" name is bad as its purpose is simple enough to have a name and users want to use it for other purposes.

Perhaps we can change it to `disable-outside-detected-project` ?
The option has existed in the past but unfortunately cannot be passed in a compatible way. I don't know how to solve the compatibility issue other than making the driver (emacs/vim/vscode plugin) aware of the tool's version.

lib-rpc-server/ocamlformat_rpc.ml
Julow: @rgrinberg suggested that the RPC could have a new error case for when the file hasn't been formatted due to no detected project.

Seems more interesting than returning the input unchanged to me.

gpetiot:
That's what I thought, if we name it `disable-outside-detected-project` we would go full circle. But that name is confusing as well. There is a long explanation in the docs (`ocamlformat --help`) about what it does, how is a project detected (dune-project, .ocamlformat) etc. that is a bit confusing. I personally find `on_save` easier to explain without giving too much details about what ocamlformat will do or check.

For the compatibility, as long as the new option mirrors what the old one does and we keep the old option deprecated long enough it should be enough. We should expect everyone to massively update once we finally release ocamlformat.1.0 🤞

Julow:
> I personally find on_save easier to explain without giving too much details about what ocamlformat will do or check.

It is easier to explain and to maintain over time (eg. we could change the reasons for not formatting later).
But users want to configure their workflows, even the on-save one. The concept of "detected project" cannot be entirely abstract away and must be documented. If the concept is exposed, let's name the option after it.

gpetiot:
Here is our plan:


|.      | default mode | RPC mode |
|-----|--------------|---------|
| currently | disabled outside detected project | disabled outside detected project |
| release N + 1 | disabled outside detected project + warning | disabled outside detected project + warning |
| release N + 2 | enabled outside detected project | disabled outside detected project |

We scrapped the plan to rename the option or add a new one, and will keep `enable-outside-detected-project` and `disable-outside-detected-project` to not make the situation harder.

My last commit displays warnings, and adds the `disable-outside-detected-project` option once again (it was deprecated/removed).

gpetiot:
I removed the warnings and just kept the cleanup.

Julow (approved):
Looks good :) I've added a test and a changelog entry.

PR #2442, opened 2023-09-18
Improve begin/end position in match branches

### LINKED ISSUE #2438: Feature request: require begin to align with end
Given this snippet:
```OCaml
match x with
  | EProj (rcd, fld) -> (
      match infer' env rcd with
      | TRecord (name, rec_ty) -> (
          match List.assoc_opt fld rec_ty with
          | Some ty -> ty
          | _ -> raise (missing_field fld name))
      | ty -> raise (type_error "TRecord" (ty_kind ty)))
```
if i format this after setting `exp-grouping = preserve`, and change the parens to `begin`/`end`, I see
```OCaml
match x with
  | EProj (rcd, fld) -> begin
      match infer' env rcd with
      | TRecord (name, rec_ty) -> begin
          match List.assoc_opt fld rec_ty with
          | Some ty -> ty
          | _ -> raise (missing_field fld name)
        end
      | ty -> raise (type_error "TRecord" (ty_kind ty))
    end
```
Because the `begin` is not aligned with the `end`, it is awkward to see how the `end` is supposed terminate the `match`. There is this form that occurs.
```
  match
  |
  |
end
```
Provide the ability to require that the `begin` align with the `end`, so it can look something like this:
```OCaml
match x with
  | EProj (rcd, fld) ->
      begin match infer' env rcd with
      | TRecord (name, rec_ty) ->
          begin match List.assoc_opt fld rec_ty with
          | Some ty -> ty
          | _ -> raise (missing_field fld name)
          end
      | ty -> raise (type_error "TRecord" (ty_kind ty))
      end
```

gpetiot:
Fix #2438

The same thing would need to be done for if-then-else constructs so maybe some factorization would be in order.

### DIFF
--- a/test/passing/tests/cases_exp_grouping.ml.ref
+++ b/test/passing/tests/cases_exp_grouping.ml.ref
@@ -29,7 +29,7 @@ let _ =
           fooooooooooooo
       | D ->
           fooooooooooooo
-      end
+    end
 [@@ocamlformat "break-cases=nested"]
 
 let _ =
--- a/test/passing/tests/cases_exp_grouping.ml.ref
+++ b/test/passing/tests/cases_exp_grouping.ml.ref
@@ -2,92 +2,101 @@ let _ =
   match x with
   | A -> begin match B with A -> fooooooooooooo end
   | A -> begin match B with A -> fooooooooooooo | B -> fooooooooooooo end
-  | A -> begin
-    match B with
-    | A -> fooooooooooooo
-    | B -> fooooooooooooo
-    | C -> fooooooooooooo
-    | D -> fooooooooooooo
-  end
+  | A ->
+    begin
+      match B with
+      | A -> fooooooooooooo
+      | B -> fooooooooooooo
+      | C -> fooooooooooooo
+      | D -> fooooooooooooo
+    end
 [@@ocamlformat "break-cases=fit"]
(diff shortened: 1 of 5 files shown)


Julow (commented):
This is not an improvement. The previous formatting was intended.

Julow:
I believe #2438 is in fact asking for a new option.

test/passing/tests/cases_exp_grouping.ml.ref
```diff
@@ -2,92 +2,101 @@ let _ =
   match x with
   | A -> begin match B with A -> fooooooooooooo end
   | A -> begin match B with A -> fooooooooooooo | B -> fooooooooooooo end
-  | A -> begin
-    match B with
-    | A -> fooooooooooooo
-    | B -> fooooooooooooo
-    | C -> fooooooooooooo
-    | D -> fooooooooooooo
-  end
```
Julow, line 11: This shows that for many values of `break-cases`, the `end` is aligned with the `|` of the pattern, which is ambiguous.

gpetiot:
I'm not sure having yet-another-option is good in this case. We will do something nice for the diff-friendly profile and hope people stop using the other ones when they notice it.
Anyway closing this PR until we get a plan.

PR #2444, opened 2023-09-18
Sanitize result types

gpetiot:
Some cleanup, I also just realized `let+` and `let*` should have been swapped.


lib/bin_conf/Bin_conf.ml
```diff
@@ -615,21 +615,25 @@ type action =
   | Check of input list
   | Print_config of Conf.t
 
+let ( let* ) = Rresult.( >>= )
+
+let ( let+ ) = Rresult.( >>| )
```
Julow, line 620: We don't need `rresult`: 

- Base has these functions
- We depend on ocaml `>= 4.08`, the stdlib has these functions

```suggestion
let ( let* ) = Result.( >>= )

let ( let+ ) = Result.( >>| )
```

Julow (commented):
> odoc-parser doesn't need result?

It seems that no. Anyway, we depend on ocaml `>= 4.08` and the stdlib has this module.

Julow (approved):
LGTM!

Issue #2406, opened 2023-07-21
Bug: infix operator dropped


hhugo:
when using `ocp-indent-compat = true`
I see the following invalid rewriting.

taken from ocsigen-start.6.1.0:tools/sort_deps.ml

```diff
let add_edge target dep =
  if target <> dep
  then (
    Hashtbl.replace edges dep
-      (target :: (try Hashtbl.find edges dep with Not_found -> []));
+      (target (try Hashtbl.find edges dep with Not_found -> []));
    Hashtbl.replace edge_count target
-      (1 + try Hashtbl.find edge_count target with Not_found -> 0);
+      (1 try Hashtbl.find edge_count target with Not_found -> 0);
    if not (Hashtbl.mem edge_count dep) then Hashtbl.add edge_count dep 0)
```

gpetiot:
I couldn't fix it but the issue is located in https://github.com/ocaml-ppx/ocamlformat/blob/main/lib/Fmt_ast.ml#L1615-L1623

The way the breaks are added is very flimsy

hhugo:
The following diff seems to fix the issue.
```diff
diff --git a/lib/Fmt_ast.ml b/lib/Fmt_ast.ml
index 36d43f5d..d544d3de 100644
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2361,6 +2361,7 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
         if c.conf.fmt_opts.leading_nested_match_parens.v then (false, None)
         else (parenze_exp xpc_rhs, Some false)
       in
+      fmt_opt epi $
       Params.Exp.wrap c.conf ~parens ~disambiguate:true
         (hvbox 2
            ( hvbox 0
```

Julow:
Indeed. I went into the rabbit hole of fixing `~pro`, `~epi` and `~box` in `fmt_expression` in every cases but never got out: https://github.com/Julow/ocamlformat/tree/ocp-indent-op-removed

We might want to use `~pro` instead of `~epi` and remove the later, as its use doesn't correspond to its name and is used in very few cases.

Issue #2416, opened 2023-08-24
Bug: AST changed on nested with constraints


rleshchinskiy:
**Describe the bug**
Running ocamlformat on
```
module N : S with module type T = (U with module M = M) = struct end
```
results in
```
ocamlformat: Cannot process "/tmp/x/U.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: ast changed.
```

**How to Reproduce**
Paste above into a `/tmp/x/U.ml`, run `ocamlformat --enable-outside-detected-project /tmp/x/U.ml`

Julow:
Thanks for the report, should be fixed by https://github.com/ocaml-ppx/ocamlformat/pull/2419

PR #2445, opened 2023-09-19
Remove incorrect `fmt_expression ~epi` mechanism

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2406

This argument doesn't work similarly to the `~epi` argument of other functions. It meant different things depending on the expression:

- `Pexp_match` and `Pexp_apply`: it was similar to `~pro`.
- `Pexp_constant`: like the usual `~epi` but was passed a break.
- all other cases: it was not used.

These rules are now handled in `fmt_args_grouped`, the only place where they made sense. The `~epi` argument is removed.

In the first case, the `~pro` argument is used instead. This requires refactoring `fmt_expression` to make sure that every cases use `~pro` exactly once.


lib/Fmt_ast.ml
```diff
@@ -1714,19 +1724,20 @@ and fmt_expression c ?(box = true) ?pro ?epi ?eol ?parens ?(indent_wrap = 0)
...
-                  ( str "%"
-                  $ hovbox 2
-                      ( fmt_str_loc c name $ str " fun "
-                      $ fmt_attributes c ~suf:" " call.pexp_attributes
-                      $ fmt_fun_args c xargs $ fmt_opt fmt_cstr $ fmt "@ ->"
-                      )
-                  $ fmt "@ " $ fmt_expression c xbody ) )
-           $ fmt "@ ;@ "
-           $ list grps " ;@;<1000 0>" fmt_grp ) )
+      pro
```
Julow, line 1713: Ideally the `pro` would be pushed as far as possible into the boxes (and perhaps the opening parenthese should be appended to the `pro`) and the outer box should be removed.
But this is out of the scope of this PR.

lib/Fmt_ast.ml
```diff
@@ -1454,15 +1458,22 @@ and fmt_args_grouped ?epi:(global_epi = noop) c ctx args =
...
+      | _ -> None
+    in
     let epi =
-      match (lbl, last) with
-      | _, true -> None
-      | Nolabel, _ -> Some (fits_breaks "" ~hint:(1000, -1) "")
-      | _ -> Some (fits_breaks "" ~hint:(1000, -3) "")
+      match ast.pexp_desc with
+      | Pexp_constant _ when not last -> fits_breaks "" ~hint:(1000, -2) ""
+      | _ -> noop
```
Julow, line 1458: I found out that the first rule was useless. Only this one remains.

The `break` code below could probably handle un-grouping `Pexp_constant` better. This introduces hard to control indentation.

Julow: This rule is actually more complex than that: it doesn't apply to every kinds of constants, when `break-string-literals=never` and in other cases that I couldn't find.

The regression are down to +126 -80 and are easily explained: Always break after multi-line string argument.

Julow:
Test branch shows a new regression:
```diff
diff --git a/stdlib/marshal.mli b/stdlib/marshal.mli
index 507af732b3..7aa6859768 100644
--- a/stdlib/marshal.mli
+++ b/stdlib/marshal.mli
@@ -53,7 +53,8 @@ type extern_flags =
   | No_sharing  (** Don't preserve sharing *)
   | Closures  (** Send function closures *)
   | Compat_32  (** Ensure 32-bit compatibility *)
-  | Compression  (** Compress the output if possible
+  | Compression
+      (** Compress the output if possible
                      @since 5.1 *)
 
 val to_channel : out_channel -> 'a -> extern_flags list -> unit
```

lib/Fmt_ast.ml
Julow: The regressions are due to hard to control implicit rules. I suggest making these explicit in https://github.com/ocaml-ppx/ocamlformat/pull/2448 (by introducing more regressions but on purpose)

Julow: This change is merged in https://github.com/ocaml-ppx/ocamlformat/pull/2453 and this PR has no more regressions!

hhugo:
You should add a regression test for #2416

Julow:
Good idea, added.

Julow:
I think this is good to merge.

PR #2446, opened 2023-09-19
Add a page about how the configuration is computed

gpetiot:
Re-purposing some docs I wrote 2 years ago but nothing changed I think.

### DIFF
--- a/doc/configuration.mld
+++ b/doc/configuration.mld
@@ -16,8 +16,10 @@ Ocamlformat fetches the following files on the FS:
 
 {1 How the configuration is built}
 
-+ The initial configuration is equal to the [default] (or [conventional]) profile
-+ The options passed through the configuration files are applied. The configuration files are considered in sequence, starting from the root of the project, and down to the directory of the file to format, overriding one or many options at each application.
+The configuration files are considered in sequence, starting from the root of the project, and down to the directory of the file to format, overriding one or many options at each application.
+
++ The initial configuration is equal to the [default] (or [conventional]) profile.
++ The options passed through the configuration files are applied.
 + The options passed through the [OCAMLFORMAT] environment variable are applied, overriding one or many options at a time.
 + The options passed through the command line are applied, overriding one or many options at a time.
 
--- a/doc/configuration.mld
+++ b/doc/configuration.mld
@@ -2,7 +2,7 @@
 
 {1 Configuration files}
 
-Ocamlformat fetches the following files on the FS:
+Ocamlformat fetches the following files on the file system:
 
 1. [.git], [.hg] or [dune-project]
 2. [.ocamlformat] and [.ocp-indent]
--- a/doc/configuration.mld
+++ b/doc/configuration.mld
@@ -39,7 +39,8 @@ y [@@ocamlformat "option=value,option=value"];;
 
 All "formatting options" (listed in the {{!page-manpage_ocamlformat}manpage}) can be set in attributes, among the non-formatting options only [enable]/[disable] can be set in floatting attributes.
 
-When the option [--enable-outside-detected-project] is set, [.ocamlformat] files outside of the project (including the one in [XDG_CONFIG_HOME]) are read.
+When the option [--enable-outside-detected-project] is set, [.ocamlformat] files outside of the project are read, if no [.ocamlformat] file has been found then then apply the global configuration [$XDG_CONFIG_HOME/ocamlformat] (if defined). The global configuration file is ignore in any other case.
+
 When this option is not set, [.ocamlformat] files outside of the project are ignored.
 
 If no configuration file is found, the formatting is disabled.
--- a/doc/configuration.mld
+++ b/doc/configuration.mld
@@ -26,7 +26,12 @@ The configuration files are considered in sequence, starting from the root of th
 Note that some options can be overriden directly in the source, with attributes like:
 
 {@ocaml[
-[@ocamlformat "option=value,option=value"]
-]}
+(* attributes attached to algebraic constructs *)
+x [@ocamlformat "option=value,option=value"];;
 
-But this is not part of the external configuration.
+(* item attributes, attached to "blocks" *)
+y [@@ocamlformat "option=value,option=value"];;
+
+(* floating attributes, standalone *)
+[@@@ocamlformat "option=value,option=value"]
+]}
--- a/doc/configuration.mld
+++ b/doc/configuration.mld
@@ -10,7 +10,7 @@ Ocamlformat fetches the following files on the FS:
 
 (1.) files are used to determine the {b project root}, which can be overriden by the [--root] option, they are looked up from the path of the file to format, and upwards following parent directories, until the first one is found.
 
-(2.) (3.) and (4.) are looked up from the path of the file to format, and following up the parents up to the project root.
+(2.) and (3.) are looked up from the path of the file to format, and following up the parents up to the project root.
 
 (2.) are the {b configuration files}, they  contain the options used to configure ocamlformat.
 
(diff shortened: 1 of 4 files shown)


doc/configuration.mld
```diff
@@ -0,0 +1,35 @@
+{0 How OCamlFormat computes its configuration}
+
+{1 Configuration files}
+
+Ocamlformat fetches the following files on the FS:
+
++ [.git], [.hg] or [dune-project]
++ [.ocamlformat] and [.ocp-indent] files
++ [.ocamlformat-ignore] files
++ [.ocamlformat-enable] files
```
Julow, line 10: Numbers are meaningful and should be written down. Also merging 3 and 4 that causes repeating text later.
```suggestion
1. [.git], [.hg] or [dune-project]
2. [.ocamlformat] and [.ocp-indent]
3. [.ocamlformat-ignore] and [.ocamlformat-enable]
```

doc/configuration.mld
```diff
@@ -0,0 +1,35 @@
...
+
+(1.) files are used to determine the {b project root}, which can be overriden by the [--root] option, they are looked up from the path of the file to format, and upwards following parent directories, until the first one is found.
+
+(2.) (3.) and (4.) are looked up from the path of the file to format, and following up the parents up to the project root.
```
Julow, line 14:
```suggestion
(2.) and (3.) are looked up from the path of the file to format, and following up the parents up to the project root.
```

doc/configuration.mld
```diff
@@ -0,0 +1,35 @@
...
+
+(2.) are the {b configuration files}, they  contain the options used to configure ocamlformat.
+
+(3.) contain a list of files to exclude from the formatting, if the formatting is enabled ([disable = false]).
+
+(4.) contain a list of files to include in the formatting, if the formatting is disabled ([disable = true]).
```
Julow, line 20: I think we can drop this and go directly to the next section as it doesn't say much and the next section is much more interesting.

doc/configuration.mld
```diff
@@ -0,0 +1,35 @@
...
+
+{1 How the configuration is built}
+
++ The initial configuration is equal to the [default] (or [conventional]) profile
++ The options passed through the configuration files are applied. The configuration files are considered in sequence, starting from the root of the project, and down to the directory of the file to format, overriding one or many options at each application.
```
Julow, line 25: This long sentence should be its own paragraph and placed before the list.

doc/configuration.mld
```diff
@@ -0,0 +1,35 @@
...
++ The options passed through the [OCAMLFORMAT] environment variable are applied, overriding one or many options at a time.
++ The options passed through the command line are applied, overriding one or many options at a time.
+
+Note that some options can be overriden directly in the source, with attributes like:
+
+{@ocaml[
+[@ocamlformat "option=value,option=value"]
+]}
+
+But this is not part of the external configuration.
```
Julow, line 35: "external configuration" is not really a thing and is not needed in this document. I suggest describing what the attributes do instead.

doc/configuration.mld
```diff
@@ -0,0 +1,35 @@
...
+
++ The initial configuration is equal to the [default] (or [conventional]) profile
++ The options passed through the configuration files are applied. The configuration files are considered in sequence, starting from the root of the project, and down to the directory of the file to format, overriding one or many options at each application.
++ The options passed through the [OCAMLFORMAT] environment variable are applied, overriding one or many options at a time.
++ The options passed through the command line are applied, overriding one or many options at a time.
+
+Note that some options can be overriden directly in the source, with attributes like:
+
+{@ocaml[
+[@ocamlformat "option=value,option=value"]
```
Julow, line 32: Example is not working. It should show both of `[@@@ocamlformat ...]` and `<item> [@@ocamlformat ...]` in valid syntax.

The smallest scope attribute `[@ocamlformat ...]` is advanced use but could be mentioned as well.

doc/getting_started.mld
```diff
@@ -75,6 +75,8 @@ Options can be modified by the means of:
 [.ocamlformat] files in the containing and all ancestor directories for each input file are used, as well as the global [.ocamlformat] file defined in [$XDG_CONFIG_HOME/ocamlformat]. The global [.ocamlformat] file has the lowest priority, then the closer the directory is to the processed file, the higher the priority.
 
 When the option [--enable-outside-detected-project] is set, [.ocamlformat] files outside of the project (including the one in [XDG_CONFIG_HOME]) are read. The project root of an input file is taken to be the nearest ancestor directory that contains a [.git] or [.hg] or [dune-project] file.
+For more details, read about {{!page-configuration}how OCamlFormat finds its root project and computes its configuration}.
```
Julow, line 78: I think this sentence should entirely replace the previous three paragraphs. Users only need to know about `.ocamlformat` at the root to get started.

Julow (commented):
Nice!

doc/configuration.mld
```diff
@@ -0,0 +1,34 @@
...
+(4.) contain a list of files to include in the formatting, if the formatting is disabled ([disable = true]).
+
+{1 How the configuration is built}
+
++ The initial configuration is equal to the [default] (or [conventional]) profile
++ The options passed through the configuration files are applied. The configuration files are considered in sequence, starting from the root of the project, and down to the directory of the file to format, overriding one or many options at each application.
++ The options passed through the [OCAMLFORMAT] environment variable are applied, overriding one or many options at a time.
++ The options passed through the command line are applied, overriding one or many options at a time.
+
+Note that some options can be overriden directly in the source, with attributes like:
```
EmileTrotignon, line 28: Is the list of such options documented somewhere ?

gpetiot: I added a note

doc/configuration.mld
```diff
@@ -0,0 +1,45 @@
...
+
+1. [.git], [.hg] or [dune-project]
+2. [.ocamlformat] and [.ocp-indent]
+3. [.ocamlformat-ignore] and [.ocamlformat-enable]
+
+(1.) files are used to determine the {b project root}, which can be overriden by the [--root] option, they are looked up from the path of the file to format, and upwards following parent directories, until the first one is found.
+
+(2.) and (3.) are looked up from the path of the file to format, and following up the parents up to the project root.
+
+(2.) are the {b configuration files}, they  contain the options used to configure ocamlformat. A global [.ocamlformat] file can also be used: [$XDG_CONFIG_HOME/ocamlformat] (if defined).
```
Julow, line 15: A typo:

```suggestion
(2.) are the {b configuration files}, they contain the options used to configure ocamlformat. A global [.ocamlformat] file can also be used: [$XDG_CONFIG_HOME/ocamlformat] (if defined).
```

I think the global configuration should be explained somewhere at the bottom, maybe in a different section. It's also important to mention when it's used and when it's not.

doc/configuration.mld
```diff
@@ -0,0 +1,45 @@
...
+{1 How the configuration is built}
+
+The configuration files are considered in sequence, starting from the root of the project, and down to the directory of the file to format, overriding one or many options at each application.
+
++ The initial configuration is equal to the [default] (or [conventional]) profile.
++ The options passed through the configuration files are applied.
++ The options passed through the [OCAMLFORMAT] environment variable are applied, overriding one or many options at a time.
++ The options passed through the command line are applied, overriding one or many options at a time.
+
+Note that some options can be overriden directly in the source, with attributes like:
```
Julow, line 27: I think this should be in a different section. It's should be more than a note and this section already contain a lot of information.

Julow (approved):
Nice! I have some comments but it's already more than good enough to be merged.

PR #2450, opened 2023-09-21
Consistent spacing between items with extension

Julow:
Structure items with an extension were formatted in a compact way when `module-item-spacing=compact` even though the similar items without an extension would not.

```ocaml
exception%foo X
module%foo M = M
module type%foo S = S
include%foo M

exception X

module M = M

module type S = S

include M
```

This is a problem in future work where the `Pstr_extension` node is no longer used.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ profile. This started with version 0.26.0.
 ### Fixed
 
 - Remove trailing space inside a wrapping empty signature (#2443, @Julow)
+- Fix extension-point spacing in structures (#2450, @Julow)
 
 ## 0.26.1 (2023-09-15)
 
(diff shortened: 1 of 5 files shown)


lib/Ast.ml
```diff
@@ -383,9 +383,11 @@ module Structure_item = struct
        |Pstr_modtype _, Pstr_modtype _
        |Pstr_class _, Pstr_class _
        |Pstr_class_type _, Pstr_class_type _
-       |Pstr_attribute _, Pstr_attribute _
-       |Pstr_extension _, Pstr_extension _ ->
+       |Pstr_attribute _, Pstr_attribute _ ->
           true
+      | ( Pstr_extension ((_, PStr [n1]), _attrs1)
+        , Pstr_extension ((_, PStr [n2]), _attrs2) ) ->
+          allow_adjacent (n1, cI) (n2, cJ)
```
Julow, line 390: `cI` and `cJ` should be recomputed from the extensions' attributes but this is meant to be reverted soon and is not worth it.

CHANGES.md
```diff
@@ -14,6 +14,7 @@ profile. This started with version 0.26.0.
 ### Fixed
 
 - Remove trailing space inside a wrapping empty signature (#2443, @Julow)
+- Fix extension-point spacing in structures (#2450, @Julow)
```
Julow, line 17: I do not consider this a breaking change as it changes very rare code.

PR #2451, opened 2023-09-22
Attributes now stay where the user put them

EmileTrotignon:
```ocaml
module [@attr] M = struct end
```
used to be formatted into 
```ocaml
module = struct end [@@attr]
```
This PR make it so that the user can choose between the two syntaxes, ocamlformat will not change anything.

It contains the first commit of #2247


lib/Ast.ml
```diff
@@ -507,7 +525,9 @@ module Lb = struct
 end
 
 module Mb = struct
-  let has_doc itm = List.exists ~f:Attr.is_doc itm.pmb_attributes
+  let has_doc itm =
+    List.exists ~f:Attr.is_doc
+      (itm.pmb_ext_attrs.attrs_before @ itm.pmb_ext_attrs.attrs_after)
```
Julow, line 530: We might want a `ext_attrs_has_doc` function.

lib/Fmt_ast.ml
```diff
@@ -396,20 +400,23 @@ let docstring_epi ~standalone ~next ~epi ~floating =
...
+(** Accept up to two docstrings, in reverse order of appearance. *)
 let fmt_docstring_around_item' ?(is_val = false) ?(force_before = false)
-    ?(fit = false) c doc1 doc2 =
-  match (doc1, doc2) with
-  | Some _, Some _ ->
-      ( fmt_docstring c ~epi:(fmt "@\n") doc1
-      , fmt_docstring c ~pro:(fmt "@\n") doc2 )
-  | None, None -> (noop, noop)
-  | None, Some doc | Some doc, None -> (
+    ?(fit = false) c docs =
```
Julow, line 413: The new signature is very unsafe. I would prefer to keep the previous one. `extract_doc_attrs` should be the only place these values are produced and already does the required logic.

test/passing/tests/attributes.ml
```diff
@@ -65,7 +65,7 @@ module type M = sig
 
   module T : [%ext] [@test7]
 
-  module T = T [@@test8]
+  module [@test8] T = T
```
Julow, line 68: It would be nice to have both the new and old modules.
- To not suggest it's a regression
- Both are useful

test/passing/tests/js_source.ml.ocp
```diff
@@ -180,12 +180,14 @@ class%foo x = x [@@foo]
...
-module%foo M = M [@@foo]
 
-module%foo rec M : S = M [@@foo]
-and M : S = M [@@foo]
+module%foo [@foo] M = M
+
+module%foo [@foo] rec M : S = M
+and [@foo] M : S = M
+
+module type%foo [@foo] S = S
```
Julow, line 189: This regression is not due to this PR and is being fixed here: https://github.com/ocaml-ppx/ocamlformat/pull/2450

lib/Fmt_ast.ml
```diff
@@ -178,6 +178,10 @@ let update_config_maybe_disabled c loc l f =
   let c = update_config c l in
   maybe_disabled c loc l f
 
+let update_config_maybe_disabled_attrs c loc attrs f =
+  let l = attrs.attrs_before @ attrs.attrs_after in
+  maybe_disabled c loc l f
```
Julow, line 183: This causes the bug in `test/passing/tests/protected_object_types.ml`
```suggestion
  update_config_maybe_disabled c loc l f
```

vendor/parser-extended/docstrings.ml
```diff
@@ -118,6 +118,11 @@ let add_docs_attrs docs attrs =
   in
   attrs
 
+(** {!add_docs_attrs} but operate on [ext_attrs]. *)
+let add_docs_attrs' docs attrs' =
+  let open Parsetree in
+  { attrs' with attrs_before = add_docs_attrs docs attrs'.attrs_before }
```
Julow, line 124: `attrs_before` is for short-syntax attributes ? I'm not sure it makes sense to put docstrings there.

vendor/parser-extended/parser.mly
```diff
@@ -1375,19 +1373,40 @@ structure_item:
...
-  args = functor_args
-  body = module_binding_body
-  attrs2 = post_item_attributes
-    { let docs = symbol_docs $sloc in
-      let loc = make_loc $sloc in
-      let attrs = attrs1 @ attrs2 in
-      let body = Mb.mk name args body ~attrs ~loc ~docs in
-      Pstr_module body, ext }
+  ext_attrs (
+    MODULE,
```
Julow, line 1402: The ext_attrs macro is nice but the large diffs to the parser are problematic. 
Upstream changes more and more often and I don't want backporting to be slowed down.

I'd prefer to keep the rule as before and to change only the `let attrs` line in the body.

vendor/parser-extended/parser.mly
```diff
@@ -1357,12 +1361,6 @@ structure_item:
         { pstr_typext $1 }
     | str_exception_declaration
         { pstr_exception $1 }
-    | module_binding
-        { $1 }
-    | rec_module_bindings
-        { pstr_recmodule $1 }
-    | module_type_declaration
-        { let (body, ext) = $1 in (Pstr_modtype body, ext) }
```
Julow, line 1365: I prefer code to be commented-out than removed in the parser. This greatly helps backporting upstream changes.
If possible with a very short explanation.

Julow (approved):
Very nice!

vendor/parser-extended/docstrings.ml
EmileTrotignon: It is extracted afterward anyways, it does not make a difference, except a moral one. I'm fine either way.

vendor/parser-extended/parser.mly
EmileTrotignon: It is not removed but moved outside of the wrap_mkstr_ext. When this is done, maybe the diff will more legible because wrap_mkstr_ext will not exist anymore

test/passing/tests/attributes.ml
Julow: The test case is changed and I had to check it wasn't changed by ocamlformat. By "old" I meant the test case as it was before your changes.

Julow: The new version of the test case is also useful, we should keep both.

vendor/parser-extended/docstrings.ml
Julow: I want this to change. The intent behind the code is always important and will be interpreted by future contributors.

test.ml
```diff
@@ -0,0 +1,11 @@
+type%foo t += T [@@foo]
```
Julow, line 1: We don't want that file :)

vendor/parser-extended/parser.mly
Julow: This is still here. Did you forget to check something in your last commit ?

lib/Fmt_ast.ml
EmileTrotignon: DOne !

lib/Fmt_ast.ml
EmileTrotignon: I tried to change it, but it a bit unwieldy for fmt_docstring_around_item_attrs lower. I'm fine either way, and I think you wrote this code, it should easier for you to change it.

lib/Ast.ml
```diff
@@ -409,30 +423,40 @@ module Signature_item = struct
...
-    | Psig_recmodule
-        ( {pmd_type= {pmty_attributes= atrs1; _}; pmd_attributes= atrs2; _}
-        :: _ )
+    (* two attribute list *)
+    | Psig_modtype
+        {pmtd_ext_attrs= {attrs_before= atrs1; attrs_after= atrs2; _}; _}
+     |Psig_modtypesubst
+        {pmtd_ext_attrs= {attrs_before= atrs1; attrs_after= atrs2; _}; _}
+     |Psig_modsubst
+        {pms_ext_attrs= {attrs_before= atrs1; attrs_after= atrs2; _}; _}
```
Julow, line 442: Same as in the other module, I'd prefer not to unpack the `ext_attrs` here. This is going to grow and the existing rules need to stay readable.

vendor/parser-extended/parser.mly
Julow: Fine to me.

lib/Fmt_ast.ml
Julow: I'm preparing a patch.

Julow (commented):
Nice work :) A few tweaks and it's ready to merge.

lib/Fmt_ast.ml
Julow: I've pushed this directly to your branch. The change in docstring formatting is reverted.

lib/Ast.ml
```diff
@@ -318,27 +324,35 @@ module Structure_item = struct
...
-     |Pstr_exception
+    (* two attribute lists *)
+    | Pstr_exception
         { ptyexn_attributes= atrs1
         ; ptyexn_constructor= {pext_attributes= atrs2; _}
         ; _ }
-     |Pstr_module
-        {pmb_attributes= atrs1; pmb_expr= {pmod_attributes= atrs2; _}; _} ->
+     |Pstr_modtype
+        {pmtd_ext_attrs= {attrs_before= atrs1; attrs_after= atrs2; _}; _}
```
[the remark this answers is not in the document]
EmileTrotignon: This is done

CHANGES.md
```diff
@@ -133,6 +133,7 @@ profile. This started with version 0.26.0.
 - JaneStreet profile: doesn't align infix ops with open paren (#2204, @gpetiot)
 - Re-use the type let_binding from the parser instead of value_binding, improve the spacing of let-bindings regarding of having extension or comments (#2219, @gpetiot)
 - The `ocamlformat` package now only contains the binary, the library is available through the `ocamlformat-lib` package (#2230, @gpetiot)
+- Modules attributes are now formatted right after the `module` keyword, instead of at the end of the module definition (#2247, @emiletrotignon)
```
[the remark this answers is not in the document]
EmileTrotignon: I think in future patches where we preserve more attributes we will edit the line instead of adding a new one, I updated with something that fits what this PR does.

Julow (approved):
This is good to go, thanks !

There's a small regression related to https://github.com/ocaml-ppx/ocamlformat/pull/2450 but it's not worth blocking on that.

PR #2452, opened 2023-09-26
Update issue template

gpetiot:
- The label of feature requests was incorrect so issues were not tagged.
- `blank_issues_enabled: false` should help reduce spam

I considered rewriting the templates using the yaml format like in https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository that allows to specify different fields, but that's probably not that useful :D

Should we add a template (and a tag) for simple questions? or for style suggestions? for now there is only Bugs vs Feature requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: OCaml Community Forum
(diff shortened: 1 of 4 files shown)


Julow (approved):
LGTM

> Should we add a template (and a tag) for simple questions? or for style suggestions?

I'm personally OK with this.

PR #2455, opened 2023-10-05
Fixes bug in bc91e42246a7fe184377ab650f66f629e8e378be

EmileTrotignon:
This fixes a bug that was introduced in commit bc91e42246a7fe184377ab650f66f629e8e378be (pr #2451 )

The test shown was formatted the following way : 
```ocaml
module rec A : A = struct end

(** about b *)
and  (** floatting *) B : B = struct end
```


test/passing/tests/rec_str_item_comment.ml
```diff
@@ -0,0 +1,10 @@
+module rec A : A =
```
Julow, line 1: This would fit nicely in the existing test `comment_in_modules.ml` instead of a new one.

Julow (commented):
Nice detective work!

You can mention this PR in the same change entry as https://github.com/ocaml-ppx/ocamlformat/pull/2451

Julow (approved):
Looks good, thanks! I'll merge when the CI is green.

PR #259, opened 2018-08-02
Improve: add a flag to skip all checks about comments

hhugo:
This is (unsafe and) useful to automatically format files triggering w50.

PR #2456, opened 2023-10-05
Better ignore doc comment

### LINKED ISSUE #2408: Feature request: improve the behavior of `--no-comment-check`
Ideally, `--no-comment-check` should not complain if an unattached doc comment (warning 50) becomes attached.
The current behavior is that we see an "Ast changed" error because there is an extra "doc" atttribute corresponding to that comment.

hhugo:
Fix #2408
Floating doc comment were not ignored when using `--no-comment-check`, only attached doc-comment were.

This change can result in empty `Ptop_def []` which are now ignored while mapping.

For the record, I've tested this PR on the all sources inside opam-respository
```
find ../all-sources -name "*.ml" -type f -o -name "*.mli" -type f | parallel --progress _build/default/bin/ocamlformat/main.exe --enable-outside-detected-project -q --no-comment-check --no-version-check --ignore-invalid-option > /dev/null
```

I don't see any `Ast changed` bug anymore

### DIFF
--- a/lib/Normalize_std_ast.ml
+++ b/lib/Normalize_std_ast.ml
@@ -152,10 +152,50 @@ let make_mapper conf ~ignore_doc_comments =
     let typ = {typ with ptyp_loc_stack= []} in
     Ast_mapper.default_mapper.typ m typ
   in
+  let structure =
+    if ignore_doc_comments then fun (m : Ast_mapper.mapper) l ->
+      List.filter l ~f:(function
+        | {pstr_desc= Pstr_attribute a; _} -> not (is_doc a)
+        | _ -> true )
+      |> Ast_mapper.default_mapper.structure m
+    else Ast_mapper.default_mapper.structure
+  in
+  let signature =
+    if ignore_doc_comments then fun (m : Ast_mapper.mapper) l ->
+      List.filter l ~f:(function
+        | {psig_desc= Psig_attribute a; _} -> not (is_doc a)
+        | _ -> true )
+      |> Ast_mapper.default_mapper.signature m
+    else Ast_mapper.default_mapper.signature
+  in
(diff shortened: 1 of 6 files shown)


hhugo:
The purpose of `--no-comment-check` is to ignore comments and doc-comments !

Here is what the ocamlformat doc says

> Control whether to check comments and documentation comments. Unsafe to turn off. May be set in .ocamlformat. The flag is set by default.

The flag was introduce for debug/dev purposes so that one case focus on `Ast changed` errors. I use this in #985

Julow:
Doc comments are part of the AST and I don't consider them problematic like comments. I think doc comments issues are of higher priority than other `Ast changed` errors.
I'd prefer to fix these first rather than to ignore them.
What do you think?

hhugo:
Please looks at what's done currently. Please look at the documentation for this option. This option is not supposed to be used by regular users. It currently already ignores some doc-comments. This PR just make the behavior consistent.

I don't really understand your pushback on this developers option. 

This change allows to run ocamlformat on files triggering warning 50 and spot `Changed ast` without the noise caused by comments.

hhugo:
Also see PR introducing the feature https://github.com/ocaml-ppx/ocamlformat/pull/259

hhugo:
This PR is restoring the feature introduced in #259 that was altered at some point later

hhugo:
The regression was introduced by https://github.com/ocaml-ppx/ocamlformat/pull/1672
The following statement was invalid

> or double checking things like ignore_doc_comments (checked by the attributes method anyway), maybe these were considered to be optimizations but I didn't observe any slowdown;

Julow:
Alright, you convinced me it's useful, let's merge.

PR #2459, opened 2023-10-06
Fixed bug with attributes on sub-expressions of infix operators.

### LINKED ISSUE #2458: Bug: Attributes on sub-expressions of n-ary infix operators are dropped.
**Describe the bug**
If you have adjacent infix operators of the same precedence (level and direction), and attach an attribute to the sub-expression that would be evaluated first by precedence direction, then the attribute gets dropped, causing a "BUG: ast changed." error.

**How to Reproduce**
Example program (either of these lines triggers the bug):
```
(* Left-associative *)
let _ = (x + y) [@attr] + z

(* Right-associative *)
let _ = x ^ (y ^ z) [@attr]
```

Error:
```
ocamlformat: Cannot process "a.ml".
  BUG: ast changed.
```

This happens regardless of option choices.

Ocamlformat version: commit d664fa2254479195f9fe0099c5847cb79299d17f

tdelvecchio-jsc:
Fixes #2458.

The issue is that sugaring of n-ary infix operators doesn't keep the original sub-expression around, which carries the attribute with it. This fix just prevents sugaring of n-ary infix operations if an attribute is reached.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ profile. This started with version 0.26.0.
 - Fix extension-point spacing in structures (#2450, @Julow)
 - \* Consistent break after string constant argument (#2453, @Julow)
 - Fix invalid syntax generated with `ocp-indent-compat` (#2445, @Julow)
+- Fixed bug with attributes on sub-expressions of infix operators (#2459, @tdelvecchio-jsc)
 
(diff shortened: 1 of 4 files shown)


gpetiot (approved):
I added an entry in the changelog, LGTM, thanks!

PR #2460, opened 2023-10-06
Janestreet profile: do not break `fun _ -> function`

tdelvecchio-jsc:
In Jane Street's production version of ocamlformat, we format a program as such:
```
let _ =
  List.mapi my_list ~f:(fun idx -> function
    | Some x -> x
    | None -> idx)
;;
```

When we pass this through the current main branch of ocaml-ppx with the janestreet profile, it is instead styled as:
```
let _ =
  List.mapi my_list ~f:(fun idx ->
      function
      | Some x -> x
      | None -> idx)
;;
```

This PR adds an option to revert the janestreet profile to the old behavior.


gpetiot (approved):
I took the liberty of simplifying the code, since we don't need a parameter for that: we just remove the special handling of ocp-indent-compat for `Pexp_function`, which is now handled the same as `Pexp_fun` and `Pexp_newtype`.

I also added an entry in the changelog.

LGTM now, thanks!

PR #2463, opened 2023-10-06
Fix cinaps comment formatting to not change multiline string contents.

### LINKED ISSUE #2462: Bug: Contents of multi-line strings in cinaps comments are changed.
**Describe the bug**
When a cinaps comment has its indentation changed, the entire contents are indented. This leads to a change in program semantics if there is a multi-line string in the cinaps comment.

**How to Reproduce**
Any profile (below example uses default profile).

Unstyled file:
```
(*$                                    
  let x = String.find y in
  let z = {|
   aaa (* <-- 3 spaces *)
 |} in
  ()
*)

let () =
  Aaaaaaaa.Aaaaaaaa.aaaa
    ~aaaa
    (*$
      {| aaaa
       aaaa |} (* <-- 7 spaces *)
    *)
    ~aaaa
```

Styled file:
```
(*$
  let x = String.find y in
  let z = {|
    aaa (* <-- 3 spaces *)
  |} in
  ()
*)

let () =
  Aaaaaaaa.Aaaaaaaa.aaaa
    ~aaaa
      (*$
        {| aaaa
         aaaa |} (* <-- 7 spaces *)
      *)
    ~aaaa
```

Ocamlformat version: commit d664fa2254479195f9fe0099c5847cb79299d17f

tdelvecchio-jsc:
Fixes #2462.

Rather than formatting comment code blocks into a string, and then converting that string back into an `Fmt.t`, we instead just generate the `Fmt.t` directly, which allows the styling to know about current indentation without needing to manually indent/unindent the formatted text.

This change has some rough edges to it (with the addition of `set_margin` as an argument to `fmt_code`, in particular), so any feedback would be greatly appreciated!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ profile. This started with version 0.26.0.
 - Fix extension-point spacing in structures (#2450, @Julow)
 - \* Consistent break after string constant argument (#2453, @Julow)
 - Fix invalid syntax generated with `ocp-indent-compat` (#2445, @Julow)
+- \* Fix cinaps comment formatting to not change multiline string contents (#2463, @tdelvecchio-jsc)
 
 ## 0.26.1 (2023-09-15)
 
--- a/lib/Fmt_odoc.ml
+++ b/lib/Fmt_odoc.ml
@@ -14,7 +14,11 @@ open Odoc_parser.Ast
 module Loc = Odoc_parser.Loc
 
 type fmt_code =
-  Conf.t -> offset:int -> string -> (string, [`Msg of string]) Result.t
+     Conf.t
+  -> offset:int
+  -> set_margin:bool
+  -> string
+  -> (Fmt.t, [`Msg of string]) Result.t
 
 type c = {fmt_code: fmt_code; conf: Conf.t}
 
@@ -119,8 +123,8 @@ let fmt_code_block c s1 s2 =
   match s1 with
   | Some ({value= "ocaml"; _}, _) | None -> (
     (* [offset] doesn't take into account code blocks nested into lists. *)
-    match c.fmt_code c.conf ~offset:2 original with
-    | Ok formatted -> fmt_code formatted
+    match c.fmt_code c.conf ~offset:2 ~set_margin:true original with
+    | Ok formatted -> formatted |> Format_.asprintf "%a" Fmt.eval |> fmt_code
     | Error (`Msg message) ->
         ( match message with
         | "" -> ()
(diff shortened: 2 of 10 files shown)


gpetiot (approved):
I added an entry in the changelog.

Everything looks good to me, thanks! Before merging, @Julow let's just confirm there is nothing conflictual with what you had planned for Jane Street in #2314 or #2371

lib/Fmt_odoc.ml
```diff
@@ -14,7 +14,11 @@ open Odoc_parser.Ast
 module Loc = Odoc_parser.Loc
 
 type fmt_code =
-  Conf.t -> offset:int -> string -> (string, [`Msg of string]) Result.t
+     Conf.t
+  -> offset:int
+  -> set_margin:bool
```
Julow, line 19: We might get rid of the `set_margin` by not using `Fmt.eval` when formatting code blocks in docstrings. That might require some time so I suggest to do it later and not to block this PR. The extra argument is not too bad as a temporary solution.

test/passing/tests/cinaps.ml.ref
```diff
@@ -22,7 +22,7 @@ let y = 2
   #use "import.cinaps" ;;
 
   List.iter all_fields ~f:(fun (name, type_) ->
-      printf "\nexternal get_%s\n: unit -> %s = \"get_%s\"" name type_ name )
+      printf "\nexternal get_%s\n  : unit -> %s = \"get_%s\"" name type_ name )
```
Julow, line 25: This might be intended but I'd like confirmation. The original comment is:
```ocaml
(*$ ;; #use "import.cinaps"

  ;; List.iter all_fields ~f:(fun (name, type_) -> printf "\nexternal get_%s
  : unit -> %s = \"get_%s\"" name type_ name) *)
```
Are the two spaces on the last line part of the string constant ?
It was my understanding that cinaps would strip any indentation before parsing the code (like Odoc does with code blocks).

Julow (approved):
I have a question but otherwise good to go in :)

CHANGES.md
```diff
@@ -19,6 +19,7 @@ profile. This started with version 0.26.0.
 - \* Consistent break after string constant argument (#2453, @Julow)
 - Fix invalid syntax generated with `ocp-indent-compat` (#2445, @Julow)
 - Fixed bug with attributes on sub-expressions of infix operators (#2459, @tdelvecchio-jsc)
+- \* Fix cinaps comment formatting to not change multiline string contents (#2463, @tdelvecchio-jsc)
```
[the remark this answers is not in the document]
gpetiot: I mean this PR changed the formatting of cinaps comments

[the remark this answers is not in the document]
Julow: I don't think so. The new output looks more like the input. I don't think this patch would change already formatted code.

test/passing/tests/cinaps.ml.ref
tdelvecchio-jsc: Yes, they are part of the string constant:
```
(*$ ;; open! Core
  ;; let all_fields = ["a", "b"; "c", "d"]

  ;; List.iter all_fields ~f:(fun (name, type_) -> printf "\nexternal get_%s
  : unit -> %s = \"get_%s\"" name type_ name) *)
external get_a
  : unit -> b = "get_a"
external get_c
  : unit -> d = "get_c"(*$*)
```

Issue #2401, opened 2023-07-17
Backport upstream parser changes about function arity


Julow:
https://github.com/ocaml/ocaml/pull/12236 has been merged in ocaml and contains changes to the parsetree.
The changes are helping the work on the CST (https://github.com/ocaml-ppx/ocamlformat/pull/2213).

This requires some more changes: (thanks @ncik-roberts for the heads up)

- `fun x -> fun y ->` must no longer be rewritten to `fun x y ->`, as the two have different semantics.
- `fun x -> function y -> x + y` is also different from `fun x -> (function y -> x + y)`. so parens around function are now meaningful.

@ncik-roberts adds:

In the case where the programmer writes `fun x -> fun y -> x + y`, it might be nice to visually exaggerate the fact that the inner function is separate now, with something like `fun x -> (fun y -> x + y)`. (Now that this PR is merged, `fun x -> fun y -> x + y` has different runtime characteristics than `fun x y -> x + y`, where the former is an arity-1 function returning an arity-1 function, and the latter is an arity-2 function.)

Julow:
Done in https://github.com/ocaml-ppx/ocamlformat/pull/2544

PR #2467, opened 2023-10-23
Remove unnecessary Sugar.polynewtype

gpetiot:
@Julow I was surprised that this code had no impact, I couldn't find a regression example, it looks like the parser never introduces this pattern, but a second review to confirm would be useful!

This makes #2401 easier, since there are heavy changes on the poly/newtypes constructors, and this code is a headache to maintain.

### DIFF
--- a/lib/Sugar.ml
+++ b/lib/Sugar.ml
@@ -201,42 +201,6 @@ let mod_with pmty =
   let l_rev, m = mod_with_ pmty in
   (List.rev l_rev, m)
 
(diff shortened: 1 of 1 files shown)


Julow (approved):
This code is indeed unnecessary since https://github.com/ocaml-ppx/ocamlformat/pull/2412
`polynewtype` was called when formatting let-op bindings and I don't think the AST it matches is outputted by the parser.
Let's merge!

Issue #2468, opened 2023-10-25
Bug: Dropped comment in some cinaps code.


tdelvecchio-jsc:
**Describe the bug**
The following test causes ocamlformat to exit with code 1, and an error about a comment being changed:

```
(*$
  let _ =
    [ x (*
      *)
    ; y
    ]
  ;;
*)
```

ocamlformat version: 181da779ad5cb6a19014b3c26cf13a0d244e9e4f (ocaml-ppx/main as of post creation)
command: `dune exec bin/ocamlformat/main.exe -- example.ml`

Julow:
This is fixed by https://github.com/ocaml-ppx/ocamlformat/pull/2371

PR #2469, opened 2023-11-02
Fix bug multiline comment in cinaps

tdelvecchio-jsc:
Fixes https://github.com/ocaml-ppx/ocamlformat/issues/2468.

Previously, if a comment got reformatted within cinaps code, the check that comments are preserved would fail. Normalization logic exists to allow formatting of code in cinaps comments, but normalization did not descend into comments within cinaps code.

This patch fixes the issue and adds tests for it.

Note: this issue is also fixed by https://github.com/ocaml-ppx/ocamlformat/pull/2371, so likely doesn't need to be merged, but I am putting this here in case that pull request gets blocked.


Julow:
The normalization of comments changed in https://github.com/ocaml-ppx/ocamlformat/pull/2371 and this is no longer needed to fix the issue.
Thanks!

PR #2476, opened 2023-11-06
Fix Ppat_cons parenthesization in function parameter position

### LINKED ISSUE #2475: Bug: Error on Ppat_cons in function parameter
**Describe the bug**
When a Ppat_cons appears as the pattern for a function parameter, ocamlformat currently prints it with no parentheses, which is a syntax error.

**How to Reproduce**
Tip of ocaml-ppx/main as of issue creation (d43aba374e58d0087882862f5b0a9f284945649e).
```$ echo "let f (x :: y) = x" | dune exec -- bin/ocamlformat/main.exe - --impl```

tdelvecchio-jsc:
Fixes #2475.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -1890,6 +1890,7 @@ end = struct
       | Ppat_cons _ -> true
       | Ppat_construct _ | Ppat_record _ | Ppat_variant _ -> false
       | _ -> true )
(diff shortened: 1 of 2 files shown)


gpetiot (approved):
LGTM, thanks!

This has been introduced in #2471, so no need to add a changelog entry, but I added the test case from the issue.

PR #2477, opened 2023-11-08
WIP: fix conflict over the Ocaml_common library

### LINKED ISSUE #2441: Feature request: using ocamlformat as library together with ppxlib
I am trying to use ocamlformat as a library, mainly adding `ocamlformat-lib` to dune `libraries`, and calling `Translation_unit.parse_and_format` to have some strings formatted (side note: I wish I could send a parse tree directly, but I was not able find any function that operates at that lower level).

There seems to be some issues when I try to use the library in executables that also depend on ppxlib, either indirectly, like here with `reason`:

```
Error: The files /home/me/code/rebind/_opam/lib/ocamlformat-lib/ocamlformat_lib__Translation_unit.cmi
       and src/js/.rebind_js.eobjs/byte/dune__exe__Reason_utils.cmi
       make inconsistent assumptions over interface Ocaml_common
```

or when the executable directly depends on ppxlib itself:

```
Error: The files /home/me/code/rebind/_opam/lib/ppxlib/ast/ppxlib_ast__Versions.cmi
       and /home/me/code/rebind/_opam/lib/ocamlformat-lib/ocamlformat_lib__Translation_unit.cmi
       make inconsistent assumptions over interface Ocaml_common
```

Is there a way to work around this problem? Thanks!

gpetiot:
Will fix #2441 (not yet)

I tried to change the public name of the `ocaml-common` library vendored in ocamlformat, no success so far, I added a test showing that.

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -28,7 +28,7 @@
   (backend bisect_ppx))
  (libraries
   format_
-  ocaml_common
+  ocamlformat-lib._ocaml-common
   parser_standard
   parser_extended
   ocamlformat_stdlib
--- a/test/install/ocamlformat-lib/dune
+++ b/test/install/ocamlformat-lib/dune
@@ -1,0 +1,9 @@
+(executable
+ (name foo)
+ (libraries ocamlformat-lib ppxlib))
+
+(rule
+ (alias runtest)
+ (deps ./foo.exe)
+ (action
+  (run %{deps})))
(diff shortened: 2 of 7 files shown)


hhugo:
I would expect the private name to be  the issue, not the public one.

Julow:
I think the only fix is to rename all the libraries, which I did in https://github.com/ocaml-ppx/ocamlformat/pull/2481

gpetiot:
Replaced by #2481

PR #2480, opened 2023-11-15
Always treat multiline `{|delimited strings|}` as though they are long

dvulakh:
Currently, ocamlformat determines whether a delimited string is "long" enough to split a box across multiple lines simply by its length. However, a multiline string should always be treated as long enough to split a box, even if it doesn't have so many characters. This feature implements the desired behavior.

We implement the fix by exposing `str_as` in `Fmt.mli` so that we can use it to append short strings that get treated as very long after multiline strings.


test/passing/tests/crlf_to_crlf.ml.ref
```diff
@@ -1,4 +1,5 @@
-let _ = {|
+let _ =
+  {|
```
Julow, line 2: The previous output was expected. We call this "docking" in other contextes.

test/passing/tests/extensions-indent.ml.ref
```diff
@@ -185,11 +185,13 @@ let this_function_has_a_long_name plus very many arguments =
 
 [%%expect {||}] ;;
 
-[%expect {|
+[%expect
+     {|
```
Julow, line 189: Same as with `let`, the previous output is expected.

test/passing/gen/gen.ml
```diff
@@ -89,8 +89,10 @@ let register_file tests fname =
 let cmd should_fail args =
   let cmd_string = String.concat " " args in
   if should_fail then
-    spf {|(with-accepted-exit-codes 1
-       (run %s))|} cmd_string
+    spf
+      {|(with-accepted-exit-codes 1
+       (run %s))|}
+      cmd_string
```
Julow, line 95: This output can be achieved by tweaking `fmt_args_grouped` instead of adding breaks as experimented in https://github.com/ocaml-ppx/ocamlformat/pull/2448

lib/Fmt_ast.ml
```diff
@@ -261,8 +261,11 @@ let fmt_constant c ?epi {pconst_desc; pconst_loc= loc} =
       str lit $ opt suf char
   | Pconst_char (_, s) -> wrap "'" "'" @@ str s
   | Pconst_string (s, loc', Some delim) ->
-      Cmts.fmt c loc'
-      @@ wrap_k (str ("{" ^ delim ^ "|")) (str ("|" ^ delim ^ "}")) (str s)
+      Cmts.fmt c loc' @@ str (Format_.sprintf "{%s|%s|%s}" delim s delim)
+      (* If a multiline string has newlines in it, it should get treated as a
+         "long" box element. To do so, we append a length-1000 empty
+         string. *)
+      $ fmt_if_k (String.mem s '\n') (str_as 1000 "")
```
Julow, line 268: It's generally not a good idea to put a break at the end of a box as it interferes with the formatting of the containing expression:
```ocaml
let _ =
  {|
    foo
    bar
  |}
  ;
  ""
```

Instead, we write rules that can be followed when formatting the containing expression. For example, `is_simple`, `break_between`, and more defined in `Ast`. This is more or less done in `fmt_args_grouped`.

test/passing/tests/crlf_to_crlf.ml.ref
dvulakh: I'll restrict the change to the janestreet profile, then. It is surprising to me, though, that the property used to determine whether to "dock" is the number of characters in the string and not something that more closely relates to how it "looks", like whether the first line is empty, the number of lines, or the length of the longest line.

lib/Fmt_ast.ml
dvulakh: Thanks for pointing this out, I'll take a look at the PR you linked above and try to come up with a fix.

dvulakh: It turns out that, as far as I can tell, semicolons after multiline strings still format fine in the janestreet profile.

I have pushed a commit that restricts the new behavior to the janestreet profile, though.

test/passing/tests/crlf_to_crlf.ml.ref
Julow: It's also surprising to me. I tried various rules in https://github.com/ocaml-ppx/ocamlformat/pull/2448 without success, as the changes to existing code is always large and the rule is always arbitrary. Thanks for the ideas, I'll try to come back to this change later.

test/passing/tests/js_source.ml.ocp
```diff
@@ -3154,7 +3154,8 @@ module FM_valid = F (struct
     type t = int
   end)
 
-[%%expect {|
+[%%expect
+  {|
```
Julow, line 3158: The docking behavior could be specially supported again if needed.

Julow (approved):
Let's merge :)

PR #2481, opened 2023-11-24
Prefix all vendored libraries

Julow:
This fixes https://github.com/ocaml-ppx/ocamlformat/issues/2441 by making sure that all libraries are prefixed with `ocamlformat_`.

The vendored libraries need some cleanup (eg. merge ocamlformat-stdlib, ocaml-common and ocamlformat_support) but I plan to do that in an other PR.


gpetiot (approved):
LGTM, thanks!
We don't need to add any changelog entry if the public names of the library didn't change, right?

By the way I added a test in #2477 (test/install/ocamlformat-lib/), should we add it to this PR?

gpetiot:
I added the test, feel free to ignore this last commit if you think that's not necessary.

Julow:
The test was very useful to validate the work but I'd prefer not adding more test-dependencies so I removed.
Thanks!

PR #2486, opened 2023-12-07
Add type constraint to binding_op

gpetiot:
Contributing to #2401 

The main positive is removing `Sugar.type_cstr` (all this code is inconvenient for the polynewtypes shenanigans) by doing this work in the parser.

A minor side effect is that we don't add useless parentheses around:
 - `let* ({ state; server; pool } as t) = xxx`
 - `let+ (Ok x) = xxx`

It's a fix, but not important enough to add a line in the changelog.


gpetiot:
I think I fixed all regressions, test_branch doesn't report any diff.

Julow (approved):
Looks good :)

PR #2487, opened 2023-12-10
`enable_outside_detected_project=true` for rpc

kohlivarun5:
The rpc api doesn't allow for enabling `ocamlformat` outside detected project. 
In situations where we want the editor (`ocaml-lsp`) to respect the global config (in `~/.config/.ocamlformat`), it is useful to enable formatting without having a `.ocamlformat` in the project

This change enables it for `rpc`, with the assumption that if a client is triggering the `rpc` explicitly for formatting, they expect it to be picked up 

Related to https://github.com/ocaml-ppx/ocamlformat/issues/2092

### DIFF
--- a/lib-rpc-server/ocamlformat_rpc.ml
+++ b/lib-rpc-server/ocamlformat_rpc.ml
@@ -64,7 +64,7 @@ let run_config conf c =
 
 let run_path path =
   match
-    Bin_conf.build_config ~enable_outside_detected_project:false ~root:None
+    Bin_conf.build_config ~enable_outside_detected_project:true ~root:None
       ~file:path ~is_stdin:false
   with
   | Ok _ as ok -> ok


Julow (commented):
The lsp doesn't know whether formatting should be enabled or not and this will format files that shouldn't be in projects that do not use ocamlformat.

`false` is the right default for now as we decided not to implement https://github.com/ocaml-ppx/ocamlformat/issues/2092 yet (related https://github.com/ocaml-ppx/ocamlformat/pull/2439).
Instead, this should be an option that the lsp could be configured to pass or the lsp could handle specially projects that do not have a `.ocamlformat`.

kohlivarun5:
> Instead, this should be an option that the lsp could be configured to pass or the lsp could handle specially projects that do not have a .ocamlformat.

Is it possible for the lsp to control this on the client side ? 
My understanding of the code was that this option is not exposed and just hard-coded

Julow:
The lsp cannot pass this option at the moment and that's what we should fix.

PR #2488, opened 2023-12-13
Fix indentation of tuples in payloads

Julow:
Extracted from https://github.com/ocaml-ppx/ocamlformat/pull/2314 as it changes the default profile.


gpetiot (approved):
Looks good to me  (and test_branch output as well), thanks!

Let's add a changelog entry and it will be ready to go!

PR #2500, opened 2023-12-21
Factorize pretty printing of class_infos, open_infos and include_infos

gpetiot:


Julow (commented):
Let's be careful that we need to backport changes from upstream into this file.

PR #2503, opened 2023-12-22
Distinguish fun_param for expr/class, unify parsing of val/newtype fun_param

gpetiot:
Part of #2401 
- the types are better because #2401 didn't go as far as improving the class parameters.
- merge the fun_param and the Pexp_newtype rules in the parser for the expr_fun_param
- split the formatting of fun_params in Fmt_ast
- the normalization of the Pexp_newtype parameters (`(type a) (type b)` becoming `(type a b)`) is moved from Sugar to Extended_ast (where the normalizations already are)
- improve the location of comments

@Julow  what do you think, can you have a look at the types in the parsetree (we don't have much freedom, as almost all types are mutually recursive, but maybe I missed something) and the normalization?

### DIFF
--- a/vendor/parser-extended/parser.mly
+++ b/vendor/parser-extended/parser.mly
@@ -68,6 +68,7 @@ let mkstr ~loc d = Str.mk ~loc:(make_loc loc) d
 let mkclass ~loc ?attrs d = Cl.mk ~loc:(make_loc loc) ?attrs d
 let mkcty ~loc ?attrs d = Cty.mk ~loc:(make_loc loc) ?attrs d
 let mkconst ~loc c = Const.mk ~loc:(make_loc loc) c
+let mkfunparam ~loc x = { pparam_loc = make_loc loc; pparam_desc = x }
 
 let pstr_typext (te, ext) =
   (Pstr_typext te, ext)
@@ -858,6 +859,8 @@ The precedences must be listed from low to high.
     { mkcf ~loc:$sloc $1 }
 %inline mkclass(symb): symb
     { mkclass ~loc:$sloc $1 }
+%inline mkfunparam(symb): symb
+    { mkfunparam ~loc:$sloc $1 }
 
 %inline wrap_mkstr_ext(symb): symb
     { wrap_mkstr_ext ~loc:$sloc $1 }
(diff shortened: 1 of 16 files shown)


vendor/parser-extended/parser.mly
```diff
@@ -2266,10 +2269,8 @@ expr:
         Pexp_letopen(od, $7), $4 }
   | FUNCTION ext_attributes match_cases
       { Pexp_function $3, $2 }
-  | FUN ext_attributes fun_param fun_def
+  | FUN ext_attributes expr_fun_param fun_def
       { Pexp_fun($3, $4), $2 }
-  | FUN ext_attributes LPAREN TYPE lident_list RPAREN fun_def
-      { (mk_newtypes ~loc:$sloc $5 $7).pexp_desc, $2 }
```
Julow, line 2272: This diff might be hard to maintain, I think I'd prefer some code duplication instead.

vendor/parser-extended/parsetree.mli
```diff
@@ -488,38 +488,29 @@ and binding_op =
...
   {
     pparam_loc : Location.t;
-    pparam_desc : function_param_desc;
+    pparam_desc : 'a;
   }
 
+and expr_function_param =
+  [ `Param_val of param_val | `Param_newtype of param_newtype ] function_param
+
+and class_function_param = [ `Param_val of param_val ] function_param
```
Julow, line 512: To reduce the diff a little:

```suggestion
and expr_function_param =
  param_val_or_newtype function_param

and class_function_param = param_val function_param
```

Julow (approved):
More of Sugar that can be removed ! Perhaps this could be proposed upstream too ?

vendor/parser-extended/parsetree.mli
gpetiot: I modified the types in 69f1210, let me know what you think

vendor/parser-extended/parser.mly
gpetiot: Theses cases are already merged upstream, there are new rules like `fun_param_as_list` and `fun_params`. I will try to get closer to these parser rules.

vendor/parser-extended/parsetree.mli
Julow: The new types look good!

Issue #2504, opened 2024-01-02
Bug: Non-stabilization caused by comment between functor name and arg.


tdelvecchio-jsc:
**Describe the bug**
Ocamlformat fails to stabilize in the very particular case shown below. Changing the module type of `M` to `(M : S)` does lead to stabilization, as well as if you add more arguments to the functor before or after `M`.

**Reproduction**
Ocamlformat version: https://github.com/ocaml-ppx/ocamlformat/commit/fff20162dfdd36b76f4db60689acab799e9d616e (ocaml-ppx/ocamlformat:main as of Jan 2, 2024)

```
$ dune exec bin/ocamlformat/main.exe -- --impl - <<EOF
module F (* test *) (M : sig type t end) : S = struct end
EOF
main: Cannot process "<standard input>".           
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: formatting did not stabilize after 10 iterations.
EXIT STATUS 1

$ dune exec bin/ocamlformat/main.exe -- --intf - <<EOF
module F (* test *) (M : sig type t end) : S
EOF
main: Cannot process "<standard input>".          
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: formatting did not stabilize after 10 iterations.
EXIT STATUS 1
```

Julow:
Sorry for the slow reply, it's holiday season here. This should be fixed by https://github.com/ocaml-ppx/ocamlformat/pull/2506

PR #2506, opened 2024-01-08
Fix unstable comment around docked functor arg

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2504


Julow:
No diff with test_branch, let's merge!

PR #2507, opened 2024-01-11
Fix unwanted alignment after comment

Julow:
This was found when looking at the regressions of https://github.com/ocaml-ppx/ocamlformat/pull/2314

This removes the alignment in expressions like:

    let _ = try_with (fun () -> (* comment before *)
                                a ; b (* after b *) )

But this also adds regressions. This used to fit on one line:

    try%lwt
      (* fooooooooooo *)
      Lwt.return 2
    with _ -> assert false

I don't think it's worth supporting this again as the inline comment reduces readability and would break again with slight modifications of the code.

Comments after the `then` keyword now break:

    let foo =
      if cmp < 0 then
        (* ast higher precedence than context: no parens *)
        false
      else if cmp > 0 then
        (* context higher prec than ast: add parens *)
        true
      else ...

This could be supported again if needed. It's changed because it depended on implicit rules.


test/passing/tests/comments-no-wrap.ml.ref
```diff
@@ -252,8 +252,11 @@ let%map
     _ =
   ()
 
-type t = < (* a *)
-         a: int [@atr] (* b *) ; b: int (* c *) >
+type t =
+  < (* a *)
+  a: int [@atr] (* b *)
+  ; b: int
+  (* c *) >
```
Julow, line 259: This is equally broken as before. I wouldn't slow this PR down due to this.

test/passing/tests/comments-no-wrap.ml.ref
```diff
@@ -296,8 +301,7 @@ type t =
 
 type t =
   (* | B *)
-  | A
-  (* A *)
+  | A (* A *)
```
Julow, line 304: This is a fix, the comment is placed as in the input.

test/passing/tests/comments_args.ml.ref
```diff
@@ -8,7 +8,8 @@ let emit_wrapper_function =
     false (* is_pair_generator *)
     hoisted true (* no_injection *)
     true (* inout_wrapper *)
-    is_interceptable false (* is_memoize_impl *)
+    is_interceptable false
+    (* is_memoize_impl *)
```
Julow, line 12: This is a fix, the comment is placed as in the input.

test/passing/tests/ite-compact_closing.ml.ref
```diff
@@ -152,3 +154,14 @@ let _ =
...
+let _ =
+  if
+    (* foo *)
+    foo
+  then 0
+  else if
+    (* bar *)
+    bar
+  then 1
+  else 2
```
Julow, line 167: Before this PR, this formatted as (with every profiles):
```
let _ = if (* foo *)
           foo then 0 else if (* bar *)
                              bar then 1 else 2
```

test/passing/tests/js_source.ml.ocp
```diff
@@ -113,7 +113,8 @@ let () =
   [%foo new x [@foo]];
   [%foo
     match[@foo] () with
-    | [%foo? (* Pattern expressions *)
+    | [%foo?
+             (* Pattern expressions *)
```
Julow, line 117: Note: This is ocp-indent output.

test/passing/tests/let_module-sparse.ml.ref
```diff
@@ -57,5 +57,6 @@ let () =
 
 let f () =
   let module (* comment *)
-  M = struct end in
+  M =
+  struct end in
```
Julow, line 61: This is a regression.

test/passing/tests/module.ml
```diff
@@ -111,9 +111,11 @@ let _ =
   let module M =
     ( val (* aa *) m (* bb *)
         : (* cc *)
-        M with type t = t (* dd *)
+        M
+        with type t = t (* dd *)
```
Julow, line 115: This is a regression.

test/passing/tests/let_module-sparse.ml.ref
Julow: Fixed. Now respect the input more.

test/passing/tests/ite-kw_first_no_indicate.ml.ref
```diff
@@ -159,3 +161,12 @@ let _ =
   else
     fun xxxxxxxxxxxxxxxxx yyyyyyyyyyyyyyyyyyyy zzzzzzzzzzz ->
     xxxxxxxxx yyyyyyyyyy zzzzzzzzzzzz
+
+let _ =
+  if (* foo *)
+     foo
+  then 0
+  else if (* bar *)
+          bar
```
gpetiot, line 170: This one could be improved.

test/passing/tests/ite-kw_first_closing.ml.ref
```diff
@@ -175,3 +177,12 @@ let _ =
   else
     fun xxxxxxxxxxxxxxxxx yyyyyyyyyyyyyyyyyyyy zzzzzzzzzzz ->
     xxxxxxxxx yyyyyyyyyy zzzzzzzzzzzz
+
+let _ =
+  if (* foo *)
+     foo
+  then 0
+  else if (* bar *)
+          bar
```
gpetiot, line 186: This one could be improved.

test/passing/tests/ite-kw_first_no_indicate.ml.ref
Julow: This seems to be the intended behavior.

gpetiot: After having a look at the diff, I'm not sure it looks better on "real" code, we can discard this commit.

test/passing/tests/module.ml
gpetiot: Fixed in cea6493

gpetiot (approved):
LGTM, thanks!

08934b4 is optional though

Julow:
Has 08934b4 changes code that do not contain comments, could you open a new PR for it ? That would improve debugging later.

Julow:
Many thanks for your help! I was pulling my hairs out on this one.

PR #2510, opened 2024-01-13
Backport 5.2 compilerlibs changes

gpetiot:
The code copied from compiler-libs.common that is used in the parser, location, error printing has changed (I had to copy more code for the Style etc.).

We lost the contextual highlighting like: (not updated in this PR)
```diff
 File "tests/option.ml", line 21, characters 3-14:
-21 | [@@ocamlformat 1, "if-then-else=keyword-first"]
-        ^^^^^^^^^^^
 Warning 47 [attribute-payload]: illegal payload for attribute 'ocamlformat'.
 Invalid format: String expected
```

@Julow I think you had a look at this code from the compiler in the past, maybe you will see something I missed.

For context we want to have error messages as close as possible to the ones from the compiler, but if the compiler changes its stuff every version we cannot be consistent with every version. Shall we just have our own stuff and cut the ties with the compiler's code?

### DIFF
--- a/test/cli/repl_file_errors.t/run.t
+++ b/test/cli/repl_file_errors.t/run.t
@@ -5,25 +5,25 @@ Make sure the locations of errors in repl files are right.
   $ ocamlformat --repl-file line5.repl
   ocamlformat: ignoring "line5.repl" (syntax error)
   File "line5.repl", line 5, characters 12-12:
-  Error: Syntax error: a toplevel phrase must end with `;;`. expected.
+  Error: Syntax error: "a toplevel phrase must end with `;;`." expected.
   [1]
 
   $ ocamlformat --repl-file line6.repl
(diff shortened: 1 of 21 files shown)


gpetiot:
I have a suggestion to improve the situation with the currently vendored compilerlibs.common stuff, we should have a look at `ocaml-compiler-libs` (the 5.2-compatible version is not released yet though https://github.com/ocaml/opam-repository/pull/25029)

Julow (approved):
Looks good :)

Colors work fine in my testing. `isatty` is always false when run from a cram test but is `true` when running from a terminal.

Julow:
While working on backporting parser-standard, I've found some more things that need to be backported and included them in this PR.

Julow (approved):
Let's merge :)

PR #2511, opened 2024-01-16
Avoid alignment in if-then-else

Julow:
These are some commits made by @gpetiot on top of https://github.com/ocaml-ppx/ocamlformat/pull/2507

This avoids alignment in if-then-else that were not introduced in #2507 but are now more noticeable. For example:

```ocaml
f
  ( if and_ even
         loooooooooooooooooooooooooooooooooooooooooooooooooooooooooonger
    then ()
    else () )
```

```ocaml
let _ =
  if (* foo *)
     foo
  then 0
  else ...
```


Julow:
No diff with test_branch, let's merge!

gpetiot:
Thank you!

PR #2513, opened 2024-01-22
Bug fix: `Pexp_constraint` with attributes should get parenthesized

alanechang:
There's a bug in formatting `f ((1 : int) [@a])` where the outer pair of parentheses gets stripped. This PR tries to fix it.

Test cases for the bug are added to `attributes.ml`.

`tools/test_branch.sh` was ran locally and resulted in no changes.


lib/Fmt_ast.ml
```diff
@@ -2156,14 +2156,16 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
           c.conf
           (fmt_constant c const $ fmt_atrs)
   | Pexp_constraint (e, t) ->
+      let outer_parens = has_attr && parens in
```
Julow, line 2159: `has_attr` doesn't imply parentheses, as seen for example in the `override.ml.ref` test.
Whether to output parentheses is specified in `Ast.parenze_exp` (which is how `parens` is computed) but in this case, it returns true too often.
I improved the Ast rules in this branch (2 commits): https://github.com/Julow/ocamlformat/tree/2513-tweak-ast Could you merge it ?

Julow (commented):
This has a small regression in override exprs (rarely used) that I have a fix for.

Julow:
Also, could you write an entry in CHANGES.md ?

alanechang:
Thanks for the feedback and suggestions! I cherry picked your commits and added an entry to CHANGES.md

I did not mark the changelog entry with an asterisk due to the minimal change to existing code (only one pair of parentheses removed in the override exprs test and no diff from `tools/test_branch.sh`). Let me know if you disagree with this and I'm happy to change it!

Julow (approved):
Many thanks for debugging and fixing this issue !

I agree that this shouldn't be considered a breaking change.

PR #2525, opened 2024-02-12
Fixes a bug in class type when trying to disable ocamlformat.

EmileTrotignon:
Currently then following snippet:
```ocaml
class type c = let open [@ocamlformat "disable"] Z
  in
  z
```
is printed as:
```ocaml
let open [@ocamlformat "disable"] Z
  in
  z
```

This PR fixes that. My implementation is a bit hackish, doing something better will require effort to ensure that we do not break the formatting in the absence of @disable.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2874,55 +2874,57 @@ and fmt_class_type ?(pro = noop) c ({ast= typ; _} as xtyp) =
   protect c (Cty typ)
   @@
   let {pcty_desc; pcty_loc; pcty_attributes} = typ in
-  update_config_maybe_disabled c pcty_loc pcty_attributes
-  @@ fun c ->
-  let doc, atrs = doc_atrs pcty_attributes in
-  let parens = parenze_cty xtyp in
-  let ctx = Cty typ in
-  let pro ~cmt =
-    pro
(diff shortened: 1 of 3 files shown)


lib/Fmt_ast.ml
```diff
@@ -2874,55 +2874,57 @@ and fmt_class_type ?(pro = noop) c ({ast= typ; _} as xtyp) =
...
-        pro ~cmt:true
-        $ fmt_open_description c ~keyword:"let open" ~kw_attributes:atrs
-            popen
-        $ str " in"
-        $ Params.Pcty.break_let_open c.conf ~rhs:cl
-      in
-      fmt_class_type c ~pro (sub_cty ~ctx cl) $ epi ~attrs:false
+  let c = update_config c pcty_attributes in
+  (if c.conf.opr_opts.disable.v then pro else noop)
+  $ maybe_disabled c pcty_loc pcty_attributes
```
Julow, line 2879: I like the approach but I think it'd be cleaner if it were implemented directly in `update_config_maybe_disabled`. In fact, it ultimately calls `maybe_disabled_k`, which supports this with its last argument.

Julow:
This is worth a change entry.

Julow (approved):
Looks good, thanks!

PR #2526, opened 2024-02-16
use `force_break` value instead of `fmt "@;<1000 0>"`

EmileTrotignon:
I thought the constant occurances of random values like "1000" in the code looked bad, so I replaced every occurence that I could by a new `force_break` value that is equal to `break 1000 0`.

I replaced `break 1000 0` and `fmt "@;<1000 0>"` by `force_break`.

I could not use `force_break` when used in function that expect a format string and not an Fmt.t like `fmt_or`. I am sure it is possible to do, but I wanted this Pr to be the easiest possible to review and merge.

### DIFF
--- a/lib/Fmt.mli
+++ b/lib/Fmt.mli
@@ -54,6 +54,9 @@ val protect : t -> on_error:(exn -> unit) -> t
 val break : int -> int -> t
 (** Format a break hint. *)
 
+val force_break : t
+(** [force_break] forces a break with indentation 0. *)
+
 val cbreak : fits:string * int * string -> breaks:string * int * string -> t
 (** Format a custom break.
 
--- a/lib/Fmt.mli
+++ b/lib/Fmt.mli
@@ -55,7 +55,7 @@ val break : int -> int -> t
 (** Format a break hint. *)
 
 val force_break : t
-(** [force_break] forces a break with indentation 0. *)
+(** [force_break] forces a break with indentation 0. Equivalent to [break 1000 0].*)
 
 val cbreak : fits:string * int * string -> breaks:string * int * string -> t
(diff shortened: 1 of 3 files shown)


lib/Fmt.mli
```diff
@@ -54,6 +54,9 @@ val protect : t -> on_error:(exn -> unit) -> t
 val break : int -> int -> t
 (** Format a break hint. *)
 
+val force_break : t
+(** [force_break] forces a break with indentation 0. *)
```
Julow, line 58: Mention that is is equivalent to `break 1000 0` ? Written like this, it could be confused with `@\n`.

Julow (approved):
Thanks for improving the code :)

`fmt_or` has an alternative that takes `Fmt.t` as arguments: `fmt_or_k`. Similarly, there's `fmt_if_k` and `list_k`.

It's already a nice thing to have so I'd merge if you don't want to go through all these cases.

You need to format.

EmileTrotignon:
> fmt_or has an alternative that takes Fmt.t as arguments: fmt_or_k. Similarly, there's fmt_if_k and list_k.

Its an easy fix so ill do it

Julow (approved):
Thanks, let's merge!

gpetiot:
Thanks, that's useful! What do you think of defining `double_linebreak` for `str '\n' $ force_break`? that is also used a lot.

EmileTrotignon:
@gpetiot 
> Thanks, that's useful! What do you think of defining double_linebreak for str '\n' $ force_break? that is also used a lot.

I am not against that, but my main goal is eliminating format strings like `fmt "@ "` which I think are far worst than `str "\n"`. i will keep this in mind though.

PR #2528, opened 2024-02-23
Removes format syntax

EmileTrotignon:
This PR is a big refactoring that completely removes format syntax such as `"@;<1000 -2>" from ocamlformat.

Instead, documented combinators are used.

The first commit should be more readable, because `dune fmt` was not called on it, and as such the diff is smaller than the total.

### DIFF
--- a/lib/Fmt.mli
+++ b/lib/Fmt.mli
@@ -11,9 +11,6 @@
 
 (** Formatting combinators *)
 
-(** Format strings that accept no arguments. *)
-type s = (unit, Format_.formatter, unit) format
-
 (** Format thunks. *)
 type t
 
@@ -57,6 +54,18 @@ val break : int -> int -> t
 val force_break : t
 (** [force_break] forces a break with indentation 0. Equivalent to [break 1000 0].*)
 
+val space_break: t
+(** fmt "@ " *)
+
+val cut_break: t
+(** fmt "@," *)
+
+val flush_newline: t
+(** fmt "@." *)
+
+val force_newline: t
+(** force_newline *)
+
 val cbreak : fits:string * int * string -> breaks:string * int * string -> t
 (** Format a custom break.
 
@@ -67,9 +76,6 @@ val cbreak : fits:string * int * string -> breaks:string * int * string -> t
 
 val noop : t
 (** Format nothing. *)
-
-val fmt : s -> t
-(** Format a format string. *)
 
 (** Primitive types -----------------------------------------------------*)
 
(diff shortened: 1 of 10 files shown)


lib/Fmt.mli
```diff
@@ -87,10 +101,6 @@ val str_as : int -> string -> t
 val opt : 'a option -> ('a -> t) -> t
 (** Format an option using provided formatter for the element. *)
 
-val list : 'a list -> s -> ('a -> t) -> t
```
gpetiot, line 90: We could have `val list : 'a list -> sp -> ('a -> t) -> t`, I think most separators are actually spaces anyway

Julow: I was quite annoyed in the past with the limited `list` function. I'd suggest instead that we remove all the `_k` in the function names and have only functions taking a `t`.

I think the extra work to pass just a space is worth the consistency: one way to build a space, one way to build a list, the list can evolve in a more complex form later.

EmileTrotignon: Another option is to have an infix operator for `str`. This could remove some of the parenthesis if the operator has higher priority than str. It would otherwise make the code a bit higher-context : `str` is self-explanatory, `!$` is not.

Overall, I think that `!$` combined with values for common strings like `"\n"` and `" "` is my preferred option. It also has the advantage of being likely easier to implement with regexp, but I am not completely against `wrap_str` either.

Julow: I don't mind the many `str " "` for now and I think we should stay focused on one change at a time.

I like the idea of the prefix operator. I don't think we need a self-explanatory name for the most important abstraction in the project, especially if the `Fmt` module is not too big.
Again, for later.

EmileTrotignon: I am happy for this kind of changes to be in an other PR. If @gpetiot is convinced as well I think we can merge.

gpetiot: Yes it's all good to me!

lib/Fmt.mli
```diff
@@ -57,6 +54,26 @@ val break : int -> int -> t
...
+(** [space_break] is either a single space or a newline.
+    See {!Stdlib.Format.print_space_break}.
+    Equivalement to format syntax ["@ "].*)
+
+val cut_break : t
+(** [cut_break] is either a newline or a {!noop}.
+    See {!Stdlib.Format.print_cut}.
+    Equivalement to format syntax ["@,"].*)
+
+val flush_newline : t
```
Julow, line 67: `flush_newline` doesn't occur in formatting code and is highly unwanted.

The only use, in `Translation_unit`, could as well be `force_newline`.

Julow (approved):
This is a very nice clean up of the code and makes the code more uniform.

I have just one comment and then I'd be happy to merge!

lib/Fmt.mli
EmileTrotignon: Okay, I will change it !

Julow: Let's merge!

PR #2529, opened 2024-02-23
Simplify representation of class field values and methods

gpetiot:
Makes #2401 easier.

@Julow some normalizations have been undone (see `js_source.ml`), but the result is closer, and some indentation has to be fixed, but this tremendously simplifies the formatting, and removes the calls to `Sugar.fun_` for `Pexp_newtype` and for `fmt_class_field_kind`.

- `object.ml.ref` can be removed once we fixed the regressions
- in the parser, `args_typ_strict_binding` can be renamed `args_typ_strict_binding` (we finally covered every case)

If you have the time to have a look at the indentation that would be very helpful!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,8 @@ profile. This started with version 0.26.0.
 - \* Don't align breaking module arguments (#2505, @Julow)
 - Improvements to ocp-indent-compat and the Janestreet profile (#2314, @Julow)
 - \* Undo let-bindings normalizations (#2523, @gpetiot)
+- \* Undo method parameters normalizations (#2529, @gpetiot)
+- \* The `break-colon` option is now taken into account for method type constraints (#2529, @gpetiot)
 
 ### Fixed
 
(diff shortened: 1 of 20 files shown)


Julow (approved):
I love the idea ! This applies the recent changes to function def to methods and cleans up the AST significantly.
The removal of `Pexp_poly` is a very good thing ! The formatting code is also a lot easier to read.

I think this patch should be proposed upstream.

The test-branch diff is large but seems to only contain improvements. Do you think it would be easy to split the formatting change and the AST change into two different PRs ? Especially the change in the formatting of the `:`, which generates most of the diffs requires a change entry in itself.

gpetiot:
I tried to split the AST change away from the formatting change, but it runs too deep. I tried to minimize the diff as much as I could instead, and I added a changelog entry for the normalizations and for taking into account the `break-colon` option.

Julow (approved):
Awesome ! This is a nice improvement to the AST and to the formatting code.

I believe the `:` change is the only change reported by test-branch.
Let's merge !

PR #2533, opened 2024-03-05
Represent the expr sequence as a list

gpetiot:
Not urgent, but it cleans up a lot of stuff in sugar/ast about sequences.

The diff is caused by:
- comments attached after the sequence were moved to after the last element, I'm not sure we should continue to do that, to me the diff looks better on this
- the special case added by Josh about adding a break before the semicolon that follows a debug/trace call, this has been factorized with the rest, with a match on the extension name. Not sure how much people would complain if we removed it? We can also only apply that for the `profile = ocamlformat` case, or just remove it.


Julow (commented):
I love this! This removes a lot of tricky code.

> the special case added by Josh about adding a break before the semicolon that follows a debug/trace call, this has been factorized with the rest, with a match on the extension name.

It would be nice to have a test case for this. I think it's fair to remove it otherwise.

test-branch finds a similar case in irmin but with `[%log.debug]`, the semicolon was broken before so it seems that the rule was more general.

Also, this might remove some rewriting that were done before, like removing parentheses around sub sequences, rewriting extensions into a shorter form. It would be nice to show the change in the tests.

gpetiot:
There is already a test of the facebook trace calls: test/passing/tests/pre_post_extensions.ml

lib/Fmt_ast.ml
```diff
@@ -1608,38 +1608,43 @@ and fmt_sequence c ?ext ~has_attr parens width xexp fmt_atrs =
...
+    match xe1.ast with
+    (* special case for Meta/Facebook *)
+    | { pexp_desc=
+          Pexp_extension
+            ( {txt= "Trace.call"; _}
+            , PStr
+                [ { pstr_desc= Pstr_eval ({pexp_desc= Pexp_fun _; _}, [])
+                  ; pstr_loc= _ } ] )
+      ; _ } ->
+        space_break $ str ";" $ space_break
```
hhugo, line 1620: `ext` is ignored here

PR #2535, opened 2024-03-25
Multi version support with binary selection

EmileTrotignon:
This gives ocamlformat multiversion support.

The workflow would be the following : 

When you `opam install ocamlformat.0.27.4`, you get two packages : `ocamlformat` with the `ocamlformat` executable that does not format anything. You also get `ocamlformat-0-27-1` that has the executable that does the formatting, named `ocamlformat-0-27-1`.

Then when you try to format a project that requires version 0.15.3, the `ocamlformat` executable goes looking for an `ocamlformat-0-15-3` executable. If it is found, it is executed, and if it is not, instruction to do `opam install ocamlformat-0-15-3` are printed.

The objective is to have the maximum number of `ocamlformat-%version%` packages be co-installable, but that might not be possible due to dependency conflicts. 

There will be some work to repackage previous version along these lines, but it should be doable.

Regarding implementation, I chose to use the same pipeline that executes ocamlformat for real to fetch the required version. This allows me to ensure that the logic to, for instance, get the config file, is the same. Minor changes where required to disable warning because they were otherwise printed twice.

I have not touched the RPC at all, I would like some help on that front.


EmileTrotignon:
The test fail if run cold after a `dune clean`. If you run `dune build` first, they work. This needs to be fixed.

lib/dune
```diff
@@ -14,8 +14,8 @@
 (ocamllex Toplevel_lexer)
 
 (library
- (name ocamlformat_lib)
- (public_name ocamlformat-lib)
+ (name ocamlformat_lib_)
```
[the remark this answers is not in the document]
EmileTrotignon: This would be quite heavy to do because of all the vendored stuff.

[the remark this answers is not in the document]
Julow: Several versioned libraries could be installed at the same time so unfortunately, we must make their name unique.

Do we have versioned libraries in your proposal ? I'm not sure this is necessary, the versioned packages could contain the library as a private library.

samoht:
What's the relationship with https://github.com/CraigFe/ocamlformat-vmux ?

EmileTrotignon:
@samoht 
> What's the relationship with https://github.com/CraigFe/ocamlformat-vmux ?

It is inspired from it, but the big difference is that here we are laying the ground work for having multiple ocamlformat versions on the same switch, which can only be done inside ocamlformat.

Our fix is also slightly less hackish : it uses the path, and therefore does not assume anything about opam switches. This could mean higher compatibility with the platform installer, or in the future, dune package management (we are going to double check for the latter).

This will allow for a familiar workflow were when you do not have the right ocamlformat version installed, you can solve the issue with a single `opam install` command (and you are given this command in the error message).

We also use ocamlformat as lib to locate the config file which makes this more maintainable : if we change the way the config file is located, the version selector will be updated as well, and we automatically cover corner-cases for that.

I think Craig's solution is great for people who dig around and are quite familiar with the ocaml ecosystem, we intend to bring a fix that is going to be usable without any prior knowledge.

RELEASE.md
```diff
@@ -0,0 +1,31 @@
...
+- `ocamlformat.X.Y.Z` which contains the `ocamlformat` executable
+- `ocamlformat-X_Y_Z.1.0` which contains the `ocamformat-X_Y_Z` executable
+
+You also have packages for the libs : `ocamlformat-lib.X.Y.Z` and
+`ocamlformat-lib-X_Y_Z.1.0`.
+
+You can only have one `ocamlformat` executable on your switch, but you can have
+multiple `ocamlformat-X_Y_Z` installed.
+
+The `ocamlformat` does not contains the formatter itself : it just reads your
```
samoht, line 15:
```suggestion
The `ocamlformat` package does not contain the formatter itself: it just reads your
```

RELEASE.md
```diff
@@ -0,0 +1,31 @@
+# Explaination of the versionning scheme
+
+OCamlformat now an unsual versionning scheme. For every version `X.Y.Z` of ocamlformat,
+there are multiple packages :
+
+- `ocamlformat.X.Y.Z` which contains the `ocamlformat` executable
+- `ocamlformat-X_Y_Z.1.0` which contains the `ocamformat-X_Y_Z` executable
+
+You also have packages for the libs : `ocamlformat-lib.X.Y.Z` and
```
samoht, line 9: Out of curiosity, why do you need to do this for the library? Are you expecting to co-install multiple libraries of ocamlformat-lib in the same switch? Who are the consumers of those libraries?

EmileTrotignon: The consumers of the `ocamlformat-lib-X_Y_Z` package are `ocamlformat-lib.X.Y.Z` and `ocamlformat-X_Y_Z.1.0`.

The executable needs it for obvious reasons, and then we want to package the lib for general use, as it has been packaged before, under the name `ocamlformat-lib`.
This felt like the simplest way to make this lib accessible to both packages. The versionned lib itself is not meant to be used by itself. An alternative would to have two equivalent libraries with a dune rule that copy files. Then the `ocamlformat-lib-X_Y_Z` would be private.

I do not really like copying source files around like that (especially because its not all in one folder, the vendor folder is also part of the `ocamlformat-lib-X_Y_Z package), but what we have now is not ideal either. I will think more about and I am happy to receive input on that topic.

samoht: I am a bit confused - I though the `ocamlformat-X_Y_Z` would only contain the `ocamlfomat` binary and be self-contained (ie. without external dependencies, so you could easily co-install all of these packages in the same switch without dependency conflicts). If you do this, why would they need to depend on `ocamlformat-lib-X_Y_Z`. If nobody is using `ocamlformat-lib_X_Y_Z` I would just not build more maintenance burden by releasing them :-)

leostera:
As I mentioned in our conversation, it seems this would be addressed naturally as Dune package management arrives, since Dune would do the "multi-version management" for you.

If you implement this, Dune will likely need to find and use the `ocamlformat-<vsn>` binary/package directly anyway.

RELEASE.md
EmileTrotignon: `ocamlformat-X_Y_Z` has a real dependency on `ocamlformat-lib-X_Y_Z`, thats how it is implemented. We might do vendoring to hide it, but this will require duplicating sources (maybe this is actually fine ?)
I will try to find a satisfying solution for this, and in the meantime, I have removed the separate package (now you have `ocamlformat-X_Y_Z.ocamlformat-lib`).
As for dependencies, I think you are right, and we should vendor them. It would be very hard to remove them for all ocamlformat previous versions, but vendoring should work and make sure everything always work. It is going to be a bit more work for each version but I think its worth it.
For forward versions we will try to remove dependencies, especially on base.

EmileTrotignon:
This would be too difficult to maintain and dune package management would solve the issue, so I am closing it.

Issue #2536, opened 2024-04-09
BUG: generating invalid ocaml syntax.


nojb:
`ocamlformat.0.26.1` chokes on:
```
let _ =
  if x && try () with _ -> () then ()
```
when `.ocamlformat` contains
```
ocp-indent-compat
```

Julow:
Thanks for the report. The bug doesn't occur on the main branch, could you check ? This seems to be fixed since #2445.

nojb:
> Thanks for the report. The bug doesn't occur on the main branch, could you check ? This seems to be fixed since #2445.

I tried master but now I got another BUG :)
```
let _ =
  try
    ()
  with _ ->
    (match () with _ -> ())
```
with
```
indicate-multiline-delimiters=closing-on-separate-line
```

nojb:
And another:

```
let _ =
  try () with _ ->
  try () with _ -> () in
()
```
also with
```
indicate-multiline-delimiters=closing-on-separate-line
```

Julow:
Thanks! This is hopefully a recent bug. Should be fixed in https://github.com/ocaml-ppx/ocamlformat/pull/2537

nojb:
> Thanks! This is hopefully a recent bug. Should be fixed in #2537

Thanks, this fixes it indeed, and the original bug reported is fixed in `master` already. Closing. Thanks!

Julow:
Thanks for checking.

PR #2537, opened 2024-04-11
Fix missing parenthesis after try..with

Julow:
Fix commit https://github.com/ocaml-ppx/ocamlformat/issues/2536#issuecomment-2045943473

When indicate-multiline-delimiters=closing-on-separate-line, the closing parenthesis at the end of a try..with expression is missing since #2528.


nojb:
I confirm this fixes https://github.com/ocaml-ppx/ocamlformat/issues/2536#issuecomment-2045943473. Thanks!

Julow:
Thanks for checking! Let's merge.

PR #2540, opened 2024-04-15
Fix comments around underscore in record patterns

Julow:
Remove the added empty line and the extraneous space with `break-separator=after`.
Fix unstable comment with `break-separator=before`.


nojb:
Thanks!

PR #2544, opened 2024-04-18
Backport Pexp_function from OCaml 5.2

Julow:
This is a work in progress and is required for 5.2 compatibility.

The old `Pexp_function` and `Pexp_fun` are formatted in different ways in different contextes (eg. as an argument, in a let binding, after an infix operator, etc..), which are not easy to rewrite as the AST is very different.


Julow:
The diffs reported by test-branch are not very reasonable. `+495 -406` on a total of a million line of code, most of which appear as bug fixes.
Let's merge !

Issue #2549, opened 2024-05-11
Bug: `--parse-docstrings` violates margin


mmottl:
**Describe the bug**

Passing `--parse-docstrings` may lead to output that violates the configured margin.

```sh
❯ ocamlformat --version
0.26.2
```

**How to Reproduce**

Formatted file `foo.ml`:
```ocaml
(** xxxxx xx xxxxxx xxxxxx xxxxxx xxxxx xxxxxxx xxx xxxxxx xxx xxxx
    xxxxxxxxxx. *)
```

Command line (explicitly sets a margin of 80):
```sh
ocamlformat --enable-outside-detected-project -m 80 --parse-docstrings foo.ml
```

Output has an 82 character line:
```ocaml
(** xxxxx xx xxxxxx xxxxxx xxxxxx xxxxx xxxxxxx xxx xxxxxx xxx xxxx xxxxxxxxxx. *)
```

Julow:
Thanks for reporting. Should be fixed by https://github.com/ocaml-ppx/ocamlformat/pull/2550

PR #2550, opened 2024-05-14
Fix doc-comments '*)' exceeding the margin

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2549

This change applies to doc-comments and regular comments parsed as doc-comments.


Julow:
Test branch shows that this bug is very common. All diffs seem reasonable.

PR #2553, opened 2024-05-27
Box debug not showing break_unless_newline and fits_or_breaks

Julow:
Format the box_debug output as length 0 with pp_print_as and remove the indentation in the output HTML.
This was interfering with many instances of break_unless_newline and fits_or_breaks and caused hair loss during previous debugging sessions.


AuntyEms:
A+

PR #2554, opened 2024-05-28
vscode debugger config

EmileTrotignon:
This adds config for the debuger provided by the earlybird vscode extension.
There might be some issues with enabling bytecode mode, but I do not know of any.

Its very very useful if you use vscode, I have managed to understand behaviours that were hard to debug-print with this.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,13 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "OCaml earlybird (experimental)",
+            "type": "ocaml.earlybird",
+            "request": "launch",
+            "program": "${workspaceRoot}/_build/default/bin/ocamlformat/main.bc",
+            "arguments": ["-g", "${workspaceRoot}/test.ml"],
+            "stopOnEntry": true
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,13 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "OCaml earlybird (experimental)",
+            "type": "ocaml.earlybird",
+            "request": "launch",
+            "program": "${workspaceRoot}/_build/default/bin/ocamlformat/main.bc",
```
Julow, line 8: It would be great if this called `dune build _build/default/bin/ocamlformat/main.bc` before every runs.

EmileTrotignon:
I would like this to be merged without extra work because it works, and is useful as is, and I need it. I would be happy to do the research necessary to also build the project through an vscode action, but I remember I tried when I first wrote this, and it did not work, so I think its not going to be a 10min job.

Julow (approved):
Ok, let's merge.

Julow:
This was partly reverted in https://github.com/ocaml-ppx/ocamlformat/pull/2637 due to problem during the release.

PR #2557, opened 2024-05-30
Stricter AST checks

Julow:
The AST check was previously trigerring if both the std AST and the
extended AST were different. This is not the intended behavior.

The first commit changes the normalization of code blocks in docstrings to correctly handle repl phrases. This fixes an AST error that was hidden by the previous check.


hhugo:
I think this PR means that we no longer check ast match for Repl files ? Because Repl are not "implemented" Std_ast.

```ocaml
  (* not implemented *)
  | Repl_file : unit t
```

@Julow, can you confirm ?

Julow:
@hhugo I think that's right, we don't test repl output.

PR #2566, opened 2024-07-29
add box_debug to new fmt combinators for breaks

EmileTrotignon:
This adds output to box_debug for the following combinators :
- space_break
- cut_break
- force_newline


Julow (approved):
Nice!

PR #2567, opened 2024-07-29
Debug_stacktrace

EmileTrotignon:
This add stacktraces to the html debug output. The stacktrace is only present on hover.

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -1400,7 +1400,9 @@ module Operational = struct
   let debug =
     let doc = "Generate debugging output." in
     Decl.flag ~default ~names:["g"; "debug"] ~doc ~kind
-      (fun conf elt -> update conf ~f:(fun f -> {f with debug= elt}))
+      (fun conf elt ->
+        if elt.v then Box_debug.enable_stacktraces := true ;
+        update conf ~f:(fun f -> {f with debug= elt}) )
       (fun conf -> conf.opr_opts.debug)
 
   let disable =
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -1400,9 +1400,7 @@ module Operational = struct
   let debug =
     let doc = "Generate debugging output." in
     Decl.flag ~default ~names:["g"; "debug"] ~doc ~kind
-      (fun conf elt ->
-        if elt.v then Box_debug.enable_stacktraces := true ;
-        update conf ~f:(fun f -> {f with debug= elt}) )
+      (fun conf elt -> update conf ~f:(fun f -> {f with debug= elt}))
       (fun conf -> conf.opr_opts.debug)
 
   let disable =
--- a/lib/Fmt.ml
+++ b/lib/Fmt.ml
@@ -97,13 +97,15 @@ let force_break = break 1000 0
 
 let space_break =
   (* a stack is useless here, this would require adding a unit parameter *)
+  let stack = "" in
   with_pp (fun fs ->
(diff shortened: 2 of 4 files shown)


EmileTrotignon:
image

lib/Conf.ml
```diff
@@ -1401,7 +1401,9 @@ module Operational = struct
   let debug =
     let doc = "Generate debugging output." in
     Decl.flag ~default ~names:["g"; "debug"] ~doc ~kind
-      (fun conf elt -> update conf ~f:(fun f -> {f with debug= elt}))
+      (fun conf elt ->
+        if elt.v then Box_debug.enable_stacktraces := true ;
```
Julow, line 1405: This means that enabling the debug output implies enabling the stacktraces. Is the purpose to have the debugging output without having the stacktraces ?

lib/Fmt.ml
```diff
@@ -88,30 +88,36 @@ let with_box_debug k = with_pp (Box_debug.with_box (fun fs -> eval fs k))
...
 let space_break =
+  (* a stack is useless here, this would require adding a unit parameter *)
+  let stack = "" in
   with_pp (fun fs ->
-      Box_debug.space_break fs ;
+      Box_debug.space_break ~stack fs ;
       Format_.pp_print_space fs () )
 
 let cut_break =
+  let stack = Box_debug.get_stack () in
```
Julow, line 106: Should be the same as in the space case, no trace ? Same in `force_newline`.

lib/Fmt.ml
```diff
@@ -88,30 +88,36 @@ let with_box_debug k = with_pp (Box_debug.with_box (fun fs -> eval fs k))
...
   with_pp (fun fs ->
-      Box_debug.break fs n o ;
+      Box_debug.break fs n o ~stack ;
       Format_.pp_print_break fs n o )
 
 let force_break = break 1000 0
 
 let space_break =
+  (* a stack is useless here, this would require adding a unit parameter *)
+  let stack = "" in
```
Julow, line 100: Should be an option.

lib/Fmt.ml
EmileTrotignon: Yes this is true

lib/Conf.ml
EmileTrotignon: Enabling the stacktraces is a bit costly in terms of performances, and they have to be collected early. This allows to disable them when they are not going to be used, that is when there is no debug run that is going to be performed.

Julow: There's already a bool variable for that: `Fmt.debug`. According to what you said, `enable_stacktraces` is unnecessary.

lib/Fmt.ml
Julow: There's still the issue in `force_newline`.

lib/Conf.ml
EmileTrotignon: No, fmt_debug is set much later

EmileTrotignon: Box_debug.debug is set by the following function :
```
val with_box_debug : t -> t
```

When this function is called, the `t` you give as input is already generated, so its stack traces are either in memory or they are lost forever.

Julow: Makes sense then!

Julow: Though, to have clearer dependencies between the modules, I'd prefer if that variable was set from `Translation_unit`.

Julow (approved):
This is helping a lot, thanks!

PR #2577, opened 2024-09-17
Allow custom commands for running ocamlformat

gridbugs:
Currently it's possible to customize the path or name of the ocamlformat executable that gets run by the emacs ocamlformat package. However sometimes it's desirable to run ocamlformat by running a multi-word command, such as `opam exec ocamlformat --`. This change adds support running coamlformat with custom multi-word commands by allowing the `ocamlformat-command` variable to be set to a list of strings.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ profile. This started with version 0.26.0.
 - Support OCaml 5.2 syntax (#2519, @Julow)
   This includes:
   + Local open in types.
+- Allow a custom command to be used to run ocamlformat in the emacs plugin (#2577, @gridbugs)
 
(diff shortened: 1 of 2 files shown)


gridbugs:
The context for this PR is that I'm in the process of adding a feature to dune that will download, build, and then run ocamlformat from a single dune command. I've been testing the command in different editors and this patch is necessary in order to use this feature in emacs.

gpetiot (approved):
I'm not an expert but it looks reasonable to me. Thanks!

gridbugs:
Updated the changelog

Julow:
Thanks :)

PR #2589, opened 2024-10-16
Improve comment placement after a `then` or `else`

Julow:
Allow comments on the same line as `then` and `else`, as it was the case in 0.26.2. Also, make sure to avoid formatting any code after a comment in that position.

This was broken since #2507.


Julow:
To avoid generating a large number of diffs, I added the locations of `then` and `else` keywords to the AST. This new information is used to preserve different cases:
```ocaml
if cond then (* attached to then *)
  expr;

if cond then
  (* attached to expr *)
  expr;

if cond then (* attached to expr *) expr;
```

PR #2590, opened 2024-10-16
Fix 'function' indent in methods with no-wrap-fun-args

Julow:
Fix a recent bug that impacts js_of_ocaml.


Julow:
This bug was introduced in https://github.com/ocaml-ppx/ocamlformat/pull/2544

PR #2591, opened 2024-10-16
parser-extended: Add missing Ast_mapper rule

Julow:
Found by test-branch on very recent code. This caused a problem during normalization with this code:
```ocaml
(** {[
      f
        (fun ~f -> x)
    ]} *)
```


Julow:
test-branch is failing because the main branch contains the bug. ocaml-ci results are the same as on main.

PR #2595, opened 2024-10-21
Make ocamlformat compatible with 5.3

adamchol:
This doesn't look right because it probably breaks compatibility with older versions but I don't understand how compiling against different OCaml versions works here.


Julow:
We do this by vendoring the functions that changed. I took your commit in https://github.com/ocaml-ppx/ocamlformat/pull/2603
Thanks :)

PR #2596, opened 2024-10-23
Represent 'let f _ = function' in the CST

Julow:
Currently, the CST represents the arguments of 'let f _ =' in a way that
is incompatible with the changed function representation from OCaml 5.2.
These two were represented the same way in the CST:

    let f _ = function ...
    let f _ = (function ...)

To fix this problem, the representation for let bindings bodies is
changed to 'function_body', matching the 5.2 representation.

The `ctx` used while formatting let bindings is changed and this caused a large number of changes in the code and removed a few unnecessary parentheses in the output.


Julow:
This fixes a critical bug in https://github.com/ocaml-ppx/ocamlformat/pull/2544 and fixes bad indentation and unwanted parentheses in let bindings.

PR #2600, opened 2024-10-25
Add compatibility with emacs ocaml tree-sitter modes

thibautbenjamin:
I use a tree-sitter based mode in emacs for ocaml (https://github.com/dmitrig/ocaml-ts-mode).

I made a small change to make the `before-save-hook` compatible with these modes, let me know if you are interested in merging


gpetiot (approved):
LGTM, thanks!
Can you add a changelog entry? Then we can merge it.

Julow (approved):
Merging, thanks!

PR #2608, opened 2024-10-31
odoc: Don't add unecessary space in references and links

Julow:
Avoid adding a space before the text of a reference or link:

    {{!ref}text}
    {{:url}text}

As regular comments might be formatted as docstrings, the space must be preserved if it appears in the source to avoid crashing. odoc-parser is changed to preserve this information.


Julow:
This slightly changes the formatting of references and links for the better. Now is the best time to do this change as this release will also enable formatting of docstrings by default:
```diff
     {{:https://github.com/mirage/irmin/blob/main/README_PPX.md} documentation
-      for [ppx_irmin]})*)
+     for [ppx_irmin]})*)
 ...
-    of similar bindings themselves, by using the appropriate {{!Merge.MSet}
-    multi-sets}. *)
+    of similar bindings themselves, by using the appropriate
+    {{!Merge.MSet} multi-sets}. *)
```

PR #2512, opened 2024-01-22
Backport 5.2 changes to parser-standard

Julow:
AST changes, perhaps due to parser-extended not being uptodate.
This might depend on https://github.com/ocaml-ppx/ocamlformat/pull/2503

PR #2619, opened 2024-11-14
5.2 support: Raw identifiers

Julow:
This is the last missing piece for 5.2 support.

Thanks to @ccasin who made the work in https://github.com/janestreet/ocamlformat/pull/88

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -592,7 +592,7 @@ let fmt_type_var s =
   (* [' a'] is a valid type variable, the space is required to not lex as a
      char. https://github.com/ocaml/ocaml/pull/2034 *)
   $ fmt_if (String.length s > 1 && Char.equal s.[1] '\'') (str " ")
-  $ str s
+  $ ident s
 
 let rec fmt_extension_aux c ctx ~key (ext, pld) =
   match (ext.txt, pld, ctx) with
@@ -1451,7 +1451,7 @@ and fmt_param_newtype c = function
   | names ->
       cbox 0
         (Params.parens c.conf
-           (str "type " $ list names space_break (fmt_str_loc c)) )
+           (str "type " $ list names space_break (fmt_ident_loc c)) )
 
 and fmt_expr_fun_arg c fp =
   let ctx = Fpe fp in
@@ -3415,6 +3415,11 @@ and fmt_type_declaration c ?(pre = "") ?name ?(eq = "=") {ast= decl; _} =
     let fit = Tyd.is_simple decl in
     fmt_docstring_around_item_attrs ~force_before ~fit c ptype_attributes
   in
+  let type_name =
+    match name with
+    | Some name -> fmt_longident_loc ~constructor:false c name
+    | None -> ident txt
+  in
   let box_manifest k =
     hvbox c.conf.fmt_opts.type_decl_indent.v
       ( str pre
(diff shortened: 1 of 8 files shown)


lib/Fmt_ast.ml
```diff
@@ -894,29 +908,33 @@ and fmt_core_type c ?(box = true) ?pro ?(pro_space = true) ?constraint_ctx
...
-        ( hovbox 0 (str "module" $ space_break $ fmt_longident_loc c id)
+        ( hovbox 0
+            ( str "module" $ space_break
+            $ fmt_longident_loc c ~constructor:false id )
         $ fmt_package_type c ctx cnstrs
         $ fmt_attributes c attrs )
   | Ptyp_open (lid, typ) ->
       hvbox 2
-        ( hvbox 0 (fmt_longident_loc c lid $ str ".(")
+        ( hvbox 0 (fmt_longident_loc c ~constructor:true lid $ str ".(")
```
hhugo, line 937: Why is it `~constructor:true` here ?

hhugo:
We need raw idents to be preserved. if I use \#effect in a 5.3 switch, I don't want it to be rewritten in a 5.2 switch.
The current implementation depends on Lexer.is_keyword that is version dependent.

hhugo:
Thinking more about it, you could even imagine a mode that would help migrating to newer ocaml version by escaping idents that are becoming keyword.

ccasin:
Sorry, I should have mentioned it, but I think my PR in the janestreet repo has a bug: escaped operators in types need special treatment.

The compiler itself also has this bug, see the report in ocaml/ocaml#13603 and a fix in ocaml/ocaml#13604.  I'm in the process of adapting that fix in my ocamlformat or to our branch.

Julow:
I agree that we should preserve the raw identifier marker and not try to add it again later. I'm now investigating passing that information through a table, like we do for comments, as changing the AST for this might brings huge changes to the lexer and parser, making future backports harder.

> Thinking more about it, you could even imagine a mode that would help migrating to newer ocaml version by escaping idents that are becoming keyword.

That should be possible on top of this work.

hhugo:
> changing the AST for this might brings huge changes to the lexer and parser, making future backports harder.

Can't you keep the escaping prefix during lexing ? Without touching the ast ?

hhugo:
> Sorry, I should have mentioned it, but I think my PR in the janestreet repo has a bug: escaped operators in types need special treatment.

If we go with preserving escaped idents, we no longer have special treatments

hhugo:
Don't we need to update the lexer in parser-standard as well ?

hhugo:
The upstream OCaml lexer is able to recognize different set of keywords. One can pass `-keywords 5.2` on the command line

hhugo:
> > Don't we need to update the lexer in parser-standard as well ?
> 
> No, I'll let the parser-standard behave as it does upstream (`\#foo` is represented the same as `foo`).
> 
> >

Ok, turns out the standard lexer was already updated in #2512

hhugo:
> I'm currently investigating changing the AST to represent how identifiers looked in the source. If that's too much changes in the parser and AST, I'll switch to the table approach.

In https://github.com/ocaml-ppx/ocamlformat/pull/2619#issuecomment-2477213347, I was suggesting to not change the Ast and just store the escaped ident in the LIDENT token. What do you think ?

Julow:
I pushed the Parsetree approach here: https://github.com/ocaml-ppx/ocamlformat/pull/2620
The support in `Fmt_ast` is trivial in exchange for a +233 -198 change in the vendored parser. There's +88 -72 line changes in `parser.mly`, which will negatively impact future backports but which is less than I expected.

Julow:
> In #2619 (comment), I was suggesting to not change the Ast and just store the escaped ident in the LIDENT token. What do you think ?

That would have been simpler indeed. I'll investigate how that looks.

Julow:
The lexer approach is in this PR: https://github.com/ocaml-ppx/ocamlformat/pull/2621
I'll close this, as the other approach is simpler and more complete.

PR #2620, opened 2024-11-18
5.2 support: Raw identifiers (by changing the Parsetree)

Julow:
Retry of https://github.com/ocaml-ppx/ocamlformat/pull/2619 by keeping the original syntax of identifiers in the parsetree.

Strings in the Parsetree are replaced by the new `ident` type, which represent the different kinds of identifiers: `\#raw`, `( >>= )`, `[]`, `true`.
This removes the need to guess whether an identifier needs parentheses and greatly simplifies the support of raw identifiers. Changes in `Fmt_ast` were mostly type-directed.


Julow:
The changes to the parser can be avoided entirely: https://github.com/ocaml-ppx/ocamlformat/pull/2621
I'll close this as the other approach avoids divergence in the vendored parser.

PR #2622, opened 2024-11-20
5.3 support: Backport utf8 identifiers

Julow:


Julow:
In `test_branch`, parsing errors in `infer` are ignored because of https://github.com/ocaml/ocaml/issues/13626

Issue #2601, opened 2024-10-28
Bug: AST changed with strings containing \r\n


Julow:
A reproduction can be found in `test-extra/code/ocaml/testsuite/tests/lexing/newlines.ml`. This is new since 0.26.2.

Julow:
This bug is noticeable since https://github.com/ocaml-ppx/ocamlformat/pull/2557

hhugo:
The issue comes from https://github.com/ocaml/ocaml/pull/12514, `lib/Literal_lexer.mll` needs to be updated accordingly.

hhugo:
The diff below should be enough to fix the issue with OCaml 5.2 and above - but will cause issues with older version. 


```diff
$ git diff
diff --git a/lib/Literal_lexer.mll b/lib/Literal_lexer.mll
index 95ed7eec..e05ec78e 100644
--- a/lib/Literal_lexer.mll
+++ b/lib/Literal_lexer.mll
@@ -81,7 +81,9 @@ and string_aux mode = parse
         string_aux mode lexbuf
       }
   | newline
-      { store_string (Lexing.lexeme lexbuf);
+      { (match Lexing.lexeme lexbuf with
+        | "\n" -> store_string "\n"
+        | s -> store_string (String.sub s 1 (String.length s - 1)));
         string_aux mode lexbuf }
   | eof
       { raise Parse_error }
```

PR #2626, opened 2024-11-25
Don't normalize newlines in parser-standard

Julow:
This was causing a changed AST error, perhaps due to an interaction with Format's handling of '\r\n'.

'newlines.ml' is from the compiler's testsuite.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ profile. This started with version 0.26.0.
 
 ### Highlight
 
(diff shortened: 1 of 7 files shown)


hhugo:
Have you looked at my comment in https://github.com/ocaml-ppx/ocamlformat/issues/2601#issuecomment-2478226513 ? The reason for the changed AST error is that `lib/Literal_lexer.mll` has not been updated.

Julow:
I totally missed that, thanks a lot! Fixed in https://github.com/ocaml-ppx/ocamlformat/pull/2628
The current PR is not good as it doesn't preserve string literals, while your patch does.

PR #2638, opened 2024-12-13
Add flag to control space around high precedence infix

giltho:
Hello,

With version 0.27, and specifically #2580, things like this following code are formatted a bit weird: https://github.com/ocaml-ppx/ocamlformat/blob/5bac2e7f71d9b0a06bd1908dda9b13da1649eee1/bench/test/source.ml#L7371-L7375
As follows: https://github.com/ocaml-ppx/ocamlformat/blob/5bac2e7f71d9b0a06bd1908dda9b13da1649eee1/test/passing/refs.ocamlformat/extensions.ml.ref#L498-L500

It happens that a lot of my code looks like this (in fact this code is mine 🤣 - #1371). 0.27 makes these expressions very hard to read while they are supposed to look like arithmetics (or a OCaml DSL to build an AST).

This PR proposes a flag to disable this new behaviour.

Now, note that I understand where this comes from, and I acknowledge that this is somewhat necessary: ocamlformat rewrites `some_function some_arg (a #+ b)` into `some_function some_arg a #+ b`, and this is also hard to read. Another option from this PR would be to have the following conf instead:
```
high-precedence-operators={compact|parens}
```
where `compact` is the 0.27 choice and parens would do something like
```
some_function some_arg (a #: b)
```
so that precedence is made clear without sacrificing readability, but I have to admit I haven't quite been able to implement it with my limited knowledge of the codebase. Happy to discuss this :)


Julow:
The syntax of your DSL is surprising. For example, these two expressions are the same:
```
  Expr.typeof x #== Lit (Type IntType)
  Expr.typeof (x #== Lit) (Type IntType)
```
Also, ocamlformat doesn't apply disambiguation parentheses or doesn't try to preserve existing parentheses, which makes the DSL code harder to read:
```diff
-  ((Expr.typeof x)#==(Lit (Type IntType))#&&(x#<=(Expr.int 4)))#&&((Expr.int 0)#<x)
+  (Expr.typeof x)#==(Lit (Type IntType))#&&(x#<=(Expr.int 4))#&&((Expr.int 0)#<x)
```

I don't think OCamlformat should render these similarly to infix operators as they clearly parse differently. An option would solve your issue without changing what ocamlformat does by default and the `high-precedence-operators=parens` idea seems even better, as the output will be less confusing to read.
But I wonder if the easiest solution wouldn't be to change your DSL to use a more ocaml-like syntax ?

Julow:
As @v-gb suggested in https://github.com/ocaml-ppx/ocamlformat/pull/2580, I also think that `&&<something else>` would work better. OCamlformat would function better and also you'll get the same precedence as the similar OCaml operators. I mean that `&&$` has the same precedence as `&&` and has a different precedence than `==$`. Also, with the symbol at the end, the connotation problem perhaps goes away ?

v-gb:
This change is probably not needed anymore, given the merge https://github.com/GillianPlatform/Gillian/pull/327 ? Instead, maybe the tests of the layout of `#==` and similar operators should be dropped.

giltho:
Yes, closing this PR :)

PR #2644, opened 2025-01-06
add a Pattern constructor to Extended_ast.t

v-gb:
So it's easier to print patterns, which seems like the main major syntactic category missing from the sum type.

I think the only way to print pattern otherwise is to print a `fun $pat -> whatever` node and chop off the unwanted bits of the string.


Julow (approved):
Looks good :)

v-gb:
Thanks!

Issue #2642, opened 2025-01-06
Question: What does wrap-comments do?


mbarbin:
This is a question about the formatting option `wrap-comments=true`.

I think I am misunderstanding its intent. I thought this would be an option to force inserting newline characters in regular non-docstring comments, so they fit into the margin. However, experimenting with this, this doesn't seem to be the case. I am not sure whether I should file a bug report, or I misunderstood how the option can be used.

Here is my config:

```yaml
version = 0.27.0
profile = janestreet
wrap-comments = true
```

Here was an attempt at writing a long comment, with the hope that it would be automatically wrapped on `fmt`:

```ocaml
(* Hello, this file has [wrap-comments = true] enabled via its [.ocamlformat] config, but I am not sure what changes does it make. This comment for example is not wrapped when this file is formatted. Sounds like the setting is ignored. Or how long should a comment for it to be wrapped? I am keeping this comment here for the time being to monitor whether there'd be changes on this front. *)
```

mbarbin:
Thank you for looking into the issue!

## Asterisk

I built a local version of ocamlformat including this fix:
```sh
$ ocamlformat --version
0.27.0-1-g4c94d48
```

and then tried it out with this config:
```yaml
version=0.27.0-1-g4c94d48
ocaml-version=5.2
profile=janestreet
parse-docstrings=true
wrap-comments=true
```

I confirm now that a file like the following is *stable* through fmt. Thank you!
```ocaml
(* Hello.
 * This is a comment with asterisk chars
 * at the beginning of every lines.
 *
 * Yay!
 *)
let () = ()
```

Incidentally as I was editing the file manually, I came across this case:
```ocaml
(* Hello.
   * This is a comment with asterisk chars
 * at the beginning of every lines.
 *
 * Yay!
 *)
 let () = ()
```
```sh
$ ocamlformat mini_fmt.ml
ocamlformat: Cannot process "mini_fmt.ml".
  Please report this bug at https://github.com/ocaml-ppx/ocamlformat/issues.
  BUG: comment changed.
File "mini_fmt.ml", lines 1-6, characters 0-3:
Error: comment (*  Hello.
   * This is a comment with asterisk chars
 * at the beginning of every lines.
 *
 * Yay!
  *) dropped.
  BUG: comment changed.
File "mini_fmt.ml", lines 1-6, characters 0-2:
Error: comment (*  Hello.
   * This is a comment with asterisk chars
   * at the beginning of every lines.
   *
[...]

PR #2645, opened 2025-01-07
Fix 'wrap-comments' not working with the janestreet profile

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2642
Fix a bug added in https://github.com/ocaml-ppx/ocamlformat/pull/2371


Julow:
I can only spot differences on asterisk-prefixed comments in test-branch. Let's merge!

PR #2648, opened 2025-02-03
nest-match=align now works with match%ext.

EmileTrotignon:
match%ext used to not be treated the same by nest-match-align. This is now fixed.

Requested in Issue #2081 

Seems to be a very lightweight patch, and that the bug was not present at all in the test base.

### DIFF
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -380,7 +380,7 @@ let get_or_pattern_is_nested ~ctx pat =
       not
         (List.exists bindings.pvbs_bindings ~f:(function
           | {pvb_body= Pfunction_cases (cases, _, _); _} -> check_cases cases
-          | _ -> false ))
+          | _ -> false ) )
   | _ -> true
 
 let get_or_pattern_sep ?(cmts_before = false) ?(space = false) (c : Conf.t)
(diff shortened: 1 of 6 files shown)


lib/Params.ml
```diff
@@ -432,6 +432,17 @@ let get_cases (c : Conf.t) ~ctx ~first ~last ~cmts_before
...
+    | ( Pexp_extension
+          ( ext
+          , PStr
+              [ { pstr_loc= _
+                ; pstr_desc=
+                    Pstr_eval
+                      ({pexp_desc= Pexp_match _ | Pexp_try _; pexp_loc; _}, _)
+                } ] )
+      , `Align )
+      when Source.extension_using_sugar ~name:ext ~payload:pexp_loc ->
```
Julow, line 444: This function is used quite often, at some point we could consider putting this information in the `Pexp_extension` node.

EmileTrotignon:
Why does the linux build fail because of lacking sudo priviledges ? I had hoped to see the result of test-branch.

Julow (commented):
Awesome, thanks! The CI is having difficulties, can you rebase on top of https://github.com/ocaml-ppx/ocamlformat/pull/2649 ?

lib/Params.ml
EmileTrotignon: Yes of course, but that turns a simple patch into a complicated project. I don't think my employer would be willing to let me do it.

EmileTrotignon:
I think this is good to go

lib/Params.ml
Julow: I'm mentioning this in case it spawns ideas. Of course, I want this fix to go in first.

Julow (approved):
No diffs with test-branch, let's merge!

PR #2651, opened 2025-02-11
generate better asts for function bindings

v-gb:
Currently:

```ocaml
let foo1 : _ = function () -> ()
let foo2 x : _ = function () -> ()
```

are parsed into these value_bindings:

```ocaml
  (* foo1 *)
  { pvb_args = []
  ; pvb_constraint = Some "_"
  ; pvb_body = Pfunction_cases ...
  }
  (* foo2 *)
  { pvb_args = ["x"]
  ; pvb_constraint = Some "_"
  ; pvb_body = Pfunction_cases ...
  }
```

I expect instead:

```ocaml
  (* foo1 *)
  { pvb_args = []
  ; pvb_constraint = Some "_"
  ; pvb_body = Pfunction_body (Pexp_function ([], None, Pfunction_cases ...))
  }
  (* foo2 (no changes here) *)
  { pvb_args = ["x"]
  ; pvb_constraint = Some "_"
  ; pvb_body = Pfunction_cases ...
  }
```

I think the ast for foo1:

- is confusing
- creates a needless distinction between `let f : _ = function () -> ()` vs `let f : _ = (function () -> ())`, unlike, say, `1 + function () -> ()` vs `1 + (function () -> ())`.
- is essentially an invariant violation. The type of value_bindings in ocamlformat should be understood to be the union of a non-function let-binding + an inline pexp_function node. But the node for foo1 corresponds to the syntax of neither a non-function let-binding (because of body = Pfunction_cases _), nor an inline pexp_function (because pexp_function can't have a type_constraint with an empty list of params).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@ profile. This started with version 0.26.0.
 
 - Fixed `nested-match=align` not working with `match%ext` (#2648, @EmileTrotignon)
 
+- Fixed the AST generated for bindings of the form `let pattern : type = function ...`
+  (#2651, @v-gb)
+
 ## 0.27.0
 
 ### Highlight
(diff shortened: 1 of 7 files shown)


Julow (approved):
This makes total sense :) This also slightly reduce the diff with OCaml's parser and that's a good thing too.

Thanks!

v-gb:
Thanks!

PR #2653, opened 2025-02-14
print valid syntax for the corner case (1).a

v-gb:
Not that I need this syntax, but I was wondering if my use of ocamlformat was adding parens properly, tried this as a tricky example, only to realize that ocamlformat doesn't handle it correctly in the first place. Considering this is handled correctly for Pexp_index_op 5 lines above, might as well handle it here as well.


Julow (approved):
Thanks :)

PR #2654, opened 2025-02-18
Fix let bindings formatting with dummy locations

Julow:
Let bindings with dummy locations formatted like this:

    let* a = let* b = c in

             d in

This happens when constructing new AST nodes and formatting them with ocamlformat-lib.

### DIFF
--- a/lib/Source.ml
+++ b/lib/Source.ml
@@ -15,6 +15,8 @@ open Extended_ast
 (** Concrete syntax. *)
 type t = {text: string; tokens: (Parser.token * Location.t) array}
 
+let is_dummy_pos p = p.Lexing.pos_cnum < 0
+
 let create ~text ~tokens =
   let tokens =
     List.filter tokens ~f:(fun (tok, _) ->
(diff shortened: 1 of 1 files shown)


v-gb:
Just curious if you'd be receptive to other changes that make ocamlformat more robust when printing a modified ast? I have this [pile of changes](https://github.com/ocaml-ppx/ocamlformat/compare/main...v-gb:ocamlformat:r-71260dcc2b65bc6b1aa28ddce2840bdc89d979d6) on my fork of ocamlformat, in particular https://github.com/ocaml-ppx/ocamlformat/commit/2ff6986953cea2952d07ed17e32fe832b32f811f and https://github.com/ocaml-ppx/ocamlformat/commit/c458f6a93c2ca58735a2ec20c15add739f8b3f1d that are doing similar things.

Julow:
Absolutely! I have fixed 2ff6986 in a different way in https://github.com/ocaml-ppx/ocamlformat/pull/2655 but please send the others :)

v-gb:
Ok.

I think your other pull request only addresses one of the two problems addressed in https://github.com/ocaml-ppx/ocamlformat/commit/2ff6986953cea2952d07ed17e32fe832b32f811f.

I need to think about your use of loc_ghost though. I don't think I ever set that, and it seems non-standard (you use it mean "a node that's not produced by the parser" or something?). I use real locations, because otherwise what usually happens if you mad while trying to figure out why ocamlformat is moving comment next to newly introduced ast nodes.

PR #2658, opened 2025-02-27
fix Ast_mapper not mapping over `in` location in `let+ .. in ..`

v-gb:
No idea why a test is impacted, but it seems to be an improvement.

### DIFF
--- a/test/passing/refs.default/sequence-preserve.ml.ref
+++ b/test/passing/refs.default/sequence-preserve.ml.ref
@@ -112,8 +112,8 @@ let _ =
 let foo x y =
   do_some_setup x;
   let* () = do_some_setup y in
-
   (* Empty line after letop *)
+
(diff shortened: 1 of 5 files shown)


test/passing/refs.default/sequence-preserve.ml.ref
```diff
@@ -112,8 +112,8 @@ let _ =
 let foo x y =
   do_some_setup x;
   let* () = do_some_setup y in
-
   (* Empty line after letop *)
+
```
Julow, line 116: This change is totally fine. Comments are parsed separately associated to closest locations collected from the AST using the `Ast_mapper`. This comment is closer to where it was in the original source (just after the `in` and not before the next expression) and that's what we want.

Julow (approved):
Thanks!

PR #2659, opened 2025-02-27
fix a few failures when printing a modified AST

v-gb:


v-gb:
The failure might be because of carriage returns on windows. But before figuring out what to do, I'd like to know if the approach is reasonable.

Also it might be useful to see what comes out of https://github.com/ocaml/ocaml/pull/13833, because the parser is producing loc_ghost in cases that don't make sense to me, so I'm not sure that the use of loc_ghost to distinguish modifications made to the AST always work (I got a strange test change when using loc_ghost in `Fmt_ast.sequence_blank_line`).

lib/Cmts.ml
```diff
@@ -365,13 +365,16 @@ let relocate_ext_cmts (t : t) src (pre, pld) ~whole_loc =
   | PStr [{pstr_desc= Pstr_eval _; pstr_loc; _}] ->
       let kwd_loc =
         match Source.loc_of_first_token_at src whole_loc LBRACKETPERCENT with
-        | Some loc -> loc
+        | Some _ as o -> o
```
Julow, line 368: Perhaps `Source.loc_of_first_token_at` shouldn't attempt to read the source at all and always return `None` for ghost loc ? I don't see other `impossible` cases related to these functions returning `None`.

Also, this `relocate_ext_cmts` shouldn't raise errors at all as it's just using heuristics to move comments around. The `impossible` case should be removed.

v-gb: In general, I think the use of loc_ghost is not unreasonable, but the most principled solution would be to make the printer work using the ast only, without the side information from `Source.t` or `Cmts.t`.

This can be seen with this thought experimental: imagine that you're combining files with `print (parse Structure file1 @ parse Structure file2)`. In principle, this should have the same effect as `In_channel.read_all file1 ^ "\n" ^ In_channel.read file2`. But in practice, either you set loc_ghost on one ast and lose part of the formatting, or you don't and instead risk the printer raising exceptions.

Whereas if the ast was self contained, the concatenated asts would get printed the same as the concatenation of the source files, modulo the whitespace at the boundary. I don't know how complex that would be, but in this function, it would probably mean that the Pexp_extension should have a loc_of_first_token field.

Julow: Right!

> but the most principled solution would be to make the printer work using the ast only, without the side information from Source.t or Cmts.t.

That was once a goal but we're not there yet as it's a lot of work. The modified AST is closer to the concrete syntax than the upstream parser but not enough to stop reading the source yet.
Any work in that direction is very welcome :)

Comments is a much more complex story. Representing comments at every nodes in the AST was deemed too hard since the beginning.

> This can be seen with this thought experimental: imagine that you're combining files with print (parse Structure file1 @ parse Structure file2).

In theory, you can concatenate the `Cmts.t` as well and not loose the comments.
For the source part, we could make use the filename field of a location but that's just adhoc. We definitely want to get rid of it.

But keep in mind that OCamlformat is a `string -> string` function. In a [much simpler use-case](https://github.com/Julow/lwt_ppx_to_let_syntax/blob/main/lib/ocamlformat_utils/ocamlformat_utils.mli#L13-L18) than yours I made it a `string -> (ast -> ast) -> string` function and haven't had to deal with source and comments. Could that help ?

test/unit/test_fmt_ast.ml
```diff
@@ -0,0 +1,188 @@
...
+  !a;
+  a + b
+|outer}
+        in
+        (* Ideally we'd improve two things about this test:
+
+           - check the new string parses, to the same AST as the original one
+           - use ppx_expect, so we have a nicer workflow and more readable
+           errors *)
+        if String.( <> ) expected printed_ast_replaced then (
```
Julow, line 183: This test is failing on 4.14. You should use [check string "equals" expected printed_ast_replaced](https://ocaml.org/p/alcotest/latest/doc/Alcotest/index.html#val-check) so we can see why.

test/unit/test_fmt_ast.ml
```diff
@@ -0,0 +1,188 @@
...
+  a.!(b);
+  a.!(b) <- c;
+  !a;
+  a + b
+|outer}
+        in
+        (* Ideally we'd improve two things about this test:
+
+           - check the new string parses, to the same AST as the original one
+           - use ppx_expect, so we have a nicer workflow and more readable
```
Julow, line 181: I'd prefer cram tests. `ppx_expect` requires defining a library in Dune, which becomes installable and is annoying to deal with during releases. You can build an executable with Dune that uses ocamlformat-lib within a cram test.

Julow: Libraries become installable (even private libraries) when they have a `(package)` field. This field might not be needed for ocamlformat right now but I had enough problems with `ppx_expect` and so much success with cram tests that I really don't want to go back.

lib/Cmts.ml
Julow: > it only requires capturing the relevant information from the concrete syntax into the tree.

That's exactly what we called "closer to the concrete syntax" so far. What we called "CST" in older issues is not a tree full of lexer tokens and snippets of source code.

We found that having different nodes that represent things with similar semantic but different syntax to be extremely useful (Pexp_function, Pexp_let, Pexp_letop or Pexp_apply, Pexp_infix, Pexp_prefix, Pexp_indexedop).
We could change ocamlformat's AST again if that could help your usecase, though we don't want to make ocamlformat harder to maintain.

> because the position between in and the tuple are 2 lines apart, and yet every node has loc_ghost=false.

I don't think `loc_ghost` is relevant here. This is a case where we tried to not use `Source` and used line numbers instead. To fix this issue, we could perhaps record whether there's an empty line by adding a boolean into the `Pexp_let`.

PR #2663, opened 2025-03-07
fix issue 2660

### LINKED ISSUE #2660: Bug: max-ident=2 with let%bind
With `--config=profile=default,max-indent=2`:

```ocaml
let v =
  let%bind
    {
  eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee;
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
} =
    4
  in
  4
```

This was introduced in #2615

EmileTrotignon:
fix #2660 

Not the nicest implement, but it is correct and in the spirit of the previous code.

### DIFF
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -679,7 +679,8 @@ let get_list_expr (c : Conf.t) =
 let get_array_expr (c : Conf.t) =
   collection_expr c ~space_around:c.fmt_opts.space_around_arrays.v "[|" "|]"
 
-let box_pattern_docked (c : Conf.t) ~ctx ~space_around ~pat opn cls k =
+let box_pattern_docked (c : Conf.t) ~ctx ~ext_length ~space_around ~pat opn
+    cls k =
   let space = if space_around then 1 else 0 in
   let indent_opn, indent_cls =
     match (ctx, c.fmt_opts.break_separators.v) with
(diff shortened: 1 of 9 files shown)


lib/Params.ml
```diff
@@ -689,38 +690,39 @@ let box_pattern_docked (c : Conf.t) ~ctx ~space_around ~pat opn cls k =
     | Ast.Exp {pexp_desc= Pexp_let ({pvbs_bindings; _}, _, _); _}, _
       when List.exists pvbs_bindings ~f:(fun b -> phys_equal b.pvb_pat pat)
       ->
-        (-4, 0)
+        (-4 - ext_length, 0)
```
Julow, line 693: You might not need to propagate `ext_length` as the information is in the context in the corresponding binding's `pvb_attributes` field.

EmileTrotignon: I will try it

EmileTrotignon: Its done

Julow (approved):
Thanks a lot !

PR #2664, opened 2025-03-12
`begin fun x ->` now can be printed on a single line

EmileTrotignon:
```ocaml
map x 
  begin 
    fun y ->
      print x;
      y
  end
```
is now 
```ocaml
map x 
  begin fun y ->
    print x;
    y
  end
```

This is a nice improvement that partially solves one of the biggest issue we have with ocamlformat at ahrefs.

A further improvement could be 

```ocaml
map x begin fun y ->
  print x;
  y
end
```
but thats more debatable and should probably be in another PR.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2893,6 +2893,21 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
   | Pexp_indexop_access x ->
       pro $ fmt_indexop_access c ctx ~fmt_atrs ~has_attr ~parens x
   | Pexp_hole -> pro $ hvbox 0 (fmt_hole () $ fmt_atrs)
+  | Pexp_beginend
+      ( { pexp_desc= Pexp_function (args, typ, body)
+        ; pexp_attributes= function_attrs
+        ; _ } as f ) ->
+      let wrap_intro x =
+        hvbox 0
+          ( hvbox 0 (str "begin" $ fmt_extension_suffix c ext $ fmt_atrs)
+          $ break 1 2 $ x )
+        $ break 1000 0
+      in
+      hvbox 2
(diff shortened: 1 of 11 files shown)


lib/Fmt_ast.ml
```diff
@@ -2893,6 +2893,21 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
...
+        ; pexp_attributes= function_attrs
+        ; _ } as f ) ->
+      let wrap_intro x =
+        hvbox 0
+          ( hvbox 0 (str "begin" $ fmt_extension_suffix c ext $ fmt_atrs)
+          $ break 1 2 $ x )
+        $ break 1000 0
+      in
+      hvbox 2
+        (fmt_function ~wrap_intro ~box ~ctx:(Exp f) ~ctx0 ~label:Nolabel
```
Julow, line 2907: This makes the formatting of `Pexp_function` too complex by duplicating codes with subtle changes. `fmt_function` is extremely complex for exactly this reason.
This won't scale as more `begin..end` special cases are added (for example https://github.com/ocaml-ppx/ocamlformat/pull/2666) and might create a mess.
This could be solved by passing the `~pro` attribute. I have tried this in this branch but there are still some unwanted diffs: https://github.com/ocaml-ppx/ocamlformat/compare/main...Julow:ocamlformat:2664_passing_pro?expand=1
Can you have a look ?

EmileTrotignon: This is adressed.

EmileTrotignon:
The code is now quite clean and has no regressions.

I included some small rename for the code introduced by #2666 , in order to have consistent naming with the one in this PR.

In my opinion this is ready to be merged.

Julow (approved):
Awesome! It's nice that the ground work for nice begin..end formatting is here, future changes will hopefully be easier.

Julow:
Thanks !

Issue #2665, opened 2025-03-13
Bug: `begin%e try` AST change


EmileTrotignon:
The following code has an AST change : 

```ocaml
let a =
  begin%e try f x i with
  | A -> f aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  | B -> bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  end
```

it gets transformed into:

```ocaml
let a =
  begin%e
    try%e f x i with
    | A -> f aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    | B -> bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  end
```

EmileTrotignon:
fixed in #2666

PR #2666, opened 2025-03-13
overhaul begin match formatting

EmileTrotignon:
```ocaml
let a =
  begin
    match f x i with
    | A -> f aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    | B -> bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  end
```

is now 

```ocaml
let a =
  begin match f x i with
  | A -> f aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  | B -> bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  end
```

This is very frequent in our codebase at ahrefs.

also fixes issue #2665 

I wanted to also do `try` in this PR, but doing it breaks the special case of a `try` that has only one branch. More code could be written so that it doesn't, but that would be another PR.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2564,6 +2564,10 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
            && c.conf.fmt_opts.break_cases.v <> `All
            && c.conf.fmt_opts.break_cases.v <> `Vertical ) ->
       let cmts_before = Cmts.fmt_before c ?eol pexp_loc in
+      let pro_outer, pro_inner =
+        if Params.Exp.box_pro_with_match ~ctx0 ~parens then (noop, pro)
+        else (pro, noop)
+      in
       (* side effects of Cmts.fmt_before before [fmt_pattern] is important *)
       let xpc_rhs = sub_exp ~ctx pc_rhs in
       let leading_cmt = Cmts.fmt_before c pc_lhs.ppat_loc in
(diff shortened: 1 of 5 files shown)


lib/Fmt_ast.ml
```diff
@@ -2893,6 +2894,36 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
...
+            ( ext_inner
+            , PStr
+                [ ( { pstr_desc=
+                        Pstr_eval
+                          (({pexp_desc= Pexp_match (e0, cs); _} as e1), _)
+                    ; pstr_loc= _ } as stru ) ] )
+      ; _ }
+    when Source.extension_using_sugar ~name:ext_inner ~payload:e1.pexp_loc ->
+      let ctx = Str stru in
+      let xexp = sub_exp ~ctx e1 in
```
Julow, line 2919: I'd really want to avoid the code duplication. This should be moved to a new function.

Julow (commented):
`begin .. end` formatting really needs some care, thanks :) Can you also add a change entry ?

EmileTrotignon:
I think this is good to go.

test/passing/refs.default/exp_grouping.ml.ref
```diff
@@ -375,13 +375,14 @@ let x =
...
-          begin [@warning "-3"]
-            try
-              let path', dist = Dijkstra.shortest_path pgraph pv1 pv2 in
-              let path = unwrap_path path' in
-              Tbl.set dist_tbl ~key:(pv1, pv2) ~data:(path, dist);
-              Some (path, dist)
-            with Not_found | Not_found_s _ -> None
+          begin
+          [@warning "-3"]
+          try
```
Julow, line 380: I feel like this could be fixed by checking whether there are attributes in the special case for `Pexp_match`.

Julow (commented):
This is also a nice improvement to the code! I just have one last comment and then I think it's good to merge.

Julow (approved):
Awesome :) This really improves the output and the code so let's merge !

This also moves comments:
```diff
 let _ =
-  begin
-    (* foo *)
-    match foo with
-    | Fooooooooooooooooooooo -> ()
-    | Fooooooooooooooooooooo -> ()
-    | Fooooooooooooooooooooo -> ()
+  (* foo *)
+  begin match foo with
+  | Fooooooooooooooooooooo -> ()
+  | Fooooooooooooooooooooo -> ()
+  | Fooooooooooooooooooooo -> ()
   end
```
But I think it's fine given that this shorten the syntax a lot. If that's a problem, it should be easy to prevent moving the comment in `fmt_beginend`.

PR #2671, opened 2025-03-26
display `lazy begin` on one line.

EmileTrotignon:
I had to improve `fmt_beginend`. Sequences like `lazy begin fun` or `lazy begin match` are also printed on one line.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2916,21 +2916,18 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
       $ fmt_atrs
 
 and fmt_lazy c ~ctx ?(pro = noop) ~fmt_atrs ~ext ~parens e =
-  let lazy_ = str "lazy" $ fmt_extension_suffix c ext in
-  match e with
-  | {pexp_desc= Pexp_beginend _; _} as e ->
-      pro
-      $ hvbox 2
-          (Params.Exp.wrap c.conf ~parens
-             ( fmt_expression c ~pro:(lazy_ $ str " ") (sub_exp ~ctx e)
-             $ fmt_atrs ) )
-  | _ ->
-      pro
-      $ hvbox 2
-          (Params.Exp.wrap c.conf ~parens
-             ( lazy_ $ space_break
-             $ fmt_expression c (sub_exp ~ctx e)
-             $ fmt_atrs ) )
+  let lazy_ = str "lazy" $ fmt_extension_suffix c ext $ space_break in
+  let kw_outer, kw_inner =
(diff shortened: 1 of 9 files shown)


lib/Fmt_ast.ml
```diff
@@ -2914,26 +2906,46 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
...
       $ hvbox 0
           (fmt_expression c ~box ?eol ~parens:true ~indent_wrap ?ext
              (sub_exp ~ctx e) )
       $ fmt_atrs
 
-and fmt_beginend c ?(box = true) ?(pro = noop) ~ctx ~fmt_atrs ~ext
+and fmt_lazy c ~ctx ?(pro = noop) ~fmt_atrs ~ext ~parens e =
+  let lazy_ = str "lazy" $ fmt_extension_suffix c ext in
+  match e with
+  | {pexp_desc= Pexp_beginend _; _} as e ->
```
Julow, line 2921: Duplicating the code this way is not sustainable. The difference between the two cases is the usual `inner, outer` thing so the code can easily be written with more sharing.

lib/Fmt_ast.ml
```diff
@@ -2914,26 +2906,46 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
...
+      pro
+      $ hvbox 2
+          (Params.Exp.wrap c.conf ~parens
+             ( fmt_expression c ~pro:(lazy_ $ str " ") (sub_exp ~ctx e)
```
Julow, line 2925: It would be nice to add a comment explaining that `str " "` plays a role in formatting `lazy begin fun .... ->`. For `lazy begin match` of `lazy begin fun` with small argument list, `space_break` should be fine.

Julow (commented):
I have a comment about the code. The change looks good to me :)

lib/Fmt_ast.ml
EmileTrotignon: Space break is indeed fine, I just need to display the `pro` in the same box as the `begin` keyword.
I dont think `lazy begin fun` with a large list of arguments is a big concern.

lib/Fmt_ast.ml
EmileTrotignon: This is done.

lib/Fmt_ast.ml
EmileTrotignon: Alright, I tested `lazy begin fun`, and its indeed pretty ugly, so I restored to `str " "`, and added a comment.

Julow (approved):
Perfect, let's merge!

PR #2672, opened 2025-03-26
format `begin if` on one line

EmileTrotignon:
Changes where quite small for this one.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2380,8 +2380,7 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
       let parens = parens || has_attr in
       let cmts_before = Cmts.fmt_before c ?eol pexp_loc in
       let pro_outer, pro_inner =
-        if Params.Exp.ifthenelse_inner_pro c.conf ~parens ~ctx0 then
-          (noop, pro)
+        if Params.Exp.ifthenelse_inner_pro ~parens ~ctx0 then (noop, pro)
         else (pro, noop)
       in
       let last_loc =
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -1948,7 +1948,8 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
     (* Some expressions format the 'pro' and comments differently. *)
     let cmts_in_pro =
       match exp.pexp_desc with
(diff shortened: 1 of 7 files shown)


lib/Fmt_ast.ml
```diff
@@ -2376,6 +2377,12 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
         @@ wrap_if outer_parens (str "(") (str ")")
         @@ (fmt_longident txt $ Cmts.fmt_within c loc $ fmt_atrs)
   | Pexp_ifthenelse (if_branches, else_) ->
+      let parens = parens || has_attr in
+      let cmts_before = Cmts.fmt_before c ?eol pexp_loc in
+      let pro_outer, pro_inner =
+        if Params.Exp.ifthenelse_inner_pro ~parens ~ctx0 then (noop, pro)
```
Julow, line 2383: I think `pro_inner` is where the pro should be in every cases. This seems to create a regression with `--if-then-else=fit-or-vertical` but I think that option is the special case, not the context.

EmileTrotignon: It seemed to me that making less changes would be less likely to lead to regressions

EmileTrotignon: Anyway, I made the change

lib/Params.ml
```diff
@@ -328,6 +328,14 @@ module Exp = struct
       | Str _ | Lb _ | Exp {pexp_desc= Pexp_ifthenelse _ | Pexp_let _; _} ->
           false
       | _ -> true
+
+  let ifthenelse_inner_pro (c : Conf.t) ~parens ~ctx0 =
+    if parens then false
+    else if Stdlib.(c.fmt_opts.if_then_else.v = `Fit_or_vertical) then
+      match ctx0 with
+      | Exp {pexp_desc= Pexp_beginend _; _} -> true
```
Julow, line 336: Did you add this special case because you use the `if_then_else` option ? I'd really like to remove any special case here. I think it's better to change how `if_then_else` is implemented and to put the pro at the inner position without special cases.

EmileTrotignon: No, it was just to fix the regression that only showed up with fit-or-vertical. I noticed this regression was because when you had an `if` inside another, `pro` would be used.

Julow: I much prefer the new version! It makes an existing special case explicit instead of adding an other one on top. This will help whoever wants to refactor how ifthenelse are formatted in the future.

Julow (approved):
Thanks! Let's merge.

PR #2673, opened 2025-03-28
`begin if` bugfix

EmileTrotignon:
There was a bug in the formatting of `begin if` that I just introduced : 

```ocaml

let () =
  f
    begin if
            a___________________________________________________________________
          then
      b_________________________________________________________________
    else
      c_________________________________________________________________
    end
```

This PR fixes that bug.


Julow (approved):
Looks good!

PR #2675, opened 2025-04-02
make the formating of `map li begin fun` the same as `map li ( fun`.

EmileTrotignon:
As written above.  Implementation was annoying, because I had to touch `fmt_function` a little bit.

This is not as clean as the other `begin` PRs, but I don't think I have a choice, because I am trying to replicate a formatting that is already an exception. I tried to share the maximum amount of code.


EmileTrotignon:
I added no changelog because it seems to me that this change is actually already included in describing the `begin fun` change.

EmileTrotignon:
#2693 should be a better implementation.

Issue #2652, opened 2025-02-14
Style suggestion: style ppx-extended function declarations like regular function declarations


lord:
**Current formatting**
```
let _ =
  my_func ~hello ~world
    ~f:
      (fun%my_ppx x ->
        let x = 1 + 2 in
        x)
```

This formatting is similar for `function%my_ppx` as well.

**Describe the formatting you'd like**
Ideally indentation would match what the fun declaration would look like without the %my_ppx, which in the default profile looks like this:

```
let _ =
  my_func ~hello ~world ~f:(fun x ->
      let x = 1 + 2 in
      x)
```

**Additional context**
I guess the fact that these are formatted differently almost feels like a bug? Not sure why any user would want unique behavior for these kinds of function declarations, but maybe I don't know how ppxes are being used here

EmileTrotignon:
What is happening here is that the ast for `map li (fun x -> y)` is 
```
apply
  ( ident "map"
  , [ ident li
    ; function (ident x, ident y)
    ]
  )
```

but the ast for `map li (fun%e x -> y)` is

```
apply
  ( ident "map"
  , [ ident li
    ; extension ("e", function (ident x, ident y))
    ]
  )
```

There is a special formatting for "apply where the last argument is a function definition" which you saw without the extension. Alas, it is not triggered for "apply where the last argument is an extension node".

So that was the bad news. The good news is that I opened a [PR ](https://github.com/ocaml-ppx/ocamlformat/pull/2676) to fix this exact issue. It works by making the AST of  `map li (fun%e x -> y)` like that :

```
apply
  ( ident "map"
  , [ ident li
    ; function (Some "e", ident x, ident y)
    ]
  )
```

I will add your example in the tests, and this should be closed as completed when #2676 is merged.

EmileTrotignon:
I added your example in the following commit : https://github.com/ocaml-ppx/ocamlformat/pull/2676/commits/ba6e8cdc4188356c87dcd7325d0c497efeb089d7

You can see in the diff that the formatting is indeed produced by my PR.

PR #2676, opened 2025-04-04
infix attrs and expression are encoded in the AST

EmileTrotignon:
This is a nice refactoring, solving a bunch of problem at once.

Shortcut syntax for extension and attributes is now preserved universally across expressions which wasn't the case before.

`begin ... end` syntax is always preserved no matter the settings, if there are attributes or an extension attached to it.

`begin end` syntax for `()` is also now preserved, but I doubt that this will be noticed by a lot of people.


EmileTrotignon:
Solves #2652

EmileTrotignon:
I am merging this because I need for some other PR, but its still worth reviewing after the fact in my opinion

PR #2678, opened 2025-04-07
bugfix : Double parens functor constraint

EmileTrotignon:
There used to be a bug where functor application with a constraint had double parens :

```ocaml
module M = F ((A : T))
```

This PR fixes it.

```ocaml
module M = F (A : T)
```


Julow (commented):
Needs formatting! Feel free to merge if test_branch reports no diffs.

PR #2680, opened 2025-04-09
Cmdliner 2.0.0 compat

### LINKED ISSUE #2667: cmdliner 2.0.0 support
It seems that the latest version of your package on the opam repository will not compile with the upcoming cmdliner 2.0.0 release.

This is either because you are using deprecated identifiers that are removed in 2.0.0 or because you still use argument converters as pairs which has been (sadly only silently) deprecated since 2017.

There are a few instruction [here](https://github.com/dbuenzli/cmdliner/issues/206) on how to make it compatible. If you run into problem or need more help use this issue, I'm listening.

emillon:
Fixes #2667

This adds compatibility with cmdliner 2.0.0 (while keeping the existing code compatible with older versions).

The 2 changes are:
- the `conv` type becomes abstract, so we use `Arg.conv`, `Arg.conv_parser`, and `Arg.conv_printer` where necessary
- unambiguous prefixes are not selected anymore, so we use the full names (`--max-iters` instead of `max-iter`, etc) in tests

This can be tested by pinning cmdliner to its dev branch and ocp-indent to OCamlPro/ocp-indent#328.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,16 @@ profile. This started with version 0.26.0.
 
 ## unreleased
 
+### Deprecated
+
+- Starting in this release, ocamlformat can use cmdliner >= 2.0.0. When that is
+  the case, the tool no longer accept unambiguous option names prefixes. For
+  example, `--max-iter` is not accepted anymore, you have to pass the full
+  option `--max-iters`. This does not apply to the keys in the `.ocamlformat`
+  configuration files, which have always required the full name.
+  See dbuenzli/cmdliner#200.
+  (#2680, @emillon)
+
 ### Fixed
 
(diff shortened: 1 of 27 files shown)


EmileTrotignon:
I added a no changelog label, but actually I think there needs to be a changelog entry to clarify that you can't use unambiguous prefixes anymore. 

Otherwise this looks good to be merged.

emillon:
To be precise, this PR does not disallow unambiguous prefixes. Using `--max-iter` continues to work, but only under cmdliner 1.x. (I don't have a good idea about how to communicate that to users of ocamlformat)

EmileTrotignon:
Indeed, that's a little annoying, maybe say its deprecated as a general warning, and then explain the details ?

EmileTrotignon:
LGTM

PR #2681, opened 2025-04-09
Label arg begin

EmileTrotignon:
fix formatting of `f ~arg:begin ... end ~arg2`:

instead of

```ocaml
f ~arg
  ~arg2:
    begin
      ...
    end
  ~arg3
```
you get: 
```ocaml
f ~arg
  ~arg2:begin
    ...
  end
  ~arg3
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,9 +35,9 @@ profile. This started with version 0.26.0.
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


Julow (approved):
Looks good to me :) Feel free to merge this one when you are happy with it.

PR #2682, opened 2025-04-10
break-struct apply on sig

EmileTrotignon:
The `break-struct=natural` setting allows `struct ... end` syntax to be printed on one line.

However, it did not apply to `sig ... end`, which is a very similar syntax. I made it more consistent, even if it does not exactly fit the name. 

I also switched the code I touched from `fmt_or` to `if then else`, which does the same thing but without confusing you.

### DIFF
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -588,9 +588,9 @@ module Formatting = struct
     let names = ["break-struct"] in
     let all =
       [ Decl.Value.make ~name:"force" `Force
-          "$(b,force) will break struct-end phrases unconditionally."
+          "$(b,force) will break struct-end and sig-end phrases unconditionally."
       ; Decl.Value.make ~name:"natural" `Natural
-          "$(b,natural) will break struct-end phrases naturally at the \
+          "$(b,natural) will break struct-end and sig-end phrases naturally at the \
            margin." ]
     in
     Decl.choice ~names ~all ~default ~doc ~kind
--- a/lib/Conf.ml
+++ b/lib/Conf.ml
@@ -588,10 +588,11 @@ module Formatting = struct
     let names = ["break-struct"] in
     let all =
       [ Decl.Value.make ~name:"force" `Force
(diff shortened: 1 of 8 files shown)


test/passing/tests/functor.ml
```diff
@@ -110,3 +110,11 @@ module M : functor (A : S) -> functor (B : S) -> S = N
 module M : functor (A : S) (B : S) -> S = N
 module M : functor (A : S) -> functor (B : S) -> S = N
 module M : (A : S) -> functor (B : S) -> S = N
+
+[@@@ocamlformat "break-struct=natural"]
+
+module M = F ((struct type t end : sig type t end))
+
+module M = struct type t end
```
Julow, line 118: Are the tests testing anything ?

lib/Conf.ml
```diff
@@ -588,10 +588,11 @@ module Formatting = struct
     let names = ["break-struct"] in
     let all =
       [ Decl.Value.make ~name:"force" `Force
-          "$(b,force) will break struct-end phrases unconditionally."
+          "$(b,force) will break struct-end and sig-end phrases \
```
Julow, line 591: The doc is quite hard to read now. What do you think of this change: https://github.com/ocaml-ppx/ocamlformat/compare/main...Julow:ocamlformat:2682-man

test/passing/tests/functor.ml
EmileTrotignon: Yes, line 116 would be broken in multiple lines before this PR. line 118 is shere to showcase the inconsistency when that happens.

lib/Conf.ml
EmileTrotignon: I like your change.

EmileTrotignon: I have pushed your change

test/passing/tests/functor.ml
Julow: I see!

Julow (approved):
Let's merge!

PR #2683, opened 2025-04-10
fix floating doc comments interaction with `;;` expressions

EmileTrotignon:
```ocaml
a ;;

(** doc *)

b ;;
```

was formatted into 

```ocaml
a

(** cmt *) ;;

b
```

which was a syntax error.

The reason for that is the logic to add `;;` is to add it at the end of every struct item that is followed by a `Pstr_eval`.

`a` was followed by a docstring (`Pstr_attribute`), but the docstring was followed by a `Pstr_eval`.

To fix it, I changed the logic of adding `;;`. Now, `;;` is added:

- at the end of every `Pstr_eval`, except if its at the end of the struct.
- at the end of every struct item that is followed by a `Pstr_eval`, except for floating doc comments, they do not need a `;;`.

This results in a bit more `;;` being inserted because of the first condition, but to me it makes what is happening a lot clearer, and it fixes a crash.  

The diff is large in the tests, but I don't think it will be that large in real codebases.


EmileTrotignon:
This crash was found in test-branch.

EmileTrotignon:
This is has small implementation. It has mid-range impact, which is in my opinion a straightforward improvement. I am merging.

PR #2684, opened 2025-04-11
fix: handle case where doc does not lex as ocaml

### LINKED ISSUE #2679: Bug: crash when backslash is in mld file
Hi,
When a mld file contains a backslash, ocamlformat crashes with a lexing error.

For example, add `{m \Gamma}` to `test/passing/tests/doc.mld` and it will crash.

The error is:

```
ocamlformat: ignoring "doc.mld" (syntax error)
File "doc.mld", line 166, characters 3-4:
166 | {m \Gamma}
         ^
Error: Illegal character (\\)
```

This is the error from `parser-extended/lexer.mll`, which is unexpected because this shouldn't be lexed as ocaml code.
(the same comment in a docstring works)

Thanks

emillon:
Fixes #2679

Formatting `mld` files can crash when the input file cannot be lexed as ocaml. This can happen with LaTeX is found in `{m ...}` or C code in `{@c[ ... ]}`, for example.

In that case we would crash when building the list of tokens in the file; however it's not used in the `mld` case.


EmileTrotignon:
I don't really like that we catch the exception even when we are not parsing documentation, but I checked and we cannot match on fragment at this stage. Maybe an `is_documentation` bool argument could fix it ? 

Still I think its okay to merge as-is, the exception will be raised elsewhere when truly parsing an ocaml file.

emillon:
yeah I agree it's not ideal. we can't really make the `tokens` field lazier because it needs the file open, so making it optional would be nice but I'm not sure how to wire that to this call site.

EmileTrotignon (approved):

Julow (approved):
Let's merge!

PR #2685, opened 2025-04-11
`|> begin` is now printed on a single line

EmileTrotignon:
Implementation is very short.


EmileTrotignon:
No diff on test branch, because no begin end there.

Issue #2650, opened 2025-02-07
Style suggestion: Strange interaction between |> and anonymous function


NoahTheDuke:
**Current formatting**
```ocaml
let tests () =
  [
    t "calculate_hand" (fun () ->
        let a_hand_value = Alcotest.testable Hand.pp_hand_value Hand.equal_hand_value in

        let hand = Hand.create (Card.create Red Sylop) (Card.create Yellow Sylop) in
        check a_hand_value ~pos "heck" Hand.PureSabacc (Game.calculate_hand hand);

        Range.from 1 6
        |> Range.iter ~f:(fun i ->
               List.iter
                 (fun (rank : Rank.t) ->
                   let hand =
                     Hand.create (Card.create Red rank) (Card.create Yellow Sylop)
                   in
                   check a_hand_value ~pos "heck" (Hand.PairSabacc i)
                     (Game.calculate_hand hand))
                 [ Value i; Imposter (Some i) ]);
        ());
  ]
```

**Describe the formatting you'd like**
```ocaml
        Range.from 1 6
        |> Range.iter ~f:(fun i ->
             List.iter
               (fun (rank : Rank.t) ->
                 let hand =
                   Hand.create (Card.create Red rank) (Card.create Yellow Sylop)
                 in
                 check a_hand_value ~pos "heck" (Hand.PairSabacc i)
                   (Game.calculate_hand hand))
               [ Value i; Imposter (Some i) ]);
```

**Additional context**
It's strange to me that the `List.iter` and all subsequent lines are indented so heavily. I know that `|>` has some strange formatting issues, but 7 spaces from the start of `|>` is excessive in my eyes.

I'm not sure how best to format this, truth be told, but this is quite cumbersome in practice.

EmileTrotignon:
The issue is that the formatting is actually four space, but its aligned from the start of `Range.iter` instead being aligned on the start of the line.

I attempted a fix independently, but it had a bunch of regressions, I am not sure if its worth to put more effort, the formatting of functions is a very complicated piece of code, full of exception and special cases.

EmileTrotignon:
I tried again to fix this, and it appears that it was actually simpler than I thought. In some case it looks a bit ugly, but I think its the correct thing to do.

NoahTheDuke:
Thanks for figuring this out, I think the changes are worthwhile.

PR #2686, opened 2025-04-11
`|> map (fun` is indented correctly

EmileTrotignon:
from

```ocaml
let _ =
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  |>>>>>> map (fun xxxxxxx ->
              xxxxxxxxxxxxxxxxxxxxxxx ;
              aaaaaaaaaAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAaaaaaa )

```

to

```ocaml
let _ =
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  |>>>>>> map (fun xxxxxxx ->
      xxxxxxxxxxxxxxxxxxxxxxx ;
      aaaaaaaaaAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAaaaaaa )
```

In my opinion this is quite a big fix. Implementation is relatively simple.

Fixes issue #2650

### DIFF
--- a/test/passing/refs.default/js_source.ml.ref
+++ b/test/passing/refs.default/js_source.ml.ref
@@ -700,7 +700,7 @@ let _ =
 let _ =
   fooooooooooooooooooooooooooooooo
   |> fooooooooooooooooooooooooooooooo ~fooooooooooooooooooooooooooooooo
-       ~fooooooooooooooooooooooooooooooo:(fun foo -> bar)
+    ~fooooooooooooooooooooooooooooooo:(fun foo -> bar)
 
 let _ =
   fooooooooooooooooooooooooooooooo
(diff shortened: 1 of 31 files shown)


EmileTrotignon:
There is no diff on the janestreet profile by the way, thats weird. Did they have cursed logic to simulate this that we could remove ?

test/passing/refs.default/js_source.ml.ref
```diff
@@ -712,7 +712,7 @@ let _ =
   fooooooooooooooooooooooooooooooo
   |> fooooooooooooooooooooooooooooooo ~fooooooooooooooooooooooooooooooo
        ~fooooooooooooooooooooooooooooooo:(fun foo ->
-         match bar with Some _ -> foo | None -> baz)
+      match bar with Some _ -> foo | None -> baz)
```
EmileTrotignon, line 715: this is unfortunate, but in the spec: the body of the function is indented 4 characters from `|>`

test/passing/refs.default/max_indent.ml.ref
```diff
@@ -1,25 +1,25 @@
...
 
 let () =
   fooooo
   |> List.iter
        (fun some_really_really_really_long_name_that_doesn't_fit_on_the_line ->
-       let x =
-         some_really_really_really_long_name_that_doesn't_fit_on_the_line $ y
-       in
-       fooooooooooo x)
+    let x =
```
EmileTrotignon, line 11: unfortunate but also in spec

test/passing/refs.default/js_source.ml.ref
EmileTrotignon: It's now fixed, at the price of a little ugliness when using `max-indent=2`, but that ugliness was already existing when doing `map li ~f:(fun` without `|>`.=, so its consistent.

test/passing/refs.ocamlformat/max_indent.ml.ref
```diff
@@ -7,7 +7,7 @@ let () =
 let () =
   fooooo
   |> List.iter
-       (fun some_really_really_really_long_name_that_doesn't_fit_on_the_line ->
+    (fun some_really_really_really_long_name_that_doesn't_fit_on_the_line ->
     let x =
```
EmileTrotignon, line 11: Not exactly great, but the exact same thing already happens when doing `map ~f (fun` without `|>`.

EmileTrotignon: I added a test to showcase this.

EmileTrotignon:
> There is no diff on the janestreet profile by the way, thats weird. Did they have cursed logic to simulate this that we could remove ?

Alright, I finally read how this was implemented, this PR was a big waste of time

PR #2687, opened 2025-04-11
promote test from #2678 and update changelog

EmileTrotignon:
I had forgotten to promote test from #2678, and the changelog entry needed `\*`.


EmileTrotignon:
I am merging, this is a hotfix that does not require review.

PR #2689, opened 2025-04-16
no new line in `@@ match`

EmileTrotignon:
```ocaml
Lwt.run
@@ match x with
| A -> a
| B -> b
```
instead of 
```ocaml
Lwt.run
@@
match x with
| A -> a
| B -> b
```

This is good because there is one less newline, and because it is more consistent with the `@@ function` formatting:

```ocaml
Lwt.run
@@ function
| A -> a
| B -> b
```

There are two issues that need to be addressed before merging:

- with operator `||`, it looks weird:
```ocaml
my_bool
|| match x with
| A -> true
| B -> false
```

Maybe this is okay. The same weirdness could also happen with `function`, even if its less likely.
 
- when the `match` is in parens, the code says it shouldnt format like it does, but it still does, with a weird indentation: 
```ocaml
let _ =
  foo
  $$ (match group with
    | [] -> impossible "previous match"
    | [ cmt ] -> fmt_cmt t conf cmt ~fmt_code $ maybe_newline ~next cmt)
  $$ bar
;;
```
This is not necessarily bad, but the indentation should be fixed, and why this is happening should be understood before merging.

### DIFF
--- a/lib/Ast.ml
+++ b/lib/Ast.ml
@@ -423,8 +423,7 @@ module Structure_item = struct
   let break_between s cc (i1, c1) (i2, c2) =
     cmts_between s cc i1.pstr_loc i2.pstr_loc
     || has_doc i1 || has_doc i2
-    ||
-    match
+    || match
       Conf.
         (c1.fmt_opts.module_item_spacing.v, c2.fmt_opts.module_item_spacing.v)
     with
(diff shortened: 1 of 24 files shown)


lib/Ast.ml
```diff
@@ -493,8 +492,7 @@ module Signature_item = struct
   let break_between s cc (i1, c1) (i2, c2) =
     cmts_between s cc i1.psig_loc i2.psig_loc
     || has_doc i1 || has_doc i2
-    ||
-    match
+    || match
```
EmileTrotignon, line 495: thats not great, also needs to be addressed

PR #2692, opened 2025-04-16
use ubuntu latest for ci

EmileTrotignon:
ubuntu 20 is not working, trying to switch to latest.


EmileTrotignon:
Pr number update still tries to run an ubuntu 20, I don't know why

EmileTrotignon:
maybe github hasnt noticed I changed it yet, because I can see nothing else.

PR #2693, opened 2025-04-17
`map begin fun` on one line, second attempt

EmileTrotignon:
Same goal as #2675 , but with a cleaner implementation that uses `pro` instead.

The box structure is not ideal, because `fmt_function` cannot open a box in the middle of `pro`, but the only case where this is an issue is the following :

```ocaml
let _ =
  lazy begin fun
         xxxxxxxxxxxxxxxxxxxxxxx
         yyyyyyyyyyyyyyyyyyy
         zzzzzzzzzzzzzzzzzzzzzzzz
       ->
    print_endline xxxxxxxxx ;
    f xxxxxxxxxx yyyyyyyyyyyyy zzzzzzzzzzzzzzzzzzzzz
  end
```

### DIFF
--- a/test/passing/refs.default/exp_grouping.ml.ref
+++ b/test/passing/refs.default/exp_grouping.ml.ref
@@ -634,3 +634,51 @@ let _ =
     a
   end
   [@a]
+
+let () =
+  fooooo
+  |>>>>> List.iter begin fun a ->
+             let x =
+               some_really_really_really_long_name_that_doesn't_fit_on_the_line
+               $ y
+             in
+             fooooooooooo x
+           end
+
+let () =
+  fooooo
+  |>>>>> List.iter
+           begin fun aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa ->
+             let x =
+               some_really_really_really_long_name_that_doesn't_fit_on_the_line
+               $ y
+             in
+             fooooooooooo x
+           end
+
+let () =
+  fooooo
+  |>>>>> List.iter begin fun a ->
+             let x =
+               some_really_really_really_long_name_that_doesn't_fit_on_the_line
+               $ y
+             in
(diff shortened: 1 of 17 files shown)


EmileTrotignon:
I will merge to check interaction with #2694

test/passing/refs.default/exp_grouping.ml.ref
```diff
@@ -468,10 +473,11 @@ let _ =
...
-  lazy begin fun xxxxxxxxxxxxxxxxxxxxxxx
-    yyyyyyyyyyyyyyyyyyy
-    zzzzzzzzzzzzzzzzzzzzzzzz
-  ->
+  lazy begin fun
+         xxxxxxxxxxxxxxxxxxxxxxx
+         yyyyyyyyyyyyyyyyyyy
+         zzzzzzzzzzzzzzzzzzzzzzzz
+       ->
     print_endline xxxxxxxxx;
```
Julow, line 481: I'm a bit surprised that the `->` and the body don't have the same indentation baseline. The `->` seems indented relative to the column where the `begin` keyword starts and not relative to a naturally incrementing indent. 
Is it intended ?

I'd expect the `->` and the argument list to be indented 2 and 4 relative to the body of the fun:
```
  lazy begin fun
        argument list
      ->
    body
  end
```

EmileTrotignon: This is not intended. What I would have liked  is the following:

```ocaml
lazy begin fun
    argument list
  ->
    body
  end
```

This seemed like a small issue in a very rare situation though, and it might be difficult to fix because everything is in `pro`, and context does not tell you you were lazy when formatting the function

Julow: The `pro` includes the `lazy` keyword ? This seems to be the right way to make the formatting you want. The box around `begin ... ->` is unwanted and the forced break after the `fun` was not the right thing to do.

PR #2694, opened 2025-04-17
promote ocp docking for infix apply.

EmileTrotignon:
--ocp-indent-compat behaviour for `|> map ( fun` and `@@ match` is very good. This enables it for everyone.

Replaces #2689 and #2686 .

Should lead to vastly smaller indentation in a lot of cases where it was huge.

### DIFF
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -139,18 +139,18 @@ module Exp = struct
       else k
 
     let dock xarg =
-        match xarg.ast.pexp_desc with
-        | Pexp_apply (_, args) -> (
-          (* Rhs is an apply and it ends with a [fun]. *)
-          match List.last_exn args with
-          | _, {pexp_desc= Pexp_function _; _}
-           |( _
-            , { pexp_desc= Pexp_beginend ({pexp_desc= Pexp_function _; _}, _)
-              ; _ } ) ->
-              true
-          | _ -> false )
-        | Pexp_match _ | Pexp_try _ -> true
-        | _ -> false
+      match xarg.ast.pexp_desc with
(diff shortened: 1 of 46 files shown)


Julow (approved):
I much much prefer the new indentation! The implementation is good but I wonder what the community feels about the change ?

If there's a release in the coming month, should this be included ?

lib/Params.ml
```diff
@@ -250,7 +248,11 @@ module Exp = struct
             match ctx_is_apply_and_exp_is_arg ~ctx ~ctx0 with
             | Some (Nolabel, fun_exp, is_last_arg) ->
                 if begins_line fun_exp.pexp_loc then
-                  if is_last_arg then 5 else 3
+                  if is_last_arg then 5
+                  else
+                    (* TODO Is this branch dead ? Changing the value doesn't
```
Julow, line 253: I couldn't find any example when running `test_branch` with `profile=janestreet` either. This seems to indicate that it is dead code, thanks for looking at that!

EmileTrotignon: i was fiddling at the value to try and "fix" something

EmileTrotignon:
> If there's a release in the coming month, should this be included ?

Yes ! I think this is great and should out as fast as possible.

EmileTrotignon:
I have asked the community here: 
https://discuss.ocaml.org/t/ocamlformat-infix-apply-with-function-definition/16546

Julow (approved):
We should merge this!

EmileTrotignon:
The test branch was failing, but the message was a failure that already existed in 0.26.0. I don't know why it fails the ci step.

PR #2695, opened 2025-04-18
add ahrefs config in tests

EmileTrotignon:
This does not add a profile, it just tests the config used by ahrefs.


EmileTrotignon:
I am merging this for convenience of testing other PRs, if there is a reason not to have it I am okay with this being deleted later.

I think it has value for the community because testing a new option combination is going to make the coverage of the testsuite better, but it will also make the test run a little slower, which is a reason not to do it.

PR #2707, opened 2025-04-28
Dune pkg GHA: continue on failure

Sudha247:
While we want to keep an eye on this, we don't want to make it a hard requirement


Julow (approved):
Thanks!

Issue #2708, opened 2025-04-29
Bug: AST changed with type quantifier and inline records


EmileTrotignon:
```ocaml
type 'a t = A : 'a. {a:'a} -> 'a t
```

is formatted into 

```ocaml
type 'a t = A : {a: 'a} -> 'a t
```

but 
```ocaml
  type 'a t = A : 'a. 'a * 'a -> 'a t
```
works fine.

This is from the base library, as tested by the test branch, and exists at least since 0.26.0.

PR #2710, opened 2025-05-05
[bugfix] fix crash with inline record type variable

EmileTrotignon:
Fix a crash were 
```ocaml
type 'a t = A : 'b. {a: 'a} -> 'a t
```

was formatted as
```ocaml
type 'a t = A : {a: 'a} -> 'a t
```

documented in issue #2708


Julow (commented):
Looks good :)

Issue #2677, opened 2025-04-05
Feature request: module-indent option


HPRIOR:
This may be an Ocaml faux pas but I prefer 4 col/character indenting. I can't find any option to indent the contents of modules, module signatures or module types. 

Here is what ocamlformat currently outputs by default:

```ocaml
module IndentMe : sig
  type t
  type indent

  val indent_me : t -> int -> indent
end

module type IndentMe = sig
  type t
  type indent

  val indent_me : t -> int -> indent
end

module IndentMe = struct
  type t
  type indent

  let indent_me = failwith ""
end

module IndentMe : sig
  type t
  type indent

  val indent_me : t -> int -> indent
end = struct
  type t = int
  type indent = int

  let indent_me = failwith ""
end
```

What I would prefer instead is this (with 4-character indentation inside modules):

```ocaml
module IndentMe : sig
    type t
    type indent

    val indent_me : t -> int -> indent
end

module type IndentMe = sig
    type t
    type indent

    val indent_me : t -> int -> indent
end

module IndentMe = struct
    type t
    type indent

    let indent_me = failwith ""
end

module IndentMe : sig
    type t
    type indent

    val indent_me : t -> int -> indent
end = struct
    type t = int
    type indent = int

    let indent_me = failwith ""
end
```

ocp-indent does not seem to provide this either. I was hopeful that adding `base = 4` to the .ocp-indent config would help but it doesn't.

I'd love to see support for configuring the indentation width of modules, module types, and signatures. Ideally, this could be done with either:
- A general option like module-indent = COL, or
- More specific options like module-type-indent = COL, module-sig-indent = COL, etc.

This indentation level would then apply to all items inside the respective module-related structures (and be recursive for submodules).

Apologies in advance if this is already provided.

EmileTrotignon:
I am not a fan of the idea, 2 spaces is indeed rather consensual in the ocaml community, and more options is not something we are trying to do. Also, its not something that has a good rationale behind it, just a personnal preference. Because of all of this I will not spend time to implement this, and I don't think other maintainers will either.

But its probably not that hard to implement. The way I would do it is format a small example of a file you would want to be different, with flag `-g`, and check the boxes html file in `/tmp`.

Then hover over the element that create the indentation (probably something like `hvbox 2 ...`), and you should get a stacktrace of where in the code this element was created. Then you can go there and change the value conditionally.

[...]

Issue #1613, opened 2021-02-27
Feature request: Allow submodule-body indentations other than 2 spaces


ELLIOTTCABLE:
**Is your feature request related to a problem? Please describe.**
At the moment, as far as I can tell, there's no way to reliably produce output like the following:

```ocaml
module Ex = struct
        type t
        (* ... *)
end
```

Unlike, for instance, extensions, functions, let-bindings, and type-decls, submodule bodies seem to be hard-coded to always be two spaces!

**Describe the solution you'd like**
A indentation setting that applies to structs and signatures:

```
// .ocamlformat
extension-indent = 3
function-indent = 3
let-binding-indent = 3
type-decl-indent = 3

module-indent = 3
```

... producing, for the above:

```ocaml
module Ex = struct
   type t
   (* ... *)
end
```

gpetiot:
Hi, my personal opinion about this kind of options is that it was a mistake to add them in ocamlformat in the first place, as it adds noise in the code, and it goes against the purpose of ocamlformat to promote a normalized formatting for the code of the community.
There will always be some place where we forgot to create an option to customize the indentation, we could argue that it would make sense to have for consistency but it reduces the chance to have a normalized style one day, I think.

I was personally considering removing these options for the 1.0 release (*note that removing more and more options to progressively converge towards a unique style is our goal working towards the 1.0 release*, and these indentation-options are the low hanging fruits of this simplification).

Out of curiosity, why do you want to apply a 3-column indentation instead of 2? (if I'm not mistaken it seems that most people are happy with 2)

ELLIOTTCABLE:
Well, there's kinda three points to make here:

1. Regarding options *in general*, I 100% agree with you. I'm all for the Prettier approach: convention over configuration for style stuff, so we're not simply promoting style-nitpicking discussions from the source-files up to the config-file, but instead removing them entirely. Not here to argue against that! Fantastic plan!

2. Regarding spacing in specific, and I think this is the most relevant point — notice that Prettier, as well as all the other tools I'm aware of, *do* continue to offer configuration for spacing, *even after having removed almost all other configuration*. I'd posit that there's several upsides to this that make it worth an exception:

    - Users will generally have an editor-wide spacing configuration for all languages and projects (mostly, either 2, 4, or 8 spaces; although for reasons below, I don't personally think it's a good idea to hard-code support for those specific values); and no matter what "default" value you pick, your autoformatting tooling is going to then be "fighting" some users' editors.

[...]

PR #2711, opened 2025-05-05
Add module indent option

HPRIOR:
Adds indent option for module structs, types and signatures.


EmileTrotignon:
Thanks for the PR !

The implementation looks very clean, I don't think this will cause a maintenance burden in the future.

I think the option should be named `struct-indent`, and the documentation should make clear that this affects the inside of `sig ... end` and `struct ... end` blocks, currently the explanation is not precise enough. I know `struct-indent` feels a bit weird because it also affects `sig`, but there is precedent with `break-struct` also affecting `sig`.

The tests could include attributes and comments to check that everything works well when they are present. 

`dune build @gen_manpage --auto-promote` should be run to regenerate the manpage.

There also need to be an entry in `CHANGES.md`. 
The CI will stay red because test-branch is currently broken, but the Build on Linux should become green once the manpage is up to date.

EmileTrotignon:
It was requested in the past here: https://github.com/ocaml-ppx/ocamlformat/issues/1613

My opinion on introducing options is that it really depends on the option. This one is very simple, should not cause maintenance issues nor bugs.

That is not the case for other options that have a complex implementation and a hard to explain behavior (the worst one of course being ocp-indent-compat, but there are others in between).

EmileTrotignon:
As for the name, in the discussions leading up the PR, the question of this option affecting functor applications came up, so its at least a little confusing. Also, `module` excludes sig, because `sig .. end` is the syntax for module types. 

What about `struct-sig-indent` ?

HPRIOR:
Thank you for reviewing @Julow.

> We generally don't want to add options to OCamlFormat because it makes the code less consistent and harder to read across the community. It also creates an exponentially increasing number of possible configurations that we can't test and that contain countless bugs.

I understand the concern about adding too many options. I think my PR does contribute to this problem. There are several syntax elements that I’ve left out of the implementation because it’s unclear whether they should follow the module-indent setting.

For example, extension nodes:

```diff
 module M =
-  [%demo
-  module Foo = Bar
+    [%demo
+    module Foo = Bar

-  type t]
+    type t]
```

And anonymous modules:

```diff
 module Subst = Map.Make (struct
-    type t = string
+        type t = string

-    let compare = compare
-  end)
+        let compare = compare
+    end)
```

If these aren't covered by module-indent, we’d need separate options like `module-extension-node-indent` or `anon-module-indent`, which would add to the option/maintenance bloat. 

@Julow would you be supportive of a global indent option instead? Without being complemented by additional options, the current implementation is insufficient for my needs since I'd like all module items to be consistently indented. @EmileTrotignon mentioned in the discussion here https://github.com/ocaml-ppx/ocamlformat/issues/2677#issuecomment-2817082530 that there may be some hidden edge cases when increasing the indentation levels above the default. Given the formatter already supports wrapping/breaking at arbitrary widths it seems like the existing logic is capable of supporting the feature otherwise. While it may slightly increases the developement cost where edge cases are found by the minority of users using non-default indentation, it would not cause configuration/option bloat.

I don't think it's worth arguing for the merits of different indentation levels. What matters is that people have strong preferences, and there’s currently no way to reflect them in ocaml. As @ELLIOTTCABLE argued in #1613, even very opinionated formatters like Prettier allow users to choose indentation width. The widespread use of the default isn’t surprising when there’s never been a choice.

HPRIOR:
Thanks for reviewing @EmileTrotignon,

> The tests could include attributes and comments to check that everything works well when they are present.
> 
> `dune build @gen_manpage --auto-promote` should be run to regenerate the manpage.
> 
> There also need to be an entry in `CHANGES.md`. The CI will stay red because test-branch is currently broken, but the Build on Linux should become green once the manpage is up to date.

If I decide to proceed with the PR I'll make these changes, otherwise it's useful to know for the next one.

EmileTrotignon:
> While it may slightly increase the developement cost where edge cases are found by the minority of users using non-default indentation, it would not cause configuration/option bloat.

In my opinion, such a widespread configuration option would be way harder to merge than the current, or even a slightly extented version of the current PR. The reason being that this PR has a defined scope which is very important for a bug-free implementation. To me, the biggest issue with option bloat is the implementation. In this perspective, a small, clearly bugfree PR like is easy to merge, but a PR with bigger changes not so much.

That being said, I understand the appeal of the feature. A plan to get it would be to have more small indentation options 
in the future, and a master option to set them all at once (the logic for that already exists in the option for ocp-indent configuration).

ELLIOTTCABLE:
For whatever it's worth, I strongly support a simple, global `tab-width` setting that applies to _all_ types of indent. For existing category-specific configuration values, they could override the global default when explicitly set.

That said, my original issue, which this PR addresses, feels like the wrong approach if a global setting is on the table. The bloat of granular settings, as far as I'm concerned, is unnecessary maintenance burden.

If there's good reason not to include a global option, then I do support this PR as a minimal substitute. (=

HPRIOR:
I'm currently working on a provisional PR for a global indent option. There are a few complicated areas, but majority of changes are just a trivial replacement of 2 with a configuration option. 

> `if parens then 3 else 2` probably is because you want the `|` to be indented by `2` relative to `function` and not relative to `(function`. But how to be sure ?

Changing `3` to `2` gives this output in the tests:

```diff
 let _ =
   { foo =
       (fun _ -> function
-         | _ ->
-           let _ = 42 in
-           ()
-         | () -> ())
+        | _ ->
+          let _ = 42 in
+          ()
+        | () -> ())
   }
 ;;

```

In this case the logic would change to: 

```
| _ when ocp c -> (
      match Exp.ctx_is_apply_and_exp_is_arg ~ctx ~ctx0 with
      | Some (_, _, true) -> (* Last argument *) 2
      | _ -> if parens then c.fmt_opts.indent.v + 1 else c.fmt_opts.indent.v )

```

Another question would be whether changing the default indentation level breaks ocp-indent compatibility. I'm not sure about the rest but I imagine some simple maths around the configured indent level will be required. `let_binding_deindent_fun` is an interesting option and would need some thought. 

> All of this is doable, but I don't think its reasonable to do it all at once.

Aren't the tests comprehensive enough to ensure regressions aren't introduced for the default indentation level? Splitting out the indentation options could result is 20/30+ indentation options. Is this desirable?

EmileTrotignon:
> There are a few complicated areas, but majority of changes are just a trivial replacement of 2 with a configuration option.

I am not surprised by that, but the complicated areas might be rabbit holes. 
Also some spots have very poor code quality, like the indentation of function is very hard to understand, I would think it would be better to simplify it before making the code more general. Maybe its fine, I haven't tried.

> let_binding_deindent_fun is an interesting option and would need some thought.

I don't really get why we need this option, I would rather delete it. Having an indentation of 1 is so weird. Maybe I can start by making the default false instead of true.

> Aren't the tests comprehensive enough to ensure regressions aren't introduced for the default indentation level? 

Probably not.

> Splitting out the indentation options could result is 20/30+ indentation options. Is this desirable?

20 to 30 is too many, maybe there would be a way to group them so that it gets to 10.

Another issue I have with a global option is how do you unsure its complete ? With limited scope option its easy, you can test the whole scope, but if its supposed to be for everything then its more complicated.

Also keep in mind that no one is working on ocamlformat full-time, so a huge PR might just be impossible to review.

HPRIOR:
Any chance this can be merged as a stopgap between a more comprehensive indentation option?

EmileTrotignon:
I don't understand the CI failures, they seem unrelated to this PR, I am merging.

Issue #2703, opened 2025-04-24
5.5 support: Changed AST for `Pexp_poly`


Julow:
Compiler PR: https://github.com/ocaml/ocaml/pull/13974

PR #2714, opened 2025-06-06
[5.4 compat] effects keyword can be an attribute id

EmileTrotignon:
Address issue #2703 

I added a test but I think its not great


EmileTrotignon:
No changelog because before `effect` was not a keyword and could therefore be used anywhere

EmileTrotignon:
Yeah, I don't think we can have to much tests.

PR #2716, opened 2025-06-06
makes exp-grouping=preserve the default

EmileTrotignon:
There is no good reason to not preserve begin end by default. 

If the user does not want them, he does not have to use them.


Julow (approved):
Thanks !

PR #2724, opened 2025-09-22
Fix quadratic command-line parsing

let-def:
Command‑line parsing occurs in two passes: one pass parses the actions, and a second pass parses the configuration.
Because the configuration parser is invoked for every input file, the overall parsing time grows quadratically with the number of files.

This patch memoizes the configuration so that the parser runs only once. The configuration changes are still applied to each input file, preserving the original ocamlformat behaviour while making the runtime independent of the number of inputs.

Below are two plots that illustrate the performance before and after the change. They show the time spent and the number of stat calls as functions of the number of input files, clearly demonstrating the quadratic behaviour that the patch removes.

stat
time

### DIFF
--- a/lib/bin_conf/Bin_conf.ml
+++ b/lib/bin_conf/Bin_conf.ml
@@ -535,11 +535,13 @@ let global_lib_term =
         new_global.lib_conf )
     $ global_term )
 
+let global_lib_term_eval = lazy (
+  Cmd.eval_value ~err:discard_formatter ~help:discard_formatter
+    (Cmd.v info global_lib_term)
+)
+
 let update_using_cmdline config =
-  match
-    Cmd.eval_value ~err:discard_formatter ~help:discard_formatter
-      (Cmd.v info global_lib_term)
-  with
+  match Lazy.force global_lib_term_eval with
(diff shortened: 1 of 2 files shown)


lib/bin_conf/Bin_conf.ml
```diff
@@ -535,11 +535,13 @@ let global_lib_term =
         new_global.lib_conf )
     $ global_term )
 
-let update_using_cmdline info config =
-  match
-    Cmd.eval_value ~err:discard_formatter ~help:discard_formatter
-      (Cmd.v info global_lib_term)
-  with
+let global_lib_term_eval = lazy (
```
Julow, line 538: You need to format the code with `dune fmt`. Do you also want to add yourself to `CHANGES.md` ?

Julow (commented):
Awesome !

Just by curiosity, what is your use case in having 500+ input files ?

let-def:
I am working on a fuzzer for OCaml syntax, a use-case with as many input files as one wants :).

However, 80 files is enough to feel a slow-down when running multiple ocamlformat in parallel (with 80 files, there is already ~50k calls to stat, which causes a significant increases to kernel load when many processes are hammering the FS simultaneously).

PR #2727, opened 2025-10-14
Promote changes induced by Cmdliner 2.0.0

Julow:


MisterDA:
Note that cmdliner's author _explicitly_ asks not to test the output of cmdliner, as it is subject to change.
https://discuss.ocaml.org/t/reminder-opam-repository-package-tests-do-not-snapshot-test-cmdliners-end-user-outputs/17331/3

I suppose it's fine to promote its output as a man page.

Julow:
We use cmdliner in a complicated way (we represent options parsed from the config files and env variable as cmdliner Arg/Term). How can we test it otherwise ?
Even if our use is unreasonable, cmdliner allows writing custom converter, with custom errors, and I think it's unreasonable not to test that.

Issue #2731, opened 2025-10-22
Bug: labeled tuples with infix operators cannot be formatted


cod1r:
**Describe the bug**
labeled tuples with infix operators cannot be formatted

**How to Reproduce**
- What is the formatted file? 
```
let labeled_tuple () =
	~x:1,
	~y:(1 + 2)

let ~x, .. = labeled_tuple ()
```
- What are the selected options?
`-i`

I already have a fix that I want to make a PR from but I guess the contributing readme says to make an issue first. :)

PR #2732, opened 2025-10-22
consider case when a labeled tuple element has an infix expression so…

cod1r:
… that it is wrapped in parentheses

#2731

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2219,6 +2219,12 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
             ; _ }
           when not (Std_longident.is_infix id) ->
             has_attr && parens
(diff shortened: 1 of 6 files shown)


lib/Fmt_ast.ml
```diff
@@ -2219,6 +2219,12 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
             ; _ }
           when not (Std_longident.is_infix id) ->
             has_attr && parens
+        | Exp {pexp_desc= Pexp_tuple tuples; _}
```
Julow: Instead of adding more special cases into `Fmt_ast`, we handle parentheses in `Ast.ml`: https://github.com/ocaml-ppx/ocamlformat/pull/2733

cod1r: 👍 no problem. Your pull request makes a lot of sense 😄

PR #2735, opened 2025-10-24
Fix issue #2729

### LINKED ISSUE #2729: Last fn arguments shifted to the right with js profile (0.28 change)
This is a test with the new 0.28.0 version from the release.

My `.ocamlformat` for this file is like this:
```text
ocaml-version=5.4
profile=janestreet
parse-docstrings=true
```

**Current formatting**

```ocaml
module type T = sig
  val operation_name
    :  from_:(module A_LONG_MODULE_TYPE_S with type t = 'a) * A_long_module_name.t
    -> to_:(module A_LONG_MODULE_TYPE_S with type t = 'b) * A_long_module_name.t
       -> ('a -> 'b) -> 'b
end
```

**Describe the formatting you'd like**

The original, and previous formatting with 0.27.0 is to align the arguments at the end, like this:
```ocaml
module type T = sig
  val operation_name
    :  from_:(module A_LONG_MODULE_TYPE_S with type t = 'a) * A_long_module_name.t
    -> to_:(module A_LONG_MODULE_TYPE_S with type t = 'b) * A_long_module_name.t
    -> ('a -> 'b)
    -> 'b
end
```

(See how the 2 last arguments are compacted and shifted to the right in the new formatting).

**Additional context**

I have spent a little bit of time trying to reduce the example further, but I haven't been able to pin-point exactly the context in which this happens. Small changes ends up canceling this behavior, so I am not sure what happens.

Setting `ocaml-version=5.3` in the config yields the same results.

With `profile=conventional` it gets formatted as:
```ocaml
module type T = sig
  val operation_name :
    from_:(module A_LONG_MODULE_TYPE_S with type t = 'a) * A_long_module_name.t ->
    to_:(module A_LONG_MODULE_TYPE_S with type t = 'b) * A_long_module_name.t ->
    ('a -> 'b) -> 'b
end
```
So, still ends up grouping the two last arguments on one line, but without the extra indent.

Can you reproduce the same behavior? Maybe it's a bug (but maybe this was intended to group more arguments on the same line, I'm not sure). The extra indent in the janestreet style feels surprising.

EmileTrotignon:
Fixes #2729  

The problem was the new code for parsing labels was not modified correctly to fit our parsetree.

Our parsetree for `a -> b -> c` is `Arrow ([a; b], c)` but the vanilla parsetree is `Arrow(a, Arrow(b ,c))`.

In the presence of ambiguity with labelled tuple, the new parser would produce `Arrow ([a], Arrow([b], c))`

### DIFF
--- a/test/passing/refs.janestreet/labeled_tuples.ml.ref
+++ b/test/passing/refs.janestreet/labeled_tuples.ml.ref
@@ -633,6 +633,6 @@ module type T = sig
   val x
     :  from:ttttttttttttttttttttttttttttttttt * ttttttttttttttttttttttttttttttttttt
     -> tttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt
-       -> a
-       -> 'b
+    -> a
+    -> 'b
 end
--- a/test/passing/refs.ocamlformat/labeled_tuples.ml.ref
+++ b/test/passing/refs.ocamlformat/labeled_tuples.ml.ref
@@ -639,7 +639,7 @@ module type T = sig
   val x :
        from:
          ttttttttttttttttttttttttttttttttt * ttttttttttttttttttttttttttttttttttt
-    ->    tttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt
-       -> a
-       -> 'b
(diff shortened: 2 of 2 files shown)


EmileTrotignon:
No changelog because this bug was never released.

PR #2737, opened 2025-10-24
make Ast_mapper iterate over all locations inside Longident.t

v-gb:
These locations, which are new in 5.4, are sometimes traversed (e.g. Pexp_setfield) and sometimes not (e.g. Pexp_field). So clearly a bug.

The fix is concretely:

- change map_loc to map over both fields `{ loc; txt }` of `Location.loc`
- propagate the change, ie change `map_loc` to `map_string_loc`, `map_string_opt_loc` or `map_loc_lid` depending on the type. All occurrences of the latter are bug fixes.


Julow:
>  I tried OCAMLFORMAT_EXE=$(readlink -f ../_build/default/bin/ocamlformat/main.exe) make DIRS=code/ocaml/testsuite/tools test_inplace before/after the change, and I see no difference. Also, the commands hides all the output, so it's not exactly convenient.

This command should find the bug, which was fixed in #2733 according to bisect. It exits with an error status before that PR:

```
main: Cannot process "code/ocaml/testsuite/tools/testRelocation.ml".
make[1]: *** [Makefile:105: test_inplace] Error 1
```

This command is intended to find regressions, to show its output you would run the `test_diff` rule.

Issue #2741, opened 2025-10-28
Bug ? In 0.28.1, nested matchings with begin end have one less indentation level


voodoos:
In 0.28.1, nested matching with `begin end` have one less indentation level, even if the `begin` and `match` keywords are not on the same line. This means the `end` keyword is not indented as it should be as a part of the right hand-side of a match branch:

<img width="687" height="289" alt="Image" src="https://github.com/user-attachments/assets/5c0117bd-586e-48bd-a111-48e480d7523d" />

Is that the intended behavior ? It look wrong to me.

<details><summary>`--print-config`</summary>

```
❯ ocamlformat --print-config
comment-check=true
debug=false
disable=false (file .ocamlformat:2)
margin-check=false
max-iters=10
ocaml-version=4.04.0
quiet=false
disable-conf-attrs=false
version-check=true
assignment-operator=end-line
break-before-in=fit-or-vertical
break-cases=fit-or-vertical (file .ocamlformat:4)
break-collection-expressions=fit-or-vertical
break-colon=after
break-fun-decl=wrap
break-fun-sig=wrap
break-infix=wrap
break-infix-before-func=false
break-separators=after
break-sequences=true
break-string-literals=auto
break-struct=force
cases-exp-indent=2 (file .ocamlformat:6)
cases-matching-exp-indent=normal
disambiguate-non-breaking-match=false
doc-comments=before (file .ocamlformat:5)
doc-comments-padding=2
doc-comments-tag-only=default
dock-collection-brackets=false (file .ocamlformat:7)
exp-grouping=preserve (file .ocamlformat:9)
extension-indent=2
field-space=loose
function-indent=2
function-indent-nested=never
if-then-else=compact
indent-after-in=0
indicate-multiline-delimiters=no
indicate-nested-or-patterns=unsafe-no
infix-precedence=indent
leading-nested-match-parens=false
let-and=compact
let-binding-indent=2
let-binding-deindent-fun=true
let-binding-spacing=compact
let-module=compact
line-endings=lf
margin=80
match-indent=0
match-indent-nested=never
max-indent=68
module-indent=2
module-item-spacing=preserve (file .ocamlformat:10)
nested-match=wrap
ocp-indent-compat=false
parens-ite=false
parens-tuple=always
parens-tuple-patterns=multi-line-only
parse-docstrings=false (file .ocamlformat:11)
parse-toplevel-phrases=false
sequence-blank-line=preserve-one
sequence-style=terminator
[...]

PR #2742, opened 2025-10-29
Improve indentation of `end` keyword in begin end enclosed submatch

EmileTrotignon:
```ocaml
begin match () with
| () -> begin
  match
  | () -> ()
end
end
```
becomes 
```ocaml
begin match () with
| () -> begin
  match
  | () -> ()
  end
end
```

also the following bugfix, with `cases-matching-exp-indent=compact`:

```ocaml
begin match () with
| () -> begin
  f x
end
end
```
becomes

```ocaml
begin match () with
| () -> begin
    f x
  end
end
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@ profile. This started with version 0.26.0.
 
 - Fix dropped comment in `(function _ -> x (* cmt *))` (#2739, @Julow)
 
-- `cases-matching-exp-indent=compact` does not impact `begin end` nodes that
+- \* `cases-matching-exp-indent=compact` does not impact `begin end` nodes that
   don't have a match inside. (#2742, @EmileTrotignon)
   ```ocaml
   (* before *)


EmileTrotignon:
Should fix issue #2741

test/passing/refs.ocamlformat/source.ml.ref
```diff
@@ -9435,19 +9435,19 @@ let h = fun ?l:(p = 1) ?y:u ?(x = 3) -> 2
...
-      ( Array.get x 1 + Array.get [||] 0 + Array.get [|1|] 1
-      + Array.get [|1; 2|] 2 ) ;
-    ignore [String.get s 1; String.get "" 2; String.get "123" 3] ;
-    ignore (ba1.{0} + ba2.{1, 2} + ba3.{3, 4, 5}) ignore bg.{1, 2, 3, 4}
+      ignore
+        ( Array.get x 1 + Array.get [||] 0 + Array.get [|1|] 1
+        + Array.get [|1; 2|] 2 ) ;
+      ignore [String.get s 1; String.get "" 2; String.get "123" 3] ;
+      ignore (ba1.{0} + ba2.{1, 2} + ba3.{3, 4, 5}) ignore bg.{1, 2, 3, 4}
     end
```
Julow: Just to note that this is a very large change, it results in +1163 -1141 diffs in `test_branch`. I'd consider this to be a bug fix. Is here any user of the `ocamlformat` profile ?

EmileTrotignon: I use it in my projects, and I would love having this bugfix.

Maybe changes entry is warranted ?

PR #2744, opened 2025-10-29
allow shortcut begin end syntax inside match and if then else

EmileTrotignon:
Instead of 
```ocaml
match () with
| () -> begin
    match () with
    | () ->
  end
end
```
you get
```ocaml
match () with
| () ->
    begin match () with
    | () ->
    end
end
```
Same thing with if-then-else


EmileTrotignon:
Results of the poll are clearly in favour of the change

PR #2745, opened 2025-10-29
allow begin end on one line in safe cases

EmileTrotignon:
you get 
```ocaml
begin f x end
```

The only "unsafe" case is
```ocaml
f begin fun () -> x end
```
which has a hovbox and is difficult to get without allowing the following:
```ocaml
f begin fun () ->
  x ;
  y end
```
Because of this 
```ocaml
f begin fun () -> 
  x 
end
```
is forced

### DIFF
--- a/test/passing/refs.ahrefs/source.ml.err
+++ b/test/passing/refs.ahrefs/source.ml.err
@@ -1,4 +1,4 @@
-Warning: source.ml:3417 exceeds the margin
-Warning: source.ml:6551 exceeds the margin
-Warning: source.ml:6985 exceeds the margin
-Warning: source.ml:7817 exceeds the margin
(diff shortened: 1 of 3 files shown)


Julow:
The complexity came exclusively from turning parens into begin..end. Removing begin..end is very easy: https://github.com/ocaml-ppx/ocamlformat/blob/main/lib/Extended_ast.ml#L282
I think it's possible to come up with a heuristic that avoids identified cases where parentheses use more lines than begin..end.

I totally agree that this could be annoying while editing. Perhaps we could make the heuristic more complex to avoid some of these cases ? For example, don't remove begin..end on a match case if an other case has them, don't remove them in `begin fun`, etc..

PR #2746, opened 2025-10-29
Add configuration to control whether let-punning is used 

### LINKED ISSUE #2743: Bug: let punning changed in 0.27/0.28
**Describe the bug**

In 0.26, `let* foo = foo in ...` would be abbreviated to `let* foo in ... ` when `ocaml-version` was high enough. In the two more recent releases, it is now left as it is. As far as I can tell, there is no option for the old behavior, nor was this mentioned in the change logs?

**How to Reproduce**

ocamlformat 0.26.1 --print-config:

```
comment-check=true
debug=false
disable=false
margin-check=false
max-iters=10
ocaml-version=4.14 (file .ocamlformat:2)
quiet=false
disable-conf-attrs=false
version-check=true
assignment-operator=end-line
break-before-in=auto (file .ocamlformat:5)
break-cases=fit
break-collection-expressions=wrap (file .ocamlformat:6)
break-colon=after
break-fun-decl=wrap
break-fun-sig=wrap
break-infix=wrap
break-infix-before-func=false
break-separators=before (file .ocamlformat:4)
break-sequences=true
break-string-literals=auto
break-struct=force
cases-exp-indent=4
doc-comments=before-except-val (file .ocamlformat:8)
doc-comments-padding=2
doc-comments-tag-only=default
dock-collection-brackets=false (file .ocamlformat:7)
exp-grouping=parens
extension-indent=2
field-space=tight (file .ocamlformat:10)
function-indent=2
function-indent-nested=never
if-then-else=compact
indent-after-in=0
indicate-nested-or-patterns=space (file .ocamlformat:9)
infix-precedence=indent
let-and=compact
let-binding-indent=2
let-binding-spacing=compact
let-module=compact
line-endings=lf
margin=80
match-indent=0
match-indent-nested=never
max-indent=68
module-item-spacing=compact
nested-match=wrap
ocp-indent-compat=false
parens-ite=false
parens-tuple=always
parse-docstrings=false
parse-toplevel-phrases=false
sequence-blank-line=compact (file .ocamlformat:14)
sequence-style=terminator
single-case=compact
space-around-arrays=true
space-around-lists=false (file .ocamlformat:11)
space-around-records=false (file .ocamlformat:12)
space-around-variants=false (file .ocamlformat:13)
stritem-extension-indent=0
type-decl=compact
type-decl-indent=2
wrap-comments=false
wrap-fun-args=true
profile=default
```

0.28.1 --print-config:
```
comment-check=true
debug=false
disable=false
margin-check=false
max-iters=10
ocaml-version=4.14 (file .ocamlformat:2)
quiet=false
disable-conf-attrs=false
version-check=true
assignment-operator=end-line
break-before-in=auto (file .ocamlformat:5)
break-cases=fit
break-collection-expressions=wrap (file .ocamlformat:6)
break-colon=after
break-fun-decl=wrap
break-fun-sig=wrap
break-infix=wrap
break-infix-before-func=false
break-separators=before (file .ocamlformat:4)
break-sequences=true
break-string-literals=auto
break-struct=force
cases-exp-indent=4
doc-comments=before-except-val (file .ocamlformat:8)
doc-comments-padding=2
doc-comments-tag-only=default
dock-collection-brackets=false (file .ocamlformat:7)
exp-grouping=preserve
extension-indent=2
field-space=tight (file .ocamlformat:10)
function-indent=2
function-indent-nested=never
if-then-else=compact
indent-after-in=0
indicate-nested-or-patterns=space (file .ocamlformat:9)
infix-precedence=indent
let-and=compact
let-binding-indent=2
let-binding-deindent-fun=true
let-binding-spacing=compact
let-module=compact
line-endings=lf
margin=80
match-indent=0
match-indent-nested=never
max-indent=68
module-indent=2
module-item-spacing=compact
nested-match=wrap
ocp-indent-compat=false
parens-ite=false
parens-tuple=always
parse-docstrings=true (file .ocamlformat:15)
parse-toplevel-phrases=false
sequence-blank-line=compact (file .ocamlformat:14)
sequence-style=terminator
single-case=compact
space-around-arrays=true
space-around-lists=false (file .ocamlformat:11)
space-around-records=false (file .ocamlformat:12)
space-around-variants=false (file .ocamlformat:13)
stritem-extension-indent=0
type-decl=compact
type-decl-indent=2
wrap-comments=true (file .ocamlformat:16)
wrap-fun-args=true
profile=default
```

WardBrian:
Adds option `--let-binding-punning={preserve|always|never}` to control whether ocamlformat outputs
[let punning](https://ocaml.org/manual/5.0/bindingops.html#ss:letops-punning) (open to suggestions on naming).

Default is `preserve` in all profiles to match the current release's behavior. 

Closes #2743

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,9 +47,8 @@ profile. This started with version 0.26.0.
 ### Added
 
 - Added option `module-indent` option (#2711, @HPRIOR) to control the indentation
-  of items within modules. This affects modules and signatures. For example,
+  of items within modules. This affects modules and signatures. For example, 
   module-indent=4:
-
   ```ocaml
   module type M = sig
       type t
@@ -88,11 +87,10 @@ profile. This started with version 0.26.0.
   was made more precise. (#2705, @EmileTrotignon)
 
 - \* `|> begin`, `~arg:begin`, `begin if`, `lazy begin`, `begin match`,
-  `begin fun` and `map li begin fun` can now be printed on the same line, with
+  `begin fun` and `map li begin fun`  can now be printed on the same line, with
   one less indentation level for the body of the inner expression.
   (#2664, #2666, #2671, #2672, #2681, #2685, #2693, @EmileTrotignon)
   For example :
-
   ```ocaml
   (* before *)
   begin
@@ -157,12 +155,11 @@ profile. This started with version 0.26.0.
 - Fix a crash where `type%e nonrec t = t` was formatted as `type nonrec%e t = t`,
   which is invalid syntax. (#2712, @EmileTrotignon)
 
-- Fix commandline parsing being quadratic in the number of arguments
+- Fix commandline parsing being quadratic in the number of arguments 
   (#2724, @let-def)
 
 - \* Fix `;;` being added after a documentation comment (#2683, @EmileTrotignon)
   This results in more `;;` being inserted, for example:
-
   ```ocaml
   (* before *)
   print_endline "foo"
@@ -190,7 +187,6 @@ profile. This started with version 0.26.0.
   identifiers.
   To format code using the new `effect` syntax, add this option to your
   `.ocamlformat`:
-
   ```
   ocaml-version = 5.3
   ```
@@ -201,7 +197,6 @@ profile. This started with version 0.26.0.
 - \* Consistent indentation of polymorphic variant arguments (#2427, @Julow)
   Increases the indentation by one to make the formatting consistent with
   normal variants. For example:
-
   ```
     ...
     (* before *)
@@ -222,7 +217,6 @@ profile. This started with version 0.26.0.
 
 - Added option `let-binding-deindent-fun` (#2521, @henrytill)
   to control the indentation of the `fun` in:
-
   ```
   let f =
    fun foo ->
@@ -241,7 +235,6 @@ profile. This started with version 0.26.0.
 
 - \* Improve formatting of type constraints with type variables (#2437, @gpetiot)
   For example:
-
   ```
   let f : type a b c.
       a -> b -> c =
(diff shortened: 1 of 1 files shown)


doc/manpage_ocamlformat.mld
```diff
@@ -256,6 +256,12 @@ OPTIONS (CODE FORMATTING STYLE)
            not fit on a single line. The default value is 2. Cannot be set in
            attributes.
 
+       --let-binding-punning={preserve|always|never}
+           Name  punning  in  extended  let bindings preserve uses let-punning
+           only when it exists in the source. always uses let-punning whenever
+           possible. never  never  uses  let-punning.  The  default  value  is
+           preserve.
```
EmileTrotignon: I know this is not the style of the current documentation, but I think an example of what let punning is should included.

Also regarding the name, I thought a let-binding is any use of `let`, but this only concerns bindings with let operators ? This does not have to be reflected in the name, but the documentation should probably say something.

Julow: The problem is that it's not possible to add line breaks in the documentation, forcing the explanation to be concise. This is the output of cmdliner, which removes line breaks and reflow the text.

I'd be totally in favor of a new way of generating the list of option in the documentation and taking advantage of Odoc markup (paragraph, lists, text blocks). The cmdliner output must stay thought.

lib/Conf.ml
```diff
@@ -978,6 +981,22 @@ module Formatting = struct
...
+  let let_binding_punning =
+    let doc = "Name punning in extended let bindings" in
+    let names = ["let-binding-punning"] in
+    let all =
+      [ Decl.Value.make ~name:"preserve" `Preserve
+          "$(b,preserve) uses let-punning only when it exists in the source."
+      ; Decl.Value.make ~name:"always" `Always
+          "$(b,always) uses let-punning whenever possible."
+      ; Decl.Value.make ~name:"never" `Never
+          "$(b,never) never uses let-punning." ]
```
Julow: To make the difference between never and preserve clearer, I think you should say that existing code using let-punning will be rewritten without.

WardBrian:
Question: 
I have the code prepared to also have this setting work for extension lets (e.g. `let%bar x = x in ...` becomes `let%bar x in ...`.

Would this be better in this PR or in another PR? I don't currently have 'preserve' working for these, because that requires parser changes, but I think I could get that done

PR #2748, opened 2025-11-03
Update letop-punning tests, add note to CONTRIBUTING

WardBrian:
Requested by @Julow [here](https://github.com/ocaml-ppx/ocamlformat/pull/2747#discussion_r2481825304).

This fixes up the letop-punning tests to avoid duplicating the same .ml file (and adds a `test/failing` case that I will resolve in a separate PR), and adds a section to CONTRIBUTING.md that should help others in the future get onboarded with the testing system.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,35 +24,31 @@ We heartily welcome pull requests. To ensure an effective contribution, please a
 
 ### Adding a Test
 
-#### Unit Tests
-
-Unit tests are located in the `test/unit` directory and are written using [Alcotest](https://github.com/mirage/alcotest).
-
 #### Golden Tests
 
 The majority of tests for `ocamlformat` are so called "golden" or "expect" tests. These
 tests all consist of running `ocamlformat` on `.ml` files and comparing the
 output to expected output stored in reference files. The `dune` configuration
 for this is automatically generated for you based on the files present in these
-directories.
+directories the first time `dune runtest` is executed.
 
 These are separated into `test/failing` and `test/passing` directories.
 
 To add a test showing currently incorrect behavior, add a `.ml` file to
 `test/failing/tests`. If command line arguments are needed, create a
-corresponding `.opts` file with the same base name. The output of these tests
-will be stored in `.broken-ref` files.
+corresponding `.ml.opts` file with the same name. The output of these tests
+will be stored in `.ml.broken-ref` files.
 
 To add a test showing correct behavior, add a `.ml` file to `test/passing/tests`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -19,8 +19,57 @@ We heartily welcome pull requests. To ensure an effective contribution, please a
...
+tests all consist of running `ocamlformat` on `.ml` files and comparing the
+output to expected output stored in reference files. The `dune` configuration
+for this is automatically generated for you based on the files present in these
+directories.
+
+These are separated into `test/failing` and `test/passing` directories.
+
+To add a test showing currently incorrect behavior, add a `.ml` file to
+`test/failing/tests`. If command line arguments are needed, create a
+corresponding `.opts` file with the same base name. The output of these tests
```
Julow: Base name is ambiguous here, I think you should use entire file names. For example, it's not `mytest.opts` but `mytest.ml.opts`

CONTRIBUTING.md
```diff
@@ -19,8 +19,57 @@ We heartily welcome pull requests. To ensure an effective contribution, please a
...
-4. If you've added code that should be tested, supplement it with tests located in the `test/` directory.
-5. Ensure that the test suite passes (see [Running the tests](#running-the-tests) for instructions).
+4. Ensure that the test suite passes (see [Running the tests](#running-the-tests) for instructions). If there are changes in the tests,
+   verify that they are expected. If no existing test output changed, you likely need to add new tests (see [Adding a test](#adding-a-test)).
+
+### Adding a Test
+
+#### Unit Tests
+
+Unit tests are located in the `test/unit` directory and are written using [Alcotest](https://github.com/mirage/alcotest).
```
Julow: There's a minority of unit tests, I think this should be moved down.

CONTRIBUTING.md
```diff
@@ -19,8 +19,57 @@ We heartily welcome pull requests. To ensure an effective contribution, please a
...
+
+#### Golden Tests
+
+The majority of tests for `ocamlformat` are so called "golden" or "expect" tests. These
+tests all consist of running `ocamlformat` on `.ml` files and comparing the
+output to expected output stored in reference files. The `dune` configuration
+for this is automatically generated for you based on the files present in these
+directories.
+
```
Julow: I think it would be useful to mention running `dune runtest` here. The linked "Running the Tests" section talks about several things but tests added following this guide is specifically run with `dune runtest`.

WardBrian:
The test failures on those platforms have been coming and going on different commits, most of which are only updating the markdown, so I agree.

Let me know if there is anything else needed to merge this!

PR #2750, opened 2025-11-15
ocaml version is set to 5.4.0 by default to allow effects

EmileTrotignon:
Right now you need to have --ocaml-version=5.3.0 or 5.4.0 to be able to use the effect syntax.

This is a big issue because otherwise you get a very annoying syntax error message and its difficult to understand what is happening.

This causes a bunch of diffs elsewhere in the tests. Its mostly improvements in my opinion. Some of it is related to let-punning, we should check it works correctly.

### DIFF
--- a/test/passing/refs.ahrefs/.ocamlformat
+++ b/test/passing/refs.ahrefs/.ocamlformat
@@ -1,5 +1,4 @@
 profile=default
-ocaml-version=4.14.0
 
 break-cases=toplevel
 break-infix-before-func=false


EmileTrotignon:
> I'm not fond of releasing often, as each release has a cost for the community. Each release create incompatibilities for people that install OCamlformat using Opam.

Agreed, releasing often is an issue, but when I found this, I thought for a while that we actually did not have compatibility with 5.4, that we forgot to add some of the syntax or something.

It took looking at the tests to understand what was happening, and I don't think a regular user of ocamlformat would be able to, and he would just get the impression that ocamlformat is not compatible with the new syntax. Even though releasing often is problematic, I think this is worse.

I will check the tests.

Issue #2646, opened 2025-01-14
Bug: Incorrectly added any-pattern into signature of functor


nikolaushuber:
**Describe the bug**
`ocamlformat` seems to insert an any-pattern into the signature of a functor, which results in a syntax error.

**How to Reproduce**
The following code is accepted by the OCaml compiler:
```ocaml
module type X = sig
  val x : int
end

module M : X -> X =
  functor (X : X) -> struct
    let x = X.x
end 
```

Following is the output of running `ocamlformat` on the above code (both if called directly or through `dune fmt`):
```ocaml
module type X = sig
  val x : int
end

module M : (_ : X) -> X =
functor
  (X : X)
  ->
  struct
    let x = X.x
  end
```

This results in the following error when trying to compile with OCaml:
```
File "example/mini_example.ml", line 5, characters 12-13:
5 | module M : (_ : X) -> X =
                ^
Error: Syntax error
```
 I am using version `0.27.0` of `ocamlformat`, and `5.2.1` of OCaml.

sanette:
I am still having the same issue with ocamlformat 0.28.1

The syntax 
```ocaml
module M : (_ : X) -> X =
```
seems to be accepted by recent ocaml, but it should not be the default, since ocaml 5.2 is not that old!

PR #2755, opened 2025-11-21
Support the new functor type syntax `MT -> MT`

Julow:
Fix https://github.com/ocaml-ppx/ocamlformat/issues/2646

Support the unnamed functor parameters in module types:

    module type F = ARG -> S

The extended parser is changed to preserve the concrete syntax of functor arguments in module types. Notably, these three lines are no longer equivalent and the first two are no longer turned into the third:

    module M : (_ : S) -> (_ : S) -> S = N
    module M : S -> S -> S = N
    module M : (_ : S) (_ : S) -> S = N


Julow:
No regressions in test-branch, let's merge !

PR #2756, opened 2025-11-22
Add parens when it's a jsx expression

davesnx:


Julow:
I wouldn't be against merging jsx support into ocamlformat. We could guard it behind a `.ocamlformat` option and a conditional lexer rule.

PR #2787, opened 2026-03-19
Add neocaml-mode support to ocamlformat-before-save

bbatsov:
[neocaml](https://github.com/bbatsov/neocaml) is a new tree-sitter based OCaml major mode for Emacs. This adds `neocaml-mode` and `neocaml-interface-mode` to the mode whitelist in `ocamlformat-before-save` so format-on-save works in neocaml buffers.

Currently users who switch from tuareg-mode to neocaml-mode lose format-on-save silently because the `memq` check doesn't recognize the new modes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -129,6 +129,11 @@ profile. This started with version 0.26.0.
      M.foo)
     ```
 
+### Emacs mode
+
+- Add `neocaml-mode` support to `ocamlformat-before-save`
+  (#2787, @bbatsov)
+
 ## 0.28.1
 
 ### Highlight
--- a/emacs/ocamlformat.el
+++ b/emacs/ocamlformat.el
@@ -101,7 +101,8 @@ nil (default)."
 \(add-hook \\='before-save-hook \\='ocamlformat-before-save)."
   (interactive)
   (when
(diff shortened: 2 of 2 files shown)


EmileTrotignon:
Looks good to me, too bad it happened right after release

bbatsov:
Yeah, that's unfortunate. Still, I guess many people install the package from MELPA, so for them that won't be an issue.

Julow (approved):
Thanks !

Do you think we should keep a changelog for the emacs mode ? Even if it ends up being distributed in a different way, I think it would be a nice touch to have it in the main changelog.

Julow:
The CI is broken because unrelated issues (fixed in https://github.com/ocaml-ppx/ocamlformat/pull/2789 and https://github.com/ocaml-ppx/ocamlformat/pull/2790)

bbatsov:
> Do you think we should keep a changelog for the emacs mode ? Even if it ends up being distributed in a different way, I think it would be a nice touch to have it in the main changelog.

Having a changelog is never a bad idea IMO. Should add a separate file for the package or update the main changelog? I guess that's up to you.

Julow:
In the main changelog in a `### Emacs mode` sub-section or similar would be good.

bbatsov:
@EmileTrotignon Done!

PR #2791, opened 2026-03-26
Simplify Extended_ast vs Std_ast

hhugo:
While working on adding support for mll and mly files. I got very confused with the distinction between Extented_ast and Std_ast.

I started following what was done for `Repl_file` and `Documentation` in `Std_ast`, that is, "no support". But then realized that this mean that we don't check the ast is preserved. (https://github.com/ocaml-ppx/ocamlformat/pull/2557#issuecomment-4130622026)

I figured it was more natural to keep the use of `Std_ast` an implementation detail of the equivalence check for some of the file format in `Extended_ast` only.

This is an attempt at improving the current situation.

## Commits

1. **Remove unused functions from Extended_ast and Normalize_extended_ast** — drop `equal_core_type`, `equal`, `Normalize_extended_ast.equal`/`ast`
2. **Break Source -> Extended_ast dependency cycle** — `open Parsetree` instead of `open Extended_ast`
3. **Break Normalize_std_ast -> Normalize_extended_ast dependency** — make `normalize_code` a parameter instead of importing it
4. **Add paired GADT to Extended_ast** — `('a, 'b) paired` record, Parse.ast produces both ASTs, update Fmt_ast/tests
5. **Add equivalent, dump, dump_normalized, get_std** — AST preservation check using embedded std ASTs
6. **Absorb Parse_with_comments + simplify Translation_unit** — move parsing pipeline into Extended_ast, delete Parse_with_comments, remove `std_fg`/`std_parsed` threading
7. **Clean up Std_ast** — remove `Repl_file`, `Documentation`, `of_syntax`, `any_t`
8. **Document why Std_ast exists** — add docstring explaining its role in semantic preservation

## Summary

- Make `Extended_ast` the single entry point for parsing, validation, and AST operations
- Embed both extended and standard ASTs in a `('a, 'b) paired` record type, so the standard AST is produced alongside the extended one during parsing
- Absorb `Parse_with_comments` into `Extended_ast` — `Documentation` bypasses the OCaml pipeline entirely (no hash-bang, w50, or token collection)
- Add `Extended_ast.equivalent` to check AST preservation, `dump`/`dump_normalized` for debug output
- Simplify `Translation_unit`: remove `std_fg`/`std_parsed` threading, no longer references `Std_ast` directly
- Clean up `Std_ast`: remove unused constructors (`Repl_file`, `Documentation`), `of_syntax`, comment collection
- Break dependency cycles (`Source` -> `Extended_ast`, `Normalize_std_ast` -> `Normalize_extended_ast`)
- Remove dead code (`Extended_ast.equal`, `equal_core_type`, `Normalize_extended_ast.equal`/`ast`)

## Test plan

- [ ] \`dune runtest\` passes (204/204 tests)
- [ ] No behavior change for formatted output
- [ ] \`printast\` tool updated to use new API


lib/Extended_ast.ml
```diff
@@ -395,3 +424,216 @@ module Asttypes = struct
...
+type ast_check_result =
+  | Ast_preserved
+  | Docstrings_moved of Cmt.error list
+  | Ast_changed
+
+let equivalent (type a) (fg : a t) ~normalize_code conf (old_v : a)
+    (new_v : a) : ast_check_result =
+  match get_std_pair fg old_v new_v with
+  | None ->
+      (* TODO: Repl_file and Documentation have no std AST, so we skip the
```
hhugo: I believe this is a preexisting bug

lib/Extended_ast.ml
```diff
@@ -395,3 +424,216 @@ module Asttypes = struct
...
+  | Module_type -> Some (Std_pair (Module_type, v1.std, v2.std))
+  | Expression -> Some (Std_pair (Expression, v1.std, v2.std))
+  | Pattern -> Some (Std_pair (Pattern, v1.std, v2.std))
+  | Repl_file -> None
+  | Documentation -> None
+
+let dump (type a) (fg : a t) fmt (v : a) =
+  match get_std fg v with
+  | Some (Std_value (std_fg, std_v)) -> Std_ast.Printast.ast std_fg fmt std_v
+  | None -> Printast.ast fg fmt v
```
hhugo: We now properly dump ast for `Documentation` and `Repl_file`

lib/Parse_with_comments.ml
```diff
@@ -1,137 +0,0 @@
...
-        Warnings.check_fatal () ;
-        let comments =
-          let mk_cmt = function
-            | `Comment txt, loc -> Cmt.create_comment txt loc
-            | `Docstring txt, loc -> Cmt.create_docstring txt loc
-          in
-          List.map ~f:mk_cmt (Lexer.comments ())
-        in
-        let tokens =
-          (* mld files can not always be lexed using the ocaml lexer *)
```
hhugo, line 98: This is no longer true, parsing documentation no longer goes through this help function.

EmileTrotignon:
I know this part of the code was a bit convoluted, thanks for trying to simplify it. Looks a bit hard to review though

lib/Extended_ast.mli
```diff
@@ -17,36 +17,31 @@ type use_file = toplevel_phrase list
 
 type repl_file = repl_phrase list
 
+module Std_parsetree = Ocamlformat_parser_standard.Parsetree
+
+(** A parsed fragment carrying both the extended AST (used for formatting)
+    and the standard AST (used to verify formatting preserves semantics,
+    see {!Std_ast}). *)
+type ('a, 'b) paired = {extended: 'a; std: 'b}
```
Julow, line 25: I'm not sure to understand what this type provides. Do we want to keep the two AST paired anywhere outside of `Translation_unit` ?
What are the benefits from moving the code from `Parse_with_comments` into `Extended_ast` ?

hhugo:
> I know this part of the code was a bit convoluted, thanks for trying to simplify it. Looks a bit hard to review though

The PR is now split into more digest commits

lib/Extended_ast.mli
hhugo: Parse_with_comments.parse should only be used for "ocaml code", code parsed by the ocaml parser/lexer. It should not be used for `Documentation` for instance (it's wrong in the main branch).

We need to dispatch on the syntax to call it only when appropriate, the dispatch is already done in Extended_ast, it feels natural to move it there.

hhugo: > I'm not sure to understand what this type provides. Do we want to keep the two AST paired anywhere outside of `Translation_unit` ?

I've been trying to minimize behavior changes in the PR.
We could probably parse the Std_ast only when needed in `equivalent`, saving the parsing for files already properly formatted. If we go that direction, we would need to have access to the source as string so that we can parse the std_ast.

Julow: We do have the source as string in `Translation_unit`, both before and after formatting. Perhaps `Parse_with_comments` could be extended with the dispatch ? (and possibly renamed)

hhugo: I can possible move back some parsing logic into Parse_with_comments. But I'm not sure  it gives much benefit.

hhugo: Are you suggesting that the `equivalent` function also receive the two strings ?
I don't quite like this version.
 ```
val equivalent :
     'a t
  -> normalize_code:(string -> string)
  -> Conf.t
  -> 'a * string
  -> 'a * string
  -> ast_check_result
```

Comment (hhugo) [reply lib/Extended_ast.mli:25]: I've added back Parse_with_comments.. Let me know what you think

Comment (Julow) [reply lib/Extended_ast.mli:25]: I'm not sure to see the point of keeping both `Parse_with_comments` and the `paired` type. I think I prefer to keep `Parse_with_comments` because otherwise, we just add a lot of code into `Extended_ast` that in fact work with something else than the extended ast.

I don't think the `equivalent` function make much sense. It should be `'a Std_ast.t -> ... -> 'a -> 'a -> ast_check_result` and not work with the extended AST at all.
It also can reasonably be `string -> string -> ..` and we could unify `Std_ast.t` and `Extended_ast.t`:
```
val ext_parse : ('ext, 'std) t -> string -> 'ext
val std_parse : ('ext, 'std) t -> string -> 'std

val equivalent :
     ('ext, 'std) t
  -> ...
  -> string
  -> string
  (* or 'std -> 'std -> *)
  -> ast_check_result
```

Comment (hhugo) [reply lib/Extended_ast.mli:25]: The equivalent signature you propose forces to parse the same string multiple time when iterating to find a fix point. This is something Ive optimized long time ago in the early days of ocamlformat.

Comment (hhugo) [reply lib/Extended_ast.mli:25]: - `equivalent` can't take a 'std because not all format have one.
- Extracting comment is very much linked to the implementation of the parser. It seems unnatural to have the logic spread across to files.
- Parse_with_comment and paired are orthogonal.

> because otherwise, we just add a lot of code into Extended_ast that in fact work with something else than the extended ast.

The Std_ast is pretty much an implementation detail of the check of some of the file format

Comment (hhugo) [reply lib/Extended_ast.mli:25]: @Julow, think more about this. I wonder if we should inline fields from the "with_comment" record into Extended_ast.t.
- the prefix is only relevent for use_file (and structure maybe)
- comments is not relevent for documentation I think
- Source is a bit weird because it deal with "ocaml" token. which we don't exactly have/want for mll mly files for instance.

The change would mean that an Extended_ast.t contains every thing needed for printing and checking.

Comment (Julow) [reply lib/Extended_ast.mli]: Absolutely ! `'a Extended_ast.paired with_comments` is really confusing.

I checked and we call the extended and standard parsers the same number of times when there's no odoc code blocks. So my concern with enforcing this at the types disappears.

File: lib/Std_ast.mli [hunk 5]
```diff
@@ -9,7 +9,18 @@
...
+    OCamlformat uses an extended parser for formatting that supports
+    additional syntax constructs (field punning, begin..end nodes, labeled
+    tuples, etc.). To verify that formatting preserves semantics,
+    ocamlformat re-parses both the original and formatted source with the
+    standard OCaml parser and compares the resulting ASTs. If they differ,
+    formatting changed the meaning of the program — which is a bug.
+
+    The standard parser is used for this comparison because it represents
+    what the OCaml compiler sees: two programs are equivalent if the
+    compiler parses them identically. *)
```
Julow: Can you link to `Normalize_std_ast` ? We remove some information from the AST before the comparison in order to allow syntax changes that we decided were acceptable.

Julow (approved):
Sorry for the slow review.

I'm really in favor of this change! The code is definitely easier to follow. Happy to merge either now or after you made the `with_comments` change.

lib/Std_ast.mli
hhugo: added a mention

lib/Extended_ast.mli
hhugo: I've pushed the change that inline fields needed for each constructors.
```ocaml
type 'a t =
  | Structure :
      {ast: structure; std: Std_parsetree.structure; cmts: Cmts.t}
      -> structure t
  | Signature :
      {ast: signature; std: Std_parsetree.signature; cmts: Cmts.t}
      -> signature t
  | Use_file :
      { ast: use_file
      ; std: Std_parsetree.toplevel_phrase list
      ; prefix: string
      ; cmts: Cmts.t }
      -> use_file t
  | Core_type :
      {ast: core_type; std: Std_parsetree.core_type; cmts: Cmts.t}
      -> core_type t
  | Module_type :
      {ast: module_type; std: Std_parsetree.module_type; cmts: Cmts.t}
      -> module_type t
  | Expression :
      {ast: expression; std: Std_parsetree.expression; cmts: Cmts.t}
      -> expression t
  | Pattern :
      {ast: pattern; std: Std_parsetree.pattern; cmts: Cmts.t}
      -> pattern t
  | Repl_file : {ast: repl_file; cmts: Cmts.t} -> repl_file t
  | Documentation :
      Ocamlformat_odoc_parser.Ast.t
      -> Ocamlformat_odoc_parser.Ast.t t
```

lib/Cmts.mli
```diff
@@ -25,11 +25,35 @@
 
 type t
 
+val source : t -> Source.t
+(** The [Source.t] used to place comments. *)
```
Julow, line 29: Why do Cmts carry the source code ? The source is used for other things than comments. I think it should be carried by `Extended_ast` then passed separately in `Fmt_ast`.

lib/Cmts.mli
```diff
@@ -25,11 +25,35 @@
...
+
+val copy : t -> t
```
Julow, line 31: Cmts shouldn't be carried by `Extended_ast` and this functions shows why. The mutable state within Cmts is really a dependency of `Fmt_ast`.
It should be created in `Fmt_ast.fmt_file` and not be a dependency of the parser and everything else.
`Extended_ast` should just carry the list of comments.

Julow (commented):
I went over the complete diff again (because the commits are not comprehensible) and I find it hard to see where this is going. Is this new API used by something ?

lib/Extended_ast.mli
Julow: This new type is really nice, except for the `Cmts.t` field, why I mention in an other comment.

PR #2797, opened 2026-04-20
Fix formatting instability on long if conditions with `if-then-else=fit-or-vertical`

MisterDA:
I vibe-clauded a fix for #2770, which is an annoying blocker for me.

### DIFF
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -849,7 +849,8 @@ type if_then_else =
 
 let get_if_then_else (c : Conf.t) ~cmts_before_opt ~pro ~first ~last
     ~parens_bch ~parens_prev_bch ~xcond ~xbch ~expr_loc ~fmt_infix_ext_attrs
-    ~infix_ext_attrs ~fmt_cond ~cmts_before_kw ~cmts_after_kw =
+    ~infix_ext_attrs ~fmt_cond ~cmts_before_kw ~cmts_after_kw
+    ~has_cmts_before_body =
   let imd = c.fmt_opts.indicate_multiline_delimiters.v in
   let beginend_loc, infix_ext_attrs_beginend, branch_expr =
     let ast = xbch.Ast.ast in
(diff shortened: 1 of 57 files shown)


lib/Params.ml
```diff
@@ -849,7 +849,8 @@ type if_then_else =
 
 let get_if_then_else (c : Conf.t) ~cmts_before_opt ~pro ~first ~last
     ~parens_bch ~parens_prev_bch ~xcond ~xbch ~expr_loc ~fmt_infix_ext_attrs
-    ~infix_ext_attrs ~fmt_cond ~cmts_before_kw ~cmts_after_kw =
+    ~infix_ext_attrs ~fmt_cond ~cmts_before_kw ~cmts_after_kw
+    ~has_cmts_before_body =
```
Julow: We don't need to add an extra argument. We can get this information with `cmts_before_opt xbch.ast.pexp_loc` but we also need to output the comment in this function.

Julow (commented):
It's a good fix but if you still have the session, the code can be improved a bit.

Julow (approved):
Thanks a lot!

yakobowski:
I found a slightly different example that this PR fixes too, and which was quite ugly without the fix.

```shell
  cat > .ocamlformat << 'EOF'
  if-then-else=fit-or-vertical
  margin=100
  EOF
```
```shell
  cat > test.ml << 'EOF'
  let f x =
    if x = 0 then None
    else match x with 1 -> Some "one" | _ -> Some "other"
  EOF
```

Expected (trunk):
```ocaml
  let f x =
    if x = 0 then
      None
    else
      match x with
      | 1 -> Some "one"
      | _ -> Some "other"
```

  Actual (0.29.0):
```ocaml
  let f x =
    if x = 0 then
      None
    else
      match
        x
      with
      | 1 -> Some "one"
      | _ -> Some "other"
````

yakobowski:
@MisterDA just in case, I have more fixes for `if-then-else=fit-or-vertical` in #2813

PR #2798, opened 2026-04-23
Fix match in `if-then-else` branches expanding to vertical with `fit-or-vertical`

### LINKED ISSUE #2766: Bug: `match expr with` gets spread into 3 lines in 0.28.1
**Describe the bug**
When a `match expr with` is one of the two branches of `if true then ... else ...` it gets formatted into 3 lines, this behaviour started in 0.28.1 and is not present under 0.27.0. I couldn't find any explanation, or PR related to this change, and it happens for very short lines, so I'm assuming it's a bug

**How to Reproduce**
Run ocamlformat 0.28.1 with the profile from https://github.com/xapi-project/xen-api/blob/57f657fd85771d901b18658c561f8a0d592f1f0b/.ocamlformat  to format this file:
```
let name =
  if true then
    match expr with true -> () | false -> ()
  else
    match expr with true -> () | false -> ()
```
It produces:

```
let name =
  if true then
    match
      expr
    with
    | true ->
        ()
    | false ->
        ()
  else
    match
      expr
    with
    | true ->
        ()
    | false ->
        ()
```

MisterDA:
When a `match expr with` appeared as a branch of `if-then-else` with `if-then-else=fit-or-vertical`, the match keyword, scrutinee, and `with` were broken onto separate lines even when they easily fit on one line:

```ocaml
if true then
  match
    expr
  with
  | true -> ()
  | false -> ()
```

The regression was introduced by 8056fa50 ("Overhaul begin match formatting"), which changed `fmt_match` to wrap `pro` inside a nested `hvbox 0 (pro_inner $ keyword)`. When the match is inside an if-then-else branch, the `pro` carries a `break_unless_newline 1000 2` (from the Fit_or_vertical branch_pro). This inflates the inner hvbox's size to 1000+, causing the Format engine to classify the parent hovbox as non-fitting, which forces all breaks in the `match .. expr .. with` box to break.

Fix: in `match_inner_pro`, return `false` for `Pexp_ifthenelse` context (same as already done for `Pexp_infix`), so that `pro` is placed outside the match's inner hvbox as `pro_outer`, where it does not interfere with the match keyword/expr/with fitting on one line.

Fix #2766. This fix was vibe-clauded.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,11 @@ profile. This started with version 0.26.0.
 
 ### Fixed
 
-- Fix instability on long if-then-else with `if-then-else=fit-or-vertical`
+- Fix instability on long `if-then-else` with `if-then-else=fit-or-vertical`
   (#2797, @MisterDA)
+
+- Fix `match` in `if-then-else` branches expanding to vertical with
+  `fit-or-vertical` (#2798, @MisterDA)
 
 ## 0.29.0
 
@@ -1944,4 +1947,3 @@ profile. This started with version 0.26.0.
 ## 0.1 (2017-10-19)
 
 - Initial release.
-
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -382,7 +382,9 @@ module Exp = struct
   let match_inner_pro ~ctx0 ~parens =
     if parens then false
     else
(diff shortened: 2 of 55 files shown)


Julow (approved):
That's the right fix, thanks !

PR #2800, opened 2026-04-24
Fix formatting oscillation in if-then-else branches

### LINKED ISSUE #2799: Bug: Unstable comment exceeding the margin with `if-then-else=fit-or-vertical`
Here's the minimal reproducer for an ocamlformat (5ebeb8ea12a0ff1d0d3ecd4c29d226d5b0f53764) instability bug:

**MWE (7 lines):**
```ocaml
let f =
  match x with
  | A ->
    if gf then begin
      (* aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa *)
      if a then b
    end
```

**Minimal .ocamlformat:**
```
if-then-else=fit-or-vertical
```

**Key findings:**

1. The bug requires **`if-then-else=fit-or-vertical`** — this is the only config option needed.
2. The code structure is: a `match` case containing an `if ... then begin (* long comment *) if ... then ... end`.
3. The comment must be long enough to exceed the margin (80 chars). Specifically, the comment content must be ≥69 characters (making the full comment line 81+ chars with `(* ... *)` and indentation).
4. The oscillation seems related to ocamlformat wrapping/unwrapping the comment at the margin boundary in interaction with the `if-then-else=fit-or-vertical` layout decision for the nested `if`.

The MWE files are at mwe.ml. You can reproduce with:
```bash
cd /tmp/ocamlformat-mwe && printf 'if-then-else=fit-or-vertical\n' > .ocamlformat && ocamlformat mwe.ml
```

MisterDA:
When a comment sits between a keyword (then/else) and certain branch
expressions (begin...end wrapping match/try/function/ifthenelse, or bare
match/try/function/ifthenelse), the comment would alternate between
"after keyword" and "before expression" placements across formatting
iterations. This caused branch_pro to compute different indentation each
time, producing "formatting did not stabilize after 10 iterations".

The fix detects these oscillation-prone patterns and ensures stable
formatting regardless of the initial comment placement:

- In Fmt_ast, when a branch matches `needs_raw_cmts_after_kw`, extract
  "after keyword" comments without breaks and override branch_pro using
  `Params.raw_cmts_branch_pro` (which computes mode-appropriate
  break+indent+comment output).

- In Params, `branch_pro_with_cmts` detects comments "before expression"
  (the alternate placement) and forces a break, producing the same layout
  as the "after keyword" path.

- For begin...end branches, `branch_pro` checks for comments before the
  inner expression (not just the begin...end node) to handle the case
  where the comment migrates inside the begin...end scope.

Affected modes: Fit_or_vertical, Compact (profile=conventional).

Closes #2799.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2562,7 +2562,7 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                          let cmts_after_kw, raw_cmts_after_kw =
                            if Cmts.has_after c.cmts keyword_loc then
                              if
-                               Params.needs_raw_cmts_after_kw
+                               Params.is_special_or_nested_special_beginend
                                  xbch.ast.pexp_desc
                              then
                                ( None
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2559,10 +2559,18 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                          let cmts_before_kw =
                            Cmts.fmt_before c keyword_loc
                          in
-                         let cmts_after_kw =
+                         let cmts_after_kw, raw_cmts_after_kw =
                            if Cmts.has_after c.cmts keyword_loc then
-                             Some (Cmts.fmt_after c keyword_loc)
-                           else None
+                             if
+                               Params.needs_raw_cmts_after_kw
+                                 xbch.ast.pexp_desc
+                             then
+                               ( None
+                               , Some
+                                   (Cmts.fmt_after ~pro:noop ~epi:noop c
+                                      keyword_loc ) )
+                             else (Some (Cmts.fmt_after c keyword_loc), None)
+                           else (None, None)
                          in
                          let cmts_before_opt loc =
                            if Cmts.has_before c.cmts loc then
@@ -2582,6 +2590,12 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                              ~fmt_cond:(fmt_expression ~box:false c)
                              ~cmts_before_kw ~cmts_after_kw
                          in
+                         let branch_pro =
+                           match raw_cmts_after_kw with
+                           | Some cmts ->
+                               Params.raw_cmts_branch_pro c.conf cmts
+                           | None -> p.branch_pro
+                         in
                          let wrap_beginend =
                            match p.beginend_loc with
                            | Some loc -> Cmts.fmt c loc
@@ -2591,7 +2605,7 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                          p.box_branch
                            ( p.cond
                            $ p.box_keyword_and_expr
-                               ( p.branch_pro
+                               ( branch_pro
                                $ wrap_beginend
                                    (p.wrap_parens
                                       ( fmt_expression c ?box:p.box_expr
--- a/lib/Params.ml
+++ b/lib/Params.ml
@@ -512,6 +512,21 @@ let is_special_beginend exp =
   | Pexp_match _ | Pexp_try _ | Pexp_function _ | Pexp_ifthenelse _ -> true
   | _ -> false
 
+let needs_raw_cmts_after_kw = function
+  | Pexp_beginend
+      ( { pexp_desc=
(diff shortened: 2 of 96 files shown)


lib/Fmt_ast.ml
```diff
@@ -2556,6 +2556,22 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
...
                          in
+                         (* When the branch is begin...end wrapping a special
+                            expression (match/try/function/if), comments
+                            between the keyword and begin can end up attached
+                            to different locations depending on source
+                            layout. Relocate them to a consistent position to
+                            avoid oscillation. *)
+                         ( match xbch.ast.pexp_desc with
+                         | Pexp_beginend ({pexp_desc; _}, _)
+                           when match pexp_desc with
```
Julow: This when is not idiomatic. Can you rewrite it to a nested pattern ?

lib/Cmts.ml
```diff
@@ -328,6 +328,20 @@ let relocate_cmts_before (t : t) ~src ~sep ~dst =
   in
   update_cmts t `Before ~f ; update_cmts t `Within ~f
 
+let relocate_after_to_before (t : t) ~src ~dst =
```
Julow: It's not right to remove comments attached to one side but not on the other. You should use `Cmts.relocate` instead and find a suitable location to pass to `~after`.

lib/Params.ml
```diff
@@ -968,7 +968,14 @@ let get_if_then_else (c : Conf.t) ~cmts_before_opt ~pro ~first ~last
             then
               match cmts_before_opt xbch.ast.pexp_loc with
               | Some cmts -> break 1000 2 $ cmts
-              | None -> branch_pro ~begin_end_offset:0 ()
+              | None -> (
+                match xbch.ast.pexp_desc with
+                | Pexp_beginend (nested_exp, _)
+                  when is_special_beginend nested_exp.pexp_desc -> (
+                  match cmts_before_opt nested_exp.pexp_loc with
+                  | Some cmts -> break 1000 2 $ cmts
```
Julow: We seem to have all the required information here. Maybe relocate should be called from here ?

MisterDA:
Thanks for the review, here's a revised version of the patch with your comments.
Not sure where the build failures on Windows come from, and I think they're unrelated.

lib/Fmt_ast.ml
```diff
@@ -2559,10 +2559,26 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
...
+                                 ; _ }
+                               , _ ) ->
+                               true
+                           | _ -> false
+                         in
+                         let cmts_after_kw, cmts_after_kw_raw =
                            if Cmts.has_after c.cmts keyword_loc then
-                             Some (Cmts.fmt_after c keyword_loc)
-                           else None
+                             if is_special_beginend_branch then
```
Julow: Can you move this code to Params ? That's where the special cases should be

lib/Params.ml
```diff
@@ -851,7 +861,8 @@ type if_then_else =
 
 let get_if_then_else (c : Conf.t) ~cmts_before_opt ~pro ~first ~last
     ~parens_bch ~parens_prev_bch ~xcond ~xbch ~expr_loc ~fmt_infix_ext_attrs
-    ~infix_ext_attrs ~fmt_cond ~cmts_before_kw ~cmts_after_kw =
+    ~infix_ext_attrs ~fmt_cond ~cmts_before_kw ~cmts_after_kw
+    ~cmts_after_kw_raw =
```
Julow: I really don't like this `cmts_after_kw_raw` argument. I tried to get rid of it by passing `Cmts.fmt_after ~pro:noop ~epi:noop ..` to `~cmts_after_kw` but it's very hard due to the many different states this creates.

I won't merge a new argument to `get_if_then_else`.

MisterDA: Claude found another workaround. Let me know if it's satisfying or not.

MisterDA:
Gentle ping @Julow !

Julow (approved):
Works for me :) The `raw_cmts_branch_pro` thing is more complex than I'd like but I haven't been able to find a simpler way to do it.

Thanks!

PR #2801, opened 2026-04-30
chore: update to actions/checkout@v6

MisterDA:


Julow (approved):
Thanks!

Do you think you could debug the Windows failure ?

MisterDA:
> Do you think you could debug the Windows failure ?

~Is it just missing a lower-bound on odoc.3? I can't tell from the logs which version of odoc is used, maybe it's because it's installed with dune pkg?~ 
EDIT: that's not it, odoc is vendored.

PR #2803, opened 2026-05-26
Print readable error messages in doc code blocks

hhugo:
## Summary

- When parsing fails inside a `{@ocaml[...]}` block, the catch-all branch in `Fmt_ast.ml` formatted the exception with `Exn.pp`, which for `Lexer.Error` produced the raw constructor pattern `Ocamlformat_parser_extended.Lexer.Error(_, _)`. The printer registered via `Location.register_error_of_exn` lives in a separate table from `Printexc`, so it was never consulted.
- Route the catch-all through `Location.error_of_exn`: if a printer is registered, format its `main` message; otherwise fall back to `Exn.pp`. Handles `Lexer.Error` and any other registered exception without per-variant enumeration.
- Add a regression test (`doc_lexer_errors.mld`) covering the variants exercised.

## Test plan

- [x] `dune build`
- [x] `dune runtest` (new `.err` fixtures updated)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@ profile. This started with version 0.26.0.
 
 ### Fixed
 
+- Print readable messages for errors raised when parsing OCaml code blocks
+  in documentation comments, instead of the raw exception representation
+  `Ocamlformat_parser_extended.Lexer.Error(_, _)` (#PR, @hhugo)
+
 - Fix instability on long `if-then-else` with `if-then-else=fit-or-vertical`
   (#2797, @MisterDA)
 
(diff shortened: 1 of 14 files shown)


Julow (commented):
The CI fails with diffs like this in test/failing:
```diff
--- a/_build/default/test/failing/tests/comments.ml.broken-ref
+++ b/_build/default/test/failing/comments.ml.output
@@ -1,4 +1,4 @@
-ocamlformat: ignoring "tests/comments.ml" (syntax error)
+main: ignoring "tests/comments.ml" (syntax error)
File "tests/comments.ml", line 64, characters 0-6:
64 | module M_ok : sig
^^^^^^
```
I'm not sure where we print the program name and I don't reproduce this locally (even with main branch merged)

PR #2804, opened 2026-05-26
Fix Windows lexer error on CRLF code blocks

hhugo:
## Summary

Fixes the Windows CI failure (e.g. [run 25160035941](https://github.com/ocaml-ppx/ocamlformat/actions/runs/25160035941)) where `{@ocaml[...]}` blocks in `.mld` / `.mli` doc comments produce `invalid code block: Ocamlformat_parser_extended.Lexer.Error(_, _)` on CRLF-checked-out files.

In `vendor/odoc-parser/odoc_parser.ml`, `handle_last_newline` scans backward over trailing whitespace and stops on `'\n'`, returning the substring up to that index. The symmetric `handle_first_newline` already eats a leading `'\r'` before matching `'\n'`, but `handle_last_newline` did not — so the `'\r'` immediately preceding the final `'\n'` survived. `deindent` then splits on `'\n'` and rejoins with `'\n'`, which leaves intermediate `\r`s followed by `\n` (tolerated by the lexer's `\r*\n` rule) but a bare `'\r'` on the last line, which falls through to `Illegal_character`.

The fix mirrors `handle_first_newline`: when `handle_last_newline` hits `'\n'`, also drop a preceding `'\r'` so the full CRLF is consumed.

## Test plan

- [x] Local repro on Linux: convert `test/passing/tests/doc_repl.mld` to CRLF and run `ocamlformat --parse-toplevel-phrases` — reproduces the failure before the patch, succeeds after.
- [x] `dune build @runtest` clean on Linux (no regressions on LF input).
- [ ] Windows CI green.


Julow (commented):
Thanks! The build is failing in CI on linux with dune-pkg. Do you have an idea why ?

Julow (approved):
Thanks a lot! This works well on top of the other changes in master.

PR #2808, opened 2026-06-23
Rename the internal main module to ocamlformat (fix Dune PM CI)

MisterDA:
The root cause was that the ocamlformat executable's internal module name was main, so dune built `main.exe`.  The program derives its self-name (used in error messages) from `argv.(0)` at `lib/Translation_unit.ml:37`.  Under dune package management: (setup-dune@v2), `%{bin:ocamlformat}` resolves to the raw `main.exe` instead of the install-layout `ocamlformat` path, making the program call itself `main` instead of `ocamlformat`.

Now `argv.(0)` ends in `ocamlformat.exe` in every build mode, so `exe = "ocamlformat"` consistently. `dune build @test/failing/runtest` passes locally, and the change is independent of how dune resolves the binary, so it'll be stable on CI too.

The separate `bin/ocamlformat-rpc/` executable (also internally named `main`) is untouched.


Julow (approved):
Very nice, thanks!

PR #2810, opened 2026-06-24
More `if-then-else=fit-or-vertical` fixes

MisterDA:
- Fix `begin match … end` (and `begin if … end`) branch with `if-then-else=fit-or-vertical`: the `match … with` header no longer splits over several lines, and `end` is aligned with `begin`.

- Fix indentation of a bare `match`/`function`/`try` branch preceded by a comment with `if-then-else=fit-or-vertical`: the branch is no longer indented relative to the comment's end column.

### DIFF
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2593,7 +2593,12 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                          let branch_pro =
                            match raw_cmts_after_kw with
                            | Some cmts ->
-                               Params.raw_cmts_branch_pro c.conf cmts
+                               let bare_branch =
+                                 Params.is_bare_branch ~parens_bch
+                                   xbch.ast.pexp_desc
+                               in
+                               Params.raw_cmts_branch_pro ~bare_branch c.conf
+                                 cmts
                            | None -> p.branch_pro
                          in
                          let wrap_beginend =
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2581,6 +2581,7 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                          in
                          let p =
                            Params.get_if_then_else c.conf ~cmts_before_opt
+                             ~has_cmts_before:(Cmts.has_before c.cmts)
                              ~pro:(fmt_if first pro_inner) ~first ~last
                              ~parens_bch ~parens_prev_bch:!parens_prev_bch
                              ~xcond ~xbch ~expr_loc:pexp_loc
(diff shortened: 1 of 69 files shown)


lib/Fmt_ast.ml
```diff
@@ -2593,7 +2593,12 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                          let branch_pro =
                            match raw_cmts_after_kw with
                            | Some cmts ->
-                               Params.raw_cmts_branch_pro c.conf cmts
+                               let bare_branch =
```
Julow, line 2597: I don't think that's the fix for if-then-else. This does:
```diff
       if gf then
         (* aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa *)
-      begin if
-        a
-      then
-        b
-      end
+        begin if a then b
+        end
```
I think the right output would be:
```ocaml
       if gf then
         (* aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa *)
       begin
         if a then b
       end
```
because without the comment it is:
```ocaml
       if gf then begin
         if a then b
       end
```
The body of the begin-end shouldn't reindent if a comment is added around it. This should be doable with boxes instead of adding a new special case.

Julow (commented):
The match/function fix is perfect :)

lib/Fmt_ast.ml
MisterDA: the LLM seems to like
```
if gf then begin
  (* comment *)
  if a then b
end
```
and I'm not sure how to go against it

Julow: It's probably fine thought we tried to avoid this in the past. If it changes existing code (eg. in the testsuite), then it's not OK.

MisterDA: hoping that I won't offend you by copy-pasting the LLM's answer:

1. Only one test case changed, nothing else. All 39 modified ref files are the same case (`ite.ml`'s `begin … (comment) … if a then b … end` regression test from #2800), repeated across every profile × `if-then-else` variant. No test input files changed; no other test files (`exp_grouping`, etc.) changed; no other case within `ite.ml` changed. The `else begin match e with … end case` is byte-identical to before.

2. That case was never properly formatted — it's the bug being fixed. On main, `fit-or-vertical` produced the broken over-split:
```
      if gf then
        (* comment *)
      begin if
        a
      then
        b
      end
```
My output:
```
      if gf then begin
        (* comment *)
        if a then b
      end
```
which is exactly the input layout (`test/passing/tests/ite.ml:206-209`) — so it's now idempotent, where before it wasn't.

3. `@runtest exits 0` — every test case formats to its ref and is stable (the harness checks idempotency). If I'd disturbed any already-correct, stable formatting, the suite would fail.

So the rule is satisfied: the only behavioural change is to the one case that was misformatted, and it now round-trips its own input. Everything that was already correct is untouched.

Julow: No problem :) Thanks for improving ocamlformat.
This output is fine. We'll see the impact in test-branch after you pushed

MisterDA: I have already pushed ?...

yakobowski:
I just noticed this MR. I suspect we've been working on more or less the same problems. My fixes are in #2813. I will attempt a comparison.

yakobowski:
This MR was uniformly better than mine, so I have removed the corresponding commits from mine. That being said, testing on our codebase, I found another problem. Tentative fixes are: https://github.com/yakobowski/ocamlformat/commit/ab9f05a2163dfbd792fa3752ba9a8e123e0558d6 and https://github.com/yakobowski/ocamlformat/commit/ecaf57f6811e2d6c9bb36ee7f7cc0e64fe351d7e
@MisterDA

MisterDA:
> This MR was uniformly better than mine, so I have removed the corresponding commits from mine. That being said, testing on our codebase, I found another problem. Tentative fixes are: https://github.com/yakobowski/ocamlformat/commit/ab9f05a2163dfbd792fa3752ba9a8e123e0558d6 and https://github.com/yakobowski/ocamlformat/commit/ecaf57f6811e2d6c9bb36ee7f7cc0e64fe351d7e
> @MisterDA 

Perhaps you should open a new PR stacked on this one? That way you can build on the changes here and work on your own.

yakobowski:
> Perhaps you should open a new PR stacked on this one? That way you can build on the changes here and work on your own.

Sorry, my previous comment was unclear. Those 2 commits are _already_ on top of your branch, in case you wanted to have a look in an easy way.

yakobowski:
#2815 created with those changes + my additional ones.

Julow (approved):
Thank you very much :)

PR #2812, opened 2026-06-29
more compact formatting for while-loops with no body

v-gb:
Concretely, this is what changes:

```diff
 while
   loooong
     condition
-do
-  ()
-done
+do () done
```

A real example where I wished for a more compact formatting:

https://github.com/ocaml/ocaml-re/pull/599/commits/59198dd3dd379bbc5325a4443b96c10b9e2f02d7#diff-a91a869b078e8995843c8b377da588505468c962166d62d7029893c27be32833R70-R80

To explain a bit: while-loops are fairly inexpressive in ocaml due to the absence of `break`.  In a decent number of cases, it's nicer to give up on the separation between condition and body, by moving the body into the condition. In that case, it's a bit silly to waste 3 lines on the empty body.

For instance, if you wanted to write a loop where the first iteration always runs, like a do-while in C, the obvious thing to do (without auxiliary functions) is:

```ocaml
let first = ref true in
while !first || condition; do first := false; body done
```

but the simpler version is:

```ocaml
while body; condition do () done
```

Or iterating through a stack:

```ocaml
while not (Stack.is_empty s) do f (Stack.pop_exn s) done
```

can be done without double-checking that the stack is empty:

```ocaml
while
  match Stack.pop s with
  | None -> false
  | Some elt -> f elt; true
do () done
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@ profile. This started with version 0.26.0.
 
 ## unreleased
 
+### Changed
+
+- Empty bodies of while loops will be formatted on one line: `do () done`.
+  (#2812, @v-gb)
+
 ### Fixed
 
 - Fix `begin match … end` (and `begin if … end`) branches: with
--- a/lib/Fmt_ast.ml
+++ b/lib/Fmt_ast.ml
@@ -2920,6 +2920,14 @@ and fmt_expression c ?(box = true) ?(pro = noop) ?eol ?parens
                  $ fmt_core_type c (sub_typ ~ctx t2) )
              $ fmt_atrs ) )
   | Pexp_while (e1, e2, infix_ext_attrs) ->
+      let break_around_body =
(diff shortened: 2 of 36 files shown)


EmileTrotignon:
Not sure why the CI is failing, but I agree with the goal. The tests and implementation are good.

EmileTrotignon (approved):

v-gb:
Thanks! Right, the failures are unrelated to the change, though no idea either why they are occurring.

I added the PR number to the CHANGES file (it's going to incidentally retrigger builds, but I expect they will fail again).

Julow (approved):
Looks good! The CI should run if you merge the main branch.

v-gb:
After rebase, there's still some kind of unrelated failure.

Julow:
Thanks! The remaining failures are not related, I'll fix them separately.

PR #2813, opened 2026-07-04
Fix vertical breaks in let-in with comments before 'in'

yakobowski:
I'm currently migrating a big codebase which is formatted using ocamlformat 0.22. I have noticed quite a few improvements (thanks!) but also some regressions. Some of them are fixed by #2810. The remaining ones are related to comments in a let binding.

*Bug*: `let x = value (* comment *) in body` is always broken across multiple lines, even when it fits within the margin. The `force_break` prefix on comments near `in` unconditionally forced the containing hvbox to vertical mode.

*Fix*: `~pro:force_break → ~pro:(break 1 0)` in two places in fmt_value_binding. The comment now gets a regular break   (becomes a space when the box fits horizontally, still breaks when it doesn't). Also fixes a non-idempotency where the one-liner and its pre-broken form produced different outputs. Affects all profiles.


yakobowski:
#2810 is uniformly better on the if-then-else issue, so I'm dropping those commits from the MR.

Julow (approved):
Looks good! I merged the main branch and pushed to your branch to allow the CI to run.
Can you add a change entry ?

MisterDA:
:ping_pong:

Julow (commented):
Looks good to me :)
Can you add a change entry ?

This changes a lot of code so add a `\*` in the change entry. For the record, here's an example of code changed:
```diff
--- a/compiler/lib/generate.ml
+++ b/compiler/lib/generate.ml
@@ -916,10 +916,7 @@ module DTree = struct
         with Not_found -> (
           (* do we have to split again ? *)
           (* we count the number of cases, default/last case count for one *)
-          let nbcases =
-            ref 1
-            (* default case *)
-          in
+          let nbcases = ref 1 (* default case *) in
           for i = 0 to array_len - 2 do
             nbcases := !nbcases + List.length (snd array_norm.(i))
           done;
```

yakobowski:
Very sorry, I completely missed the initial notification. (I had to rebase and force-push, apologies if that's an issue.)

Julow (approved):
Thanks!

PR #2814, opened 2026-07-05
Fix "BUG: ast changed" on hand-written [@JSX] applications (#12)

### LINKED ISSUE #12: Bug: Can't format when `[@JSX]` is present

davesnx:
Fixes https://github.com/ocaml-mlx/ocamlformat-mlx/issues/12

ocamlformat-mlx's printer assumed every [@JSX]-attributed application was of kind 1 (canonical shape) and unconditionally printed it back as JSX syntax. For kind 2, that assumption broke — ~children:[] wasn't recognized (printed as a duplicate prop) and the argument order didn't survive the round trip, so the safety check rejected the output.


hhugo:
Opened against the wrong remote

PR #2180, opened 2022-10-12
Add Github Action to run test_branch.sh checks

gpetiot:

PR #2815, opened 2026-07-07
Additional fixes on top of #2810

yakobowski:
This MR fixes a few additional problems for the `if-then-else=fit-or-vertical` mode. It is stacked on top of #2180.

The first commit partially reverts some changes (`bare_branch` logic) in order to fix some additional less-than-ideal reformatting found on our codebase. I have added testcases, but I was not able to add the more interesting ones in a standalone commit, as ocamlformat without the fix does not stabilize in some configurations. (See the last commit.)


Julow (approved):
Thanks! I took the liberty to rebase on top of https://github.com/ocaml-ppx/ocamlformat/pull/2810 and https://github.com/ocaml-ppx/ocamlformat/pull/2818 and to push on your branch.

Issue #2534, opened 2024-03-24
No support for default custom config?


dunric:
**How can I have my default (fallback) configuration settings for `ocamlformat` when no per-project's one is present?**

I may be missing something or it has been chosen some weird logic for `ocamlformat`  behavior, but it seems when it detects project's root - even a false one by a sole presence of `.git` or `.hg` subdir - it doesn't pick global/default config at `$XDG_CONFIG_HOME/ocamlformat` or `$HOME/.config/ocamlformat` when local `.ocamlformat` is missing and `--enable-outside-detected-project` option is passed.

This effectively forces one to manually copy or symlink global config for each and every project to use his preferred formatting options, which I would consider tedious and unreasonable practice.

Likely related issues #2039 and #2223 .

dunric:
Hmm, it is somehow silent here …

A quick-hack solution follows, for cases when a project's root is "detected" and none project's configuration file is found. Equivalent to undetected project's root (and no config file found).

``` diff
diff --git a/lib/bin_conf/File_system.ml b/lib/bin_conf/File_system.ml
index 7e1cb5f..10e8991 100644
--- a/lib/bin_conf/File_system.ml
+++ b/lib/bin_conf/File_system.ml
@@ -64,7 +64,6 @@ let make ~enable_outside_detected_project ~disable_conf_files
         (* Outside of a detected project, only apply the global config file
            when [--enable-outside-detected-project] is set and no
            [.ocamlformat] file has been found. *)
-        assert (Option.is_none fs.project_root) ;
         if
           List.is_empty fs.configuration_files
           && enable_outside_detected_project
@@ -104,8 +103,11 @@ let make ~enable_outside_detected_project ~disable_conf_files
         in
         (* Inside a detected project, configs are applied in top-down
            starting from the project root (i.e. excluding the global config
-           file). *)
-        if is_project_root ~root dir then {fs with project_root= Some dir}
+           file).
+           In case of none config detected, attempt to fall back to a global
+           one. *)
+        if is_project_root ~root dir then
+          aux {fs with project_root= Some dir} ~segs:[]
         else aux fs ~segs:upper_segs
   in
   aux ~segs
```

I'm quite new to OCaml and don't quite understand some design decisions, like why anyone would need multiple per-directory formatting options and *merge* them together instead of a single per-project one. Similar with dune build system and questionable benefits to specify separate build stanzas instead of a writing one, common for a project.

Issue #2509, opened 2024-01-13
Bug?: `--enable-outside-detected-project` does not use global configuration in projects


lenntil2:
**Describe the bug**

When `dune-project` exists and project-wise `.ocamlformat` does not exist, unconfigured default foramtting will be used. (macOS)

**Expected Behavior**
Use global configuration if it is project directory and no `.ocamlformat` file exist.

**How to Reproduce**

`$XDG_CONFIG_HOME/ocamlformat`:
```
profile=janestreet # 'default', 'janestreet', ...
```
In project directories, i.e., with `dune-project` file, or even in directories that is not OCaml project but `.git` exists,
```console
$ ocamlformat --enable-outside-detected-project --print-config
...
space-around-variants=true
stritem-extension-indent=0
type-decl=compact
type-decl-indent=2
wrap-comments=false
wrap-fun-args=true
profile=default
```

Not in project directories,
```console
$ ocamlformat --enable-outside-detected-project --print-config
...
type-decl-indent=2 (profile janestreet (file ../../.config/ocamlformat:1))
wrap-comments=false (profile janestreet (file ../../.config/ocamlformat:1))
wrap-fun-args=false (profile janestreet (file ../../.config/ocamlformat:1))
profile=janestreet (file ../../.config/ocamlformat:1)
```

## See also
- #2223 
- [when dune-project exists and no project-wise .ocamlformat exist, unconfigured default foramtting will be used.](https://github.com/helix-editor/helix/discussions/8752#discussioncomment-8045180)

Julow:
I agree that ocamlformat should use the global configuration when `--enable-outside-detected-project` and a project is detected but has no configuration.

Issue #2223, opened 2023-01-01
Bug: .ocamlformat has no effect even with --enable-outside-detected-project


tari3x:
```
avatar@pixel:~/docs/visuals/src/spark/lib$ ll ~/.ocamlformat
-rw-r--r-- 1 avatar users 33 Dec 12 00:41 /home/avatar/.ocamlformat
avatar@pixel:~/docs/visuals/src/spark/lib$ ocamlformat --enable-outside-detected-project  server_state.ml
<no effect of settings in ocamlformat>
avatar@pixel:~/docs/visuals/src/spark/lib$ ln -s ~/.ocamlformat 
avatar@pixel:~/docs/visuals/src/spark/lib$ ocamlformat --enable-outside-detected-project  server_state.ml
<works as expected now>
```

I tried renaming ocamlformat to ~/ocamlformat and to ~/.config/ocamlformat and to ~/.config/.ocamlformat. No result. I tried setting XDG_CONFIG_HOME, no result. 

Strace shows that ocamlformat makes no attempt to read any file outside of the current project directory.

gpetiot:
Hi, that's a tough one, I have a few questions.

What version of ocamlformat do you use?

Do you get a warning saying that `--enable-outside-detected-project` is not used? or that no .ocamlformat is found? or that no root is found?

Does your .ocamlformat (or maybe in the parent directory as it seems `lib/` is likely not the project root) contain `disable`? Do you have more .ocamlformat files in your project that could mess up the configuration? (you can check with `ocamlformat --print-config`)

It is recommended to place the .ocamlformat file at the project root unless you want to override the settings in some subdirectories.

tari3x:
```
$ ocamlformat --version 
0.24.1
$ cat .ocamlformat 
profile = janestreet
margin = 75
```

I see no warnings when I run with  --enable-outside-detected-project but I do see a warning when I remove that flag. 

Looks like there are no conflicting config files. 

I have many projects and I would of course want to apply the same formatting in all of them. Thus putting the same .ocamlformat in each of them seems silly. I'd much rather have a single global config.

gpetiot:
Do you have a ocp-indent file in your project?
That would explain the situation: ocamlformat will only look for configuration files outside of the project if no configuration file (.ocamlformat files but also .ocp-indent files) is found in the project. Btw having a symlink pointing to a global .ocamlformat is not a bad solution.

tari3x:
[...]

PR #2816, opened 2026-07-08
Improved global configuration file lookup.

torimus:
Possible fix for issues #2509, #2223 and #2534 .

This patch adds a case when project's root is detected but no configuration file (`.ocamlformat`) has been found. It then attempts to use global configuration file, when lookup outside project's root is enabled.

It improves behavior described in existing documentation

> If  no  project  root and  no  ocamlformat  file  has  been  found, and if the option enable-out-side-detected-project is  set,  the  global  ocamlformat  file  defined  in $XDG_CONFIG_HOME  (or in $HOME/.config if $XDG_CONFIG_HOME is undefined) is used.

by allowing to use existing global configuration also when there is none present in a project.
I find it quite handy to have this option available when new projects are instantiated, without a need to do a tedious work to clone my preferred settings for each one.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,9 @@ profile. This started with version 0.26.0.
 
 - Fix missing parentheses in `(a, ~b:(if .. then .. else ..))` (#2811, @Julow)
 
+- Fix configuration in `$XDG_CONFIG_HOME` was not used when
+  `--enable-outside-detected-project` was passed (#2816, @torimus)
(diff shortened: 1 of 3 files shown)


torimus (commented):
Update to reflect detected project's root.

Julow (approved):
Thanks !
I caught a bug while I was adding a cram test to test this out and changed the code to make it simpler.

